PR #21, opened 2025-03-26
Automatic generation of scheduler RBAC

omer-dayan:


pkg/scheduler/cache/status_updater/default_status_updater.go
```diff
@@ -107,6 +109,9 @@ func (su *defaultStatusUpdater) Evicted(
 		message)
 }
 
+// +kubebuilder:rbac:groups="",resources=pods,verbs=create;update;patch
```
romanbaron: also create?

omer-dayan: It has create.. what do you mean?

romanbaron: I meant that I don't think the scheduler creates pods, so why does it need this permission?

omer-dayan: I am sure you right!
Its leftover from yesterday that this line was about pods + pod/status.

Changed!

romanbaron (approved):

PR #22, opened 2025-03-26
added kai prefix to cluster scoped objects

romanbaron:


.github/workflows/on-pr.yaml
```diff
@@ -54,6 +54,7 @@ jobs:
       - name: Build helm chart
         run: |
           sed -i 's#registry/local/kai-scheduler#${{ env.DOCKER_REGISTRY }}#' deployments/kai-scheduler/values.yaml
+          sed -i "s/appVersion.*/appVersion: $PACKAGE_VERSION/" deployments/kai-scheduler/Chart.yaml
```
omer-dayan: You dont need it.
Helm package command get a flag "--app-version" to override the app version

.github/workflows/on-release.yaml
```diff
@@ -53,6 +53,7 @@ jobs:
       - name: Build helm chart
         run: |
           sed -i 's#registry/local/kai-scheduler#${{ env.DOCKER_REGISTRY }}#' deployments/kai-scheduler/values.yaml
+          sed -i "s/appVersion.*/appVersion: $PACKAGE_VERSION/" deployments/kai-scheduler/Chart.yaml
```
omer-dayan: ditto

deployments/kai-scheduler/templates/binder-webhook.yaml
```diff
@@ -4,7 +4,7 @@
 apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
-  name: mutating-binder
+  name: kai-binder-mutating
```
omer-dayan: Why do we need this suffix?
When you have cluster role of scheduler you dont call it "kai-scheduler-clusterrole" because its already type of clusterrole.
So in here the type is MutatingWebhookConfiguration, we do we need the mutating at the end?

.github/workflows/on-pr.yaml
romanbaron: I actually looked for one and didn't find one.. where do you see it?

deployments/kai-scheduler/templates/binder-webhook.yaml
romanbaron: I will remove

.github/workflows/on-pr.yaml
omer-dayan: image

omer-dayan (approved):

PR #30, opened 2025-04-01
snapshot using listers and k8s objects

enoodle:
The internal ClusterInfo struct contains too many interfaces and imported structs that can't have json tags. overcoming it is more complicated than just re-creating the cluster info from the k8s objects, so we will use them for the snapshot.


Makefile
```diff
@@ -33,7 +33,7 @@ $(SERVICE_NAMES):
 	$(MAKE) docker-build-generic SERVICE_NAME=$@
 
 .PHONY: validate
-validate: generate manifests clients gen-license generate-mocks lint
+validate: generate manifests clients gen-license generate-mocks
```
davidLif: Why did you remove the lint call?

davidLif (approved):

Makefile
enoodle: This was part of the revert of earlier PR, I will put this back in

PR #33, opened 2025-04-01
Add nvidia-smi to example gpu-pod

guptaNswati:
This is to add classic nvidia-smi to example gpu-pod 

```
$ kubectl get pod
NAME      READY   STATUS    RESTARTS   AGE
gpu-pod   1/1     Running   0          6s

$ kubectl logs gpu-pod
Tue Apr  1 21:39:23 2025       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 570.86.15              Driver Version: 570.86.15      CUDA Version: 12.8     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GH200 96GB HBM3         On  |   00000009:01:00.0 Off |                    0 |
| N/A   37C    P0            101W /  900W |       1MiB /  97871MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
                                                                                         
+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|  No running processes found                                                             |
+-----------------------------------------------------------------------------------------+
```

### DIFF
--- a/docs/quickstart/pods/gpu-pod.yaml
+++ b/docs/quickstart/pods/gpu-pod.yaml
@@ -13,7 +13,7 @@ spec:
     - name: main
       image: ubuntu
       command: ["bash", "-c"]
-      args: ["nvidia-smi; trap 'exit 0' TERM; sleep 9000 & wait"]
+      args: ["nvidia-smi; trap 'exit 0' TERM; sleep infinity & wait"]
       resources:
         limits:
           nvidia.com/gpu: "1"
--- a/docs/quickstart/pods/gpu-pod.yaml
+++ b/docs/quickstart/pods/gpu-pod.yaml
@@ -12,7 +12,8 @@ spec:
   containers:
     - name: main
       image: ubuntu
-      args: ["sleep", "infinity"]
+      command: ["bash", "-c"]
+      args: ["nvidia-smi; trap 'exit 0' TERM; sleep 9000 & wait"]
       resources:
         limits:
           nvidia.com/gpu: "1"


docs/quickstart/pods/gpu-pod.yaml
```diff
@@ -12,7 +12,8 @@ spec:
   containers:
     - name: main
       image: ubuntu
-      args: ["sleep", "infinity"]
+      command: ["bash", "-c"]
+      args: ["nvidia-smi; trap 'exit 0' TERM; sleep 9000 & wait"]
```
omer-dayan: What do you think about making it `infinity` instead of `9000`?
We prefer not to make time limit so people can run it go do other things and get back directly where they left.

guptaNswati: For quick testing, i think people may just want to run a simple container like this quickly and get onto more useful things once they get a hang of it. I would prefer to increase the time to a bigger number than infinity because i had to force terminate this pod.

omer-dayan: Yeah you had to force terminate your code because `sleep` does not handle signals, no matter if its 9000 or infinity.
You handled it with the `trap` command (which is a great option), and I suggest of keeping the trap, just changing the number to infinity.
So you wont need to force terminate the pod + the guide will not be depend on times.
What do you think?

guptaNswati: Okay i will update it :)

omer-dayan (approved):

guptaNswati:
Should be good to merge.

PR #43, opened 2025-04-03
chore: running e2e tests with each PR ci

enoodle:


hack/fake-gpu-operator-values.yaml
```diff
@@ -0,0 +1,23 @@
...
+      gpuMemory: 11441
+    mig:
+      gpuProduct: A100-SXM4-40GB
+      gpuCount: 0
+      otherDevices:
+        - name: nvidia.com/mig-1g.5gb
+          count: 3
+        - name: nvidia.com/mig-2g.10gb
+          count: 2
+  nodePoolLabelKey: run.ai/simulated-gpu-node-pool
```
davidLif: Currently, KAI doesn't support sharding with the basic helm install

enoodle: These are values for the fake gpu operator, not for KAI helm chart.

davidLif (approved):

PR #46, opened 2025-04-04
Roman/resource reservation nvml discovery

romanbaron:


.github/workflows/on-pr.yaml
```diff
@@ -9,8 +9,8 @@ env:
   DOCKER_REGISTRY: "nvcr.io/nvstaging/runai"
 
 jobs:
-  test-build-and-push:
-    name: Test & Build & Push
+  test-and-build:
```
omer-dayan, line 12: You need to change repo settings as well

build/builder/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# Copyright 2025 NVIDIA CORPORATION
+# SPDX-License-Identifier: Apache-2.0
+
+FROM golang:1.23.4-bullseye as builder
```
omer-dayan, line 4: No need for the `as builder` right?

omer-dayan (approved):
Looks good!

PR #55, opened 2025-04-08
chore: improve pr ci time

enoodle:


.github/workflows/on-pr.yaml
```diff
@@ -12,6 +12,26 @@ jobs:
   test-and-build:
     name: Test & Build
     runs-on: ubuntu-latest
+    outputs:
+      package_version: ${{ steps.package_version.outputs.PACKAGE_VERSION }}
```
romanbaron: should probably be only in build, not here.

.github/workflows/on-pr.yaml
```diff
@@ -12,6 +12,26 @@ jobs:
   test-and-build:
     name: Test & Build
```
romanbaron:
```suggestion
  validate-and-test:
    name: Validate & Test
```

.github/workflows/on-pr.yaml
enoodle: :+1:

romanbaron (approved):

PR #59, opened 2025-04-08
Add a simple DRA example

guptaNswati:
This is to add a simple DRA example using the [k8s-dra-driver-gpu](https://github.com/NVIDIA/k8s-dra-driver-gpu) to show how a BindRequest is created.

### DIFF
--- a/docs/dra/gpu-imex-pod.yaml
+++ b/docs/dra/gpu-imex-pod.yaml
@@ -1,0 +1,34 @@
+# Copyright 2025 NVIDIA CORPORATION
+# SPDX-License-Identifier: Apache-2.0
+---
+apiVersion: resource.nvidia.com/v1beta1
+kind: ComputeDomain
+metadata:
+  name: imex-channel-injection
+spec:
+  numNodes: 1
+  channel:
+    resourceClaimTemplate:
+      name: imex-channel-0
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: gpu-imex-pod
+  labels:
+    runai/queue: test
+spec:
+  schedulerName: kai-scheduler
+  containers:
+    - name: main
+      image: ubuntu
+      command: ["bash", "-c"]
+      args: ["nvidia-smi; ls -la /dev/nvidia-caps-imex-channels; trap 'exit 0' TERM; sleep 9000 & wait"]
+      resources:
+        limits:
+          nvidia.com/gpu: "1"
+        claims:
+        - name: imex-channel-0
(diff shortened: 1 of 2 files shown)


docs/developer/binder.md
```diff
@@ -27,6 +27,116 @@ The scheduler and binder communicate through a custom resource called `BindReque
 
 The BindRequest API serves as a clear contract between the scheduler and binder, allowing them to operate independently.
 
+### Example BindRequest for NVIDIA GPU DRA Driver
```
omer-dayan: I see you put it under the developers docs.
Dont you prefer to have it in the quickstart examples?

And example of how to run workload with DRA

guptaNswati: Whatever you suggest. I initially thought of adding it in the quickstart examples with the other gpu-example but then i thought may be its better to put it in BindRequest as when i was looking at DRA and DRA plugin, i ended up spending some time in creating the BindRequest myself, only realizing later that its autogenerated by the Binder.

romanbaron: Thanks for adding DRA example, I think it is important enough to have a dedicated section in the docs. Probably in `docs/dra` directory.
Quickstart is a place where we point to users right after installation so they can easily try the very basic flow with KAI Scheduler.

guptaNswati: Agree. docs/dra sounds good.

docs/dra/gpu-imex-pod.yaml
```diff
@@ -0,0 +1,34 @@
...
+      name: imex-channel-0
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: gpu-imex-pod
+  labels:
+    runai/queue: test
+spec:
+  schedulerName: kai-scheduler
```
enoodle, line 21: @guptaNswati Do you think that we need to add node affinity here like in the example form  the gpu-operator that your referenced?

```
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: nvidia.com/gpu.clique
            operator: Exists

```

guptaNswati: mm i dont think its necessary for this simple example, especially if its tried on a GB200 node. This is more useful in a heterogeneous cluster

enoodle (approved):

guptaNswati:
Its ready to merge.

romanbaron (approved):

PR #71, opened 2025-04-14
docs: specify legacy v1 version of kubeflow-training-operator

tgasla:
This update corrects the documentation for PyTorchJob in batch scheduling to reflect the fact that the kubeflow-training-operator-v1 is required, rather than the v2 version. The original instructions were misleading, as the v2 version does not support the necessary functionality for scheduling jobs as described. The updated doc now correctly points to the legacy v1 version of the operator, which is compatible with gang scheduling for PyTorch jobs.

### DIFF
--- a/docs/batch/README.md
+++ b/docs/batch/README.md
@@ -13,7 +13,7 @@ This will create 2 pods that will be scheduled separately. Both pods will either
 To run in a distributed way across multiple pods, you can use PyTorchJob.
 
 ### Prerequisites
-This requires the [training-operator](https://github.com/kubeflow/trainer) to be installed in the cluster.
+This requires the [kubeflow-training-operator-v1](https://www.kubeflow.org/docs/components/trainer/legacy-v1/) to be installed in the cluster.
 
 ### Instructions
 Apply the following command to create a sample PyTorchJob with a master pod and two worker pods:


romanbaron:
Looks good, we probably want to support v2 version of Kubeflow workloads but until this support is added changing the docs is the right thing to do.

tgasla:
Exactly, that was my initial plan as well — to migrate the batch examples so they work with the v2 version of the operator. However, it turned out to be less straightforward than expected. The spec has changed significantly in v2, and unfortunately, there isn't clear documentation on how to migrate from v1 to v2. As you can see [here](https://www.kubeflow.org/docs/components/trainer/operator-guides/migration/#migration-paths), the migration paths are not well defined.

To avoid having incorrect or misleading documentation in the meantime, I think the best approach is to explicitly state that only v1 is supported for now. Once proper support and guidance for v2 becomes available, we can revisit the migration.

romanbaron:
> Exactly, that was my initial plan as well — to migrate the batch examples so they work with the v2 version of the operator. However, it turned out to be less straightforward than expected. The spec has changed significantly in v2, and unfortunately, there isn't clear documentation on how to migrate from v1 to v2. As you can see [here](https://www.kubeflow.org/docs/components/trainer/operator-guides/migration/#migration-paths), the migration paths are not well defined.
> 
> To avoid having incorrect or misleading documentation in the meantime, I think the best approach is to explicitly state that only v1 is supported for now. Once proper support and guidance for v2 becomes available, we can revisit the migration.

I believe we don't really need to migrate since we can support both v1 and v2. I can imagine it as an additional PodGrouper plugin. Similar to this one:
https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/podgrouper/podgrouper/supported_types/supported_types.go#L109

If you want to look into it further, there are docs that explain about PodGrouper plugin mechanism:
https://github.com/NVIDIA/KAI-Scheduler/blob/main/docs/developer/pod-grouper.md

And, you are welcome to open a discussion/issue on this manner and we can guide you through the process more closely. Since v2 is the official Kubeflow version, we definitely want this integration in KAI Scheduler.

romanbaron (approved):

tgasla:
I'll try to take a look at it and enable support for v2. I'll probably start an issue about that and I may need your guidance as well. Many thanks!

PR #77, opened 2025-04-17
BUG fix - handle pod grouper top owner climbing RBAC error

davidLif:


keinanDaniel (approved):

pkg/podgrouper/podgrouper/podgrouper.go
```diff
@@ -130,7 +138,7 @@ func (pg *podGrouper) getOwnerInstance(ctx context.Context, ownerRef *metav1.Own
 		Version: groupVersion[1],
 		Kind:    ownerRef.Kind,
 	})
-	err := pg.client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: ownerRef.Name}, owner)
+	err := pg.clientWithoutCache.Get(ctx, types.NamespacedName{Namespace: namespace, Name: ownerRef.Name}, owner)
```
natasharomm, line 141: I think this cannot be uncached always,
we need to check if we are familiar with the type - if so, use cached client;
only use uncached if the type is unfamiliar.

davidLif: This function is only used for the top-owner climbing. All other functions are used a cahed client. 
I don't think that this case is really worth remembering different types of GVKs just to know when to try and use the cached client - most of the objects in this case are unlikely to appear in cache

enoodle (approved):

PR #106, opened 2025-04-28
Made resource reservation parameters configurable

romanbaron:


.github/workflows/on-pr.yaml
```diff
@@ -106,10 +106,8 @@ jobs:
 
       - name: Deploy fake gpu operator
         run: |
-          helm repo add fake-gpu-operator https://fake-gpu-operator.storage.googleapis.com
-          helm repo update
```
davidLif, line 110: Why did you remove these lines?

cmd/binder/app/options.go
```diff
@@ -25,37 +29,61 @@ type Options struct {
...
+		"scheduler-name", "kai-scheduler",
+		"The scheduler name the workloads are scheduled with")
+	fs.StringVar(&options.ResourceReservationNamespace,
+		"resource-reservation-namespace", "kai-resource-reservation",
+		"Namespace for resource reservation pods")
+	fs.StringVar(&options.ResourceReservationServiceAccount,
+		"resource-reservation-service-account", "kai-resource-reservation",
+		"Service account name for resource reservation pods")
+	fs.StringVar(&options.ResourceReservationPodImage,
+		"resource-reservation-pod-image", "registry/local/kai-scheduler/resource-reservation",
```
davidLif, line 49: Where is this path pointing to? registry/local/kai-scheduler/resource-reservation

romanbaron: nowhere, it is the default registry we build KAI chart with - it can be loaded to node cache in a kind cluster or replaced with a real registry url if you have one.
This default is being overwritten in the chart:
https://github.com/NVIDIA/KAI-Scheduler/blob/main/deployments/kai-scheduler/templates/services/binder.yaml#L28C4-L28C32

.github/workflows/on-pr.yaml
romanbaron: Fake GPU operator is now available on `ghcr.io` with oci protocol, it does not function as a helm repo and can't be added as such. So only `helm upgrade` command remains.

davidLif (dismissed):

PR #107, opened 2025-04-28
fix: gpu resource device count calculation

enoodle:


davidLif (dismissed):

pkg/scheduler/api/resource_info/gpu_resource_requirment.go
```diff
@@ -158,6 +158,9 @@ func (g *GpuResourceRequirement) GPUs() float64 {
 }
 
 func (g *GpuResourceRequirement) GetNumOfGpuDevices() int64 {
+	if g.portion == 0 && g.gpuMemory == 0 {
+		return 0
+	}
```
davidLif: I think that this case should be addressed in the new NewGpuResourceRequirementWithGpus as well.

davidLif (approved):

pkg/scheduler/api/resource_info/gpu_resource_requirment.go
enoodle: :+1:

davidLif (approved):

PR #108, opened 2025-04-29
Fix typo in priority docs

Sovietaced:
Was reading up on the priority docs and just noticed this spelling error.


romanbaron (approved):
Thanks for this fix @Sovietaced

PR #112, opened 2025-05-01
Add Parent Queue Quota Checking to Scheduler.

shchennu:
# Add Parent Queue Quota and Limit Checking to KAI Scheduler

## Description
This PR enhances the KAI Scheduler with comprehensive parent queue resource management, implementing both quota and limit checking at parent queue levels. The changes improve scheduler efficiency by preventing unnecessary scheduling attempts and provide better resource management for different job types.

## Changes
- Added `checkParentQueueLimits` method to `CapacityPolicy` to check both limits and quotas at parent queue levels
- Added `getFirstPendingPod` helper function to identify the first pending pod in a job
- Modified `OnSessionOpen` to include early validation of resource limits
- Added comprehensive test cases for parent queue limit checking
- Added support for elastic jobs with minimum required resources
- Separated job priority from preemptibility handling

## Implementation Details
- Resource checks are performed for GPU, CPU, and Memory resources
- Both limits and quotas are checked, using the more restrictive value
- Checks are done at each parent queue level in the hierarchy
- Early validation prevents unnecessary scheduling attempts
- Proper error messages indicate which resource limit/quota was exceeded
- Special handling for preemptible jobs (PriorityInferenceNumber)
- Support for elastic jobs with minimum resource requirements

## Test Coverage
- Basic limit/quota checks for each resource type (GPU, CPU, Memory)
- Multi-level queue hierarchy checks
- Elastic job handling with minimum required resources
- Preemptible vs non-preemptible job behavior
- Edge cases (zero limits/quotas, missing values)
- Error message formatting
- Helper function validation

## Files Changed
- Modified: `pkg/scheduler/plugins/proportion/capacity_policy/capacity_policy.go`
- Modified: `pkg/scheduler/plugins/proportion/capacity_policy/parent_queue_test.go`

## Testing
All tests are passing:
- 37 specs from the original test suite
- 10 new test cases for parent queue limit checking
- 3 test cases for the helper function

## Impact
- Improves scheduler efficiency by preventing unnecessary scheduling attempts
- Provides clear error messages when limits/quotas are exceeded
- Better resource management for different job types (preemptible, elastic)
```

### DIFF
--- a/pkg/scheduler/plugins/proportion/capacity_policy/capacity_policy.go
+++ b/pkg/scheduler/plugins/proportion/capacity_policy/capacity_policy.go
@@ -125,119 +125,107 @@ func getFirstPendingPod(job *podgroup_info.PodGroupInfo) *pod_info.PodInfo {
 	return nil
 }
 
-// OnSessionOpen is called when a new scheduling session begins. It registers
-// the early quota checking function that prevents jobs from being considered
-// for scheduling if they would exceed their parent queues' quotas.
-func (cp *CapacityPolicy) OnSessionOpen(ssn *framework.Session) {
-	// Register early quota checks
-	ssn.AddPrePredicateFn(func(task *pod_info.PodInfo, job *podgroup_info.PodGroupInfo) error {
-		// Only check for the first pending pod to avoid duplicate checks
-		firstPending := getFirstPendingPod(job)
-		if firstPending == nil || task != firstPending {
-			return nil
-		}
-
-		// Check parent queue quotas
-		return cp.checkParentQueueQuotas(job, ssn)
-	})
+// isPreemptibleJob checks if a job is preemptible based on its priority.
+// Preemptible jobs are allowed to exceed queue limits/quotas.
+func (cp *CapacityPolicy) isPreemptibleJob(job *podgroup_info.PodGroupInfo) bool {
+	preemptiblePriorities := map[int32]bool{
+		constants.PriorityInferenceNumber:              true,
+		constants.PriorityInteractivePreemptibleNumber: true,
+		constants.PriorityTrainNumber:                  true,
+	}
+	return preemptiblePriorities[job.Priority]
 }
 
-// checkParentQueueQuotas verifies that a job's resource requirements don't
-// exceed quotas at any level in its queue hierarchy. This includes:
-// - GPU quota checks
-// - CPU quota checks
-// - Memory quota checks
+// checkParentQueueLimits verifies that a job's resource requirements don't
+// exceed limits or quotas at any level in its queue hierarchy. This includes:
+// - GPU limit/quota checks
+// - CPU limit/quota checks
+// - Memory limit/quota checks
 //
 // The function traverses up the queue hierarchy starting from the job's
-// immediate parent queue. If any quota would be exceeded, it returns an
+// immediate parent queue. If any limit or quota would be exceeded, it returns an
 // error with a detailed message.
 //
-// Note: Preemptible jobs (PriorityTrainNumber) are allowed to exceed parent
-// queue quotas, while non-preemptible jobs must strictly adhere to quotas.
-func (cp *CapacityPolicy) checkParentQueueQuotas(job *podgroup_info.PodGroupInfo, ssn *framework.Session) error {
-	// Skip quota checks for preemptible jobs
-	if job.Priority == constants.PriorityTrainNumber {
-		log.InfraLogger.V(5).Infof("Job: <%v/%v> is preemptible, skipping parent queue quota checks", job.Namespace, job.Name)
+// For preemptible jobs, the limit/quota checks are skipped as they are allowed
+// to exceed queue limits/quotas.
+func (cp *CapacityPolicy) checkParentQueueLimits(job *podgroup_info.PodGroupInfo, ssn *framework.Session) error {
+	// Skip limit/quota checks for preemptible jobs
+	if cp.isPreemptibleJob(job) {
 		return nil
 	}
 
 	// Get queue info for this job
 	queue, found := ssn.Queues[job.Queue]
 	if !found {
-		return nil // Can't check quota without queue info
+		return nil // Can't check limits/quotas without queue info
 	}
 
-	// Only check parent queues, not the job's direct queue
-	currentQueueID := queue.ParentQueue
+	// Calculate job's minimum required resources
+	jobResources := resource_info.EmptyResource()
+	for _, pod := range job.PodInfos {
+		if pod.Status == pod_status.Pending {
+			jobResources.AddResourceRequirements(pod.ResReq)
+		}
+	}
 
-	for currentQueueID != "" {
+	// Traverse up the queue hierarchy
+	for currentQueueID := queue.ParentQueue; currentQueueID != ""; currentQueueID = ssn.Queues[currentQueueID].ParentQueue {
 		parentQueue, found := ssn.Queues[currentQueueID]
 		if !found {
 			break
 		}
 
-		// Calculate job's total resource requirements
-		jobResources := resource_info.EmptyResource()
-		for _, pod := range job.PodInfos {
-			if pod.Status == pod_status.Pending {
-				jobResources.AddResourceRequirements(pod.ResReq)
-			}
+		// Check resource limits and quotas
+		resourceChecks := []struct {
+			resourceType string
+			limit        float64
+			quota        float64
+			used         float64
+		}{
+			{"GPU", float64(parentQueue.Resources.GPU.Limit), float64(parentQueue.Resources.GPU.Quota), jobResources.GPUs()},
+			{"CPU", float64(parentQueue.Resources.CPU.Limit), float64(parentQueue.Resources.CPU.Quota), jobResources.Cpu()},
+			{"Memory", float64(parentQueue.Resources.Memory.Limit), float64(parentQueue.Resources.Memory.Quota), jobResources.Memory()},
 		}
 
-		// Check GPU quota
-		if parentQueue.Resources.GPU.Quota > 0 && jobResources.GPUs() > float64(parentQueue.Resources.GPU.Quota) {
-			errorMsg := fmt.Sprintf(
-				"parent queue '%s' quota has reached the allowable limit of GPUs. "+
-					"Limit is %.0f GPUs, workload requested %.0f GPUs",
-				parentQueue.Name,
-				parentQueue.Resources.GPU.Quota,
-				jobResources.GPUs())
+		for _, check := range resourceChecks {
+			// Check if either limit or quota is exceeded
+			if (check.limit > 0 && check.used > check.limit) || (check.quota > 0 && check.used > check.quota) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by romanbaron — COMMENTED** [review]
Hi @shchennu 👋 
Thanks for suggesting this change, I want to better understand the motivation here, more specifically what is the gap with `IsJobOverQueueCapacity`? From what I see it is called before `PrePredicateFn` so there is probably something that you identified as a gap in the implementation?

Another question, what is the problem with job allocation above parent's queue quota? If the job is preemptible it can be allowed.

**Comment by shchennu** [comment]
> Hi @shchennu 👋 Thanks for suggesting this change, I want to better understand the motivation here, more specifically what is the gap with `IsJobOverQueueCapacity`? From what I see it is called before `PrePredicateFn` so there is probably something that you identified as a gap in the implementation?
> 
> Another question, what is the problem with job allocation above parent's queue quota? If the job is preemptible it can be allowed.

Hi @romanbaron 👋

Thank you for your questions! Let me address them:

1. **Gap with `IsJobOverQueueCapacity`**:
   - `IsJobOverQueueCapacity` only checks the immediate queue's capacity
   - Our new `checkParentQueueQuotas` function checks ALL parent queues up the hierarchy
   - This is important because a job might fit in its immediate queue but exceed parent queue quotas
   - By checking in `PrePredicateFn`, we fail fast and avoid unnecessary scheduling attempts

2. **Preemptible Jobs and Parent Queue Quotas**:
   - You're absolutely right! Preemptible jobs should be allowed to exceed parent queue quotas
   - I've updated the implementation to handle this:
     - Added a check for `PriorityTrainNumber` at the start of `checkParentQueueQuotas`
     - Preemptible jobs now skip quota checks entirely
     - Non-preemptible jobs still maintain strict quota enforcement
   - This change is reflected in the test cases:
     - `preemptible_job_can_exceed_parent_queue_GPU_quota`
     - `non-preemptible_job_cannot_exceed_parent_queue_GPU_quota`

The changes ensure that:
- Preemptible jobs can utilize resources beyond parent queue quotas
- Non-preemptible jobs maintain strict quota enforcement
- We still get early validation for jobs that would exceed quotas

File: pkg/scheduler/plugins/proportion/capacity_policy/capacity_policy.go [hunk 1]
```diff
@@ -83,3 +113,131 @@ func getRequiredQuota(tasksToAllocate []*pod_info.PodInfo) *podgroup_info.JobReq
...
+			return pod
+		}
+	}
+	return nil
+}
+
+// OnSessionOpen is called when a new scheduling session begins. It registers
+// the early quota checking function that prevents jobs from being considered
+// for scheduling if they would exceed their parent queues' quotas.
+func (cp *CapacityPolicy) OnSessionOpen(ssn *framework.Session) {
```
enoodle: Maybe I am missing something, but I don't think this is called by the changes in the current PR because this is not a plugin that is registered.

pkg/scheduler/plugins/proportion/capacity_policy/capacity_policy.go
```diff
@@ -1,29 +1,49 @@
...
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/resource_info"
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/constants"
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/framework"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/log"
 	rs "github.com/NVIDIA/KAI-scheduler/pkg/scheduler/plugins/proportion/resource_share"
 )
 
+// capacityCheckFn is a function type that checks if a job's requested resources
+// exceed capacity limits. It returns a SchedulableResult indicating whether the
+// job can be scheduled.
```
enoodle, line 27: Its a style comment, because this is a new project and all, but we don't have many obvious comments here (the comment really doesn't add anything that is not written in the next line)

pkg/scheduler/plugins/proportion/capacity_policy/capacity_policy.go
```diff
@@ -83,3 +113,131 @@ func getRequiredQuota(tasksToAllocate []*pod_info.PodInfo) *podgroup_info.JobReq
...
+	}
+
+	// Only check parent queues, not the job's direct queue
+	currentQueueID := queue.ParentQueue
+
+	for currentQueueID != "" {
+		parentQueue, found := ssn.Queues[currentQueueID]
+		if !found {
+			break
+		}
```
enoodle: This together with line 239 can be one line of `for` with initialization, check and step

pkg/scheduler/plugins/proportion/capacity_policy/capacity_policy.go
```diff
@@ -83,3 +113,131 @@ func getRequiredQuota(tasksToAllocate []*pod_info.PodInfo) *podgroup_info.JobReq
...
+
+		// Calculate job's total resource requirements
+		jobResources := resource_info.EmptyResource()
+		for _, pod := range job.PodInfos {
+			if pod.Status == pod_status.Pending {
+				jobResources.AddResourceRequirements(pod.ResReq)
+			}
+		}
+
+		// Check GPU quota
```
enoodle: this part seems duplicated 3 times, there are functions to avoid it in the code base. you can loop on the resource type and the . look at proportion/proportion.go for example.

pkg/scheduler/plugins/proportion/capacity_policy/capacity_policy.go
```diff
@@ -83,3 +113,131 @@ func getRequiredQuota(tasksToAllocate []*pod_info.PodInfo) *podgroup_info.JobReq
...
+		if parentQueue.Resources.GPU.Quota > 0 && jobResources.GPUs() > float64(parentQueue.Resources.GPU.Quota) {
+			errorMsg := fmt.Sprintf(
+				"parent queue '%s' quota has reached the allowable limit of GPUs. "+
```
enoodle: A queue can go over the quota, but not the limit. I would change the whole thing to address the limit.

pkg/scheduler/plugins/proportion/capacity_policy/capacity_policy.go
```diff
@@ -83,3 +113,131 @@ func getRequiredQuota(tasksToAllocate []*pod_info.PodInfo) *podgroup_info.JobReq
...
+//
+// The function traverses up the queue hierarchy starting from the job's
+// immediate parent queue. If any quota would be exceeded, it returns an
+// error with a detailed message.
+//
+// Note: Preemptible jobs (PriorityTrainNumber) are allowed to exceed parent
+// queue quotas, while non-preemptible jobs must strictly adhere to quotas.
+func (cp *CapacityPolicy) checkParentQueueQuotas(job *podgroup_info.PodGroupInfo, ssn *framework.Session) error {
+	// Skip quota checks for preemptible jobs
+	if job.Priority == constants.PriorityTrainNumber {
```
enoodle: We aim to separate the job priority and preemptibility soon, and the Train priority is not the only one that is currently preemptible.
We do have several scalars to define the resource of a queue: quota, limit and over quota weight. We should look at limit here and not quota and then it will be correct for all jobs.

enoodle (commented):
Hi,
I left a few comments, mostly about style.

I do have a question about the logic: how will this handle elastic jobs who doesn't have to schedule all pods to run? The job has a minimal number of pods that it needs to run, and you don't know how many will actually fit in the cluster / queue limits.

pkg/scheduler/plugins/proportion/capacity_policy/capacity_policy.go
```diff
@@ -83,3 +113,131 @@ func getRequiredQuota(tasksToAllocate []*pod_info.PodInfo) *podgroup_info.JobReq
...
+}
+
+// OnSessionOpen is called when a new scheduling session begins. It registers
+// the early quota checking function that prevents jobs from being considered
+// for scheduling if they would exceed their parent queues' quotas.
+func (cp *CapacityPolicy) OnSessionOpen(ssn *framework.Session) {
+	// Register early quota checks
+	ssn.AddPrePredicateFn(func(task *pod_info.PodInfo, job *podgroup_info.PodGroupInfo) error {
+		// Only check for the first pending pod to avoid duplicate checks
+		firstPending := getFirstPendingPod(job)
```
enoodle: Isn't the fact that the job needs scheduling (and then predicates are running on it) is not enough to understand that it has pending pods?

shchennu:
> Hi, I left a few comments, mostly about style.
> 
> I do have a question about the logic: how will this handle elastic jobs who doesn't have to schedule all pods to run? The job has a minimal number of pods that it needs to run, and you don't know how many will actually fit in the cluster / queue limits.

Thanks for the review! I've addressed all the comments:
- Added support for elastic jobs with minimum resource requirements
- Improved code style by removing redundant comments and combining loops
- Now checking both limits and quotas, using the more restrictive value
- Separated job priority from preemptibility
- Added test coverage

All tests are passing. Let me know if you'd like any further improvements!

enoodle:
> 1. **Gap with `IsJobOverQueueCapacity`**:
>    
>    * `IsJobOverQueueCapacity` only checks the immediate queue's capacity
>    * Our new `checkParentQueueQuotas` function checks ALL parent queues up the hierarchy
>    * This is important because a job might fit in its immediate queue but exceed parent queue quotas
>    * By checking in `PrePredicateFn`, we fail fast and avoid unnecessary scheduling attempts


@shchennu I looked at `IsJobOverQueueCapacity` and I don't agree with your observation - it is checking all the parent queues, so this addition is not needed.
https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/scheduler/plugins/proportion/capacity_policy/quota_check.go#L34

romanbaron:
@shchennu - all hierarchy levels are validated in these two places:
[resultsOverLimit](https://github.com/NVIDIA/KAI-Scheduler/blob/2a0bacdbba1444b72236373c84fc8be742e57b63/pkg/scheduler/plugins/proportion/capacity_policy/max_allowed_check.go#L19) - checks that workload scheduling won't cause any of the queues (from leaf to top) to go over limit.
[resultsWithNonPreemptibleOverQuota](https://github.com/NVIDIA/KAI-Scheduler/blob/2a0bacdbba1444b72236373c84fc8be742e57b63/pkg/scheduler/plugins/proportion/capacity_policy/quota_check.go#L34) - checks that non-preemptible workload scheduling won't cause any of the queues (from leaf to top) to go over quota.

IsJobOverQueueCapacity calls these two functions here:
https://github.com/NVIDIA/KAI-Scheduler/blob/2a0bacdbba1444b72236373c84fc8be742e57b63/pkg/scheduler/plugins/proportion/capacity_policy/capacity_policy.go#L35

Doesn't it cover it?

romanbaron:
@shchennu - can you please update on your thoughts regarding the questions above?

itayvallach:
@shchennu Is there an update here?

enoodle:
Seems irrelevant now, Please re-open if it changes.

PR #113, opened 2025-05-02
Remove repeated active departments queue creation 

bgedik:
Created this as a test MR to see what the contribution process looks like. 

The PR moves a code block that is unnecessarily repeated under a for loop. There appears to be a bug in the `buildActiveJobOrderPriorityQueues` method. The code between lines 214-221 is incorrectly placed inside the outer for loop that iterates over `jobsOrder.ssn.Queues`. This means that for every queue, it's:
 * Creating a new activeDepartments priority queue
 * Rebuilding the entire department priority queue from scratch
 * Pushing all departments into it

Looks inefficient.


davidLif (approved):

davidLif (approved):
You are right - this seems to be a mistake.

PR #116, opened 2025-05-05
Add/amend test logging to use .TestTopologyBasic.Name

ArmedGuy:
Currently some tests are trying to read TestToplogyMetadata.Name which is empty for all tests, causing go test -v to return very non-descript test cases such as `integration_tests_utils.go:52: Running test number: 0, test name: ` Since .TestTopologyBasic.Name is set for all cases, this diff changes all log lines to that instead for now.

Now they output:
`reclaim_test.go:31: Running test number: 11, test name: 1 train job of 2 GPUs from queue0 running on node0 with deserved of 1, 2 pending jobs from queue1 - reclaim,`


enoodle (approved):
Thanks!

Issue #63, opened 2025-04-09
Feature: Support Scheduling Gates


kannon92:
ref: https://github.com/NVIDIA/KAI-Scheduler/issues/29#issuecomment-2781462103.

Scheduling gates are a GA API field as part of the Kube-Scheduler. They provide a way to "pause" scheduling.

See the thread above for more motivation.

itsomri:
Should be solved now: https://github.com/NVIDIA/KAI-Scheduler/pull/122

PR #122, opened 2025-05-07
Scheduling gates

itsomri:
See https://kubernetes.io/docs/concepts/scheduling-eviction/pod-scheduling-readiness/ & #63 

We want to respect k8s pod scheduling gates, which are used to delay pending pods from being considered for scheduling until other conditions in the cluster are met.

For podgroups, gated pods will not count towards "minMember". So, for example, a podgroup with minMember=4, 4 pending pods, with one of them being gated - will be considered not ready for scheduling. However, if the podgroup has enough ungated pods to satisfy minMember - it will schedule those pods. Additional pods will get scheduled as they become ungated.

### DIFF
--- a/pkg/scheduler/api/pod_status/pod_status.go
+++ b/pkg/scheduler/api/pod_status/pod_status.go
@@ -10,6 +10,9 @@ const (
 	// Pending means the task is pending in the apiserver.
 	Pending PodStatus = 1 << iota
 
+	// Gated means the task is gated in the apiserver.
+	Gated
+
 	// Allocated means the scheduler assigns a host to it.
 	Allocated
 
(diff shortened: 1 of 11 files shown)


pkg/scheduler/api/pod_status/pod_status.go
```diff
@@ -44,10 +45,21 @@ const (
 	Deleted
 )
 
+// Status group constants using bitwise OR
+const (
+	activeUsedStatuses      = Allocated | Pipelined | Binding | Bound | Running | Releasing
```
enoodle, line 50: How is it working? Allocated is not including Pending and Gated... (3 => 11 in binary, which is Pending (01) and Gated (10) 
Can you add unit tests to all the functions that you changed here? only because it is not a trivial refactor.

itsomri: The | operator is bitwise OR, and each status is a power of two. So, say Allocated = 001, Pipelined = 010, Binding = 100, the OR between them is 111.
Added a test for `TestIsAliveStatus`, should be enough to validate this functionality

enoodle (approved):

PR #126, opened 2025-05-09
[Design] Minimum runtime before preemptions and reclaims

ArmedGuy:
Here follows a proposed design for introducing parameters for min-runtime as a concept in KAI, allowing jobs to get guarantees about how long they can execute before getting reclaimed and/or preempted. This would be configurable on a scheduler level, as well as per queue.

It covers the changes to scheduler to store startTime, sets up new plugin extension points to allow plugins to determine reclaim/preemptability, and proposes updates to solvers to pass this information on to scheduler actions.

This is a draft, so changes are expected after input from the team. :)


docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+The configuration follows a hierarchical override structure:
+
+1. **Node Pool Level**: Base configuration that applies to all reclaims/preemptions if not overridden by a queue. Default will be set to 0 which preserves existing behaviors to always reclaim/preempt.
+2. **Queue Level**: Overrides node pool configuration for reclaims/preemptions, can be further overridden by a child queue. Default will be set to unassigned, causing the node pool level value to be used.
+
+### Resolving the applicable min-runtime for reclaims and preemptions
+
+#### Reclaims (preemptor and preemptee are in different queues)
+1. Resolve the lowest common ancestor between the leaf-queues of preemptor and preemptee.
```
bugra-gedik-nv: nit: lowest common ancestor (LCA)

docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,117 @@
...
+2. From the LCA, take 1 step in the hierarchy towards the preemptee's leaf queue
```
bugra-gedik-nv: nit: move back up towards root of tree to locate a queue level override or default to ...

docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,117 @@
...
+https://github.com/NVIDIA/KAI-Scheduler/blob/420efcc17b770f30ca5b899bc3ca8969e352970a/pkg/scheduler/cache/status_updater/default_status_updater.go#L149-L154
+
+This will be a readable annotation that is set to current time when the first pod of a podgroup reaches running state.
+
+For scheduling purposes, the readable timestamp is converted to a unix timestamp when pods are snapshotted, using https://github.com/NVIDIA/KAI-Scheduler/blob/420efcc17b770f30ca5b899bc3ca8969e352970a/pkg/scheduler/api/podgroup_info/job_info.go#L81
+
+### Phase 2
+
+Prepare https://github.com/NVIDIA/KAI-Scheduler/blob/420efcc17b770f30ca5b899bc3ca8969e352970a/pkg/scheduler/framework/session_plugins.go to expose `IsPreemptible()` extension function (potentially moving existing isPreemptible to there?), as well as `IsReclaimable()`.
+These functions will return `[]*common_info.PodID`, a which is the set of PodIDs that are eligible for preemption/reclaims given the preemptor.
```
bugra-gedik-nv: nit: a which  -> which

docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+For the new Is* functions we will do set intersection between the results of each plugin returning the values, and use the result of that.
+
+`IsReclaimable()`/`IsPreemptible()` will be called in each action's victim selection filters, and will be called only AFTER a job has been considered eligible based on the fundamental filters of "reclaims" and "preemptible" (such as preemptible only being relevant for in-queue jobs). If the resulting set has length 0 the whole job is considered not preemptible/reclaimable.
```
bugra-gedik-nv: Victim selection filters are at the job level, but here we have IsReclaimable() and IsPreemptible() at the pod level. I assume you designed them at the pod level to support elastic jobs. Is that the case? If so, this requires more changes to carry candidate victim information at the pod level. How does that compare to an alternative where we keep candidate victim information still at the job level, but mark them to indicate they can only be evicted up their min and not further?

Also, why don't we define IsReclaimable() and IsPreemptible() so that they take a preeptor and preemtee and instead they take just the preemtor?

docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,117 @@
...
+https://github.com/NVIDIA/KAI-Scheduler/blob/420efcc17b770f30ca5b899bc3ca8969e352970a/pkg/scheduler/actions/reclaim/reclaim.go#L154-L158
+
+Additionally will rename existing `Reclaimable()` to `HasReclaimableResources()` to better reflect what it is doing (and its other variables/names for the same purpose). Also need to refactor PluginOptions potentially to reflect the changed name.
+https://github.com/NVIDIA/KAI-Scheduler/blob/420efcc17b770f30ca5b899bc3ca8969e352970a/pkg/scheduler/framework/session_plugins.go#L88
+
+### Phase 3
+
+Implement configuration options for (preempt|reclaim)-min-runtime in node pool and queue configurations.
+
+For node pool level, `pkg/scheduler/conf/scheduler_conf.go` seems like the appropriate place, either in `SchedulerParams` or `SchedulerConfiguration`.
```
bugra-gedik-nv: Run:ai folks to help clarify what's the difference between params and configuration.

docs/developer/designs/min-runtime/README.md
ArmedGuy: They will take preemptor and preemptee, will update the doc to make that clearer (unless you meant you would want to see something else?)

The choice of letting the Is* functions return pod-level preemptability is simply to allow for potentially more advanced scenarios in the future, whilst solving it for elastic now. Passing just the number relies heavily on TaskOrderFn being called the same way both in the plugin and in the solver, so being exact about which pods are eligible for preemption allows more freedom in designing future plugins.

bugra-gedik-nv: > unless you meant you would want to see something else?

That's what I meant. Just wording then.

> The choice of letting the Is* functions return pod-level preemptability

`IsReclaimable()`/`IsPreemptible()` may need renaming if we are returning pre-emptable pods IMO. But if we return reclaimabile status (either false, true, or true with up to x pods), then I think the naming is fine. Thinking from a plugin perspective, I could argue that the task ordering should be kept out of this plugin. If there is a concern about the ordering of tasks and not repeating that, perhaps we can compute the pods from the output of the `IsReclaimable()`/`IsPreemptible()` plugin.

ArmedGuy: I initially looked at the API being `IsReclaimable(preemptor, preemptee)` returning `(bool, int)` but ran into the ordering concern, and then I looked at `(bool, []podIds)` which still sense given the name, but then from an information standpoint `len(podIds) == 0` already conveys not preemptible status, hence why I removed the boolean.
If we want it to be explicit given the function name, I can add the bool back in.

Regarding returning specific pods or just number of pods, I would leave up to Run:ai for how they would like the API to evolve, but heres my motivation:

Basically what happens today is deep in the solver, tasks to preempt are ordered via reverse TaskOrderFn just before they are marked as victims. (https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/scheduler/api/podgroup_info/eviction_info.go#L13-L22)
So if the plugin returns just a number, the flexibility available via the plugin interface is to always only preempt pods from the end of the list. By returning podIDs, it allows plugins to say "you can specifically preempt THESE pods in these jobs", or not preempt at all (or preempt the whole job by returning the whole list).

I can imagine plugins for specialized workloads saying "you can preempt pod 2 and pod 6 because they currently arent doing any work". That could potentially be conveyed via TaskOrderFn as well, hence why id let others decide what we should move forward with.

bugra-gedik-nv: Agreed that it would be good for Run:ai folks to chime in.

I'm thinking that what pods are preferred for pre-emption from an elastic workload is not the reclaim/preemption plugin's concern. For instance, for Ray workloads it may be that the first pod is the master/coordinator and better not be pre-empted. Workload specific orderings is not something the reclaim/preemption plugin should care about.

ArmedGuy:
Outstanding thing:
While this introduces -1 as "not preemptible", this would only be in relation when checking preemptor vs preemptee to resolve the correct value. This means it could not on its own potentially replace the current Priority > constant == not preemptible check. For that we would need to be able to set a value on the job level that cannot be different with different contexts (to properly replace all existing checks for IsPreemptible).

docs/developer/designs/min-runtime/README.md
ArmedGuy: JFYI the min-runtime plugin would also make use of reverse ssn.TaskOrderFn to just select pods up until MinAvailable are left.
All registered plugins such as Ray etc with their custom TaskOrderFn registered will ensure min-runtime makes the optimal selection for them as it selects the least important pods (its agnostic to it). Hence it would not be the plugin itself making the decision of which pods, but I see the point about separation of concern. 

The proportion plugin uses this today: https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/scheduler/plugins/proportion/proportion.go#L60 (its passed to its reclaim functions but not used there, but it is used extensively for queue ordering).

docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Detailed Design
+
+### Reclaim Min Runtime Configuration
+
+The reclaim-min-runtime parameter can be configured with the following values:
+
+- **0 (default)**: Jobs are always preemptible via reclaims
+- **-1**: Jobs are never preemptible via reclaims
+- **Positive value (in seconds)**: Minimum guaranteed runtime before preemption via reclaims
```
itsomri: I don't think it's necessary to specify that we'll use seconds here, IMO we should use a format accepted by time.ParseDuration (https://www.geeksforgeeks.org/time-parseduration-function-in-golang-with-examples/), like we do for scheduling period or staleness-grace-period (see https://github.com/NVIDIA/KAI-Scheduler/blob/main/cmd/scheduler/app/options/options.go), but either way - no need to commit to anything at this point

docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Preempt Min Runtime Configuration
+
+In addition to protecting jobs from being preempted too early with reclaim-min-runtime, we also introduce preempt-min-runtime to ensure that in-queue preemptions are also protected with min-runtime.
+
+The preempt-min-runtime parameter can be configured with the following values:
+
+- **0 (default)**: Jobs can always be preempted by others (subject to reclaim-min-runtime constraints) in the same queue
+- **-1**: Jobs can never be preempted by others in the same queue
+- **Positive value (in seconds)**: Minimum guaranteed runtime before in-queue preemption
```
itsomri: Same comment as above

docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,118 @@
...
+1. Leaf queues `root.A.B.C.leaf1` (preemptor) and `root.A.B.D.leaf2` (preemptee) will use the min-runtime resolved for `root.A.B.D`.
+2. Leaf queues `root.A.B.C.leaf1` (preemptor) and `root.A.B.C.leaf2` (preemptee) will use the min-runtime resolved for `root.A.B.C.leaf2`.
+
+#### Preemptions (preemptor and preemptee are within the same leaf-queue)
+Starting from the leaf-queue, walk the tree until the first defined preempt-min-runtime is set and use that.
+
+##### Example
+1. `root.A.B` has preempt-min-runtime: 600, `root.A.B.C.leaf1` has preempt-min-runtime: 300. Job in leaf1 will have preempt-min-runtime: 300.
+
+1. `root.A.B` has preempt-min-runtime: 600, `root.A.B.C.leaf1` has preempt-min-runtime unset. Job in leaf1 will have preempt-min-runtime: 600.
```
itsomri:
```suggestion
2. `root.A.B` has preempt-min-runtime: 600, `root.A.B.C.leaf1` has preempt-min-runtime unset. Job in leaf1 will have preempt-min-runtime: 600.
```

docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+#### Reclaims (preemptor and preemptee are in different queues)
+1. Resolve the lowest common ancestor between the leaf-queues of preemptor and preemptee.
+2. From the LCA, take 1 step in the hierarchy towards the preemptee's leaf queue
+3. Use the reclaim-min-runtime from this queue, if it is set, otherwise move back up towards root of tree and default to node pool-level configuration value.
+
+The idea around the algorithm here is to isolate settings of min-runtime in the queue tree to only affect siblings in reclaim scenarios, and for the potential to distribute the administration of these values in the queue tree (such as giving a user access to change parts of the tree). 
+As a follow-up, we could also provide a setting to disable this and always use the leaf-tree resolved value in all cases. This could be favorable in a scenario where all min-runtimes in the queue tree are managed by one entity.
+
+##### Example
```
itsomri, line 46: Visuals will help with the examples. I added one for example here https://github.com/NVIDIA/KAI-Scheduler/pull/128 . Cursor & https://www.mermaidchart.com/play are good to build those and they render well on github

docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,118 @@
...
+1. Leaf queues `root.A.B.C.leaf1` (preemptor) and `root.A.B.D.leaf2` (preemptee) will use the min-runtime resolved for `root.A.B.D`.
```
itsomri: I think we need to elaborate more here. Specify that the "leafs" are jobs (if I understood correctly) and that Leaf1 is attempting to reclaim from Leaf2. Also, let's add one example where the sibling queues are higher level, say: root.A.B.C is trying to reclaim from root.A.D.E, and explain that the value will be the min reclaim of D (or whatever default value D inherited from it's ancestors)

docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Phase 3
+
+Implement configuration options for (preempt|reclaim)-min-runtime in node pool and queue configurations.
+
+For node pool level, `pkg/scheduler/conf/scheduler_conf.go` seems like the appropriate place, either in `SchedulerParams` or `SchedulerConfiguration`.
+https://github.com/NVIDIA/KAI-Scheduler/blob/420efcc17b770f30ca5b899bc3ca8969e352970a/pkg/scheduler/conf/scheduler_conf.go#L18-L43
+
+
+Since queues are defined as CRDs, the extra values will have to be implemented in `pkg/apis/scheduling/v2/queue_types.go` under `QueueSpec`.
```
itsomri, line 123: We should probably start v3alpha1 for this. I forsee more changes to the queue (& podgroup) API

docs/developer/designs/min-runtime/README.md
itsomri: This requires some careful thinking because there are a few ways to define the Reclaimable fn.
Our use case for minRuntime only requires the reclaimer job (the reclaimer queue, actually, is enough), and reclaimee job.
Other use cases, for example the proportion plugin's fair share, will need to know the actual resources being evicted vs the ones being allocated, meaning that it needs pod-level granularity, to handle elastic jobs. 
We should take a look at the roadmap and consider potential future use cases and discuss this further

docs/developer/designs/min-runtime/README.md
itsomri: In general, plugin-specific configs should go in scheduler configuration

docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+Unrestrained preemption can lead to resource thrashing, where jobs are repeatedly preempted before making meaningful progress. The min-runtime feature addresses this issue by providing configurable minimum runtime guarantees that can be set by either cluster operators or sometimes within parts of the queue to provide guarantees about minimal useful work.
+
+## Detailed Design
+
+### Reclaim Min Runtime Configuration
+
+The reclaim-min-runtime parameter can be configured with the following values:
+
+- **0 (default)**: Jobs are always preemptible via reclaims
```
romanbaron: An alternative way to represent this option is through a nullable configuration, which can either be set or left unset. If unset, it is interpreted as having no minimum runtime, allowing immediate reclamation.

docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,118 @@
...
+- **-1**: Jobs are never preemptible via reclaims
```
romanbaron: I am not sure about "-1" value, it sounds like a way to transform a preemptible job into a non-preemptible one. We already have this option - maybe it is not really needed?

docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,118 @@
...
+Add startTime to PodGroup by mimicking how staleTimestamp is set today:
+https://github.com/NVIDIA/KAI-Scheduler/blob/420efcc17b770f30ca5b899bc3ca8969e352970a/pkg/scheduler/cache/status_updater/default_status_updater.go#L149-L154
+
+This will be a readable annotation that is set to current time when the first pod of a podgroup reaches running state.
+
+For scheduling purposes, the readable timestamp is converted to a unix timestamp when pods are snapshotted, using https://github.com/NVIDIA/KAI-Scheduler/blob/420efcc17b770f30ca5b899bc3ca8969e352970a/pkg/scheduler/api/podgroup_info/job_info.go#L81
+
+### Phase 2
+
+Prepare https://github.com/NVIDIA/KAI-Scheduler/blob/420efcc17b770f30ca5b899bc3ca8969e352970a/pkg/scheduler/framework/session_plugins.go to expose `IsPreemptible(preemptor, preemptee)` extension function (potentially moving existing isPreemptible to there?), as well as `IsReclaimable(preemptor, preemptee)`.
```
enoodle: We have a problem that we use the word `preemptible` for two things in this project:
1. There is an action called `preempt` (allowing jobs to replace other jobs from the same queue with lower priority)
2. We refer to jobs as `preemptible` or `non preemptible` to also say if they are eligible for the `reclaim` action.

I think that this double usage makes it initially seem like `isReclaimable` is also needed here, but I think it is actually not needed. We have a `Reclaimable` plugin function already that is used, and if I understand the design here correctly we only need to add `isPreemptible` here.

I am also not sure about the signature, I don't think that the preemptor is relevant to the decision if the job can be preempted or not.

docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,118 @@
...
+These functions will return `[]*common_info.PodID`, a which is the set of PodIDs that are eligible for preemption/reclaims given the preemptor.
```
romanbaron: What do we gain from returning a slice of Pod IDs over a bool for the entire workload?

docs/developer/designs/min-runtime/README.md
romanbaron: Do we plan to remove any existing fields?

docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,118 @@
+# [DRAFT] Minimum runtime before preemptions and reclaims
+
+## Overview
+
+This document proposes a new feature called "reclaim-min-runtime" and "preempt-min-runtime" that provides configurable guarantees for job runtime before preemption or reclaim can occur. This feature enables administrators to define minimum runtime guarantees at various levels of the resource hierarchy: node pool, nth level queue and leaf-queue.
```
romanbaron: It may be a very minor detail, but we typically use the term 'workloads' - primarily because KAI supports all phases of the AI lifecycle.

docs/developer/designs/min-runtime/README.md
romanbaron: I agree, we should align on clear terminology. Please suggest one.

docs/developer/designs/min-runtime/README.md
ArmedGuy: Unset on queue level will mean "look at parent" and eventually "look at nodepool", hence wanted to avoid overloading the meaning.

docs/developer/designs/min-runtime/README.md
ArmedGuy: There were discussions that this could be used instead of priority to denote non-preemptible, but we can leave it out if thats not the case.

docs/developer/designs/min-runtime/README.md
ArmedGuy: @enoodle the goal was actually to rename Reclaimable to HasReclaimableResources because the functionality defined by the function is too far from what we need (and too far from a generalized solution imo), and introduce 2 new ones. I did have a look at making use of it, but it would require a bunch of refactoring of proportion plugin.

For the sake of min-runtime, we do need both preemptor and preemptee (or reclaimer and reclaimee) to resolve the min-runtime values. 
Overall the API denotes "can preemptor preempt preemptee", which is generally how preemptability is evaluated in schedulers (because the context of the preemptor matters). It is only if preemptability is controlled in a job setting/only related to the potential victim that you can make the decision in isolation.

Agreed on the terminology. From looking at the code they are really both preemptions as far as the victim is concerned, it is just different types of preemptions filtered on different jobs. One suggestion could be to call it preemption in the code, and differentiate the two via an enum like PreemptionType::InQueue and PreemptionType::Reclaim. Since the concept of preemption is well understood within scheduling (and reclaim vs preempt in KAI sense are just flavors of preemptions with filters IMO).

We do seem to want to have the ability to control the logic for the different types, so having some way to know what type of preemption is being done will still be needed for the plugin to have enough granularity.

docs/developer/designs/min-runtime/README.md
ArmedGuy: Since different plugins might report different slices that they consider okay, it was easier to just do set intersection over all the results of the API and returning whatever slice is left. If 0 elements are in the result, the workload is not preemptible/reclaimable at all. But you can also return (preemptible bool, preemptiblePods []*common_info.PodID) and only filter if the slice is non-nil.

docs/developer/designs/min-runtime/README.md
ArmedGuy: will do, the leaf1 and leaf2 were actually queues here and it is loosely defined as jobs scheduled under these queues. Will clarify.

docs/developer/designs/min-runtime/README.md
ArmedGuy: As I mentioned in a discussion further down, the existing Reclaimable fn signature is way too different (preemptor, []victims) returning list of resources, for this need. I suggested keeping it out of this design by renaming it to be more explicit. But yeah this is probably the main point that we need some clarity on.

I can draft up an MR with the suggestion in this doc if its of any help to visualize the usage.

docs/developer/designs/min-runtime/README.md
bugra-gedik-nv: Can we make Is-pre-emptable plugin agnostic of task order within a job?

docs/developer/designs/min-runtime/README.md
bugra-gedik-nv: The existing Reclaimable (proposed HasReclaimableResources) seems to check if the available reclaimable resources are sufficient to satisfy the reclaimer. One of the questions on my mind is if this is worth being a plugin or whether this should be mainline functionality.

IsPreemptable(reclaimer, reclaimee, preemptionType) 

is more of a plugin functionality that we need for minimum running time. 

I agree that this should be one of the main topics of discussion.

enoodle: I now understand the part of the preemptor in this function, thank you for the clarification.

I still think you can use only one function here because if it gets  both preemptor and preemptee then it can just look at their queues and understand if they are the same or not and branch accordingly if it changes something in the logic. I think that in this case it will be the same: get the LCA queue (if they are in the same one, then it is just that queue) step one level down (if possible) and get the minimum run time from there.

I also see those two actions (Preempt, Reclaim) as being so similar that we could actually maybe merge them into one action that is being called multiple times with different parameters (Preemption type, The set of available jobs to preempt from, hooks to verify proposed preempt scenarios etc). Maybe if we will have less such plugin points it will be easier to do in the future, but it shouldn't be part of this design - so it is up to you.

About the current reclaimable function: I am not sure about your new proposed name. Today it is getting a reclaimer and a bunch of queues with the amounts that it wants to reclaim from them, and checks that all stay withing their allowed boundaries - both the reclaiming queue and the ones being reclaimed from. 
About making it part of the mainline functionality - today only one plugin is implementing it, but we talked in the past about making an alternate proportion like plugin that will only mind the queue priority and limits and then users can choose not to use the current proportion plugin.

docs/developer/designs/min-runtime/README.md
romanbaron: Today workloads are evicted as a gang (besides elastic workloads above minMember which are evicted as separate pods). What will happen if only part of a workload pods are considered preemptible/reclaimable? I understand that with min-runtime feature it will not happen but I also wonder how it will work with other plugins that might return a subset of workload pods.

ArmedGuy: @romanbaron that is a good counter argument to the podID proposal, and does push me towards just returning number of pods like @bugra-gedik-nv suggested `(bool, numPods)` or just `numPods`. It basically means any workload that would want to allow preemption of only a subset of workload pods would always have to be configured as elastic workload, which today is already covered with TaskOrderFn and MinAvailable.

The reason why we need the ability to say "subset of pods or all pods" is so that an elastic job with min-runtime left could still have extra pods preempted up until MinAvailable pods, and only after min-runtime is passed the rest of the job would be eligible.

romanbaron: I think that in partial eviction scenarios the scheduler splits the job into several partial jobs (with only subset of the tasks). In case of elastic job it will be several jobs with single pod and then one job with all the remaining pods (minAvailable set).
https://github.com/NVIDIA/KAI-Scheduler/blob/b24df9984ec75db0dfc8b38b80356a9256573475/pkg/scheduler/actions/common/solvers/pod_scenario_builder.go#L92

So it seems to me that we can always operate at job level.

ArmedGuy:
Okay, I have updated the design with the following reconsiderations:
- `IsPreemptible(actionType, preemptor, preemptee) bool`, the IsPreemptible function will simply return job-level preemptability given an actionType, preemptor and preemptee.
- Introducing `GetAccumulatedScenarioFilters(actionType, session, pendingJob)` pluggable function that allows plugins to return a list of filters that can be used to invalidate various scenarios. This will be used by min-runtime plugin to register a filter that ensures protected elastic workloads cannot be preempted beyond MinAvailable.
- Removed mentions of min-runtime being set to -1. Will leave that out of the design for now, and if its desirable we can bring it in as a v2.
- Keeping the lastStartTime implementation with Annotations for now. This should be easily updatable to use scheduling conditions instead down the line.

bugra-gedik-nv (approved):

docs/developer/designs/min-runtime/README.md
```diff
@@ -0,0 +1,146 @@
+# [DRAFT] Minimum runtime before preemptions and reclaims
```
bugra-gedik-nv: Can we drop the [DRAFT]

bugra-gedik-nv (approved):

enoodle (approved):

ArmedGuy:
Will merge the current plan as we are in agreement, and then we can tackle changes as they come up.

kannon92 (commented):
Forgive me if I miss this.

What is to stop a bad actor on using this to make sure their workloads do not get premepted?

If someone sets a large value for min-runtime I would think that they can get by without getting their workloads preempted.

bugra-gedik-nv:
> Forgive me if I miss this.
> 
> What is to stop a bad actor on using this to make sure their workloads do not get premepted?
> 
> If someone sets a large value for min-runtime I would think that they can get by without getting their workloads preempted.

Minimum running time is *not* configured at the workload level. It is configured as part of queues (and scheduler config). So the configuration is for 'admins' and not for users submitting workloads.

kannon92:
> > Forgive me if I miss this.
> > What is to stop a bad actor on using this to make sure their workloads do not get premepted?
> > If someone sets a large value for min-runtime I would think that they can get by without getting their workloads preempted.
> 
> Minimum running time is _not_ configured at the workload level. It is configured as part of queues (and scheduler config). So the configuration is for 'admins' and not for users submitting workloads.

I see. That makes sense!

So you have an admin would configures queues and users are able to select queues that can allow for different minTime?

ArmedGuy:
@kannon92 Yep its controlled by admins, there is also a concept for having variable min-runtime depending on which queue preemptor and preemptee comes from (to give certain guarantees between hierarchical queues that are modelled after things like org structure), but it will be configurable if it should be on or not.

PR #130, opened 2025-05-12
binder: only run node sync when needed

enoodle:
Cache the gpu sharing pods in the binder so that it will now exactly when it has to run reservation pod syncing.

On clusters with very high amount of pods and actions, such api calls (pod listing with label and node filtering) can be very slow, so this will make sure that we run them only if we have to.

### DIFF
--- a/pkg/binder/binding/resourcereservation/resource_reservation.go
+++ b/pkg/binder/binding/resourcereservation/resource_reservation.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 
 	"golang.org/x/exp/slices"
@@ -43,12 +44,16 @@ const (
 	nodeIndex                      = "runai-node"
 )
 
+type gpuGroupToPodsMap map[string]map[string]bool
+
 type service struct {
 	fakeGPuNodes        bool
 	kubeClient          client.WithWatch
 	reservationPodImage string
 	allocationTimeout   time.Duration
 	gpuGroupMutex       *group_mutex.GroupMutex
+	sharedGPUPods     sync.Map
+
 	namespace           string
 	serviceAccountName  string
 	appLabelValue       string
@@ -91,6 +96,10 @@ func (rsc *service) Sync(ctx context.Context) error {
 }
 
 func (rsc *service) SyncForNode(ctx context.Context, nodeName string) error {
+	if !rsc.cachedSharedGPUPodsOnNode(nodeName) {
+		return nil
+	}
+
 	podsList := &v1.PodList{}
 	err := rsc.kubeClient.List(ctx, podsList,
 		client.HasLabels{constants.GPUGroup},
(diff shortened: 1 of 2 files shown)


pkg/binder/binding/resourcereservation/resource_reservation.go
```diff
@@ -524,3 +537,59 @@ func (rsc *service) isScalingUp(ctx context.Context) bool {
...
+			gpuGroups, found := rsc.sharedGPUPods.Load(nodeName)
+			if !found {
+				gpuGroups = make(gpuGroupToPodsMap)
+			}
+
+			if _, found := gpuGroups.(gpuGroupToPodsMap)[gpuGroup]; !found {
+				gpuGroups.(gpuGroupToPodsMap)[gpuGroup] = make(map[string]bool)
+			}
+			gpuGroups.(gpuGroupToPodsMap)[gpuGroup][pod.Name] = true
+			rsc.sharedGPUPods.Store(nodeName, gpuGroups)
```
davidLif, line 557: Are you sure you don't need to lock this section per node?
2 threads might load, constrcut their version of the list and then one will override the other.

pkg/binder/binding/resourcereservation/resource_reservation.go
```diff
@@ -524,3 +537,59 @@ func (rsc *service) isScalingUp(ctx context.Context) bool {
...
+	if _, found := podsInGroup[pod.Name]; found {
+		delete(podsInGroup, pod.Name)
+	}
+	if len(podsInGroup) == 0 {
+		delete(gpuGroups, gpuGroup)
+	}
+	if len(gpuGroups) > 0 {
+		rsc.sharedGPUPods.Store(nodeName, gpuGroups)
+	} else {
+		rsc.sharedGPUPods.Delete(nodeName)
```
davidLif, line 585: Same here

pkg/binder/binding/resourcereservation/resource_reservation.go
enoodle: There is already such locking mechanism in the functions that call this one that I relay on.

enoodle: it is called from `syncForPods` that is called from `syncForGPUgroupWithLock`

pkg/binder/binding/resourcereservation/resource_reservation.go
enoodle: it is called from `syncForPods` that is called from `syncForGPUgroupWithLock`

enoodle:
There is an issue here with some pods not being cleaned up from this cache, so I will move it back to draft until I fix it.

pkg/binder/binding/resourcereservation/resource_reservation.go
```diff
@@ -524,3 +537,59 @@ func (rsc *service) isScalingUp(ctx context.Context) bool {
...
+			}
+			gpuGroups, found := rsc.sharedGPUPods.Load(nodeName)
+			if !found {
+				gpuGroups = make(gpuGroupToPodsMap)
+			}
+
+			if _, found := gpuGroups.(gpuGroupToPodsMap)[gpuGroup]; !found {
+				gpuGroups.(gpuGroupToPodsMap)[gpuGroup] = make(map[string]bool)
+			}
+			gpuGroups.(gpuGroupToPodsMap)[gpuGroup][pod.Name] = true
```
romanbaron, line 556: You probably want to have namespeced name or use pod id here to avoid colisions

enoodle:
closing for #144

PR #132, opened 2025-05-14
scheduler: bugfix: Make pod_scenario_builder build scenarios for rest of elastic job

ArmedGuy:
While going through the solvers, @bugra-gedik-nv noticed a potential bug in pod_scenario_builder, where we jump to the next scenario if any of the potentialVictimTasks are already in recordedVictimTasks.
https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/scheduler/actions/common/solvers/pod_scenario_builder.go#L77-L82

When an elastic job is the victim, `GetTasksToEvict` will only return the last task from reverse TaskOrderFn of a victim until the job has equal or less than `MinAvailable` tasks to consider for scenario (the `jobHasMoreTasks = false` situation).

https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/scheduler/actions/common/solvers/pod_scenario_builder.go#L71-L82

When you have a reclaimer/preemptor with at least 2 tasks to schedule and no free nodes, we would successfully solve for the first task in job_solver.go by calling `solvePartialJob`, and `recordedVictimJobs` is now populated with an elastic job representation with 1 task. But when solving for 2 tasks in the next iteration of `solvePartialJob`,  the first call to `GetNextScenario()` with `recordedVictimJobs` populated  with the elastic job representation with 1 pod would match the first pod in `potentialVictimTasks` from `GetTasksToEvict()` and directly call  `GetNextScenario()` internally. 
However, because we already popped the full elastic job off `victimsJobsQueue`, the next scenario will no longer evaluate any of the other pods in that elastic job and move on to another job in the victim queue.

In the scenario where the elastic job is the only victim, no further scenario is presented and solver would consider the scenario unsolvable even though the elastic job victim could give up more tasks.

More generalized: it would mean elastic jobs can currently only have 1 pod preempted each scheduling loop.

The proposed solution is when a victimJob popped of the queue has any task that is already part of reportedVictims, we construct a job representative with all remaining tasks that are not reportedVictims, and if the number of remainingTasks is non-zero we push that job back onto the victimQueue before calling GetNextScenario.

### DIFF
--- a/pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
+++ b/pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
@@ -175,7 +175,7 @@ var _ = Describe("PodAccumulatedScenarioBuilder", func() {
 				Expect(len(sn.PotentialVictimsTasks())).To(Equal(potentialVictimsPerScenario[numberOfGeneratedScenarios]))
 				numberOfGeneratedScenarios += 1
 			}
-
+			Expect(numberOfGeneratedScenarios).To(Equal(len(potentialVictimsPerScenario)))
 		})
 	})
 
@@ -260,6 +260,7 @@ var _ = Describe("PodAccumulatedScenarioBuilder", func() {
 				Expect(len(sn.PotentialVictimsTasks())).To(Equal(potentialVictimsPerScenario[numberOfGeneratedScenarios]))
 				numberOfGeneratedScenarios += 1
 			}
+			Expect(numberOfGeneratedScenarios).To(Equal(len(potentialVictimsPerScenario)))
 		})
 	})
 })
--- a/pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
+++ b/pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
@@ -142,6 +142,41 @@ var _ = Describe("PodAccumulatedScenarioBuilder", func() {
 
 			Expect(numberOfGeneratedScenarios).To(Equal(2))
 		})
+
+		It("Returned scenarios have correct number of potential victims", func() {
+			ssn, _ = initializeSession(3, 2)
+			for _, podGroupInfo := range ssn.PodGroupInfos {
+				podGroupInfo.MinAvailable = int32(len(podGroupInfo.PodInfos))
+				podGroupInfo.PodGroup.Spec.MinMember = int32(len(podGroupInfo.PodInfos))
+			}
+			submitQueue := createQueue("team-a")
+			ssn.Queues[submitQueue.UID] = submitQueue
+			reclaimerJob, _ = createJobWithTasks(1, 1, "team-a", v1.PodPending)
+
+			var recordedVictimsJobs []*podgroup_info.PodGroupInfo
+			recordedVictimIndexes := []int{0, 2}
+			podGroupIndex := 0
+
+			for _, podGroupInfo := range ssn.PodGroupInfos {
+				if slices.Contains(recordedVictimIndexes, podGroupIndex) {
+					recordedVictimsJobs = append(recordedVictimsJobs, podGroupInfo)
+				}
+				podGroupIndex += 1
+			}
+
+			victimsQueue := utils.GetVictimsQueue(ssn, nil)
+
+			scenarioBuilder = NewPodAccumulatedScenarioBuilder(ssn, reclaimerJob, recordedVictimsJobs, victimsQueue)
+
+			numberOfGeneratedScenarios := 0
+			potentialVictimsPerScenario := []int{0, 2}
+			for sn := scenarioBuilder.GetCurrentScenario(); sn != nil; sn = scenarioBuilder.GetNextScenario() {
+				Expect(numberOfGeneratedScenarios < len(potentialVictimsPerScenario)).To(BeTrue())
+				Expect(len(sn.PotentialVictimsTasks())).To(Equal(potentialVictimsPerScenario[numberOfGeneratedScenarios]))
+				numberOfGeneratedScenarios += 1
+			}
+
+		})
 	})
 
 	Context("with recorded victims that are elastic", func() {
@@ -159,22 +194,18 @@ var _ = Describe("PodAccumulatedScenarioBuilder", func() {
 
 			var recordedVictimsJobs []*podgroup_info.PodGroupInfo
 
-			podGroupIndex := 0
 			// Only the first pod group with the last task is recordedVictimJobs
 			for _, podGroupInfo := range ssn.PodGroupInfos {
-				if podGroupIndex == 0 {
-					podIndex := 0
-					var partialTasks []*pod_info.PodInfo
-					for _, podInfo := range podGroupInfo.PodInfos {
-						// use last pod as recorded victim as sorting will be reversed
-						if podInfo.Name == "pod-2" {
-							partialTasks = append(partialTasks, podInfo)
-						}
-						podIndex += 1
+				var partialTasks []*pod_info.PodInfo
+				for _, podInfo := range podGroupInfo.PodInfos {
+					// use last pod as recorded victim as sorting will be reversed
+					if podInfo.Name == "pod-2" {
+						partialTasks = append(partialTasks, podInfo)
 					}
-					recordedVictimsJobs = append(recordedVictimsJobs, podGroupInfo.CloneWithTasks(partialTasks))
 				}
-				podGroupIndex += 1
+				recordedVictimsJobs = append(recordedVictimsJobs, podGroupInfo.CloneWithTasks(partialTasks))
+				// we only want to change the first pod group, break after this
+				break
 			}
 
 			victimsQueue := utils.GetVictimsQueue(ssn, nil)
@@ -188,6 +219,47 @@ var _ = Describe("PodAccumulatedScenarioBuilder", func() {
 			}
 
 			Expect(numberOfGeneratedScenarios).To(Equal(3))
+		})
+
+		It("Returned scenarios have correct number of potential victims", func() {
+			// run 1 job with 4 tasks, set minAvailable to 2 for elastic
+			ssn, _ = initializeSession(1, 4)
+			minAvailable := 2
+			for _, podGroupInfo := range ssn.PodGroupInfos {
+				podGroupInfo.MinAvailable = int32(minAvailable)
+				podGroupInfo.PodGroup.Spec.MinMember = int32(minAvailable)
+			}
+			submitQueue := createQueue("team-a")
+			ssn.Queues[submitQueue.UID] = submitQueue
+			reclaimerJob, _ = createJobWithTasks(1, 2, "team-a", v1.PodPending)
+
+			var recordedVictimsJobs []*podgroup_info.PodGroupInfo
+
+			// Only the first pod group with the last task is recordedVictimJobs
+			for _, podGroupInfo := range ssn.PodGroupInfos {
+				var partialTasks []*pod_info.PodInfo
+				for _, podInfo := range podGroupInfo.PodInfos {
+					// use last pod as recorded victim as sorting will be reversed
+					if podInfo.Name == "pod-3" {
+						partialTasks = append(partialTasks, podInfo)
+					}
+				}
+				recordedVictimsJobs = append(recordedVictimsJobs, podGroupInfo.CloneWithTasks(partialTasks))
+				// we only want to change the first pod group, break after this
+				break
+			}
+
+			victimsQueue := utils.GetVictimsQueue(ssn, nil)
+
+			scenarioBuilder = NewPodAccumulatedScenarioBuilder(ssn, reclaimerJob, recordedVictimsJobs, victimsQueue)
+
+			numberOfGeneratedScenarios := 0
+			potentialVictimsPerScenario := []int{0, 1, 3}
+			for sn := scenarioBuilder.GetCurrentScenario(); sn != nil; sn = scenarioBuilder.GetNextScenario() {
+				Expect(numberOfGeneratedScenarios < len(potentialVictimsPerScenario)).To(BeTrue())
+				Expect(len(sn.PotentialVictimsTasks())).To(Equal(potentialVictimsPerScenario[numberOfGeneratedScenarios]))
+				numberOfGeneratedScenarios += 1
+			}
 		})
 	})
 })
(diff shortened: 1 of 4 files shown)


ArmedGuy:
Oh and also re-added ginkgo test suite needed to run these tests, as coverage was missing.

pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
```diff
@@ -142,6 +143,53 @@ var _ = Describe("PodAccumulatedScenarioBuilder", func() {
...
+			ssn.Queues[submitQueue.UID] = submitQueue
+			reclaimerJob, _ = createJobWithTasks(1, 2, "team-a", v1.PodPending)
+
+			var recordedVictimsJobs []*podgroup_info.PodGroupInfo
+
+			podGroupIndex := 0
+			// Only the first pod group with the last task is recordedVictimJobs
+			for _, podGroupInfo := range ssn.PodGroupInfos {
+				if podGroupIndex == 0 {
+					podIndex := 0
```
bugra-gedik-nv, line 166: podIndex unused?

bugra-gedik-nv: podIndex unused?

pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
```diff
@@ -142,6 +143,53 @@ var _ = Describe("PodAccumulatedScenarioBuilder", func() {
...
+			submitQueue := createQueue("team-a")
+			ssn.Queues[submitQueue.UID] = submitQueue
+			reclaimerJob, _ = createJobWithTasks(1, 2, "team-a", v1.PodPending)
+
+			var recordedVictimsJobs []*podgroup_info.PodGroupInfo
+
+			podGroupIndex := 0
+			// Only the first pod group with the last task is recordedVictimJobs
+			for _, podGroupInfo := range ssn.PodGroupInfos {
+				if podGroupIndex == 0 {
```
bugra-gedik-nv: nit: do we need a for loop here? Just take the first of the pod group infos perhaps

pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
```diff
@@ -142,6 +143,53 @@ var _ = Describe("PodAccumulatedScenarioBuilder", func() {
...
+					podIndex := 0
+					var partialTasks []*pod_info.PodInfo
+					for _, podInfo := range podGroupInfo.PodInfos {
```
bugra-gedik-nv, line 168: nit: can we just take the last one without a for loop?

pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
```diff
@@ -142,6 +143,53 @@ var _ = Describe("PodAccumulatedScenarioBuilder", func() {
...
+
+			scenarioBuilder = NewPodAccumulatedScenarioBuilder(ssn, reclaimerJob, recordedVictimsJobs, victimsQueue)
+
+			numberOfGeneratedScenarios := 0
+			for sn := scenarioBuilder.GetCurrentScenario(); sn != nil; sn = scenarioBuilder.GetNextScenario() {
+				Expect(len(sn.RecordedVictimsJobs())).To(Equal(len(recordedVictimsJobs)))
+				numberOfGeneratedScenarios += 1
+			}
+
+			Expect(numberOfGeneratedScenarios).To(Equal(3))
```
bugra-gedik-nv, line 221: Can we assert individual scenarios to make sure they are as expected?

pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
ArmedGuy: good flag, refactored it out before as maps arent ordered. will remove

pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
ArmedGuy: You will be surprised at how annoying it is to get the "first" value of a map. It seems it might be possible to do `vals := slices.Collect(maps.Values(m))` if the project uses a new enough Golang version (Aug 2024).

But can remove the if and just break after first element in the loop in this case.

pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
enoodle: I wonder why the linter didn't fail on it

davidLif (changes requested):
I  reproduced the issue as an e2e test - https://github.com/NVIDIA/KAI-Scheduler/tree/DavidL/reclaim-elastic-to-distributed-bug.

However, I am not sure about the suggested solution. 
The `remainingTasks` will contain all the all the elastic job pods except for the "last" pod (which is part of the recorded victims). This means that all the pods for this elastic jobs will become potential victims. In some cases (for example, if the reclaimee job has min +2 pods) we want only one pod to join as the the next "potential victim".

I will look into it a bit more, but of the top of my head an implementation of `podgroup_info.GetTasksToEvict` that knows to skip spesific pods might be more closly aligned to the required result.

Thank you for your input — this is an important bug.

bugra-gedik-nv:
> I reproduced the issue as an e2e test - https://github.com/NVIDIA/KAI-Scheduler/tree/DavidL/reclaim-elastic-to-distributed-bug.
> 
> However, I am not sure about the suggested solution. The `remainingTasks` will contain all the all the elastic job pods except for the "last" pod (which is part of the recorded victims). This means that all the pods for this elastic jobs will become potential victims. In some cases (for example, if the reclaimee job has min +2 pods) we want only one pod to join as the the next "potential victim".
> 
> I will look into it a bit more, but of the top of my head an implementation of `podgroup_info.GetTasksToEvict` that knows to skip spesific pods might be more closly aligned to the required result.
> 
> Thank you for your input — this is an important bug.

@davidLif For elastic jobs, GetTasksToEvict seems to return 1 pod if num pods > MinAvailable and MinAvailable pods if num pods == MinAvailable.

> This means that all the pods for this elastic jobs will become potential victims

I don't see this, but might be wrong. Can you explain how this can happen? We should be able to see it in unit test.

@ArmedGuy  Would be good to have the unit test updated to MinAvailable = 2 and total pods to 4, so the possible scenarios after one pod is already in recorded victims are:
- evict 1 (bringing us down to 2 pods)
- evict 3 (bring us down to 0 pods)

ArmedGuy:
@davidLif It should not be a problem, because this logic of taking all tasks that aren't reportedVictims already and pushing a job representative back into the queue with them will make the job pass by `GetTasksToEvict` on the next scenario, which does the filtering down to 1 job for us. It is the same logic that is already prescribed further down when `GetTasksToEvict` returns an elastic job with `jobHasMoreTasks = true`.

pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
bugra-gedik-nv: Strictly speaking, the variable is used down below (it is incremented). However it is not needed. Perhaps the use makes it pass linting.

davidLif:
> @davidLif It should not be a problem, because this logic of taking all tasks that aren't reportedVictims already and pushing a job representative back into the queue with them will make the job pass by `GetTasksToEvict` on the next scenario, which does the filtering down to 1 job for us. It is the same logic that is already prescribed further down when `GetTasksToEvict` returns an elastic job with `jobHasMoreTasks = true`.

I stand corrected — you were right, and my earlier statement was inaccurate.

ArmedGuy:
Added 2 new test cases to validate potential victims returned in each scenario to address @bugra-gedik-nv s point. Added for both elastic and non-elastic to validate both.

pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
ArmedGuy: yep the incr caused linter to not care, I removed this.

ArmedGuy:
@davidLif Should I pull in the e2e tests to this diff so we can see fix passing them?

pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
```diff
@@ -141,6 +142,125 @@ var _ = Describe("PodAccumulatedScenarioBuilder", func() {
...
+			}
+
+			victimsQueue := utils.GetVictimsQueue(ssn, nil)
+
+			scenarioBuilder = NewPodAccumulatedScenarioBuilder(ssn, reclaimerJob, recordedVictimsJobs, victimsQueue)
+
+			numberOfGeneratedScenarios := 0
+			potentialVictimsPerScenario := []int{0, 2}
+			for sn := scenarioBuilder.GetCurrentScenario(); sn != nil; sn = scenarioBuilder.GetNextScenario() {
+				Expect(numberOfGeneratedScenarios < len(potentialVictimsPerScenario)).To(BeTrue())
```
bugra-gedik-nv, line 174: better to have a final expect outside of the loop, in case we generate less scenarios than expected.

pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
```diff
@@ -141,6 +142,125 @@ var _ = Describe("PodAccumulatedScenarioBuilder", func() {
...
+			victimsQueue := utils.GetVictimsQueue(ssn, nil)
+
+			scenarioBuilder = NewPodAccumulatedScenarioBuilder(ssn, reclaimerJob, recordedVictimsJobs, victimsQueue)
+
+			numberOfGeneratedScenarios := 0
+			for sn := scenarioBuilder.GetCurrentScenario(); sn != nil; sn = scenarioBuilder.GetNextScenario() {
+				Expect(len(sn.RecordedVictimsJobs())).To(Equal(len(recordedVictimsJobs)))
+				numberOfGeneratedScenarios += 1
+			}
+
```
bugra-gedik-nv, line 220: Any reason we don't have `potentialVictimsPerScenario` for this one?

bugra-gedik-nv (approved):

pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
ArmedGuy: This was to prevent out of bounds slice access from erroring out, giving a cryptic runtime panic.

ArmedGuy: But good point on still ensuring all scenarios were validated, will keep the expect outside as well.

pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
ArmedGuy: I separated the unit tests so the original makes sure RecordedVictimJobs is preserved (the intent of the original unit test) and the new ones check for potentialVictims instead.

pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
```diff
@@ -141,6 +142,126 @@ var _ = Describe("PodAccumulatedScenarioBuilder", func() {
 
 			Expect(numberOfGeneratedScenarios).To(Equal(2))
 		})
+
+		It("Returned scenarios have correct number of potential victims", func() {
```
bugra-gedik-nv: nit: I think the test case descriptions are made to read like a sentence:

It("returns scenarios with the correct number of potential victims")

pkg/scheduler/actions/common/solvers/pod_scenario_builder_test.go
```diff
@@ -141,6 +142,126 @@ var _ = Describe("PodAccumulatedScenarioBuilder", func() {
...
+				Expect(numberOfGeneratedScenarios < len(potentialVictimsPerScenario)).To(BeTrue())
+				Expect(len(sn.PotentialVictimsTasks())).To(Equal(potentialVictimsPerScenario[numberOfGeneratedScenarios]))
+				numberOfGeneratedScenarios += 1
+			}
+			Expect(numberOfGeneratedScenarios).To(Equal(len(potentialVictimsPerScenario)))
+		})
+	})
+
+	Context("with recorded victims that are elastic", func() {
+		It("All scenarios have the same recorded victims", func() {
```
bugra-gedik-nv: nit: same, let's keep the It(...) part consistent with the rest, so it reads like a sentence.

enoodle (dismissed):

davidLif:
> @davidLif Should I pull in the e2e tests to this diff so we can see fix passing them?

@ArmedGuy yes.

davidLif (approved):

test/e2e/suites/reclaim/reclaim_elastic_test.go
```diff
@@ -145,4 +145,86 @@ var _ = Describe("Reclaim with Elastic Jobs", Ordered, func() {
...
+		}
+		_, reclaimerPods := pod_group.CreateDistributedJob(
+			ctx, testCtx.KubeClientset, testCtx.ControllerClient,
+			reclaimerQueue, 2, reclaimerPodRequirements, "",
+		)
+		reclaimerNamespace := queue.GetConnectedNamespaceToQueue(reclaimerQueue)
+		wait.ForPodsScheduled(ctx, testCtx.ControllerClient, reclaimerNamespace, reclaimerPods)
+
+		// verify results
+		wait.ForPodsWithCondition(ctx, testCtx.ControllerClient, func(watch.Event) bool {
```
bugra-gedik-nv, line 222: don't we need to verify that the elastic reclaimee job has 1 pod remaining?

test/e2e/suites/reclaim/reclaim_elastic_test.go
```diff
@@ -145,4 +145,86 @@ var _ = Describe("Reclaim with Elastic Jobs", Ordered, func() {
...
+			pods, err := testCtx.KubeClientset.CoreV1().Pods(reclaimeeNamespace).List(ctx, metav1.ListOptions{
+				LabelSelector: fmt.Sprintf("%s=%s", podGroupLabelName, reclaimeePodGroup.Name),
+			})
+			Expect(err).To(Succeed())
+			return len(pods.Items) == 0
```
bugra-gedik-nv: should be == 1 

but then, if we want to make sure it is stable, it would be best to wait a little more to make sure it doesn't change. Not sure what's the recommended way of doing that.

ArmedGuy: have an update for stability testing, but will hold that off to another PR since its a change on the entire e2e framework.

davidLif (approved):

PR #138, opened 2025-05-14
PodGroup info caching for some of the results

davidLif:
1- Add caching to GetTasksToAllocate and GetTasksToAllocateInitResource
2- GetActivelyRunningTasksCount returns the count instead of the pod 
3- save the amount of ActiveAllocated pods in podgroup info instead of calculating them on request


pkg/scheduler/api/podgroup_info/job_info.go
```diff
@@ -76,6 +77,10 @@ type PodGroupInfo struct {
 	StalenessInfo
 
 	schedulingConstraintsSignature common_info.SchedulingConstraintsSignature
+
+	// inner cache
```
itsomri, line 80: PodStatusIndex & ActiveAllocatedCount are also types of inner cache

pkg/scheduler/api/podgroup_info/job_info.go
```diff
@@ -182,31 +192,35 @@ func (podGroupInfo *PodGroupInfo) UpdateTaskStatus(task *pod_info.PodInfo, statu
...
+	podGroupInfo.tasksToAllocateInitResource = nil
+}
+
+func (podGroupInfo *PodGroupInfo) GetActiveAllocatedTasksCount() int {
+	return podGroupInfo.ActiveAllocatedCount
 }
 
-func (podGroupInfo *PodGroupInfo) GetActivelyRunningTasks() []*pod_info.PodInfo {
-	var tasks []*pod_info.PodInfo
+func (podGroupInfo *PodGroupInfo) GetActivelyRunningTasksCount() int32 {
```
itsomri, line 227: Why specifically int32?

pkg/scheduler/api/podgroup_info/job_info_test.go
```diff
@@ -75,8 +75,9 @@ func TestAddTaskInfo(t *testing.T) {
 						case01_task4.UID: case01_task4,
 					},
 				},
-				JobFitErrors:   make(v2alpha2.UnschedulableExplanations, 0),
-				NodesFitErrors: map[common_info.PodID]*common_info.FitErrors{},
+				ActiveAllocatedCount: 3,
```
itsomri: This looks annoying and a possible source of bugs in tests. What if we make it an int pointer, and `GetActiveAllocatedTasksCount` will populate it if nil? That way it will work without having to populate it externally

davidLif: I considered that, but then I didn't want to make this change just for tests. I guess i went a bit overboard with the optimization...

pkg/scheduler/api/podgroup_info/job_info.go
davidLif: The only call for this function have been casting it to int32. In that case, why not return an int32 in the first place?

pkg/scheduler/api/podgroup_info/job_info.go
davidLif: Good point - I will move them into this section as well

pkg/scheduler/test_utils/jobs_fake/jobs.go
```diff
@@ -130,6 +130,7 @@ func BuildJobInfo(
 			},
 		},
 	}
+	_ = result.GetActiveAllocatedTasksCount()
```
enoodle, line 133: Does it have any side effect?

davidLif: Yes. It initializes the the value of the activeAllocatedCount field. This is done to help comapre the 2 pod groups with the reflect package

enoodle: OK, usually we are supposed to use NewPodInfo which does it by adding the tasks one by one

enoodle (approved):

PR #144, opened 2025-05-14
refactor(binder): add cache to the resource reservation client

enoodle:
In order to handle scale better in clusters with very large number of pods where API calls that filter pods are very slow.

### DIFF
--- a/cmd/binder/app/app.go
+++ b/cmd/binder/app/app.go
@@ -6,9 +6,12 @@ package app
 import (
 	"context"
 	"flag"
+	"fmt"
 	"time"
 
 	podmutator "github.com/NVIDIA/KAI-scheduler/pkg/binder/admission/pod-mutator"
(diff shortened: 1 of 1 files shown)


cmd/binder/app/app.go
```diff
@@ -201,3 +215,40 @@ func (app *App) Run() error {
...
 	return nil
 }
+
+func createIndexesForResourceReservation(mgr manager.Manager) error {
+
+	if err := mgr.GetFieldIndexer().IndexField(
+		context.Background(), &corev1.Pod{}, "spec.nodeName",
+		func(obj client.Object) []string {
+			nodeName := obj.(*corev1.Pod).Spec.NodeName
+			if nodeName != "" {
```
itsomri: Nitpicking but let's invert
```suggestion
			if nodeName == "" {
				return nil
			}
			return []string{nodeName}
```

itsomri (approved):

itsomri (approved):

PR #147, opened 2025-05-15
refactor(scheduler): patch pod labels concurrently

enoodle:
To support clusters with very large scale we want to move the patching of the pod node pool labels to happen concurrently in the background instead of blocking the scheduler loop.

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -227,16 +227,16 @@ func (sc *SchedulerCache) Bind(taskInfo *pod_info.PodInfo, hostname string) erro
 	log.InfraLogger.V(3).Infof(
 		"Creating bind request for task <%v/%v> to node <%v> gpuGroup: <%v>, requires: <%v> GPUs",
 		taskInfo.Namespace, taskInfo.Name, hostname, taskInfo.GPUGroups, taskInfo.ResReq)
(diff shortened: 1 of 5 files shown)


pkg/scheduler/cache/cache.go
```diff
@@ -224,8 +224,19 @@ func (sc *SchedulerCache) Bind(taskInfo *pod_info.PodInfo, hostname string) erro
...
 	defer metrics.UpdateTaskBindDuration(startTime)
 
-	err := sc.createBindRequest(taskInfo, hostname)
-	return sc.StatusUpdater.Bound(taskInfo.Pod, hostname, err, sc.getNodPoolName())
+	log.InfraLogger.V(3).Infof(
+		"Creating bind request for task <%v/%v> to node <%v> gpuGroup: <%v>, requires: <%v> GPUs",
+		taskInfo.Namespace, taskInfo.Name, hostname, taskInfo.GPUGroups, taskInfo.ResReq)
+	bindRequestError := sc.createBindRequest(taskInfo, hostname)
+
+	if bindRequestError == nil {
```
itsomri:
```suggestion
	if bindRequestError != nil {
```
handle error first, keep happy flow to the left

enoodle: good point for readability
I again can't use the exact suggestion :(

davidLif (approved):

PR #148, opened 2025-05-15
[Refactor] Reclaimable api improvements

itsomri:


pkg/scheduler/api/types.go
```diff
@@ -23,8 +22,11 @@ type PrePredicateFn func(*pod_info.PodInfo, *podgroup_info.PodGroupInfo) error
 // CanReclaimResourcesFn is a function that determines if a reclaimer can get more resources
 type CanReclaimResourcesFn func(*reclaimer_info.ReclaimerInfo) bool
 
-// EvictableFn is a function which determines if a reclaimer can reclaim a victim.
-type EvictableFn func(*reclaimer_info.ReclaimerInfo, map[common_info.QueueID][]*resource_info.Resource) bool
+// ReclaimeeFilterFn is a function which filters out jobs that cannot a reclaimee candidate for a specific reclaimer.
+type ReclaimeeFilterFn func(*reclaimer_info.ReclaimerInfo, *podgroup_info.PodGroupInfo) bool
```
bugra-gedik-nv: Nit: would be nice to add optional parameter names to function type definitions to make them clear. As in … reclaimer, … reclaimer

bugra-gedik-nv (dismissed):

ArmedGuy:
@itsomri Given the proposed generalization of the API in https://github.com/NVIDIA/KAI-Scheduler/pull/126, could we go ahead now and name ReclaimeeFilter to IsPreemptible(actionType, preemptor, preemptee)? You can swap out ReclaimerInfo for just preemptor PodGroupInfo since the vast majority of info is already in that.

EDIT: You can cherry-pick https://github.com/NVIDIA/KAI-Scheduler/commit/a673d393063f94360997664370cceba63916352e into your branch to get this change.

bugra-gedik-nv (dismissed):

itsomri:
> @itsomri Given the proposed generalization of the API in #126, could we go ahead now and name ReclaimeeFilter to IsPreemptible(actionType, preemptor, preemptee)? You can swap out ReclaimerInfo for just preemptor PodGroupInfo since the vast majority of info is already in that.
> 
> EDIT: You can cherry-pick [a673d39](https://github.com/NVIDIA/KAI-Scheduler/commit/a673d393063f94360997664370cceba63916352e) into your branch to get this change.

I did merge the function signatures, but I don't want to do the actionType at the moment, as it will result in having to resolve the relevant function for each call to the filter - this is something we're trying to avoid due to past experience.
The result of the current implementation is that plugins that have similar logic for both filters, will need to register them twice - explicitly for each action type. I think this is acceptable for now, as introducing new actions is a very rare event.  
We could revisit this decision in the future and find an efficient system to resolve action types with enums, but I think the current implementation here is a good balance for now.
What do you think?

ArmedGuy:
@itsomri sounds good, lets go with this!

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -152,10 +137,15 @@ func getOrderedVictimsQueue(ssn *framework.Session, evictingQueue common_info.Qu
 		})
 		jobs := map[common_info.PodGroupID]*podgroup_info.PodGroupInfo{}
 		for _, job := range ssn.PodGroupInfos {
-			if job.Queue != evictingQueue {
-				jobs[job.UID] = job
+			if job.Queue == reclaimerInfo.Queue {
```
ArmedGuy: nit: you can use reclaimer.Queue and remove reclaimerInfo from the fn signature

ArmedGuy (approved):
approved with nit

ArmedGuy (approved):

PR #149, opened 2025-05-15
Remove requirement for Worker when using PyTorchJob

Phlip79:
Currently, if I launch a PyTorchJob with just a "Master" pod, it will fail as a "Worker" pod is a requirement. It should not be a requirement.

It is also possible to launch just "Worker" pods without a "Master" pod (see [this flag](https://docs.run.ai/v2.19/Researcher/cli-reference/runai-submit-dist-pytorch/#-no-master) as an example).

Ultimately I am removing the requirement to have a "Worker" pod: there will be no requirement to have either a "Master" or "Worker" pod.


davidLif (approved):
Thank you for the contribution!

Phlip79:
@davidLif do you have to submit the PR? I don't see the button to do so myself.

PR #153, opened 2025-05-15
scheduler: Add LastStartTimestamp to PodGroup

ArmedGuy:
As per the design document in https://github.com/NVIDIA/KAI-Scheduler/pull/126, this PR contains phase 1 which adds LastStartTimestamp to pods, stored as an annotation on the PodGroup and retrieved during the scheduling loop.
The timestamp will be set as the time after all Bind statements have been successfully committed in the allocate action.

Overall, this functions similarly to `staleTimestamp` that is set in the stalegangeviction action.

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -43,15 +43,16 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 	for !jobsOrderByQueues.IsEmpty() {
 		job := jobsOrderByQueues.PopNextJob()
 		stmt := ssn.Statement()
+		alreadyAllocated := podgroup_info.HasTasksAllocated(job)
 		if ok, pipelined := attemptToAllocateJob(ssn, stmt, job); ok {
 			metrics.IncPodgroupScheduledByAction()
 			err := stmt.Commit()
+			if err == nil && !pipelined && !alreadyAllocated {
+				setLastStartTimestamp(job)
+			}
 			if err == nil && podgroup_info.HasTasksToAllocate(job, true) {
 				jobsOrderByQueues.PushJob(job)
 				continue
-			}
-			if err == nil && !pipelined {
-				setLastStartTimestamp(job)
 			}
 		} else {
 			stmt.Discard()
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -51,9 +51,7 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 				continue
 			}
 			if err == nil && !pipelined {
-				// Set start time for the job after statement has been successfully committed
-				timeNow := time.Now()
-				job.LastStartTimestamp = &timeNow
+				setLastStartTimestamp(job)
 			}
 		} else {
 			stmt.Discard()
@@ -93,4 +91,9 @@ func attemptToAllocateJob(ssn *framework.Session, stmt *framework.Statement, job
 	}
 
 	return true, pipelined
+}
+
+func setLastStartTimestamp(job *podgroup_info.PodGroupInfo) {
+	timeNow := time.Now()
+	job.LastStartTimestamp = &timeNow
 }
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -4,6 +4,8 @@
 package allocate
 
 import (
+	"time"
+
 	"golang.org/x/exp/maps"
 
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/actions/common"
@@ -41,11 +43,17 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 	for !jobsOrderByQueues.IsEmpty() {
 		job := jobsOrderByQueues.PopNextJob()
 		stmt := ssn.Statement()
-		if attemptToAllocateJob(ssn, stmt, job) {
+		if ok, pipelined := attemptToAllocateJob(ssn, stmt, job); ok {
 			metrics.IncPodgroupScheduledByAction()
 			err := stmt.Commit()
 			if err == nil && podgroup_info.HasTasksToAllocate(job, true) {
 				jobsOrderByQueues.PushJob(job)
+				continue
+			}
+			if err == nil && !pipelined {
+				// Set start time for the job after statement has been successfully committed
+				timeNow := time.Now()
+				job.LastStartTimestamp = &timeNow
 			}
 		} else {
 			stmt.Discard()
@@ -53,7 +61,7 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 	}
 }
 
-func attemptToAllocateJob(ssn *framework.Session, stmt *framework.Statement, job *podgroup_info.PodGroupInfo) bool {
+func attemptToAllocateJob(ssn *framework.Session, stmt *framework.Statement, job *podgroup_info.PodGroupInfo) (bool, bool) {
 	queue := ssn.Queues[job.Queue]
 
 	resReq := podgroup_info.GetTasksToAllocateInitResource(job, ssn.TaskOrderFn, true)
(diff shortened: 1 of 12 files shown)


pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -41,19 +43,23 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
...
 		stmt := ssn.Statement()
-		if attemptToAllocateJob(ssn, stmt, job) {
+		if ok, pipelined := attemptToAllocateJob(ssn, stmt, job); ok {
 			metrics.IncPodgroupScheduledByAction()
 			err := stmt.Commit()
 			if err == nil && podgroup_info.HasTasksToAllocate(job, true) {
 				jobsOrderByQueues.PushJob(job)
+				continue
+			}
+			if err == nil && !pipelined {
```
bugra-gedik-nv, line 53: attemptToAllocateJob was returning false for pipelined anyway. What was the motivation for returning the pipelined status?

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -41,19 +43,23 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
...
+				setLastStartTimestamp(job)
 			}
 		} else {
 			stmt.Discard()
 		}
 	}
 }
 
-func attemptToAllocateJob(ssn *framework.Session, stmt *framework.Statement, job *podgroup_info.PodGroupInfo) bool {
+func attemptToAllocateJob(ssn *framework.Session, stmt *framework.Statement, job *podgroup_info.PodGroupInfo) (bool, bool) {
```
bugra-gedik-nv: Should we use named output parameters to make this clear?

`(allocated bool, pipelined bool)`

You can set them to false at the top.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -41,19 +43,23 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
...
-		if attemptToAllocateJob(ssn, stmt, job) {
+		if ok, pipelined := attemptToAllocateJob(ssn, stmt, job); ok {
 			metrics.IncPodgroupScheduledByAction()
 			err := stmt.Commit()
 			if err == nil && podgroup_info.HasTasksToAllocate(job, true) {
 				jobsOrderByQueues.PushJob(job)
+				continue
+			}
+			if err == nil && !pipelined {
+				setLastStartTimestamp(job)
```
bugra-gedik-nv: We should not set this if this is an elastic job that just got a few pods bound, right?

bugra-gedik-nv: It should also be good to have a unit test that tests not setting it again.

ArmedGuy: If the job had enough pods Allocated for MinAvailable, pipelined will be set to false and the job will be considered started.

pkg/scheduler/actions/allocate/allocate.go
ArmedGuy: Can do but I will still use explicit return. Named parameter return should be avoided for slightly more complex fns.

pkg/scheduler/actions/allocate/allocate.go
bugra-gedik-nv: I mean if it already had min but got more added later.

ArmedGuy: Ah yes, that we need to guard against.

bugra-gedik-nv: I think `getNumOfAllocatedTasks(podGroupInfo) == podGroupInfo.MinAvailable` indicates this is the start.

ArmedGuy: Did it slightly differently, just checking if there were no Allocated pods in the job before we attempted to allocate.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -41,19 +43,24 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 	for !jobsOrderByQueues.IsEmpty() {
 		job := jobsOrderByQueues.PopNextJob()
 		stmt := ssn.Statement()
-		if attemptToAllocateJob(ssn, stmt, job) {
+		alreadyAllocated := podgroup_info.HasTasksAllocated(job)
+		if ok, pipelined := attemptToAllocateJob(ssn, stmt, job); ok {
 			metrics.IncPodgroupScheduledByAction()
 			err := stmt.Commit()
+			if err == nil && !pipelined && !alreadyAllocated {
+				setLastStartTimestamp(job)
```
enoodle, line 51: I don't think that this function is the best place for this:
* jobs can be scheduled in other actions (reclaim, preempt, consolidation, maybe something new in the future)
* this function is currently not doing any direct updates on the PodGroupInfo struct, so maybe it will be better to find a place that already does that.

I think that a better place will be in the cache Bind function. I know that it is binding only a specific task out of the whole job, but maybe you can look there at the task that is the job.MinAvailable that is being bound and then know that it is the right moment to add this.
https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/scheduler/cache/cache.go#L223

enoodle:
Nit: I find the name `LastStartTimestamp` a bit confusing with the value that is assigned that is the time the job started for most jobs. Maybe it can be just `StartTimestamp` ?

bugra-gedik-nv:
> Nit: I find the name `LastStartTimestamp` a bit confusing with the value that is assigned that is the time the job started for most jobs. Maybe it can be just `StartTimestamp` ?

Once a job is pre-empted, wouldn't it get rescheduled? So it may start again.

pkg/scheduler/actions/allocate/allocate.go
ArmedGuy: Jobs can be scheduled, but is there any other action today that will actually create Bind statements? From when I looked through the code in an attempt to answer this, it seemed the other functions would create pipelined jobs (which arent actually ever bound as I understood it).

Did have a look at doing it as close to the bind statements as possible, but as you mention its per task only.

I will do a second attempt at finding a better place for it so that it can cover all scenarios. It sounds like such a place should probably get an API hook to make it easier to trigger things on the condition of "job has enough pods to be started and we sent the requests for it".

itsomri: You're correct, other actions only virtually allocate jobs. The Allocate action is the only one doing real allocations.
I think it's OK to have this logic in the Allocate action, but maybe move the logic from here to `attemptToAllocateJob` - that way we don't need to add the "pipelined" return value to `attemptToAllocateJob`.

pkg/scheduler/api/podgroup_info/allocation_info.go
```diff
@@ -24,6 +24,15 @@ func HasTasksToAllocate(podGroupInfo *PodGroupInfo, isRealAllocation bool) bool
 	return false
 }
 
+func HasTasksAllocated(podGroupInfo *PodGroupInfo) bool {
```
itsomri: Instead, take a look at `PodGroupInfo.GetNumPendingTasks` or `PodGroupInfo.GetNumGatedTasks` at `pkg/scheduler/api/podgroup_info/job_info.go`. They use `podGroupInfo.PodStatusIndex` which is simpler

pkg/scheduler/actions/allocate/allocate.go
ArmedGuy: @itsomri I wanted the value to be set only if the statement commit succeeds. Should the commit be moved into allocate as well if so?

bugra-gedik-nv (dismissed):

pkg/scheduler/api/podgroup_info/allocation_info.go
ArmedGuy: PodStatusIndex stores only per-status. Is it safe enough to just look at if all tasks are in pending?
I wanted to flag any status that might indicate the Job has already gotten a LastStartTimestamp, which I suspected would contain both Allocated, Bound, Binding etc (hence using the AllocatedStatus helper fn).

bugra-gedik-nv: I took the comment to mean sum up the relevant counts without having to iterate over all the pods

bugra-gedik-nv (dismissed):

itsomri:
Please add a CHANGELOG line, something like:
``` md
- Scheduler now adds a "LastStartTimestamp" to podgroup on allocation
```

itsomri (approved):

PR #154, opened 2025-05-19
add pr coverage report

enoodle:


.github/workflows/on-pr.yaml
```diff
@@ -27,6 +28,28 @@ jobs:
...
+    name: Code Coverage Report
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    needs: validate-and-test
+    permissions:
+      contents: read
+      actions: read
+      pull-requests: write
+    steps:
+      - uses: fgrosse/go-coverage-report@v1.1.1
```
ArmedGuy:
```suggestion
      - uses: fgrosse/go-coverage-report@8c1d1a09864211d258937b1b1a5b849f7e4f2682
```

ArmedGuy: should be an allowedlisted version

ArmedGuy (changes requested):

.github/workflows/on-pr.yaml
```diff
@@ -27,6 +28,28 @@ jobs:
...
+    needs: validate-and-test
+    permissions:
+      contents: read
+      actions: read
+      pull-requests: write
+    steps:
+      - uses: fgrosse/go-coverage-report@8c1d1a09864211d258937b1b1a5b849f7e4f2682
+        with:
+          coverage-artifact-name: "code-coverage"
+          coverage-file-name: "coverage/coverage.out"
```
ArmedGuy:
```suggestion
          coverage-file-name: "coverage.out"
```

ArmedGuy (approved):
From what I understand it should be fine to merge now, the failure is due to it not running on main yet.

ArmedGuy:
ah okay, we will need something that runs make test and stores the same artifact running on normal main as well to provide the baseline. Then change `github-baseline-workflow-ref` to point to that. Supposedly we can set this variable to just `update-coverage-badge.yaml` and it will get the baseline from there.

itsomri (approved):

PR #155, opened 2025-05-19
Add min-runtime configuration to queues

ArmedGuy:
This PR adds reclaim and preempt minRuntime to the queue CRD. Because they are optional values, my understanding is they don't need a new version for the CRD at this time (because the schema is still backwards compatible).

For node-pool configuration, my new goal is actually to make use of plugin arguments via:
https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/scheduler/conf/scheduler_conf.go#L68-L69

This way it will be more like a "plugin default" to use when a queue value is not set, and the default will be set to 0.


bugra-gedik-nv (dismissed):
Is the nodepool level config going into scheduler configs?

ArmedGuy:
@bugra-gedik-nv yes, as I understand it the scheduler config contains the plugins used for the scheduler, so you can define it as an argument to the minruntime plugin directly.

itsomri (approved):

itsomri:
Please add a changelog entry, something like:
```md
- Added `PreemptMinRuntime` and `ReclaimMinRuntime` properties to queue CRD
```

itsomri (approved):

PR #156, opened 2025-05-20
refactor(taskorder): add configuration for task order label

enoodle:


omer-dayan (commented):
From the beggining we did not make it configurable because:
1. Its not used as part of the program (No injection of this label at `runai-engine` / `runai`)
2. Its not documented anywhere (No runai docs / No where else)

This means its an API of a nice feature, but its not official API.

We would prefer then break this private API and change the name to be "kai" always because ultimately we would like `runai-engine` to use "kai" labels and not "runai" prefix, all the time.
So no need to make it configurable (tech-debt) to resolve it later

itsomri (approved):

PR #162, opened 2025-05-22
minruntime Plugin

### LINKED ISSUE #136: Min Run Time per workload

Support Min Run Time per workloads.

### Why is this needed?

Unrestrained preemption can lead to resource thrashing, where workloads are repeatedly preempted before making meaningful progress. The min-runtime feature addresses this issue by providing configurable minimum runtime guarantees that can be set by either cluster operators or sometimes within parts of the queue to provide guarantees about minimal useful work.

ArmedGuy:
This implements a plugin that protects jobs with minruntime.

The e2e tests are not completed yet and will currently fail, but we can start reviewing the rest of the code while it gets fixed.

Refactored the plugins so that ReclaimerInfo isnt needed anywhere, and proportion plugin builds that datastructure inside itself, then cleaned up the naming of the functions and added a missing one.

For the plugin, it implements the rough design we talked about, with a few extra bits such as being able to select which resolver type to use for reclaims (LCA method or victim queue).

Added 2 types of caches both for resolving min-runtime between queues and also checking if jobs are protected or not. This should reduce the amount of double-resolving that could happen for elastic workloads, but also for other jobs.

Also, made most e2e functions have a period of steady state (needed for the minruntime, but good for the rest). This ensures we didnt hit a state on one scheduling loop and then regressed the next one. Will be extra important as we start implementing more advanced prioritization etc.

Closes #136

TODO: fix e2e tests, add changelog

### DIFF
--- a/docs/plugins/minruntime.md
+++ b/docs/plugins/minruntime.md
@@ -1,0 +1,121 @@
+# MinRuntime Plugin
+
+## Overview
+
+The MinRuntime plugin for KAI-Scheduler provides runtime protection for jobs by preventing preemption or resource reclamation of jobs until they have run for a specified minimum time duration. This ensures workloads can complete critical initialization or checkpointing before being interrupted.
+
+## Key Features
+
+- **Preemption Protection**: Prevents jobs from being preempted until they have run for a specified minimum duration
+- **Reclamation Protection**: Prevents elastic jobs from having resources reclaimed until they have run for a specified minimum duration
+- **Queue-based Configuration**: Configure different minimum runtime durations for different queues in your scheduling hierarchy
+- **Hierarchical Inheritance**: Minimum runtime settings cascade down from parent queues to leaf queues
+- **Flexible Resolution Methods**: Two methods for reclaim minimum runtime resolution:
+  - Queue-based: Evaluates the min-runtime protection based on the victim's queue configuration
+  - LCA (Lowest Common Ancestor): Uses the common ancestor of reclaimer and victim queues to determine min-runtime protection
+
+## Usage
+
+The minruntime plugin is configured in two ways:
+
+1. **Queue Definition**: Set minimum runtime values in Queue configurations
+2. **Plugin Arguments**: Configure default values and resolution methods in the scheduler configuration
+
+### Queue Configuration
+
+Queues can specify the following minimum runtime parameters:
+
+- `preemptMinRuntime`: Minimum runtime before a job in this queue can be preempted
+- `reclaimMinRuntime`: Minimum runtime before a job in this queue can have resources reclaimed
+
+Example Queue definition:
+
+```yaml
+apiVersion: scheduling.run.ai/v2
+kind: Queue
+metadata:
+  name: production
+spec:
+  preemptMinRuntime: "20s"
+  reclaimMinRuntime: "30s"
+```
+
+### Plugin Configuration
+
+In the scheduler configuration (`scheduler-config` ConfigMap), add the minruntime plugin with its arguments:
+
+```yaml
+tiers:
+- plugins:
+  # other plugins...
+  - name: minruntime
+    arguments:
+      defaultPreemptMinRuntime: "10m"
+      defaultReclaimMinRuntime: "10m"
+      reclaimResolveMethod: "lca"  # or "queue"
+```
+
+### Configuration Parameters
+
+| Parameter | Description | Default |
+|-----------|-------------|---------|
+| `defaultPreemptMinRuntime` | Default minimum runtime before preemption if not specified in queue | "0s" |
+| `defaultReclaimMinRuntime` | Default minimum runtime before resource reclamation if not specified in queue | "0s" |
+| `reclaimResolveMethod` | Method to resolve reclaim minimum runtime ("lca" or "queue") | "lca" |
+
+0s means workloads are instantly reclaimable/preemptible.
+
+## Resolution Methods
+
+### Preemption Resolution
+
+For preemption, the minimum runtime is determined by starting from the victim's queue and walking up the queue hierarchy until a `preemptMinRuntime` value is found.
+
+### Reclamation Resolution
+
+For reclamation, two methods are supported:
+
+1. **Queue-based Resolution** (`reclaimResolveMethod: "queue"`):
+   - Similar to preemption, looks at the victim's queue and walks up the hierarchy until a `reclaimMinRuntime` value is found.
+   - If no value is found at all, use the default value for the plugin.
+
+2. **LCA Resolution** (`reclaimResolveMethod: "lca"`):
+   - Identifies the Lowest Common Ancestor (LCA) in the queue hierarchy between the preemptor and victim
+   - Walks one step down from the LCA towards the victim and uses that queue's `reclaimMinRuntime` value
+   - If no value is found, moves up towards the root queue until one is found.
+   - If no value is found at all, use the default value for the plugin.
+
+The LCA method is the default method if none is specified. The purpose of the LCA method is to follow how policies might be set for queue hierarchy, allowing users in sub-queues to set min-runtime values that are honored by their siblings, whilst not affecting cousin queues.
+
+## Elastic Jobs Handling
+
+For elastic jobs (where `MinAvailable < len(PodInfos)`), the plugin:
+
+1. Allows the job to be considered for preemption/reclamation in the filter phase
+2. Validates in the scenario validator phase that the job will maintain its minimum number of pods if min-runtime has not been reached.
+
+## Implementation Details
+
+The plugin implements the following functions:
+
+- `preemptFilterFn`: Filters victims that shouldn't be preempted due to minimum runtime
+- `reclaimFilterFn`: Filters victims that shouldn't have resources reclaimed due to minimum runtime
+- `preemptScenarioValidatorFn`: Validates preemption scenarios for elastic jobs
+- `reclaimScenarioValidatorFn`: Validates reclamation scenarios for elastic jobs
+
+## Example Workflow
+
+1. A job starts running in a queue with `reclaimMinRuntime: "30s"`
+2. At 20 seconds of runtime, another job attempts to reclaim resources
+3. The minruntime plugin identifies that the minimum runtime has not been reached
+4. The victim job is protected from reclamation until it runs for at least 30 seconds
+
+## Caching
+
+The plugin maintains caches to improve performance:
+- `preemptMinRuntimeCache`: Caches preemption minimum runtime values by queue
+- `reclaimMinRuntimeCache`: Caches reclamation minimum runtime values by queue pair
+- `preemptProtectionCache`: Tracks jobs protected from preemption
+- `reclaimProtectionCache`: Tracks jobs protected from reclamation
+
+These caches are reset at the beginning of each scheduling session.
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -12,7 +12,6 @@ import (
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/common_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/pod_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/podgroup_info"
-	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/reclaimer_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/framework"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/log"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/metrics"
@@ -48,9 +47,7 @@ func (ra *reclaimAction) Execute(ssn *framework.Session) {
 
 	for !jobsOrderByQueues.IsEmpty() {
 		job := jobsOrderByQueues.PopNextJob()
-
-		reclaimerInfo := buildReclaimerInfo(ssn, job)
-		if !ssn.CanReclaimResources(reclaimerInfo) {
+		if !ssn.CanReclaimResources(job) {
 			continue
 		}
 
@@ -69,7 +66,7 @@ func (ra *reclaimAction) Execute(ssn *framework.Session) {
 			}
 		}
 		metrics.IncPodgroupsConsideredByAction()
-		succeeded, statement, reclaimeeTasksNames := ra.attemptToReclaimForSpecificJob(ssn, job, reclaimerInfo)
+		succeeded, statement, reclaimeeTasksNames := ra.attemptToReclaimForSpecificJob(ssn, job)
 		if succeeded {
 			metrics.IncPodgroupScheduledByAction()
 			log.InfraLogger.V(3).Infof(
@@ -88,7 +85,7 @@ func (ra *reclaimAction) Execute(ssn *framework.Session) {
 }
 
 func (ra *reclaimAction) attemptToReclaimForSpecificJob(
-	ssn *framework.Session, reclaimer *podgroup_info.PodGroupInfo, reclaimerInfo *reclaimer_info.ReclaimerInfo,
+	ssn *framework.Session, reclaimer *podgroup_info.PodGroupInfo,
 ) (bool, *framework.Statement, []string) {
 	queue := ssn.Queues[reclaimer.Queue]
 	resReq := podgroup_info.GetTasksToAllocateInitResource(reclaimer, ssn.TaskOrderFn, false)
@@ -100,30 +97,19 @@ func (ra *reclaimAction) attemptToReclaimForSpecificJob(
 	feasibleNodes := common.FeasibleNodesForJob(maps.Values(ssn.Nodes), reclaimer)
 	solver := solvers.NewJobsSolver(
 		feasibleNodes,
-		reclaimableScenarioCheck(ssn, reclaimerInfo),
+		reclaimableScenarioCheck(ssn, reclaimer),
 		getOrderedVictimsQueue(ssn, reclaimer),
 		framework.Reclaim)
 	return solver.Solve(ssn, reclaimer)
 }
 
 func reclaimableScenarioCheck(ssn *framework.Session,
-	reclaimerInfo *reclaimer_info.ReclaimerInfo) solvers.SolutionValidator {
+	reclaimer *podgroup_info.PodGroupInfo) solvers.SolutionValidator {
 	return func(
 		_ *podgroup_info.PodGroupInfo,
 		victimJobs []*podgroup_info.PodGroupInfo,
 		victimTasks []*pod_info.PodInfo) bool {
-		return ssn.ReclaimScenarioValidator(reclaimerInfo, victimJobs, victimTasks)
-	}
-}
-
-func buildReclaimerInfo(ssn *framework.Session, reclaimerJob *podgroup_info.PodGroupInfo) *reclaimer_info.ReclaimerInfo {
-	return &reclaimer_info.ReclaimerInfo{
-		Name:          reclaimerJob.Name,
-		Namespace:     reclaimerJob.Namespace,
-		Queue:         reclaimerJob.Queue,
-		IsPreemptable: reclaimerJob.IsPreemptibleJob(ssn.IsInferencePreemptible()),
-		RequiredResources: podgroup_info.GetTasksToAllocateInitResource(
-			reclaimerJob, ssn.TaskOrderFn, false),
+		return ssn.ReclaimScenarioValidator(reclaimer, victimJobs, victimTasks)
 	}
 }
 
--- a/pkg/scheduler/plugins/minruntime/minruntime.go
+++ b/pkg/scheduler/plugins/minruntime/minruntime.go
@@ -66,8 +66,11 @@ func New(arguments map[string]string) framework.Plugin {
 	if plugin.reclaimResolveMethod == "" {
 		plugin.reclaimResolveMethod = resolveMethodLCA
 	}
+	// setup caches on plugin init, but they will be reset on session open anyway
 	plugin.preemptProtectionCache = make(map[common_info.PodGroupID]map[common_info.PodGroupID]bool)
 	plugin.reclaimProtectionCache = make(map[common_info.PodGroupID]map[common_info.PodGroupID]bool)
+	plugin.preemptMinRuntimeCache = make(map[common_info.QueueID]metav1.Duration)
+	plugin.reclaimMinRuntimeCache = make(map[common_info.QueueID]map[common_info.QueueID]metav1.Duration)
 	return plugin
 }
 
@@ -76,17 +79,24 @@ func (mr *minruntimePlugin) Name() string {
 }
 
 func (mr *minruntimePlugin) OnSessionOpen(ssn *framework.Session) {
-	mr.queues = ssn.Queues
 	ssn.AddReclaimVictimFilterFn(mr.reclaimFilterFn)
 	ssn.AddPreemptVictimFilterFn(mr.preemptFilterFn)
 	ssn.AddReclaimScenarioValidatorFn(mr.reclaimScenarioValidatorFn)
 	ssn.AddPreemptScenarioValidatorFn(mr.preemptScenarioValidatorFn)
+	mr.queues = ssn.Queues
+	// setup caches on session open
+	mr.preemptProtectionCache = make(map[common_info.PodGroupID]map[common_info.PodGroupID]bool)
+	mr.reclaimProtectionCache = make(map[common_info.PodGroupID]map[common_info.PodGroupID]bool)
+	mr.preemptMinRuntimeCache = make(map[common_info.QueueID]metav1.Duration)
+	mr.reclaimMinRuntimeCache = make(map[common_info.QueueID]map[common_info.QueueID]metav1.Duration)
 }
 
 func (mr *minruntimePlugin) OnSessionClose(ssn *framework.Session) {
 	mr.queues = nil
 	mr.preemptProtectionCache = nil
 	mr.reclaimProtectionCache = nil
+	mr.preemptMinRuntimeCache = nil
+	mr.reclaimMinRuntimeCache = nil
 }
 
 func (mr *minruntimePlugin) reclaimFilterFn(pendingJob *podgroup_info.PodGroupInfo, victim *podgroup_info.PodGroupInfo) bool {
--- a/pkg/scheduler/plugins/minruntime/minruntime.go
+++ b/pkg/scheduler/plugins/minruntime/minruntime.go
@@ -102,6 +102,7 @@ func (mr *minruntimePlugin) OnSessionClose(ssn *framework.Session) {
 func (mr *minruntimePlugin) reclaimFilterFn(pendingJob *podgroup_info.PodGroupInfo, victim *podgroup_info.PodGroupInfo) bool {
 	protected := mr.isReclaimMinRuntimeProtected(pendingJob, victim)
 	// always return true for elastic jobs, but cache the result for scenario validator
+	log.InfraLogger.V(3).Infof("preemptFilterFn: pendingJob: %s, victim: %s, protected: %t", pendingJob.Name, victim.Name, protected)
 	if victim.MinAvailable < int32(len(victim.PodInfos)) {
 		return true
 	}
@@ -112,8 +113,8 @@ func (mr *minruntimePlugin) reclaimFilterFn(pendingJob *podgroup_info.PodGroupIn
 func (mr *minruntimePlugin) preemptFilterFn(pendingJob *podgroup_info.PodGroupInfo, victim *podgroup_info.PodGroupInfo) bool {
 	protected := mr.isPreemptMinRuntimeProtected(pendingJob, victim)
 	// always return true for elastic jobs, but cache the result for scenario validator
+	log.InfraLogger.V(3).Infof("preemptFilterFn: pendingJob: %s, victim: %s, protected: %t", pendingJob.Name, victim.Name, protected)
 	if victim.MinAvailable < int32(len(victim.PodInfos)) {
-
 		return true
 	}
 	return !protected
--- a/pkg/scheduler/plugins/minruntime/minruntime.go
+++ b/pkg/scheduler/plugins/minruntime/minruntime.go
@@ -4,6 +4,8 @@ import (
 	"time"
 
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/common_info"
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/pod_info"
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/podgroup_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/queue_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/framework"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/log"
@@ -13,13 +15,22 @@ import (
 const (
 	defaultReclaimMinRuntimeConfig = "defaultReclaimMinRuntime"
 	defaultPreemptMinRuntimeConfig = "defaultPreemptMinRuntime"
-	actionTypeReclaim              = "reclaim"
-	actionTypePreempt              = "preempt"
+	reclaimResolveMethod           = "reclaimResolveMethod"
+	resolveMethodLCA               = "lca"
+	resolveMethodQueue             = "queue"
 )
 
 type minruntimePlugin struct {
 	defaultReclaimMinRuntime *metav1.Duration
 	defaultPreemptMinRuntime *metav1.Duration
+	reclaimResolveMethod     string
+	queues                   map[common_info.QueueID]*queue_info.QueueInfo
+
+	preemptMinRuntimeCache map[common_info.QueueID]metav1.Duration
+	reclaimMinRuntimeCache map[common_info.QueueID]map[common_info.QueueID]metav1.Duration
+
+	preemptProtectionCache map[common_info.PodGroupID]map[common_info.PodGroupID]bool
+	reclaimProtectionCache map[common_info.PodGroupID]map[common_info.PodGroupID]bool
 }
 
 func New(arguments map[string]string) framework.Plugin {
@@ -40,9 +51,11 @@ func New(arguments map[string]string) framework.Plugin {
 				duration = time.Duration(0 * time.Second)
 			}
 			plugin.defaultPreemptMinRuntime = &metav1.Duration{Duration: duration}
+		case reclaimResolveMethod:
+			plugin.reclaimResolveMethod = value
 		}
-	}
 
+	}
 	// Initialize with default values if not provided
 	if plugin.defaultReclaimMinRuntime == nil {
 		plugin.defaultReclaimMinRuntime = &metav1.Duration{Duration: time.Duration(0 * time.Second)}
@@ -50,7 +63,11 @@ func New(arguments map[string]string) framework.Plugin {
 	if plugin.defaultPreemptMinRuntime == nil {
 		plugin.defaultPreemptMinRuntime = &metav1.Duration{Duration: time.Duration(0 * time.Second)}
 	}
-
+	if plugin.reclaimResolveMethod == "" {
+		plugin.reclaimResolveMethod = resolveMethodLCA
+	}
+	plugin.preemptProtectionCache = make(map[common_info.PodGroupID]map[common_info.PodGroupID]bool)
+	plugin.reclaimProtectionCache = make(map[common_info.PodGroupID]map[common_info.PodGroupID]bool)
 	return plugin
 }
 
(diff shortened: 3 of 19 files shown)


docs/plugins/minruntime.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Resolution Methods
+
+### Preemption Resolution
+
+For preemption, the minimum runtime is determined by starting from the victim's queue and walking up the queue hierarchy until a `preemptMinRuntime` value is found.
+
+### Reclamation Resolution
+
+For reclamation, two methods are supported:
```
itsomri, line 76: Maybe let's add some examples here: https://github.com/NVIDIA/KAI-Scheduler/pull/169

docs/plugins/minruntime.md
```diff
@@ -0,0 +1,121 @@
...
+   - Identifies the Lowest Common Ancestor (LCA) in the queue hierarchy between the preemptor and victim
+   - Walks one step down from the LCA towards the victim and uses that queue's `reclaimMinRuntime` value
+   - If no value is found, moves up towards the root queue until one is found.
+   - If no value is found at all, use the default value for the plugin.
+
+The LCA method is the default method if none is specified. The purpose of the LCA method is to follow how policies might be set for queue hierarchy, allowing users in sub-queues to set min-runtime values that are honored by their siblings, whilst not affecting cousin queues.
+
+## Elastic Jobs Handling
+
+For elastic jobs (where `MinAvailable < len(PodInfos)`), the plugin:
```
itsomri: Feels like this part should be accessible for users who may not know the internals of the scheduler, and will not be familiar with "PodInfos"
```suggestion
For elastic jobs (where `MinAvailable < number of pods`), the plugin:
```

test/e2e/modules/wait/watcher/watcher.go
```diff
@@ -28,24 +28,22 @@ type Interface interface {
 type CheckCondition func(watch.Event) bool
 
 var (
-	FlowTimeout = 5 * time.Minute
+	FlowTimeout         = 5 * time.Minute
+	SteadyStateDuration = 10 * time.Second
```
itsomri: Why is this needed? Did you observe flakiness?

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -0,0 +1,239 @@
...
+	preemptMinRuntimeCache map[common_info.QueueID]metav1.Duration
+	reclaimMinRuntimeCache map[common_info.QueueID]map[common_info.QueueID]metav1.Duration
+
+	preemptProtectionCache map[common_info.PodGroupID]map[common_info.PodGroupID]bool
+	reclaimProtectionCache map[common_info.PodGroupID]map[common_info.PodGroupID]bool
+}
+
+func New(arguments map[string]string) framework.Plugin {
+	plugin := &minruntimePlugin{}
+	for key, value := range arguments {
```
itsomri: You can clean most of the code in this fn with a simple helper method:
```Go
func getMinRuntime(rawDuration string, defaultMinRuntime time.Duration) *metav1.Duration {
	if key == "" {
		return &metav1.Duration{Duration: defaultMinRuntime}
	}
	duration, err := time.ParseDuration(rawDuration)
	if err != nil {
		log.InfraLogger.Errorf("Failed to parse %v: %v, using default value 0", rawDuration, err)
		duration = time.Duration(0 * time.Second)
	}
	return &metav1.Duration{Duration: duration}
}
```
And then use it:
```Go
plugin.defaultReclaimMinRuntime = getMinRuntime(arguments[defaultReclaimMinRuntimeConfig], time.Duration(0*time.Second))
```
Note that the arg for default duration uses the time.Duration, not the metav1.Duration - did you specifically mean to use the k8s metav1? (For easier serialization I guess?)

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -0,0 +1,239 @@
...
+	if victim.MinAvailable < int32(len(victim.PodInfos)) {
+		return true
+	}
+	return !protected
+}
+
+func (mr *minruntimePlugin) reclaimScenarioValidatorFn(reclaimer *podgroup_info.PodGroupInfo, victims []*podgroup_info.PodGroupInfo, tasks []*pod_info.PodInfo) bool {
+	for _, victim := range victims {
+		protected := mr.isReclaimMinRuntimeProtected(reclaimer, victim)
+		if protected {
```
itsomri: Try to avoid nesting logic as much as possible:
```suggestion
		if !protected {
			continue
		}
```

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -0,0 +1,239 @@
...
+			numVictimTasks := countVictimTasks(victim, tasks)
```
itsomri: Will it be easier if the scenario will provide the expected jobs and tasks victims structured better? Iterating over all the victim pods for each podgroup can be expensive: remember that the number of scenarios can be huge for each pending job

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -0,0 +1,239 @@
...
+
+func (mr *minruntimePlugin) cachePreemptProtection(pendingJob *podgroup_info.PodGroupInfo, victim *podgroup_info.PodGroupInfo) {
+	if mr.preemptProtectionCache[pendingJob.UID] == nil {
+		mr.preemptProtectionCache[pendingJob.UID] = make(map[common_info.PodGroupID]bool)
+	}
+	mr.preemptProtectionCache[pendingJob.UID][victim.UID] = true
+}
+
+func (mr *minruntimePlugin) cacheReclaimProtection(pendingJob *podgroup_info.PodGroupInfo, victim *podgroup_info.PodGroupInfo) {
+	if mr.reclaimProtectionCache[pendingJob.UID] == nil {
```
itsomri, line 197: How can a single bool represent the protection status of a podgroup from reclaim?
Say I'm a podgroup under queue A.B.C. The reclaim min time for A is 60s, for B is 30s, for C 90s.
After 40s my status is: protected from siblings of A, not protected for siblings of B, protected for siblings of C.
Seems to me that we need to cache this info for every ancestor queue of each podgroup

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -0,0 +1,239 @@
...
+	if mr.reclaimProtectionCache[pendingJob.UID][victim.UID] {
+		return true
+	}
+	// Get the appropriate queue objects for both jobs
+	pendingQueue, foundPending := mr.getQueueForJob(pendingJob)
+	victimQueue, foundVictim := mr.getQueueForJob(victim)
+
+	// If we can't find the queues, use default min runtime
+	var minRuntime metav1.Duration
+	if !foundPending || !foundVictim {
```
itsomri: If we can't find any of the queues it's very concerning: it can mean that some data in the session was corrupted. Highly unlikely, but if it does happen, we should at the very least log this error.

With this in mind, and considering that we usually prefer to not nest happy flow ("Keep happy flow to the left"), I would refactor this part to something along the lines of:
1. having getReclaimMinRuntime return optional error value (that way we can know exactly which queue was missing or what went wrong)
```Go
minRuntime, err := mr.getReclaimMinRuntime(mr.reclaimResolveMethod, pendingQueue, victimQueue)
if err != nil {
	log.InfraLogger.Errorf("Failed to get reclaim min runtime: %v", err)
	minRuntime = *mr.defaultReclaimMinRuntime
}
```
This is cleaner and more informative in case something goes wrong. Also, IMO it's more readable: if we keep happy flow to the left as much as possible, it allows the reader to focus on the important areas of the code first, and only look at special cases when necessary.

pkg/scheduler/plugins/minruntime/resolver.go
```diff
@@ -0,0 +1,241 @@
...
+		return *mr.defaultReclaimMinRuntime
+	}
+
+	// Get hierarchy paths for both queues
+	preemptorPath := mr.getQueueHierarchyPath(preemptorQueue)
+	preempteePath := mr.getQueueHierarchyPath(preempteeQueue)
+
+	// Handle case where both queues are from different top-level hierarchies
+	// Consider them as siblings with a shadow parent
+	if len(preemptorPath) > 0 && len(preempteePath) > 0 {
```
itsomri: Here, instead of nesting importing logic, just handle the case of missing preemptor path early and get it out of the way:
```suggestion
	if len(preemptorPath) == 0 {
		// Handle error - return default value? return error?
	}
	// Do the same for preempteePath (we can combine this to a single `if` statement, but then we will loose some info for the error
	
	// proceed to happy flow
```

pkg/scheduler/plugins/minruntime/resolver.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	// If no reclaim-min-runtime is found in the hierarchy, use default
+	mr.cacheReclaimMinRuntime(preemptorQueue, preempteeQueue, *mr.defaultReclaimMinRuntime)
+	return *mr.defaultReclaimMinRuntime
+}
+
+// getQueueHierarchyPath returns the queue hierarchy from top-level queue to the specified queue
+// Note: For top-level queues (empty parent), the path will only contain the queue itself
+func (mr *minruntimePlugin) getQueueHierarchyPath(
```
itsomri: Some thoughts - not necessarily in the scope of this PR:
Finding a common ancestor for queues will probably be useful in the core scheduler. We can save the hierarchy path for queues on snapshot instead of having the plugin do it. We can also implement the logic of finding the common ancestor in the core scheduler

test/e2e/modules/wait/watcher/watcher.go
ArmedGuy: It was a request from @bugra-gedik-nv to overall test for stability as some of the scenarios we have will return true immediately as the condition is met, but there is no guarantee its actually staying there (for example partially reclaiming an elastic job doesnt test that the rest of the job wasnt reclaimed). But also using it to test min-runtime by ensuring the job stays pending for min-runtime minutes before it can be scheduled.

pkg/scheduler/plugins/minruntime/minruntime.go
ArmedGuy: Yeah this current way is problematic imo. Maybe we can change the tasks into map[podgroupID][]podinfo?
Overall, scenario validator seems expensive as it does validation post-simulation. At first I thought we were going to extend scenario filters (like IdleGpu) instead which at least filters out invalid scenarios before the solve, but it requires a lot of refactoring to do (I tried).

pkg/scheduler/plugins/minruntime/minruntime.go
ArmedGuy: The cache is specifically between a pendingJob and a victim, which will always resolve to the same result within a scheduling cycle. This is so that scenario validator doesn't need to do the lookup again when new scenarios with same victims but more tasks are presented to it (which will happen with elastic jobs). However I should also add the possibility of caching false so the lookup doesn't need to be done when its not protected as well.

pkg/scheduler/plugins/minruntime/resolver.go
```diff
@@ -0,0 +1,241 @@
...
+
+import (
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/common_info"
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/queue_info"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// getPreemptMinRuntime resolves min-runtime for preemptions
+// Starting from the leaf-queue, walk the tree until the first defined preempt-min-runtime is set and use that
+func (mr *minruntimePlugin) getPreemptMinRuntime(
```
bugra-gedik-nv: just a nit: it is usually cleaner if the outermost method only does: check the cache, if there, return it, if not, call another function that resolves the min runtime (no cache involved).

pkg/scheduler/plugins/minruntime/resolver.go
bugra-gedik-nv: Can getQueueHierarchyPath return 0 length array? I thought it always adds the queue itself (given the non null check above). So perhaps we should just assert and not have this `if`.:D

pkg/scheduler/plugins/minruntime/resolver.go
```diff
@@ -0,0 +1,241 @@
...
+	preempteePath := mr.getQueueHierarchyPath(preempteeQueue)
+
+	// Handle case where both queues are from different top-level hierarchies
+	// Consider them as siblings with a shadow parent
+	if len(preemptorPath) > 0 && len(preempteePath) > 0 {
+		topLevelPreemptor := preemptorPath[0]
+		topLevelPreemptee := preempteePath[0]
+
+		// If both are top-level queues (empty parent) and different
+		if topLevelPreemptor.ParentQueue == "" && topLevelPreemptee.ParentQueue == "" &&
```
bugra-gedik-nv: Why the 'topLevelPreemptor.ParentQueue == ""' check? Isn't the `getQueueHierarchyPath` method already going all the way to the top?

pkg/scheduler/plugins/minruntime/resolver.go
```diff
@@ -0,0 +1,241 @@
...
+
+	for i := 0; i < minLength; i++ {
+		if preemptorPath[i].UID != preempteePath[i].UID {
+			break
+		}
+		lcaIndex = i
+	}
+
+	// If we found a common ancestor
+	if lcaIndex < len(preemptorPath) && lcaIndex < len(preempteePath) &&
```
bugra-gedik-nv: lcaIndex must be smaller than len(preemptorPath) and len(preempteePath) by construction. 

We don't need this condition IMO. Because, we know from earlier than at index 0 they have the same UID. So lcaIndex must contain a common ancestor for sure.

For these things that should hold by construction, it is better to use asserts and not create nested if scopes.

pkg/scheduler/plugins/minruntime/resolver.go
```diff
@@ -0,0 +1,241 @@
...
+			childQueue := preempteePath[lcaIndex+1]
+
+			// Check if this queue has reclaim-min-runtime set
+			if childQueue.ReclaimMinRuntime != nil {
+				mr.cacheReclaimMinRuntime(preemptorQueue, preempteeQueue, *childQueue.ReclaimMinRuntime)
+				return *childQueue.ReclaimMinRuntime
+			}
+
+			// Otherwise move back up towards root and select first available override
+			for i := lcaIndex; i >= 0; i-- {
```
bugra-gedik-nv: nit: would be nice to avoid repeating this code below by moving it out of the if/else

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -0,0 +1,239 @@
...
+	mr.queues = nil
+	mr.preemptProtectionCache = nil
+	mr.reclaimProtectionCache = nil
+	mr.preemptMinRuntimeCache = nil
+	mr.reclaimMinRuntimeCache = nil
+}
+
+func (mr *minruntimePlugin) reclaimFilterFn(pendingJob *podgroup_info.PodGroupInfo, victim *podgroup_info.PodGroupInfo) bool {
+	protected := mr.isReclaimMinRuntimeProtected(pendingJob, victim)
+	// always return true for elastic jobs, but cache the result for scenario validator
```
bugra-gedik-nv: nit: move the comment one down?

pkg/scheduler/plugins/minruntime/resolver.go
bugra-gedik-nv: we can just do `if topLevelPreemptor.UID != topLevelPreemptee.UID`

pkg/scheduler/plugins/minruntime/resolver.go
```diff
@@ -0,0 +1,241 @@
...
+	// If no preempt-min-runtime is set in the queue tree, use default
+	mr.preemptMinRuntimeCache[queue.UID] = *mr.defaultPreemptMinRuntime
+	return *mr.defaultPreemptMinRuntime
+}
+
+// getReclaimMinRuntime resolves min-runtime for reclaims
+// Depending on the resolveMethod, it will use either:
+// 1. queue: Starting from the leaf-queue, walk the tree (similar to preempt)
+// 2. lca: Use the lowest common ancestor approach
+func (mr *minruntimePlugin) getReclaimMinRuntime(
```
bugra-gedik-nv: nit: Would be nice to have the resolver as a separate class, that keeps its own caches, so the minruntime plugin code is a bit cleaner.

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -58,7 +59,7 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 	pp.calculateResourcesProportion(ssn)
 	pp.taskOrderFunc = ssn.TaskOrderFn
 	pp.reclaimablePlugin = rec.New(ssn.IsInferencePreemptible(), pp.taskOrderFunc)
-
+	pp.isInferencePreemptible = ssn.IsInferencePreemptible()
 	capacityPolicy := cp.New(pp.queues, ssn.IsInferencePreemptible())
```
bugra-gedik-nv, line 65: may be use `pp.isInferencePreemptible` in this line as well?

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -0,0 +1,239 @@
...
+	log.InfraLogger.V(3).Infof("preemptFilterFn: pendingJob: %s, victim: %s, protected: %t", pendingJob.Name, victim.Name, protected)
+	if victim.MinAvailable < int32(len(victim.PodInfos)) {
+		return true
+	}
+	return !protected
+}
+
+func (mr *minruntimePlugin) reclaimScenarioValidatorFn(reclaimer *podgroup_info.PodGroupInfo, victims []*podgroup_info.PodGroupInfo, tasks []*pod_info.PodInfo) bool {
+	for _, victim := range victims {
+		protected := mr.isReclaimMinRuntimeProtected(reclaimer, victim)
```
bugra-gedik-nv: Can we first check if this is elastic job, if not, we don't need to do anything further

bugra-gedik-nv: Otherwise, we don't even need the filter

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -0,0 +1,239 @@
...
+		if protected {
+			numVictimTasks := countVictimTasks(victim, tasks)
+			currentlyRunning := victim.GetActivelyRunningTasksCount()
```
bugra-gedik-nv: I thought`victim` could be a partial job (based on scenario generator code), no? 

Also, why do we need to do `countVictimTasks` (given victims are partial jobs that contain the victim tasks only)?

bugra-gedik-nv: Perhaps what we need is to go over all tasks once, create a job -> task count mapping and make sure for all jobs (currently running - victim task count) is >= minAvailable.

This way we make a single scan of the tasks. But the currently running should not be using partial job info.

bugra-gedik-nv: How about we look at only elastic jobs, and the group by elastic job and add the task counts in the jobs (not iterate over victim tasks).

pkg/scheduler/plugins/minruntime/minruntime.go
ArmedGuy: yeah was committed debug output, the log will go

pkg/scheduler/plugins/minruntime/minruntime.go
ArmedGuy: Have been using metav1 because the values tend to come from Queue which uses metav1.Duration, but internally in the plugin it matters a bit less.

pkg/scheduler/plugins/proportion/proportion.go
ArmedGuy: keeping as is for now, this will be removed completely with [#170](https://github.com/NVIDIA/KAI-Scheduler/issues/170)

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -113,36 +102,51 @@ func (mr *minruntimePlugin) reclaimFilterFn(pendingJob *podgroup_info.PodGroupIn
...
 
-func (mr *minruntimePlugin) reclaimScenarioValidatorFn(reclaimer *podgroup_info.PodGroupInfo, victims []*podgroup_info.PodGroupInfo, tasks []*pod_info.PodInfo) bool {
-	for _, victim := range victims {
+func (mr *minruntimePlugin) reclaimScenarioValidatorFn(reclaimer *podgroup_info.PodGroupInfo, victimRepresentatives []*podgroup_info.PodGroupInfo, tasks []*pod_info.PodInfo) bool {
+	for _, victimRepresentative := range victimRepresentatives {
+		victim, found := mr.podGroupInfos[victimRepresentative.UID]
+		if !found {
+			continue
+		}
+		if victim.IsElastic() {
```
bugra-gedik-nv: Did you mean `!victim.IsElastic()`?

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -113,36 +102,51 @@ func (mr *minruntimePlugin) reclaimFilterFn(pendingJob *podgroup_info.PodGroupIn
...
-		if protected {
-			numVictimTasks := countVictimTasks(victim, tasks)
-			currentlyRunning := victim.GetActivelyRunningTasksCount()
-			if victim.MinAvailable > currentlyRunning-numVictimTasks {
-				return false
-			}
+		if !protected {
+			continue
+		}
+		numVictimTasks := int32(len(victim.PodInfos))
```
bugra-gedik-nv: Hmm, not sure about this. What we need is the number of tasks in `victimRepresentative`. Actually, since there can be multiple victimRepresentatives for the same `victim`, we need a first pass to collect victim -> victimTaskCount.

bugra-gedik-nv:
```
victim2VictimTaskCount := make(map[common_info.JobID]int)
for _, victimRepresentative := range victimRepresentatives {
    victim, found := mr.podGroupInfos[victimRepresentative.UID]
    if found and victim.IsElastic() {
        victim2VictimTaskCount[victim.UID] += len(victimRepresentative.PodInfos)
    } 
}
for victimId, victimTaskCount := range victim2VictimTaskCount {
    victim, _ := mr.podGroupInfos[victimId]
    currentlyRunning := victim.GetActivelyRunningTasksCount()
    if victim.MinAvailable > currentlyRunning-victimTaskCount {
       return false
    }
}
return true
```

pkg/scheduler/plugins/minruntime/minruntime.go
ArmedGuy: ah yeah, pushed this too early. had fixed the other one already.

pkg/scheduler/plugins/minruntime/minruntime.go
ArmedGuy: yeah forgot to change this back to victimRepresentative when I switched victim to be full job, not rep.

But why would there be multiple victimRepresentatives? Afaik there will be only one?
Even though this code is deceptive, it seems to deduplicate so there is only 1 victim job if im reading it correctly:
https://github.com/NVIDIA/KAI-Scheduler/blob/6a1fc6571af1dcacf9f6baf08bf87717c4fafa56/pkg/scheduler/actions/common/solvers/by_pod_solver.go#L258-L282

But also the victimRepresentative's tasks does not seem to be reflective of the actual scenario either because its just the first representative for the first pod that has matched for a job (from my understanding of the above function), so thats why we need the tasks list separately. I will look at changing the list of tasks to a map[PodGroupInfo][]PodInfo so we get what we need.

bugra-gedik-nv: My bad on duplicates, you are right.

ArmedGuy:
Pushed the old getVictimCount thats inefficient for now, will await @itsomri s suggestions around changing the API or otherwise improving the way we get the correct tasks that would be affected. This API and especially jobRepresentatives are very confusing.

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -100,30 +97,19 @@ func (ra *reclaimAction) attemptToReclaimForSpecificJob(
...
-
-func buildReclaimerInfo(ssn *framework.Session, reclaimerJob *podgroup_info.PodGroupInfo) *reclaimer_info.ReclaimerInfo {
-	return &reclaimer_info.ReclaimerInfo{
-		Name:          reclaimerJob.Name,
-		Namespace:     reclaimerJob.Namespace,
-		Queue:         reclaimerJob.Queue,
-		IsPreemptable: reclaimerJob.IsPreemptibleJob(ssn.IsInferencePreemptible()),
-		RequiredResources: podgroup_info.GetTasksToAllocateInitResource(
-			reclaimerJob, ssn.TaskOrderFn, false),
+		return ssn.ReclaimScenarioValidator(reclaimer, victimJobs, victimTasks)
```
romanbaron: There might be a subtle difference with this approach. The reclaimerInfo was initially constructed prior to the simulation, meaning that the allocated resources, tasks to allocate, and related calculations reflected the pre-simulation state. Unless this is handled elsewhere, there's a risk that the proportion plugin is building reclaimerInfo based on the post-simulation state, which could lead to different result.

ArmedGuy: Good point. This could be solved by taking a copy of reclaimer right here and passing that in instead right?
Though I am curious, why would we not validate a scenario based on the resources needed for that specific scenario (which supposedly would mean passing in the variable that is currently _ which has the most up to date state of what the job needs) rather than taking the resources that were needed at the initial pass of the job?

Need to read a bit more into this.

ArmedGuy: Also the PodGroupInfos are used and abused throughout the simulation with various states, which makes it hard to write code around which one is "real" or not (especially job representatives). Would it make sense to flag instances of PodGroupInfo that are not the "real" one somehow?

ArmedGuy: So, because `GetTasksToAllocateInitResource` now has a cache on podGroup, after the first call (happening in many different places) it will never diverge from the initial state. Thats not maybe the nicest way it would be handled, but it does cover this scenario. Regardless, I will break out these changes to a separate PR isntead, and rebase this one on top of it so we can review them in isolation.

ArmedGuy:
Okay, I broke out the ReclaimerInfo refactor into #172 and for now set this branch to merge into that to get the effect of "stacked" PRs. This way we can review the changes separately.

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -0,0 +1,230 @@
...
+			continue
+		}
+		if !victim.IsElastic() {
+			continue
+		}
+		protected := mr.isPreemptMinRuntimeProtected(preemptor, victim)
+		if !protected {
+			continue
+		}
+		numVictimTasks := getVictimCount(victimRepresentative, tasks)
```
bugra-gedik-nv: Can we just collect the representatives in the loop? And if there are any, we can then scan the tasks once to collect task counts for each job representative. This way, we'll do single scan of tasks.

test/e2e/modules/wait/watcher/watcher.go
```diff
@@ -74,7 +74,18 @@ func ForEventCustomTimeout(ctx context.Context, client runtimeClient.WithWatch,
...
+				now := time.Now()
+				if satisfiedSince == nil {
+					satisfiedSince = &now
+				} else {
+					if now.Sub(*satisfiedSince) > steadyStateTime {
+						return true
+					}
+				}
+			} else {
+				if satisfiedSince != nil {
```
bugra-gedik-nv: Wouldn't this be a flip? At this point it is better to fail.

test/e2e/modules/wait/watcher/watcher.go
```diff
@@ -89,7 +100,18 @@ func ForEventCustomTimeout(ctx context.Context, client runtimeClient.WithWatch,
...
```
bugra-gedik-nv: same

test/e2e/modules/wait/pod.go
```diff
@@ -130,6 +131,22 @@ func ForPodsWithCondition(
...
+	duration time.Duration, listOptions ...runtimeClient.ListOption) {
+	condition := func(event watch.Event) bool {
+		_, ok := event.Object.(*v1.PodList)
+		if !ok {
+			return false
+		}
+		return checkCondition(event)
+	}
+	pw := watcher.NewGenericWatcher[v1.PodList](client, condition, listOptions...)
+	if !watcher.ForEventCustomTimeout(ctx, client, pw, watcher.FlowTimeout+duration, duration) {
```
bugra-gedik-nv: `watcher.FlowTimeout+duration, duration` is off IMO. Steady state duration is not necessarily == duration

ArmedGuy: I set it to that because otherwise you can easily have a timeout thats lower than the steady state duration. So I pad the default timeout with the custom steady state duration just to be safe.

bugra-gedik-nv: Didn't get it. This function, `ForPodsWithConditionFor` takes 'duration' as a parameter. I assume the goal of the function is to a) wait for some time for the condition to become true and then b) wait for some more time for the condition to continue to stay true. Isn't that the case?

If so, shouldn't it have 2 different timeouts. May be I misinterpreted what it does.

ArmedGuy: Not quite, the first duration (`watcher.FlowTimeout+duration`) is the total timeout of how long it is allowed to wait for test overall. So while it kind of serves as "wait for somet ime for the condition to become true", it doesnt stop counting after that. Because its e2e timeout of the wait, it needs to always be bigger than the second duration that is now the steady state duration, as otherwise the test will time out before steady state duration is met.

bugra-gedik-nv: Ok, got it, in that case perhaps we can do some naming improvement to`ForPodsWithConditionFor`. `duration` is not the `timeoutDuration`, it is the `steadyStateDuration` in this case. Perhaps there could be a variation of this function that takes both `steadyStateDuration` and `timeoutDuration`.

Also, `watcher.FlowTimeout+duration` is also a bit interesting. Perhaps ensuring `steadyStateDuration < timeoutDuration` is a better approach, whether `timeoutDuration` is specified or is the default of `watcher.FlowTimeout`.

test/e2e/modules/wait/watcher/watcher.go
ArmedGuy: It would be a flip, but apparently its needed for a bunch of places during teardown (this code is also used after tests). Will keep it for now.

.github/workflows/on-pr.yaml
```diff
@@ -150,4 +150,4 @@ jobs:
 
       - name: Run e2e tests
         run: |
-          ginkgo -r --keep-going --randomize-all --randomize-suites --trace -vv ./test/e2e/suites --label-filter '!autoscale', '!scale'
+          ginkgo -r --timeout 3h --keep-going --randomize-all --randomize-suites --trace -vv ./test/e2e/suites --label-filter '!autoscale', '!scale'
```
bugra-gedik-nv: why?

ArmedGuy: With default steady-state duration of just 5 seconds, the default global timeout for ginkgo is just 1h which we blow through. This were the TIMEDOUT issues we saw with e2e tests before.

itsomri: Then we need to rethink this change - this is too long and will hurt development velocity. Maybe the time-dependent tests will run selectively nightly or something.

test/e2e/suites/reclaim/reclaim_test.go
```diff
@@ -243,6 +243,37 @@ var _ = Describe("Reclaim", Ordered, func() {
...
+			testCtx.InitQueues([]*v2.Queue{parentQueue, reclaimeeQueue, reclaimerQueue})
+
+			pod := createPod(ctx, testCtx, reclaimeeQueue, 1)
+			wait.ForPodScheduled(ctx, testCtx.ControllerClient, pod)
+
+			reclaimee := createPod(ctx, testCtx, reclaimeeQueue, 3)
+			wait.ForPodScheduled(ctx, testCtx.ControllerClient, reclaimee)
+
+			reclaimer := createPod(ctx, testCtx, reclaimerQueue, 1)
+			wait.ForPodsWithConditionSteadyState(ctx, testCtx.ControllerClient, func(event watch.Event) bool {
```
itsomri: I think we don't actually need the "SteadyState".
You can use `ForPodUnschedulable`. This one will only evaluate as true if the pod was *unschedulable* - meaning the scheduler tried a full scheduling cycle (including reclaim & preempt) and wasn't able to schedule it. It's what we use in other tests to assert that a reclaim/preempt failed. It's also more reliable, as it will only pass if the scheduler actually tried to schedule the pod - while in the steady state implementation it can give a false positive if the scheduling cycle was delayed by a minute for some reason (unlikely in this specific test, but could happen in large-scale envs)

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -0,0 +1,212 @@
...
+
+func New(arguments map[string]string) framework.Plugin {
+	plugin := &minruntimePlugin{}
+
+	plugin.defaultReclaimMinRuntime = parseMinRuntime(arguments, defaultReclaimMinRuntimeConfig)
+	plugin.defaultPreemptMinRuntime = parseMinRuntime(arguments, defaultPreemptMinRuntimeConfig)
+
+	validResolveMethods := []string{resolveMethodLCA, resolveMethodQueue}
+	plugin.reclaimResolveMethod = arguments[reclaimResolveMethod]
+	if len(plugin.reclaimResolveMethod) == 0 || !slices.Contains(validResolveMethods, plugin.reclaimResolveMethod) {
```
itsomri, line 57: In cases like this I prefer to log explicitly that the config was missing/bad and the system is defaulting to something: imagine if a user tried to configure "queue" but had a typo - it's better to be noisy and let them know you are using LCA - otherwise it can create some frustrating situations when debugging

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -0,0 +1,212 @@
...
+func (mr *minruntimePlugin) Name() string {
+	return "minruntime"
+}
+
+func (mr *minruntimePlugin) OnSessionOpen(ssn *framework.Session) {
+	ssn.AddReclaimVictimFilterFn(mr.reclaimFilterFn)
+	ssn.AddPreemptVictimFilterFn(mr.preemptFilterFn)
+	ssn.AddReclaimScenarioValidatorFn(mr.reclaimScenarioValidatorFn)
+	ssn.AddPreemptScenarioValidatorFn(mr.preemptScenarioValidatorFn)
+	// store session data used for lookups
```
itsomri: IMO these comments are not needed, but it's a matter of taste so up to you

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -0,0 +1,212 @@
...
+	mr.queues = nil
+	mr.preemptProtectionCache = nil
+	mr.reclaimProtectionCache = nil
+	mr.resolver = nil
+}
+
+func (mr *minruntimePlugin) reclaimFilterFn(pendingJob *podgroup_info.PodGroupInfo, victim *podgroup_info.PodGroupInfo) bool {
+	protected := mr.isReclaimMinRuntimeProtected(pendingJob, victim)
+	// always return true for elastic jobs, but cache the result for scenario validator
+	if victim.IsElastic() {
```
itsomri, line 92: This is an easier evaluation, so maybe do it first?

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+	}
+	return true
+}
+
+func (mr *minruntimePlugin) isReclaimMinRuntimeProtected(pendingJob *podgroup_info.PodGroupInfo, victim *podgroup_info.PodGroupInfo) bool {
+	if cached, ok := mr.reclaimProtectionCache[pendingJob.UID][victim.UID]; ok {
+		return cached
+	}
+	// Get the appropriate queue objects for both jobs
```
itsomri: Unnecessary comment IMO - up to you

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -0,0 +1,212 @@
...
+	pendingQueue := mr.queues[pendingJob.Queue]
+	victimQueue := mr.queues[victim.Queue]
+
+	minRuntime, err := mr.resolver.getReclaimMinRuntime(mr.reclaimResolveMethod, pendingQueue, victimQueue)
```
itsomri, line 151: We should probably give some more context here - for example, reclaimer and reclaimee. Imagine if something goes wrong here due to some corruption in the session data (unlikely, admittedly) - you could default to a default min runtime, which could be 0 - meaning that the victim has no protection at all - this could have serious implications!
It's a very unlikely case, but let's be expressive as possible if it does happen. Maybe even resolve not allow the reclaim? (in cases of uncertainty, maybe it's better to "do no harm"?)

pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -0,0 +1,212 @@
...
+	resolveMethodQueue             = "queue"
+)
+
+type minruntimePlugin struct {
+	defaultReclaimMinRuntime metav1.Duration
+	defaultPreemptMinRuntime metav1.Duration
+	reclaimResolveMethod     string
+	queues                   map[common_info.QueueID]*queue_info.QueueInfo
+
+	preemptProtectionCache map[common_info.PodGroupID]map[common_info.PodGroupID]bool
```
itsomri: Don't we only need `map[common_info.PodGroupID]bool` in this case? For preemption, we are guaranteed that the preemptor will be from the same leaf queue - so the protection time will be identical no matter the preemptor, no?

pkg/scheduler/plugins/minruntime/resolver.go
```diff
@@ -0,0 +1,255 @@
...
+	minRuntime := r.defaultPreemptMinRuntime
+	// Walk up the tree until we find a set preempt-min-runtime or reach the root
+	for currentQueue != nil {
+		if currentQueue.PreemptMinRuntime != nil {
+			minRuntime = *currentQueue.PreemptMinRuntime
+			break
+		}
+
+		// Move to parent queue if it exists
+		if currentQueue.ParentQueue != "" {
```
itsomri: Actually I think you can do something much more elegant here:
```go
currentQueue := r.queues[currentQueue.ParentQueue]
// No need to evaluate if "found": if not found, the value of currentQueue will be nil - breaking the loop
```

pkg/scheduler/plugins/minruntime/minruntime.go
ArmedGuy: The original thought was that the result of this would be cached for elastic jobs for later, but tbh its not much of a speedup here so yes these can be moved.

pkg/scheduler/plugins/minruntime/minruntime.go
ArmedGuy: yeah thats correct. will change.

pkg/scheduler/plugins/minruntime/minruntime.go
ArmedGuy: I think minruntime should be seen as a best-effort protection rather than a hard guarantee. Because its arbitrarily set there is nothing that guarantees the job actually got to do its super-important initial work (it may have been delayed, slowed down by something that pushes the critical timeline outside of min-runtime), so the protection is best-effort as long as there is no scheduler<->workload communications to indicate readyness.
Ill add some more context, but I dont think we should refuse the reclaim.

itsomri: Ok, I agree regarding the level of guarantee, so using the default value is fine. But let's add the details of the reclaimee (namespace & name of the PG) to the error log for easier debugging

.github/workflows/on-pr.yaml
```diff
@@ -150,4 +149,4 @@ jobs:
...
```
itsomri: Can we lower this back to 1h?

PR #163, opened 2025-05-22
Scheduler pod status update instead of patch

davidLif:


pkg/scheduler/cache/status_updater/concurrency.go
```diff
@@ -98,7 +98,11 @@ func (su *defaultStatusUpdater) processPayload(ctx context.Context, payload *upd
 
 	switch payload.objectType {
 	case podType:
-		su.updatePod(ctx, payload.key, updateData.patchData, updateData.subResources, updateData.object)
+		if len(updateData.subResources) == 1 && updateData.subResources[0] == podStatusSubresource {
```
enoodle, line 101: I think it is better to move this into the `updatePod` function and there choose which function to use. It will save some duplication of the logic.

pkg/scheduler/cache/status_updater/default_status_updater.go
```diff
@@ -308,6 +301,7 @@ func (su *defaultStatusUpdater) recordUnschedulablePodsEvents(job *podgroup_info
 		msg = addNodePoolPrefixIfNeeded(job, msg)
 		log.InfraLogger.V(6).Infof("setting message for task: %v, %v", taskInfo.Name, msg)
 		updatePodCondition := utils.GetMarkUnschedulableValue(job.PodGroup.Spec.MarkUnschedulable)
+		updatePodCondition = updatePodCondition && !k8s_internal.IsPodScheduled(taskInfo.Pod)
```
enoodle, line 304: This is not needed. The pod will never be scheduled here as this function is not running concurrently to the main loop.

pkg/scheduler/cache/status_updater/concurrency.go
enoodle: Also there is already a flag for doing exactly that , `updateStatus`. it is sent to to `updatePodGroup` function for example.

pkg/scheduler/cache/status_updater/default_status_updater.go
```diff
@@ -263,24 +265,15 @@ func (su *defaultStatusUpdater) updatePodCondition(pod *v1.Pod, condition *v1.Po
...
 		su.pushToUpdateQueue(
 			&updatePayload{
 				key:        su.keyForPayload(pod.Name, pod.Namespace, pod.UID) + "-Status",
 				objectType: podType,
 			},
 			&inflightUpdate{
 				object:       pod,
-				patchData:    patchData,
-				subResources: []string{"status"},
+				patchData:    nil,
```
enoodle, line 275: use the `updateStatus` field.

pkg/scheduler/cache/status_updater/concurrency.go
```diff
@@ -120,12 +124,26 @@ func (su *defaultStatusUpdater) loadInflighUpdate(payload *updatePayload) (*infl
 	return data.(*inflightUpdate), true
 }
 
-func (su *defaultStatusUpdater) updatePod(
-	ctx context.Context, key updatePayloadKey, patchData []byte, subResources []string, object runtime.Object,
+func (su *defaultStatusUpdater) updatePodStatus(
+	ctx context.Context, key updatePayloadKey, object runtime.Object,
+) {
+	pod := object.(*v1.Pod)
+	_, err := su.kubeClient.CoreV1().Pods(pod.Namespace).UpdateStatus(
```
enoodle, line 131: I think it is less optimal to use this update function because it is very likely to fail if someone else updated the pod. I understand that the goal of this change is for this function to fail in a specific case, but it seems a waste of time to send it and wait for the failure if we can know that it will already fail.
My suggestion is that instead of this change, we will delete the updateData in the statusUpdater when calling to the Bound function.

pkg/scheduler/cache/status_updater/concurrency_test.go
```diff
@@ -223,6 +223,11 @@ var _ = Describe("Status Updater Concurrency", func() {
 				wg.Done()
 				return true, nil, nil
 			})
+			kubeClient.CoreV1().(*fakecorev1.FakeCoreV1).PrependReactor("update", "pods", func(action faketesting.Action) (handled bool, ret runtime.Object, err error) {
```
enoodle, line 226: is the `patch` function above still needed?

enoodle (changes requested):
This is trying to solve a race condition between the scheduler and the binder/kubelet.
The scheduler's status update can take some time to update the status of a pod, and in that time the pod might already be scheduled and bound to a node, which means the kubelet will update the pod scheduled condition to true.

pkg/scheduler/cache/status_updater/concurrency.go
enoodle: Maybe for the best results we should do both things ? both use `UpdateStatus` and also delete the updateData to prevent most of the failing attempts.

PR #165, opened 2025-05-22
docs: topology awarness design

enoodle:


docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,287 @@
...
+There are two possibilities as to where the struct should be built and kept:
+  * Plugin - We will use a plugin that will build it in `OnSessionOpen` and will use hooks similar to DRA plugin to keep tabs on all the changes made in the cluster.
+  * ClusterInfo - This can be done with TopologyInfo structs in the ClusterInfo package, the plugin will later read this from the cache to make node order decision and will update this struct with simulated allocations.
+Because this might be resource intensive we will leave users with the option to totally disable it and only reference it in the plugin.
+
+
+
+### Topology constraint request on a job
+
+The pod-grouper will look for topology annotations on the top owner of pods and set it to fields on the PodGroup that it creates.
```
itsomri: Can we discover inter-pod affinity that references the known keys of topology and include those as well? This has the benefits of:
1. Using the "Weights" factor in preferred inter-pod affinity
2. Better compatibility with the rest of the k8s ecosystem - some workload controllers might choose to use the native k8s APIs instead of use kueue - this allows us to respect these, in cases where the topologyKey was declared by the admin

enoodle: I am not sure how to use it with the current design, it raises some questions that seemed to complicated to answer:
-  If not all of the pods in a pod group have this set, should the others also be  constrained by it?
- Shat if there is more than one group of inter pod affinity pods defined in one pod group?
- How to handle anti affinity
- What is the relation between the weight of the pod affinity and the topology tree

I think this can be a good future enhancement, but currently it will make this feature too complicated. It seems to me like they can't be used together without some compromises.

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,287 @@
...
+
+#### Annotations for Topology Constraints
+
+To support topology-aware scheduling, we'll define the following annotations that can be added to job resources (e.g., Job, BatchJob, MPIJob):
+
+1. `scheduling.kai.nvidia.com/topology-constraint-type`: Specifies whether the topology constraint is "Required" or "Preferred"
+2. `scheduling.kai.nvidia.com/topology-level`: The minimal level of hierarchy that the job will care for
```
ArmedGuy: This might be an addition to this, but I would suggest to go further to enable dictating some requirements for how the job should be placed in the topology. Jobs that want to ensure optimal placement on topologies such as NVL72 to make use of certain model parallelism needs to be able to dictate to the scheduler that its pods are placed in segments of N on various parts of the topology. This is critical to properly make use of the new topology efficiently.

For example, I would suggest if this was described as follows:
`scheduling.kai.nvidia.com/topology-placement: "rack=4,datacenter"` - place my jobs however you want inside one datacenter, but pods must be allocated in fours on the rack level. A rack can host 4, 8, 12, 16 of the pods, but not 3 on one rack and 5 on the other.
`scheduling.kai.nvidia.com/topology-placement: "rack=4,zone=16,datacenter"` - place my jobs so that pods are allocated in fours on the rack level, and 16ths on the zone level, and keep it within a datacenter.
`scheduling.kai.nvidia.com/topology-placement: "rack"` - place my jobs so that pods are all allocated within a rack.
The values must be evenly divisible by the job size to be valid.

Additionally, for required vs preferred. Im wondering if it would be of value to be able to describe a preferred placement which is more optimal for the job, and then let the scheduler fall back to a required placement if it can't.
As an example:

`scheduling.kai.nvidia.com/topology-preferred-placement: "rack"`
`scheduling.kai.nvidia.com/topology-required-placement: "zone"`

Meaning place all my pods inside the rack if you can, but inside a zone if you must.

ArmedGuy: The downside of such constraints is that you need to hold the job when it cant be satisfied, and if you dont have head of line blocking with backfill support, you may have a hard time placing it unless you can preempt/reclaim the space.

bugra-gedik-nv: @ArmedGuy Would your suggestion cover tree and block topology in a single framework? I think that's quite powerful.

ArmedGuy: Yep thats the idea. Its how I would have implemented block topology if I could, because it also creates support for the other thing I wanted for topology/tree (being able to force that a job is placed in a certain scope of a larger topology, such as one zone or datacenter in a multi-datacenter deployment).

Something like this would be especially powerful not only for the NVL72 case, but also for more advanced mixes of model parallelisms where you start doing cross-datacenter or even cross-region training and want to ensure your job has balanced placement in a very large topology. It might even allow for more expression than block because you get to select segment sizes at different levels.

enoodle: This seems like an important use case that we should address.
It will complicate the algorithm , and also requires the scheduler to make some assumptions on the pod groups:
- All the pods have similar resource requirements and roles.
- The pods will know to find their location and group after they start, or learn it from environment variables that will be set (maybe end users can use k8s downward api for that).
I think these are fair assumptions if someone want to get to use this sophisticated scheduling.

It will require using some kind of subtree matching algorithm that will be more time consuming.
This algorithm will run everytime we simulate allocation of a job that asking for this topology aware scheduling, so in the reclaim action (or if we implement some "plane ahead" action that will solve the starving of large complicated to schedule jobs) it will be called many times. maybe it is OK because if users request this kind of scheduling they will be aware of the potential scheduling delays.

BTW I want to update the annotation to remove the `nvidia.ocm` part, I hope it will not hide this conversation, so I will wait a few days with this.

ArmedGuy: @enoodle yeah the types of jobs making these requests will fall into those buckets. If we need to be dynamic, we might want to figure out a way to flag "all these pods need this placement strategy, but these pods dont need any" at some point. This could be for jobs that schedule a coordinator on cpu resources, and N pods on GPU resources.
But generally from what I have seen coordination is just done on rank 0 (potentially because of scheduler's lack of heterogenous gangs).

For the discovery part, env variables is generally how its communicated, but I am curious to know if Kubernetes might do it differently. In SLURM, one thing to signal locality is by giving all tasks under one leaf the same rank (from SLURM perspective), so when the the world is being constructed, they know they are directly connected. But also each task gets its topology path in another env variable.

And topology is indeed one of the most time consuming steps we have today, and we don't even give good guarantees about it.

bugra-gedik-nv: > The pods will know to find their location and group after they start

This is quite important indeed. Location, I assume, is the node, rack, zone, data center, etc. information to locate a pod's placement within the topology.

In the case of groupings, such as "rack=4,zone=16,datacenter", we also need the group information for each level. As in, which rack level group the pod belongs (4 GPU groups), which rack level group the pod belongs, which zone level group it belongs, etc.

enoodle: The reason that I think we will need those assumptions is that they allow us to relax a lot of things in the algorithm and know that sorting the pods or changing the groupings will not matter.

Currently in kubernetes you can send down to the pods mostly the node name, but using configmaps we can send any data that we like, and we could send all those labels.

What about the possibility that more than one group will be allocated in the same rack? if a zone has 3 racks, but I can allocate 8 pods in each, I could use only 2 racks. Is that OK to assume that they will know to split themselves into 4 groups? 

https://kubernetes.io/docs/concepts/workloads/pods/downward-api/

ArmedGuy: @enoodle yeah thats fine, what we need to tell them is which pods would belong to which segments. This is always the case with topology because otherwise the pods might arrange communications in a way that actually makes the topology hurt more than help.

The way I would pass metadata is in a few following ways:

Job with `rack=4,zone=16,datacenter`, scheduler does strict placement with desired layout:
Pod 0
```
KAI_RANK=0
KAI_TOPOLOGY_RANK=0
KAI_TOPOLOGY_PLACEMENT=rack=4,zone=16,datacenter
```
Pod 1
```
KAI_RANK=1
KAI_TOPOLOGY_RANK=0
KAI_TOPOLOGY_PLACEMENT=rack=4,zone=16,datacenter
```
Pod 4
```
KAI_RANK=4
KAI_TOPOLOGY_RANK=1
KAI_TOPOLOGY_PLACEMENT=rack=4,zone=16,datacenter
```
Pod 10
```
KAI_RANK=10
KAI_TOPOLOGY_RANK=2
KAI_TOPOLOGY_PLACEMENT=rack=4,zone=16,datacenter
```

Job with just `datacenter`, scheduler does a best effort placement within datacenter scope:
Pod 0 (on leaf switch 1)
```
KAI_RANK=0
KAI_TOPOLOGY_RANK=0
KAI_TOPOLOGY_PLACEMENT=datacenter
```
Pod 1 (on leaf switch 1)
```
KAI_RANK=1
KAI_TOPOLOGY_RANK=0
KAI_TOPOLOGY_PLACEMENT=datacenter
```
Pod 2 (on leaf switch 2)
```
KAI_RANK=2
KAI_TOPOLOGY_RANK=1
KAI_TOPOLOGY_PLACEMENT=datacenter
```
Pod 3 (on leaf switch 3)
```
KAI_RANK=3
KAI_TOPOLOGY_RANK=2
KAI_TOPOLOGY_PLACEMENT=datacenter
```

This way the consuming job will know that all pods with the same topology rank are of the same segment (be it via strict placement or best effort), and can use the placement variable to resolve where in the hierarchy each pod is. For `rack=4,zone=16,datacenter`, the job knows the topology ranks 0s, 1s, 2s and 3s are separate segments, but also that them together make up the first zone (because you sort ranks based on topology). So 0-3 is the first zone, 4-7 is the second zone etc.

enoodle: @ArmedGuy 
To be honest, I didn't fully understand your example. I am not sure what is the exact meaning of `KAI_TOPOLOGY_RANK` value.

What do you think about a vector for those ranks?
If you asked for `rack=4,zone=16,datacenter` you will get something like:
`KAI_RANK=0,1,1,0` means you are the first pod of the 2nd  rack, 2nd zone, 1st datacenter

You will also get exactly the same information if you asked only for `datacenter` as the topology placement.
It will be simpler to implement this because it won't depend on the request, only on the allocation.

bugra-gedik-nv: >  KAI_RANK=0,1,1,0 means you are the first pod of the 2nd rack, 2nd zone, 1st datacenter

@enoodle  You meant 'on the first node', right? 

We have 2 kinds of info:
- place within the topology
- group ranks of the pods (if there are grouping constraints)

For place within the topology, given that the topology can change, should we use names instead.
`TOPOLOGY_PLACEMENT={"node":"Node-C1", "rack":"Rack-A1", "zone": "Zone-A", "datacenter":"Datacenter"}`

When we have a requirement like `rack=4,zone=16,datacenter`, then there is also group rank information that we need to make the pod aware of. Which 4-pod  group within rack do you belong to? Which 16-pod group in zone you belong to?

Assume a 32 GPU job:

pods 0-3:
`TOPOLOGY_GROUP_RANKS={"rack": 0, "zone": 0}`
pods 4, 7:
`TOPOLOGY_GROUP_RANKS={"rack": 1, "zone": 0}`
pods 8, 11:
`TOPOLOGY_GROUP_RANKS={"rack": 2, "zone": 0}`
pods 12, 15:
`TOPOLOGY_GROUP_RANKS={"rack": 3, "zone": 0}`
pods 16-19:
`TOPOLOGY_GROUP_RANKS={"rack": 0, "zone": 1}`
pods 20, 23:
`TOPOLOGY_GROUP_RANKS={"rack": 1, "zone": 1}`
pods 24, 27:
`TOPOLOGY_GROUP_RANKS={"rack": 2, "zone": 1}`
pods 28, 31:
`TOPOLOGY_GROUP_RANKS={"rack": 3, "zone": 1}`

Group ranks are not about which rack or zone you are in, it is about the group you are in.

enoodle: Thanks, I now understand your example better, I will think of how to implement it here as it is not so simple but seems like important use case that we don't want to miss.

This looks more and more like a hierarchy of pod groups to me. :thinking:

ArmedGuy: @enoodle It is and it isnt, these types of jobs are always submitted as a flat gang, and its up to the scheduler to place them according to requested topology and then inform the job of how it did. The code is the same for all pods, and then it comes down to which rank (index) a pod got that determines exactly which types of communications goes to what node.

What @bugra-gedik-nv described would also work. We should see if there is already some standard for these, and ideally omit using KAI_ prefix like I mentioned first.
SLURM describes topology as an address such as datacenter1.zone4.rack3, but as mentioned that doesn't actually help the job in any way to establish appropriate communications.

For how the data is fed, its an implementation detail (but potentially standard setting), so we just mention high level the data the job needs and then research more for specifics.

I was opting for the simplest from a data perspective, because a lot of code still mostly relies on being given a list of nodes/ranks that are sorted/setup according to the topological placement.
Before sorting by switch rank existed in Slurm, iirc you had to name your nodes so that they represented the topology (node1 and node2 on first leaf switch in tree, node 3 and 4 on the next one etc) because jobs would use the list of nodes sorted by string comparison to identify which tasks in a job were closest and were best suited to have heavier cross-node communications.

Actually by giving each pod which global rank (literally index of pod based on sorted topology, what I called `KAI_RANK`), they can self-identify the rest of the topology if they are told what the resulting topology was and a sorted list of nodes/pod addresses. Group rank can be helpful, but isnt actually needed now that I think about it.

On that note, here you can read a bit about how NVL72 scheduling was solved in Slurm (and we can have a look at the code for inspiration): https://slurm.schedmd.com/SLUG24/NVIDIA-Craig_Tierney.pdf

I dont fully agree with the design because it basically throws away the existing tree and remodels the tree into blocks, which makes you lose out on other details (such as being able to describe tradeoffs and limitations between different levels, and arbitrary sizing of a level). Thats why I would want this similar scheduling but using tree topology instead.

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,287 @@
...
+- Leverage Kueue's topology CRD concepts to keep an industry standard
+   - Support hierarchical topology representations (e.g., datacenter/rack/node/device)
+- Provide flexible configuration options for topology-based placement
+- Support both Required and Preferred topology constraint request
+
+## Non-Goals
+
+- Automatically detecting topology without user input
+- Ensure backward compatibility with existing scheduling mechanisms
+- Fixing KAI-scheduler scheduling issues with Jobs that use PodAffinity assignments
```
romanbaron: Can you explain the second point?

Also, I am not sure that results based on PodAffinity is an issue, it simply can lead to a sub-optimal result and topology aware scheduling is the way to achieve a better result. I can see how in most cases these features have different goals. What do you think?

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,287 @@
...
+
+#### Approach 1: Topology Tree with resource sums to estimate allocation probability and effect node order by task
+
+Similar to Kueue, this approach involves:
+
+1. Building a tree representation of all topology hierarchies in the cluster
+  * This tree will be updated with each allocation / preemption that is done in simulations
+  * We will need to save and restore it in simulation checkpoints
+2. Summing the resources in each topology tree node
+3. Choosing the topology with the node with the most resources that fits current pod group
```
romanbaron: let's continue calling it tree node to avoid confusion.

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,287 @@
...
+    RB2 --> NB4[Node-B4: 2 GPUs]
+
+    RC1 --> NC1[Node-C1: 2 GPUs]
+    RC1 --> NC2[Node-C2: 2 GPUs]
+    RC2 --> NC3[Node-C3: 2 GPUs]
+    RC2 --> NC4[Node-C4: 2 GPUs]
+```
+
+* The distance from node NC4 to NC3 is 2, the distance between RB2 to NA1 is 5.
+* When using hard constraints (required) this will not always find a solution even if one exists in the cluster due to fragmentation of the resources between the nodes
```
romanbaron:
```suggestion
* When using hard constraints (required) this will not always find a solution even if one exists in the cluster due to fragmentation of the resources between the nodes or other scheduling constraints
```

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,287 @@
...
+
+### Topology constraint request on a job
+
+The pod-grouper will look for topology annotations on the top owner of pods and set it to fields on the PodGroup that it creates.
+
+#### Annotations for Topology Constraints
+
+To support topology-aware scheduling, we'll define the following annotations that can be added to job resources (e.g., Job, BatchJob, MPIJob):
+
+1. `scheduling.kai.nvidia.com/topology-constraint-type`: Specifies whether the topology constraint is "Required" or "Preferred"
```
romanbaron: maybe `kai.scheduler/` so it will be consistent with other annotation prefixes?

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,287 @@
...
+6. Performance evaluation and optimization
+7. User experience improvements and documentation
+
+## Alternatives Considered
+
+- Using existing Kubernetes topology mechanisms like topology spread constraints, pod affinity
+
+These alternatives were considered but rejected due to breaking away from what end users already know and expect them to do and the need to create something more limiting to allow the algorithms to be more simple.
+
+## Future Work
```
romanbaron: Since these are not concrete, I wonder why mentioning them at all?

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,287 @@
...
+    RC1 --> NC2[Node-C2: 2 GPUs]
+    RC2 --> NC3[Node-C3: 2 GPUs]
+    RC2 --> NC4[Node-C4: 2 GPUs]
+```
+
+* The distance from node NC4 to NC3 is 2, the distance between RB2 to NA1 is 5.
+* When using hard constraints (required) this will not always find a solution even if one exists in the cluster due to fragmentation of the resources between the nodes
+* It will allow for soft/preferred constraints to still ensure it schedules pods "close" to the rest of the group even if the chosen topology can't accommodate the whole job.
+
+#### Approach 2: Simulation-based Evaluation
```
romanbaron: It sounds more like a different phase in this project, and not an alternative approach - right? If so, let's call it phase

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,287 @@
...
+
+    // TopologyLevel defines the minimal level in the topology hierarchy
+    // that this constraint applies to (e.g., "rack", "zone", "datacenter").
+    // Jobs will be scheduled to maintain locality at this level when possible.
+    TopologyLevel string `json:"topologyLevel"`
+
+    // Topology specifies the name of the topology CRD that defines the 
+    // physical layout to use for this constraint. This allows for supporting
+    // multiple different topology configurations in the same cluster.
+    Topology string `json:"topology,omitempty"`
```
romanbaron: I wonder if it should be defined at PodGroup level or at Nodepool/Scheduling shard level. I assume that single nodepool will not have multiple types of topologies.

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,287 @@
...
+2. **Scoring Mechanism**: Add topology-based scoring to node ranking
+3. **API Extensions**: Provide APIs for specifying topology preferences in workloads
+
+### User Experience
+
+Users will be able to:
+
+1. Define topology constraints in their workload specifications
+2. Set preferences for topology-aware placement
+3. Configure the relative importance of topology vs. other scheduling factors
```
romanbaron: How this configuration is applied?

docs/developer/designs/topology-awareness/README.md
enoodle: I meant here that jobs that use JobAffinity are not meant to be supported with this solution.
There is an issue with KAI that when distributed jobs use PodAffinity many possible solutions will not be checked and the scheduler might not allocate the job even though there are available resources in the cluster.
I created an example E2E test that replicates it in this branch: `erez/pod-afinity-issue` https://github.com/NVIDIA/KAI-Scheduler/tree/erez/pod-affinity-issue

From my point of view PodAffinity is usually used for similar goal to topology aware scheduling. The issue with supporting PodAffinity is that this is very expressive, and can have many different conditions. One pod group could have several sub groups of affinity and anti-affinity and those groups might also be somehow connected between them. It will be very difficult to support all of it.

docs/developer/designs/topology-awareness/README.md
enoodle: :+1: I mentioned somewhere that I want to fix it, but not yet because I don't want the change to hide other conversations.

docs/developer/designs/topology-awareness/README.md
enoodle: Good question. This was something I thought about but didn't go through with, I will remove it.
My thought was that if the topology aware scheduling is set to "preferred" then it will be only a factor in ordering the nodes, and maybe it can be less important than other plugins, but now I think that if topology aware scheduling is requested it won't make sense to not to make it the most important node ordering factor.

docs/developer/designs/topology-awareness/README.md
enoodle: The goal was to allow more than one topology. I don't see why this should be restricted.

docs/developer/designs/topology-awareness/README.md
enoodle: :+1:

docs/developer/designs/topology-awareness/README.md
enoodle: You are right, it is now written like it is a different phase, but I didn't mean it this way.
I am planning to re-write a lot of this to be more clear and accommodate new requirements that came up, so I will update the description here as well. 
I mean that the first approach will use the plugin hooks for node filtering and ordering to guide the allocation to the chosen solution and this approach was more to implement it in the action itself and to go more in depth to check the viability of each topology using all the other predicates.

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,306 @@
...
+The pods then need to get some hints to their respective location in this internal job topology.
+
+## Goals
+
+- Implement topology-aware scheduling in KAI-Scheduler
+- Leverage Kueue's topology CRD concepts to keep an industry standard
+   - Support hierarchical topology representations (e.g., datacenter/rack/node/device)
+- Provide flexible configuration options for topology-based placement
+- Support both Required and Preferred topology constraint request
+- Checking more possible allocations than is possible for PodAffinity pod groups: When the scheduler tries to scheudler a pod group with multiple pods that have a pod affinity requirement it finds a suitable node for the  first pods and that node's label is then used to compare the pod affinity of the next pods in the pod group. If the rest of the pods can't be allocated with that node affinity then the first pod's allocation is not attempted on another node. We aim to fix this issue
```
romanbaron:
```suggestion
- Checking more possible allocations than is possible for PodAffinity pod groups: When the scheduler tries to schedule a pod group with multiple pods that have a pod affinity requirement it finds a suitable node for the  first pods and that node's label is then used to compare the pod affinity of the next pods in the pod group. If the rest of the pods can't be allocated with that node affinity then the first pod's allocation is not attempted on another node. We aim to fix this issue
```

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,306 @@
...
+
+## Non-Goals
+
+- Automatically detecting topology without user input
+- Ensure backward compatibility with existing scheduling mechanisms
```
romanbaron: I still don't understand the second point, now that you specifically mentioned that TAS is going to be a better alternative for PadAffinity - what existing scheduling mechanisms are not going to supported with TAS?

docs/developer/designs/topology-awareness/README.md
romanbaron: Got it, if we are not going to implement both approaches -- let's decide on one and remove the other.

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,306 @@
...
+
+    // Total allocated resources in this domain
+    AllocatedResources v1.ResourceList
+
+    // Depth in the tree from root (0 for root)
+    Depth int
+}
+```
+
+There are two possibilities as to where the struct should be built and kept:
```
romanbaron: let's choose one.

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,306 @@
...
+  * Plugin - We will use a plugin that will build it in `OnSessionOpen` and will use hooks similar to DRA plugin to keep tabs on all the changes made in the cluster.
+  * ClusterInfo - This can be done with TopologyInfo structs in the ClusterInfo package, the plugin will later read this from the cache to make node order decision and will update this struct with simulated allocations.
+Because this might be resource intensive we will leave users with the option to totally disable it and only reference it in the plugin.
+
+#### Topology Awareness Plugin
+
+##### Setup
+Maintain topology tree with available resources using the Allocate/Deallocate event handlers (similar to the dynamicresources plugin)
```
romanbaron: Since every workload can have a different topology - do you plan to maintain and update all of them here?

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,306 @@
...
+
+##### Pre Process
+Register the PrePredicate function (Possibly change its name to PreFilter ? ) to create a job-specific and look for allocation in that tree.
+
+####  Node Ordering and Filtering
```
romanbaron, line 275: Do you think we can leverage FeasibleNodes and attempt multiple scenarios (each scenario with a different set of nodes) from the Pre-Process step?

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,306 @@
...
+#### Pod Grouper Implementation
+
+The pod-grouper component will be enhanced to:
+
+1. Look for topology-related annotations on the top owner of pods
+2. Extract and validate topology constraint information
+3. Set the corresponding fields in the PodGroup that it creates
+4. Support both Required and Preferred constraints
+
+For pods that are part of the same job but have different topology requirements, the pod-grouper will use the most restrictive constraints to ensure all pods can be scheduled.
```
romanbaron: Since we said that the topology information is extracted from the top owner of the pod, pods should probably not be part of it.

docs/developer/designs/topology-awareness/README.md
enoodle: We will no support it through d affinity, only through the new mechanics. It is not designed to work with pod affinity

romanbaron: Can we say exactly that so it will not sound as if there is no BC here?

docs/developer/designs/topology-awareness/README.md
enoodle: Here a tree of te cluster topology is built, without relation to any specific job.
It is later adapted to the jobs

docs/developer/designs/topology-awareness/README.md
enoodle: I think that we could use this to implement feasible nodes faster, but not the other way around.

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,306 @@
...
+
+1. Define topology constraints in their workload specifications
+2. Define the way a job can be split into several topologies
+3. Set preferences for topology-aware placement - strict (Only schedule the job if it can get the topology requirements) or soft (Do your best, but I prefer to start running than to wait for the perfect setup)
+4. View topology information for their running workloads
+
+### Topology Representation
+
+Following Kueue's approach, topology will be represented as a hierarchical structure that describes the physical layout of resources in the cluster.
+We will be using Kueue Topology CRD as is or import its definition into KAI scheduler to simplify integrations and support it as a standard.
```
bugra-gedik-nv, line 56: Who would populate this CRD? Is Topography service supporting CR population or is the idea to use something else to populate it?

enoodle: The user is responsible to create the instance of this  CRD with the knowledge of the cluster topology. This is probably something that doesn't change often after the cluster is brought up.

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,312 @@
+# Topology Aware Scheduling
+
+## Summary
+
+This design document outlines an approach for implementing topology aware scheduling in KAI-Scheduler, leveraging the topology CRD concepts from Kueue. Topology awareness enables more efficient scheduling of workloads by considering the physical placement of resources within a cluster, leading to improved performance for applications that benefit from resource locality.
```
itsomri, line 5: We should provide a reference to kueue's design, either here or in the "Goals" section

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,312 @@
...
+The pods then need to get some hints to their respective location in this internal job topology.
+
+## Goals
+
+- Implement topology-aware scheduling in KAI-Scheduler
+- Leverage Kueue's topology CRD concepts to keep an industry standard
+   - Support hierarchical topology representations (e.g., datacenter/rack/node/device)
+- Provide flexible configuration options for topology-based placement
+- Support both Required and Preferred topology constraint request
+- Checking more possible allocations than is possible for PodAffinity pod groups: When the scheduler tries to schedule a pod group with multiple pods that have a pod affinity requirement it finds a suitable node for the  first pods and that node's label is then used to compare the pod affinity of the next pods in the pod group. If the rest of the pods can't be allocated with that node affinity then the first pod's allocation is not attempted on another node. We aim to fix this issue
```
itsomri, line 35: This one is a bit too wordy IMO - it suffices to say that the current inter pod affinity is not podgroup-aware, and that it's impractical to make it so - the kueue approach simplifies things by narrowing down the labels to a few topology levels
(Only my opinion, we can keep it like this if you prefer)

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,312 @@
...
+
+## Design Details
+
+### User Experience
+
+Users will be able to:
+
+1. Define topology constraints in their workload specifications
+2. Define the way a job can be split into several topologies
+3. Set preferences for topology-aware placement - strict (Only schedule the job if it can get the topology requirements) or soft (Do your best, but I prefer to start running than to wait for the perfect setup)
```
itsomri:
```suggestion
3. Set preferences for topology-aware placement - strict (Only schedule the job if it can get the topology requirements) or preferred (Do your best, but I prefer to start running than to wait for the perfect setup)
```

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,312 @@
...
+    // physical layout to use for this constraint. This allows for supporting
+    // multiple different topology configurations in the same cluster.
+    Topology string `json:"topology,omitempty"`
+}
+
+type PodGroupSpec struct {
+    // ...existing fields...
+    
+    // TopologyConstraint defines the topology constraints for this PodGroup
+    TopologyConstraint TopologyConstraint `json:"topologyConstraints,omitempty"`
```
itsomri, line 111: Could a podgroup have several overlapping constraints for more than one topologies in the cluster?

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,312 @@
...
+- No order or rank is predefined on the pods.
+
+### Implementation Stages
+
+We will consider two main implementation approaches:
+
+#### Stage 1: Plugin for node order / filter
+The plugin will maintain an internal topology tree with resource sums to estimate allocation probability and effect node order by task.
+
+Similar to Kueue, this stage involves:
```
romanbaron: Saying it is similar to Kueue might be confusing, let's just explain the flow.

docs/developer/designs/topology-awareness/README.md
romanbaron: PodGroup points to a specific topology the scheduler needs to take into account when the PodGroup is scheduled. So it seems like there is more than one hierarchy tree. Does it mean that the plugin maintains and updates all these trees?

docs/developer/designs/topology-awareness/README.md
```diff
@@ -0,0 +1,312 @@
...
+
+1. Building a tree representation of all topology hierarchies in the cluster
+  * This tree will be updated with each allocation / preemption that is done in simulations
+  * We will need to save and restore it in simulation checkpoints
+2. Summing the resources in each topology tree node
+3. Choosing the topology with the tree node with the most resources that fits current pod group
```
romanbaron: In some cases a sum of the available resources is a good representation, they might be fragmented and might not fit the pod requirements.
Maybe you can check how many of the workload pods can be scheduled on the leaf nodes and then summarize it to understand how many of them can be scheduled in every sub-tree and use this information to decide?

docs/developer/designs/topology-awareness/README.md
enoodle: There is a reference at the end of the document to Kueue documentation.

docs/developer/designs/topology-awareness/README.md
enoodle: I had a comment that more details were needed here, so we will leave it like that :-)

docs/developer/designs/topology-awareness/README.md
enoodle: In this design - No

enoodle: Can you come up with the user story for this?

docs/developer/designs/topology-awareness/README.md
enoodle: Yes

enoodle: Yes, I will clarify that.

docs/developer/designs/topology-awareness/README.md
enoodle: :+1: I added that

docs/developer/designs/topology-awareness/README.md
itsomri: I'm thinking of a cluster with regular topology, and some other form of connectivity between the nodes, similar to NVLink (or any other implementation for accelerator connectivity), which might be cross-racks. Then, you might have some workloads which are only sensitive to classic internet connectivity (for example, CPU preprocessing), some that only care about the accelerator connectivity, and some that care about both.
I don't know that this is something we need to support right away - just keep in mind that it could be relevant in the future

itsomri (approved):

romanbaron (approved):

PR #167, opened 2025-05-22
Changed Queue order tie-breaker to be based on workload creation time…

romanbaron:
…stamp


pkg/scheduler/framework/session_plugins.go
```diff
@@ -232,11 +232,39 @@ func (ssn *Session) QueueOrderFn(lQ, rQ *queue_info.QueueInfo, lJob, rJob *podgr
...
 	}
 
-	// If no queue order funcs, order queue by CreationTimestamp first, then by UID.
-	if lQ.CreationTimestamp.Equal(&rQ.CreationTimestamp) {
-		return lQ.UID < rQ.UID
+	// Queue order not resolved, fallback to job CreationTimestamp first, then by UID.
+	lJ, rJ := ssn.getJobsForQueueOrder(lJob, rJob, lVictims, rVictims)
+	if lJ != nil && rJ != nil {
+		if !lJ.CreationTimestamp.Equal(&rJ.CreationTimestamp) {
+			lJ.CreationTimestamp.Before(&rJ.CreationTimestamp)
```
enoodle: Here you are calling `Before` but not using the return value, it happens again later.

pkg/scheduler/framework/session_plugins.go
```diff
@@ -232,11 +232,39 @@ func (ssn *Session) QueueOrderFn(lQ, rQ *queue_info.QueueInfo, lJob, rJob *podgr
...
+func (ssn *Session) getJobsForQueueOrder(
+	lJob, rJob *podgroup_info.PodGroupInfo, lVictims, rVictims []*podgroup_info.PodGroupInfo,
+) (*podgroup_info.PodGroupInfo, *podgroup_info.PodGroupInfo) {
+	if lJob != nil && rJob != nil {
+		return lJob, rJob
+	}
+	return ssn.getOldestJob(lVictims), ssn.getOldestJob(rVictims)
+}
+
+func (ssn *Session) getOldestJob(jobs []*podgroup_info.PodGroupInfo) *podgroup_info.PodGroupInfo {
```
enoodle: I think that this function is going to be called many times and it has implications on scale.

pkg/scheduler/framework/session_plugins.go
```diff
@@ -232,11 +232,39 @@ func (ssn *Session) QueueOrderFn(lQ, rQ *queue_info.QueueInfo, lJob, rJob *podgr
...
+	lJ, rJ := ssn.getJobsForQueueOrder(lJob, rJob, lVictims, rVictims)
+	if lJ != nil && rJ != nil {
+		if !lJ.CreationTimestamp.Equal(&rJ.CreationTimestamp) {
+			lJ.CreationTimestamp.Before(&rJ.CreationTimestamp)
+		}
+		return lJ.UID < rJ.UID
+	}
+
+	// Queue CreationTimestamp and UID as final tie-breaker
+	if !lQ.CreationTimestamp.Equal(&rQ.CreationTimestamp) {
```
enoodle: Is it possible to get to this line? I think that the only possible way is if one of the jobs is `nil` but then this is a weird situation to compare queues.

romanbaron: I agree, but the code allows that. Maybe some further refactoring is needed here.
When this function received only two jobs as parameters it was much cleaner.

romanbaron:
Closing, not sure it is the right change at the moment.

PR #179, opened 2025-05-25
Scheduler and PodGrouper use configurable nodepool label key

romanbaron:


cmd/podgrouper/app/options.go
```diff
@@ -28,7 +27,7 @@ func (o *Options) AddFlags(fs *flag.FlagSet) {
 	fs.StringVar(&o.MetricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	fs.StringVar(&o.ProbeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	fs.BoolVar(&o.EnableLeaderElection, "leader-elect", false, "Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
-	fs.StringVar(&o.NodePoolLabelKey, "nodepool-label-key", constants.NodePoolNameLabel, "The label key for node pools")
+	fs.StringVar(&o.NodePoolLabelKey, "nodepool-label-key", "", "The label key for node pools")
```
itsomri, line 30: Why is the default ""?

itsomri (approved):

cmd/podgrouper/app/options.go
romanbaron: The default in the scheduler service is an empty string, and I thought it would be weird to have different defaults in these two services.

PR #191, opened 2025-05-27
Refactor scenario validators

itsomri:
This PR aims to introduce a common api for a "Scenario" which will include the necessary information for plugins to efficiently filter or allow reclaim/preempt scenario.

### DIFF
--- a/pkg/scheduler/actions/common/solvers/by_pod_solver.go
+++ b/pkg/scheduler/actions/common/solvers/by_pod_solver.go
@@ -17,11 +17,7 @@ import (
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/log"
 )
 
-type SolutionValidator func(
-	pendingJob *podgroup_info.PodGroupInfo,
-	victimJobs []*podgroup_info.PodGroupInfo,
-	victimTasks []*pod_info.PodInfo,
-) bool
+type SolutionValidator func(scenario *api.ScenarioInfo) bool
 
 type simulationVictims struct {
 	preemptedVictims []*pod_info.PodInfo
--- a/pkg/scheduler/actions/common/solvers/by_pod_solver.go
+++ b/pkg/scheduler/actions/common/solvers/by_pod_solver.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/actions/common"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/actions/common/solvers/scenario"
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/common_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/node_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/pod_info"
@@ -17,7 +18,7 @@ import (
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/log"
 )
 
-type SolutionValidator func(scenario *api.ScenarioInfo) bool
+type SolutionValidator func(scenario api.ScenarioInfo) bool
 
 type simulationVictims struct {
 	preemptedVictims []*pod_info.PodInfo
@@ -175,9 +176,6 @@ func (s *byPodSolver) evictPotentialVictimsFromNode(
 func (s *byPodSolver) handleScenarioSolution(
 	scenario *scenario.ByNodeScenario, statement *framework.Statement, solutionVictims *simulationVictims,
 ) *solutionResult {
-
-	pendingJob := scenario.PendingJob()
-
 	victimsTasks := make([]*pod_info.PodInfo, len(solutionVictims.preemptedVictims))
 	for i := 0; i < len(solutionVictims.preemptedVictims); i++ {
 		victimsTasks[i] = solutionVictims.preemptedVictims[i]
@@ -188,7 +186,7 @@ func (s *byPodSolver) handleScenarioSolution(
 	actualVictimJobs := getVictimJobsFromVictimTasks(victimsTasks, scenario)
 
 	if s.solutionValidator != nil {
-		validSolution := s.solutionValidator(pendingJob, actualVictimJobs, victimsTasks)
+		validSolution := s.solutionValidator(scenario)
 		if !validSolution {
 			statement.Discard()
 			return &solutionResult{false, nil, nil, nil}
--- a/pkg/scheduler/actions/common/solvers/scenario/base_scenario.go
+++ b/pkg/scheduler/actions/common/solvers/scenario/base_scenario.go
@@ -6,20 +6,26 @@ package scenario
 import (
 	"strings"
 
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/common_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/pod_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/podgroup_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/framework"
 )
 
+var _ api.ScenarioInfo = &BaseScenario{}
+
 type BaseScenario struct {
 	session *framework.Session
 
+	preemptor             *podgroup_info.PodGroupInfo
 	pendingTasks          []*pod_info.PodInfo
-	pendingTasksAsJob     *podgroup_info.PodGroupInfo
 	potentialVictimsTasks []*pod_info.PodInfo
 	recordedVictimsJobs   []*podgroup_info.PodGroupInfo
 	recordedVictimsTasks  []*pod_info.PodInfo
+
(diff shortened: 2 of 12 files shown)


pkg/scheduler/actions/consolidation/consolidation.go
```diff
@@ -115,12 +115,10 @@ func attemptToConsolidatePreemptor(
...
 }
 
-func allPodsReallocated(_ *podgroup_info.PodGroupInfo,
-	consolidatedJobs []*podgroup_info.PodGroupInfo,
-	_ []*pod_info.PodInfo) bool {
-	for _, consolidatedJob := range consolidatedJobs {
-		for _, consolidatedTask := range consolidatedJob.PodInfos {
-			if consolidatedTask.Status == pod_status.Releasing {
+func allPodsReallocated(scenario api.ScenarioInfo) bool {
+	for _, victim := range scenario.GetVictims() {
```
bugra-gedik-nv, line 119: is there a concern around GetVictims() being computed each time? Is it worth caching it?

pkg/scheduler/api/scenario_info.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+type ScenarioInfo interface {
+	GetPreemptor() *podgroup_info.PodGroupInfo
+	GetVictims() map[common_info.PodGroupID]*VictimInfo
+}
+
+type VictimInfo struct {
+	Job               *podgroup_info.PodGroupInfo
+	RepresentativeJob *podgroup_info.PodGroupInfo
```
bugra-gedik-nv: is the representative job needed?

itsomri: For now it is - I want to get rid of this concept, but some plugins still rely on it.

pkg/scheduler/actions/consolidation/consolidation.go
itsomri: GetVictims is a simple getter - we calculate the victims on insertion

bugra-gedik-nv: Was not like that when I reviewed :), but good now.

pkg/scheduler/framework/session_plugins.go
```diff
@@ -96,13 +96,9 @@ func (ssn *Session) ReclaimVictimFilter(reclaimer *podgroup_info.PodGroupInfo, v
 	return true
 }
 
-func (ssn *Session) ReclaimScenarioValidator(
-	reclaimer *podgroup_info.PodGroupInfo,
-	reclaimees []*podgroup_info.PodGroupInfo,
-	victimsTasks []*pod_info.PodInfo,
-) bool {
+func (ssn *Session) ReclaimScenarioValidator(scenario api.ScenarioInfo) bool {
```
ArmedGuy: nit: the other calls are suffixed with Fn, if we want to do that as well?

pkg/scheduler/actions/common/solvers/scenario/by_node_scenario.go
```diff
@@ -7,24 +7,27 @@ import (
...
 	"golang.org/x/exp/slices"
 
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/common_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/pod_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/podgroup_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/framework"
 )
 
+var _ api.ScenarioInfo = &ByNodeScenario{}
```
ArmedGuy, line 17: is this needed?

pkg/scheduler/actions/common/solvers/by_pod_solver.go
```diff
@@ -17,11 +18,7 @@ import (
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/log"
 )
 
-type SolutionValidator func(
-	pendingJob *podgroup_info.PodGroupInfo,
-	victimJobs []*podgroup_info.PodGroupInfo,
-	victimTasks []*pod_info.PodInfo,
-) bool
+type SolutionValidator func(scenario api.ScenarioInfo) bool
```
ArmedGuy, line 21: Would it be better for these functions to pass a pointer than by value? All the underlying functions return pointers to the original objects anyways.

pkg/scheduler/actions/common/solvers/scenario/by_node_scenario.go
itsomri: I like using [interface guards](https://rednafi.com/go/interface_guards/), it's a matter of taste, but I like to make things explicit

pkg/scheduler/actions/common/solvers/by_pod_solver.go
itsomri: It's an interface so I think it's by reference implicitly in go

pkg/scheduler/framework/session_plugins.go
itsomri: Fixed

pkg/scheduler/actions/common/solvers/by_pod_solver.go
ArmedGuy: Golang doesnt have pass by reference as I understand it, but I see if the underlying instance is a pointer then its passed as a pointer (which scenario would be). So all good.

pkg/scheduler/actions/common/solvers/scenario/base_scenario.go
```diff
@@ -109,6 +112,18 @@ func (s *BaseScenario) appendTasksAsVictimJob(tasks []*pod_info.PodInfo) {
...
+
+	victimTasks := make([]*pod_info.PodInfo, 0)
+	victim, found := s.victims[job.UID]
+	if found {
+		victimTasks = victim.Tasks
+	}
+	victimTasks = append(victimTasks, tasks...)
+	s.victims[job.UID] = &api.VictimInfo{
+		Job:               job,
+		RepresentativeJob: originalJob,
```
ArmedGuy: Aren't these two flipped around? I would expect job to be RepresentativeJob as it contains only a subset of tasks, and and originalJob to be Job.

Attempted to use victimInfo.Job in my code and it would fail as I need the whole job to get actively running tasks.

pkg/scheduler/api/scenario_info.go
davidLif: Can you add some comments here about the difference between the two?

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -98,27 +101,73 @@ func (pp *proportionPlugin) CanReclaimResourcesFn(reclaimer *podgroup_info.PodGr
...
+		totalVictimsResources[victim.Job.Queue] = append(
+			totalVictimsResources[victim.Job.Queue],
+			totalJobResources...,
 		)
 	}
 
 	return pp.reclaimablePlugin.Reclaimable(pp.jobSimulationQueues, reclaimerInfo, totalVictimsResources)
 }
 
+func (pp *proportionPlugin) getVictimResources(victim *api.VictimInfo) []*resource_info.Resource {
```
davidLif, line 124: If we can avoid having this calculation in the proportion plugin we should - the scenario victims calculation isn't spesific for the proportion

pkg/scheduler/actions/common/solvers/scenario/base_scenario.go
davidLif: I agree. At line 38 we see that the preemptor is set to point at the `originalJob` variable

pkg/scheduler/actions/common/solvers/scenario/base_scenario.go
```diff
@@ -109,6 +112,18 @@ func (s *BaseScenario) appendTasksAsVictimJob(tasks []*pod_info.PodInfo) {
 	job := originalJob.CloneWithTasks(tasks)
 
 	s.victimsJobsTaskGroups[job.UID] = append(s.victimsJobsTaskGroups[job.UID], job)
+
+	victimTasks := make([]*pod_info.PodInfo, 0)
+	victim, found := s.victims[job.UID]
+	if found {
+		victimTasks = victim.Tasks
+	}
```
davidLif: Does these lines handle the elastic case?

pkg/scheduler/api/scenario_info.go
itsomri: Added

pkg/scheduler/plugins/proportion/proportion.go
itsomri: We will move to vector representation of the resources soon

pkg/scheduler/actions/common/solvers/scenario/base_scenario.go
itsomri: It's handled and tested

pkg/scheduler/actions/common/solvers/scenario/base_scenario.go
itsomri: You're right, good catch, this actually caught a fundamental issue with the victim representation: each victim can have more than one representative, depending on elasticity.
I changed the API to be a slice of representatives. Obviously this is very far from ideal - I plan on dumping the representative victim jobs completely, but this will take some time.

davidLif (approved):

PR #200, opened 2025-06-03
remove redundant `replicas` value for binder

slaupster:
see https://github.com/NVIDIA/KAI-Scheduler/pull/200#issuecomment-2938958845

This now simply removes the unused replicas key from the `values.yaml` as its not used and not supported anyway apparently.

Original description:

`replicas` is a binder value in the `values.yaml` but the service just uses `replicas: 1`.  I think its intentional that you can configure replicas as you might want multiple in a large cluster and binding [can be slow](https://github.com/NVIDIA/KAI-Scheduler/blob/main/docs/developer/binder.md#:~:text=API%20calls%20and-,can%20be%20slow,-%2C%20especially%20when%20handling).

This PR updates the `binder.yaml` to use the `replicas` value for `binder`.

Should be a no-op because the default value is `1` but allows configuration if needed.


romanbaron:
Hi @slaupster - thanks for suggesting this change 🙌 
Do you have an actual use case where multiple binder replicas are needed?
Mainly asking because there are several things that need to be changed in the binder service to support multiple replicas and there is `max-concurrent-reconciles` parameter that can be changed to support different number of go routines in the reconcile loop. Can it address your needs?

slaupster:
> g becaus

hey @romanbaron - no I don't is the honest answer. Am just standing this up as we speak and having a play. My first draft was simply "deleting unused value" then I read about the binder and saw that "it can be slow" and assumed that it was intentional that the binder was scalable and the bug was fixed by this PR but happy to change it to simply removing the value if you like ?

slaupster:
> > g becaus
> 
> hey @romanbaron - no I don't is the honest answer. Am just standing this up as we speak and having a play. My first draft was simply "deleting unused value" then I read about the binder and saw that "it can be slow" and assumed that it was intentional that the binder was scalable and the bug was fixed by this PR but happy to change it to simply removing the value if you like ?

I've update the PR

Issue #47, opened 2025-04-04
Cannot create pods with fractional GPU. OCI runtime create failed: could not apply required modification to OCI specification


zhiming-shen:
When trying to create a pod with fractional GPU, the pod enters `StartError` status. The error message is as following:

```
Error: failed to create containerd task: failed to create shim task: OCI runtime create failed: could not apply required modification to OCI specification: error modifying OCI spec: failed to inject CDI devices: unresolvable CDI devices management.nvidia.com/gpu=GPU-213ef228-8121-3a14-4880-6fcf31e97653: unknown
```

Pod yaml file:
```
apiVersion: v1
kind: Pod
metadata:
  name: kai-gpu-test
  labels:
    runai/queue: test
  annotations:
    gpu-fraction: "0.5"
spec:
  schedulerName: kai-scheduler
  restartPolicy: Never
  tolerations:
  - key: "nvidia.com/gpu"
    operator: "Equal"
    value: "present"
    effect: "NoSchedule"
  containers:
  - name: nvidia-container
    image: nvidia/cuda:12.2.0-base-ubuntu22.04  # /usr/local/nvidia/bin
    command: ["sleep", "infinity"]
```

Kubernetes: GKE 1.32.2-gke.1182001
NVIDIA GPU Operator: v25.3.0
KAI installation command: `helm upgrade -i kai-scheduler nvidia-k8s/kai-scheduler -n kai-scheduler --create-namespace --set "global.registry=nvcr.io/nvidia/k8s" --set "global.gpuSharing=true"`

I can see that the reservation pod was started correctly. But the real testing pod failed to start. 

```
# kubectl describe pod kai-gpu-test
Name:             kai-gpu-test
Namespace:        default
Priority:         0
Service Account:  default
Node:             gke-kai-test-gpu-pool-8ada50fa-mh1v/10.128.15.235
Start Time:       Fri, 04 Apr 2025 23:01:31 +0000
Labels:           runai-gpu-group=2992ada3-d208-43fd-aec8-9f43822f4aaf
                  runai/queue=test
Annotations:      gpu-fraction: 0.5
                  pod-group-name: pg-kai-gpu-test-a65737cf-41a3-4523-8f23-ae8c64601446
                  received-resource-type: Fraction
                  runai-job-id: a65737cf-41a3-4523-8f23-ae8c64601446
                  runai-node: gke-kai-test-gpu-pool-8ada50fa-mh1v
[...]

PR #205, opened 2025-06-05
cdi flag added

christophemacabiau:
Related to #47 
I didn't find how to enable cdi when installing the chart, and it's required to use kai in gke.

### DIFF
--- a/deployments/kai-scheduler/values.yaml
+++ b/deployments/kai-scheduler/values.yaml
@@ -6,6 +6,7 @@ global:
   securityContext: {}
   imagePullSecrets: []
   gpuSharing: false
+  cdi: false
   clusterAutoscaling: false
   nodeSelector: {}
   affinity: {}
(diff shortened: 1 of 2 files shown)


deployments/kai-scheduler/values.yaml
```diff
@@ -6,6 +6,7 @@ global:
   securityContext: {}
   imagePullSecrets: []
   gpuSharing: false
+  cdi: false
```
romanbaron, line 9: I'm considering whether this should reside in the global section or be scoped directly within binder. The rationale for placing it in global is that it's an environment-level parameter that could potentially impact multiple services. However, at present, it only affects binder, which may justify keeping it localized there. What are your thoughts?

romanbaron:
> Related to #47 I didn't find how to enable cdi when installing the chart, and it's required to use kai in gke.

Thanks for suggesting this change 🙏

deployments/kai-scheduler/values.yaml
christophemacabiau: I agree!

christophemacabiau:
Sorry, git branch error, I opened a new PR: #209

PR #208, opened 2025-06-05
bug: Elastic job cannot be reclaimed by 2 jobs

ArmedGuy:
Below is an e2e test similar to what I have for minruntime that used to be working, but after I rebased onto `omric/refactor-scenario-validators` the second reclaimer never schedules. Here is a shortened e2e test without the min runtime parts on main that also seems to fail.

From my understanding, this test should pass? (Since it did before).

I have no clue what might be wrong yet, but will have a look at if something like https://github.com/NVIDIA/KAI-Scheduler/commit/bac4866ab5bba39cc76e900ced0c1da31339a4a1 might be a cause.


ArmedGuy:
Seems like there is a unit test failing as well. Since it didnt fail on other PRs it might be flaky?

ArmedGuy:
Okay nvm the e2e test is actually wrong, it should be `createQueues(3, 0, 3)` for this scenario. How this might have passed before is puzzling.

PR #210, opened 2025-06-06
Add plugin for Grove PodGangs

sanjaychatterjee:
This PR creates a PodGrouper plugin for Grove PodGangs. The Grove PodGang API can be found here:
https://github.com/NVIDIA/grove/blob/main/scheduler/api/core/v1alpha1/podgang.go

The plugin builds on top of defaultgrouper and then sets values for priorityClassName and minAvailable in the KAI podgroup. Currently, the KAI PodGroup is a represents a logically flattened Grove PodGang structure. Hence, the value for minAvailable in the KAI PodGroup is the sum of all the minReplicas from the slice of PodGroups defined in the Grove PodGang.

### DIFF
--- a/pkg/podgrouper/podgrouper/hub/hub.go
+++ b/pkg/podgrouper/podgrouper/hub/hub.go
@@ -85,6 +85,8 @@ func NewPluginsHub(kubeClient client.Client, searchForLegacyPodGroups,
 	sparkGrouper := spark.NewSparkGrouper(defaultGrouper)
 	podJobGrouper := podjob.NewPodJobGrouper(defaultGrouper, sparkGrouper)
 
+	podgangGrouper := podgang.NewPodGangGrouper(defaultGrouper)
+
 	table := map[metav1.GroupVersionKind]grouper.Grouper{
 		{
 			Group:   "apps",
@@ -236,6 +238,11 @@ func NewPluginsHub(kubeClient client.Client, searchForLegacyPodGroups,
 			Version: "v1",
 			Kind:    "SPOTRequest",
 		}: spotrequest.NewSpotRequestGrouper(defaultGrouper),
+		{
+			Group:   "scheduler.grove.io",
+			Version: "v1alpha1",
+			Kind:    "PodGang",
+		}: podgangGrouper,
 	}
 
 	skipTopOwnerGrouper := skiptopowner.NewSkipTopOwnerGrouper(kubeClient, defaultGrouper, table)
--- a/pkg/podgrouper/podgrouper/plugins/podgang/podgang_grouper.go
+++ b/pkg/podgrouper/podgrouper/plugins/podgang/podgang_grouper.go
@@ -1,0 +1,72 @@
+// Copyright 2025 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+package podgang
+
+import (
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/NVIDIA/KAI-scheduler/pkg/podgrouper/podgroup"
+	"github.com/NVIDIA/KAI-scheduler/pkg/podgrouper/podgrouper/plugins/constants"
+	"github.com/NVIDIA/KAI-scheduler/pkg/podgrouper/podgrouper/plugins/defaultgrouper"
+)
+
+type PodGangGrouper struct {
+	*defaultgrouper.DefaultGrouper
+}
+
+func NewPodGangGrouper(defaultGrouper *defaultgrouper.DefaultGrouper) *PodGangGrouper {
+	return &PodGangGrouper{
+		defaultGrouper,
+	}
+}
+
+func (pgg *PodGangGrouper) Name() string {
+	return "PodGang Grouper"
+}
+
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch
+// +kubebuilder:rbac:groups=apps,resources=deployments/finalizers,verbs=patch;update;create
+
+func (pgg *PodGangGrouper) GetPodGroupMetadata(
+	topOwner *unstructured.Unstructured, pod *v1.Pod, _ ...*metav1.PartialObjectMetadata,
+) (*podgroup.Metadata, error) {
+	metadata, err := pgg.DefaultGrouper.GetPodGroupMetadata(topOwner, pod)
+	if err != nil {
+		return nil, err
+	}
+
+	priorityClassName, found, err := unstructured.NestedString(topOwner.Object, "spec", "priorityClassName")
+	if err != nil {
+		return nil, err
+	}
+	if found {
+		metadata.PriorityClassName = priorityClassName
+	}
+
+	minAvailable := 0
+	pgs, found, err := unstructured.NestedSlice(topOwner.Object, "spec", "podgroups")
+	if err != nil {
+		return nil, err
+	}
+	for _, v := range pgs {
+		pgr, ok := v.(map[string]interface{})
+		if !ok {
+			return nil, fmt.Errorf("invalid podgang structure")
+		}
+		minReplicas, found, err := unstructured.NestedInt64(pgr, "minReplicas")
+		if err != nil {
+			return nil, err
+		}
+		if found {
+			minAvailable += minReplicas
+		}
+	}
+	metadata.MinAvailable = int32(minAvailable)
+
+	return metadata, nil
+}
--- a/pkg/podgrouper/podgrouper/plugins/podgang/podgang_grouper_test.go
+++ b/pkg/podgrouper/podgrouper/plugins/podgang/podgang_grouper_test.go
@@ -1,0 +1,130 @@
+// Copyright 2025 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+package podgang
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/NVIDIA/KAI-scheduler/pkg/podgrouper/podgrouper/plugins/constants"
+	"github.com/NVIDIA/KAI-scheduler/pkg/podgrouper/podgrouper/plugins/defaultgrouper"
+)
+
+const (
+	queueLabelKey    = "kai.scheduler/queue"
+	nodePoolLabelKey = "kai.scheduler/node-pool"
+)
+
+func TestGetPodGroupMetadata(t *testing.T) {
+	podgang := &unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"kind":       "PodGang",
+			"apiVersion": "scheduler.grove.io/v1alpha1",
+			"metadata": map[string]interface{}{
+				"name":      "pgs1",
+				"namespace": "test-ns",
+				"uid":       "1",
+				"labels": map[string]interface{}{
+					"test_label": "test_value",
+				},
+				"annotations": map[string]interface{}{
+					"test_annotation": "test_value",
+				},
+			},
+			"spec": map[string]interface{}{
+				"podgroups": []map[string]interface{}{
+					{
+						"podReferences": []map[string]interface{}{
+							{
+								"namespace": "test-ns"
+								"name": "pgs1-pga1"
+							},
+							{
+								"namespace": "test-ns"
+								"name": "pgs1-pga2"
+							},
+							{
+								"namespace": "test-ns"
+								"name": "pgs1-pga3"
+							},
+							{
+								"namespace": "test-ns"
+								"name": "pgs1-pga4"
+							},
+						},
+						"minReplicas": 4,
+					},
+					{
+						"podReferences": []map[string]interface{}{
+							{
+								"namespace": "test-ns"
+								"name": "pgs1-pgb1"
+							},
+							{
+								"namespace": "test-ns"
+								"name": "pgs1-pgb2"
+							},
+							{
+								"namespace": "test-ns"
+								"name": "pgs1-pgb3"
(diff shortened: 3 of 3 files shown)


pkg/podgrouper/podgrouper/hub/hub.go
```diff
@@ -85,6 +86,8 @@ func NewPluginsHub(kubeClient client.Client, searchForLegacyPodGroups,
 	sparkGrouper := spark.NewSparkGrouper(defaultGrouper)
 	podJobGrouper := podjob.NewPodJobGrouper(defaultGrouper, sparkGrouper)
 
+	podgangGrouper := podgang.NewPodGangGrouper(defaultGrouper)
```
romanbaron:
```suggestion
	podGangGrouper := podgang.NewPodGangGrouper(defaultGrouper)
```

pkg/podgrouper/podgrouper/hub/hub.go
```diff
@@ -236,6 +239,11 @@ func NewPluginsHub(kubeClient client.Client, searchForLegacyPodGroups,
 			Version: "v1",
 			Kind:    "SPOTRequest",
 		}: spotrequest.NewSpotRequestGrouper(defaultGrouper),
+		{
+			Group:   "scheduler.grove.io",
+			Version: "v1alpha1",
+			Kind:    "PodGang",
+		}: podgangGrouper,
```
romanbaron:
```suggestion
		}: podGangGrouper,
```

pkg/podgrouper/podgrouper/plugins/podgang/podgang_grouper.go
```diff
@@ -0,0 +1,79 @@
...
+		defaultGrouper,
+	}
+}
+
+func (pgg *PodGangGrouper) Name() string {
+	return "PodGang Grouper"
+}
+
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch
+// +kubebuilder:rbac:groups=apps,resources=deployments/finalizers,verbs=patch;update;create
```
romanbaron:
```suggestion
// +kubebuilder:rbac:groups=scheduler.grove.io,resources=podgangs,verbs=get;list;watch
// +kubebuilder:rbac:groups=scheduler.grove.io,resources=podgangs/finalizers,verbs=patch;update;create
```

romanbaron: Once changed, please run `make manifests` to generate podgrouper clusterrole

pkg/podgrouper/podgrouper/plugins/podgang/podgang_grouper.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+		if !found {
+			return nil, fmt.Errorf("unexpected podgroup structure")
+		}
+		minReplicas, found, err := unstructured.NestedInt64(pgr, "minReplicas")
+		if err != nil {
+			return nil, err
+		}
+		if found && int(minReplicas) != len(podSlice) {
+			fmt.Printf("Unsupported minReplicas: expected: %v, found: %v", len(podSlice), minReplicas)
```
romanbaron: Please use logger instead of fmt. And I wonder if we want the podgroup not to be created, otherwise it might lead to an unexpected behavior during scheduling.

pkg/podgrouper/podgrouper/plugins/podgang/podgang_grouper_test.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	grouper := NewPodGangGrouper(defaultgrouper.NewDefaultGrouper(queueLabelKey, nodePoolLabelKey))
+	metadata1, err1 := grouper.GetPodGroupMetadata(podgang1, pod1)
+	assert.Nil(t, err1)
+	assert.Equal(t, int32(12), metadata1.MinAvailable)
+	assert.Equal(t, constants.InferencePriorityClass, metadata1.PriorityClassName)
+	assert.Equal(t, "test_queue", metadata1.Queue)
+
+	podgang2 := &unstructured.Unstructured{
```
romanbaron: If there is a difference between these two podGangs, can you maybe split them into separate tests?

pkg/podgrouper/podgrouper/plugins/podgang/podgang_grouper.go
sanjaychatterjee: Agreed. The only supported value for minReplicas currently is the total number of pods in the PodGroup. Hence, based on your suggestion maybe returning an error would avoid any ambiguities. Will update the code.

pkg/podgrouper/podgrouper/plugins/podgang/podgang_grouper_test.go
sanjaychatterjee: The difference is that the second PodGang does not define any minReplicas. Since, we do not want to support any such configuration, we can remove the second test.

sanjaychatterjee: I have added a negative test to check for the expected error when minReplicas don't match the expected value.

romanbaron (approved):

pkg/podgrouper/podgrouper/plugins/grove/grove_grouper.go
```diff
@@ -0,0 +1,117 @@
...
+// +kubebuilder:rbac:groups=scheduler.grove.io,resources=podgangs,verbs=get;list;watch
+// +kubebuilder:rbac:groups=scheduler.grove.io,resources=podgangs/finalizers,verbs=patch;update;create
+
+func (gg *GroveGrouper) GetPodGroupMetadata(
+	_ *unstructured.Unstructured, pod *v1.Pod, _ ...*metav1.PartialObjectMetadata,
+) (*podgroup.Metadata, error) {
+	if len(pod.Spec.SchedulingGates) > 0 {
+		return nil, fmt.Errorf("Ignoring pod %s/%s due to scheduling gates: %v",
+			pod.Namespace, pod.Name, pod.Spec.SchedulingGates)
+	}
```
romanbaron: This part is already covered in the scheduler.
https://github.com/NVIDIA/KAI-Scheduler/blob/b66691d1efd35b0ea04fd0fe5232742714b784e4/pkg/scheduler/api/pod_info/pod_info.go#L332

I think you can safely create a podgroup and it won't be scheduled as long as it has a gated pod.

pkg/podgrouper/podgrouper/hub/hub.go
```diff
@@ -236,6 +239,11 @@ func NewPluginsHub(kubeClient client.Client, searchForLegacyPodGroups,
 			Version: "v1",
 			Kind:    "SPOTRequest",
 		}: spotrequest.NewSpotRequestGrouper(defaultGrouper),
+		{
+			Group:   "grove.io",
+			Version: "v1alpha1",
+			Kind:    "PodGangSet",
+		}: groveGrouper,
```
romanbaron, line 248: Can you explain how owner references of grove pods look like? Starting from pod level up to the top owner.

sanjaychatterjee: The chain of owner references for Grove pods looks like the following: 
Pod -> [PodClique](https://github.com/NVIDIA/grove/blob/main/operator/api/core/v1alpha1/podclique.go#L33) -> [PodGangSet](https://github.com/NVIDIA/grove/blob/main/operator/api/core/v1alpha1/podgangset.go#L33).

pkg/podgrouper/podgrouper/plugins/grove/grove_grouper.go
sanjaychatterjee: Ok. Then I think it should be fine to remove this check. Will update PR. Thanks.

pkg/podgrouper/podgrouper/hub/hub.go
sanjaychatterjee: cc @unmarshall

romanbaron: In this case, I think you should probably choose one of the options below:
1. Add RBAC for PodClique so PodGropuer will be able to read it when it searches for the pod top owner (PodGangSet).
2. Stop searching at PodClique. The PodGrouper searches up the owner reference chain until it finds an object without an owner or it cannot search to a higher level due to missing RBAC, in this case the highest entity found is considered the top owner. At this point, PodGrouper chooses the matching plugin based on the owner's GVK and executes it. Since Grove plugin doesn't really access the top owner instance, you just want to check that the pod has a Grove related owner, which can be either PodGangSet or PodClique.

Both options are probably equal, and in both options you probably need to add RBAC for PodClique.

romanbaron (dismissed):

pkg/podgrouper/podgrouper/plugins/grove/grove_grouper.go
```diff
@@ -0,0 +1,117 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/NVIDIA/KAI-scheduler/pkg/podgrouper/podgroup"
+	"github.com/NVIDIA/KAI-scheduler/pkg/podgrouper/podgrouper/plugins/defaultgrouper"
+)
+
+const (
+	podGangNameKey = "grove.io/podgang"
```
unmarshall: Can you rename this field to indicate that this is a label key?

pkg/podgrouper/podgrouper/plugins/grove/grove_grouper.go
```diff
@@ -0,0 +1,117 @@
...
+// +kubebuilder:rbac:groups=grove.io,resources=podgangsets/finalizers,verbs=patch;update;create
+// +kubebuilder:rbac:groups=grove.io,resources=podcliques,verbs=get;list;watch
+// +kubebuilder:rbac:groups=grove.io,resources=podcliques/finalizers,verbs=patch;update;create
+// +kubebuilder:rbac:groups=scheduler.grove.io,resources=podgangs,verbs=get;list;watch
+// +kubebuilder:rbac:groups=scheduler.grove.io,resources=podgangs/finalizers,verbs=patch;update;create
+
+func (gg *GroveGrouper) GetPodGroupMetadata(
+	_ *unstructured.Unstructured, pod *v1.Pod, _ ...*metav1.PartialObjectMetadata,
+) (*podgroup.Metadata, error) {
+	if pod.Labels == nil {
```
unmarshall: the subsequent check `podGangName, ok := pod.Labels[podGangNameKey]` should be good enough. In golang key access on nil maps is a safe operation.

pkg/podgrouper/podgrouper/plugins/grove/grove_grouper.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	metadata, err := gg.DefaultGrouper.GetPodGroupMetadata(podGang, pod)
+	if err != nil {
+		return nil, err
+	}
+
+	priorityClassName, found, err := unstructured.NestedString(podGang.Object, "spec", "priorityClassName")
+	if err != nil {
+		return nil, err
```
unmarshall: would you want to introduce usage of sentinel errors? It will only make sense if the consumer wishes to react to certain type of errors. If not then generally it is a good practice to wrap the error using`fmt.Errorf` and provide semantic context. See if this is required in general.

pkg/podgrouper/podgrouper/plugins/grove/grove_grouper.go
sanjaychatterjee: Done

pkg/podgrouper/podgrouper/plugins/grove/grove_grouper.go
sanjaychatterjee: Done.

pkg/podgrouper/podgrouper/plugins/grove/grove_grouper.go
sanjaychatterjee: Updated.

pkg/podgrouper/podgrouper/plugins/grove/grove_grouper.go
```diff
@@ -0,0 +1,122 @@
...
+// +kubebuilder:rbac:groups=grove.io,resources=podcliques/finalizers,verbs=patch;update;create
+// +kubebuilder:rbac:groups=scheduler.grove.io,resources=podgangs,verbs=get;list;watch
+// +kubebuilder:rbac:groups=scheduler.grove.io,resources=podgangs/finalizers,verbs=patch;update;create
+
+func (gg *GroveGrouper) GetPodGroupMetadata(
+	_ *unstructured.Unstructured, pod *v1.Pod, _ ...*metav1.PartialObjectMetadata,
+) (*podgroup.Metadata, error) {
+	podGangName, ok := pod.Labels[labelKeyPodGangName]
+	if !ok {
+		return nil, fmt.Errorf("Label for podgang name (key: %s) not found in pod %s/%s",
```
romanbaron:
```suggestion
		return nil, fmt.Errorf("label for podgang name (key: %s) not found in pod %s/%s",
```

pkg/podgrouper/podgrouper/plugins/grove/grove_grouper.go
```diff
@@ -0,0 +1,122 @@
...
+			return nil, fmt.Errorf("missing podReferences in spec.podgroup[%v] of PodGang %s/%s",
+				idx, pod.Namespace, podGangName)
+		}
+		minReplicas, found, err := unstructured.NestedInt64(pgr, "minReplicas")
+		if err != nil {
+			return nil, fmt.Errorf("failed to get minReplicas from spec.podgroup[%v] of PodGang %s/%s : %w",
+				idx, pod.Namespace, podGangName, err)
+		}
+		if found && int(minReplicas) != len(podSlice) {
+			return nil, fmt.Errorf("Unsupported minReplicas in spec.podgroup[%v] of PodGang %s/%s : expected: %v, found: %v",
```
romanbaron:
```suggestion
			return nil, fmt.Errorf("unsupported minReplicas in spec.podgroup[%v] of PodGang %s/%s : expected: %v, found: %v",
```

romanbaron (approved):

romanbaron (approved):

PR #216, opened 2025-06-08
Default priority class per workload type, read from configmap

natasharomm:
podgrouper: added a feature of optionally providing default priority class name per workload (top owner) type: 

* Optionally providing a configmap name and namespace to the podgrouper (in the command line)
* When creating a podgroup, if priority class is not provided in pod.Spec or as a label, the podgrouper will read the configmap data and take the default PriorityClassName by type (if the configmap name and namespace provided)
* Of course, the PriorityClass objects need to exist as well
* The configmap structure should be like in the following example:

data:
      types: '[{"TypeName":"PyTorchJob","PriorityName":"low"},{"TypeName":"CronJob","PriorityName":"very-high"},{"TypeName":"Deployment","PriorityName":"very-high"}]'

### DIFF
--- a/pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
+++ b/pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
@@ -148,12 +148,12 @@ func (dg *DefaultGrouper) calculateQueueName(topOwner *unstructured.Unstructured
 
 func (dg *DefaultGrouper) CalcPodGroupPriorityClass(topOwner *unstructured.Unstructured, pod *v1.Pod,
 	defaultPriorityClassForJob string) string {
-	if len(pod.Spec.PriorityClassName) != 0 {
-		return pod.Spec.PriorityClassName
-	} else if priorityClassName, found := topOwner.GetLabels()[constants.PriorityLabelKey]; found {
+	if priorityClassName, found := topOwner.GetLabels()[constants.PriorityLabelKey]; found {
 		return priorityClassName
 	} else if priorityClassName, found = pod.GetLabels()[constants.PriorityLabelKey]; found {
 		return priorityClassName
+	} else if len(pod.Spec.PriorityClassName) != 0 {
+		return pod.Spec.PriorityClassName
 	} else {
 		groupKind := topOwner.GroupVersionKind().GroupKind()
 		return dg.getDefaultPriorityClassNameForKind(&groupKind, defaultPriorityClassForJob)
--- a/pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
+++ b/pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
@@ -162,8 +162,8 @@ func (dg *DefaultGrouper) CalcPodGroupPriorityClass(topOwner *unstructured.Unstr
 
 // getDefaultPriorityClassNameForKind - returns the default priority class name for a given group kind.
 func (dg *DefaultGrouper) getDefaultPriorityClassNameForKind(groupKind *schema.GroupKind, defaultPriorityClassFallback string) string {
-	if groupKind == nil {
-		logger.V(3).Info("Unable to get default priority class name: GroupKind is nil, using default priority class fallback")
+	if groupKind == nil || groupKind.String() == "" || groupKind.Kind == "" {
+		logger.V(3).Info("Unable to get default priority class name: GroupKind is empty, using default priority class fallback")
 		return defaultPriorityClassFallback
 	}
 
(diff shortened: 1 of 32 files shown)


davidLif (commented):
I think that changing/setting the priorityClass based on the configMap would be cleaner if you use the PluginsHub struct rather then making the changes inside all of the podrgouper plugin

pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
```diff
@@ -36,6 +43,14 @@ func NewDefaultGrouper(queueLabelKey, nodePoolLabelKey string) *DefaultGrouper {
 	}
 }
 
+func (dg *DefaultGrouper) SetDefaultPrioritiesConfigMapParams(
+	defaultPrioritiesConfigMapName, defaultPrioritiesConfigMapNamespace string, kubeReader client.Reader,
+) {
+	dg.defaultPrioritiesConfigMapName = defaultPrioritiesConfigMapName
+	dg.defaultPrioritiesConfigMapNamespace = defaultPrioritiesConfigMapNamespace
+	dg.kubeReader = kubeReader
```
davidLif, line 51: Why not combine this with NewDefaultGrouper ?

pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
```diff
@@ -144,6 +159,91 @@ func (dg *DefaultGrouper) CalcPodGroupPriorityClass(topOwner *unstructured.Unstr
...
+	configMap := &v1.ConfigMap{}
+	err := dg.kubeReader.Get(context.Background(), client.ObjectKey{
+		Name:      dg.defaultPrioritiesConfigMapName,
+		Namespace: dg.defaultPrioritiesConfigMapNamespace,
+	}, configMap)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get default priorities configmap: %w", err)
+	}
+
+	return parseConfigMapDataToDefaultPriorities(configMap)
```
davidLif, line 213: Maybe this should happen once at the start of the podGrouper instead of every time a podgroup is created?

pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
```diff
@@ -144,6 +159,91 @@ func (dg *DefaultGrouper) CalcPodGroupPriorityClass(topOwner *unstructured.Unstr
...
+		return nil, fmt.Errorf("default priorities configmap is empty, cannot parse default priorities")
+	}
+
+	data, ok := cm.Data[constants.DefaultPrioritiesConfigMapTypesKey]
+	if !ok {
+		return nil, fmt.Errorf("default priorities configmap Data does not contain <%s> key", constants.DefaultPrioritiesConfigMapTypesKey)
+	}
+
+	var configs []workloadTypePriorityConfig
+	err := json.Unmarshal([]byte(data), &configs)
```
davidLif, line 244: Do we need to unmarshal it into a struct? for somethig so simple, we can just json traverse, and the it won't break if anybody adds more fields to the priority config

pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
natasharomm: that's how it was before and then, as I understand, we agreed I would change it.
should I change it back?

pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
natasharomm: this configmap can change - if the user is updating the defaults per type - that's why we need to read it every time.
but with a cached client it's really no big deal.

perhaps, if the podgrouper is not reading any other configmaps, and I think it's not,
I can use cacheconfig to limit the cache for configmaps to only this 1 namespace. I will do it.

pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
natasharomm: what do you mean json traverse?
it's an array of objects, I am not sure how we can do that...
but anyway it won't break, we already have more fields actually, json unmarshal just ignores them - I tested this.

pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
davidLif: I would prefer not to change the current cache settings - it might have some other implication. You can revert this ochange and I will resolve this

davidLif (approved):

PR #224, opened 2025-06-10
Vendor neutrality migration guide

romanbaron:


docs/backwardcompatibility/v0.6.0/README.md
```diff
@@ -0,0 +1,23 @@
...
+kubectl get pods -n runai-reservation
+```
+
+## Scheduling Queue Label Key
+The label key for a scheduling queue was changed from `runai/queue` to `kai.scheduler/queue`.
+In order to adopt the new label key, all workloads must have the new label with the name of the respective queue before the upgrade.
+The following command should result without any existing pods:
+```
+kubectl get pods -A -l 'runai/queue'
+```
```
davidLif: Maybe we should add a script that creates the appropriate `kai.scheduler/queue` label on existing workloads, based on the `runai/queue` value?
It's not a perfect solution, but it might help to migrate running workloads

romanbaron: I thought about this approach, the problem is that it has to be labeled on the workload level. and if someone added this label in the pod template, it might restart the workload pods. Since there is more than one option here, I am not sure it can cover all cases and not affect the running workloads.

deployments/kai-scheduler/templates/services/binder.yaml
```diff
@@ -23,12 +23,15 @@ spec:
           image: "{{ .Values.global.registry }}/{{ .Values.binder.image.name }}:{{ .Chart.Version }}"
           imagePullPolicy: {{ .Values.binder.image.pullPolicy }}
           args:
-            - --webhook-addr={{ .Values.binder.ports.webhookPort }}
+            - "--webhook-addr={{ .Values.binder.ports.webhookPort }}"
```
omer-dayan, line 26: Why?

deployments/kai-scheduler/templates/services/resourcereservation-namespace.yaml
```diff
@@ -1,11 +1,12 @@
 # Copyright 2025 NVIDIA CORPORATION
 # SPDX-License-Identifier: Apache-2.0
 
-{{ if not (lookup "v1" "Namespace" "" "runai-reservation") -}}
+{{- $ns := .Values.global.resourceReservation.namespace }}
+{{ if not (lookup "v1" "Namespace" "" $ns) -}}
```
omer-dayan, line 5: What is this?

deployments/kai-scheduler/templates/services/binder.yaml
romanbaron: Just because all other args have these format

deployments/kai-scheduler/templates/services/resourcereservation-namespace.yaml
romanbaron: The purpose here is to lookup for the resource reservation namespace with the name in values file. It can be `kai-resource-reservation` or `runai-reservation` if someone still uses the legacy configs.

omer-dayan: Ohh got it

PR #225, opened 2025-06-10
avoid creating `nodescaleadjuster` artefacts when disabled

slaupster:
the `Deployment` already [has this](https://github.com/NVIDIA/KAI-Scheduler/blob/a1f3cd1a074503e5dbddd5b1803fc1bd7625f836/deployments/kai-scheduler/templates/services/nodescaleadjuster.yaml#L4) guard. It should be extended to include its IAM and service.


romanbaron:
Thanks for suggesting this change, it makes a lot of sense.
The reason the validation test fails is because nodescaleadjuster ClusterRole is generated here:
https://github.com/NVIDIA/KAI-Scheduler/blob/main/Makefile#L65
And the generated version doesn't have the `if guard` that you added.

slaupster:
> Thanks for suggesting this change, it makes a lot of sense. The reason the validation test fails is because nodescaleadjuster ClusterRole is generated here: https://github.com/NVIDIA/KAI-Scheduler/blob/main/Makefile#L65 And the generated version doesn't have the `if guard` that you added.

I have to say I'm not following what the manifests target is intending to do - just generate what is missing ? If I just remove it like last commit we accept going forward none of it is generated because it has go template for the helm ?

romanbaron:
All ClusterRoles yaml files are generated from kubebuilder annotations, it assists with keeping track of where the resources in the ClusterRole are actually being used. Here is an example:
https://github.com/NVIDIA/KAI-Scheduler/blob/238759247051ba0c7ff6450debb2b8b74ddecfcc/pkg/podgrouper/pod_controller.go#L57
So these files cannot be manually modified. In addition 'make validate' step in CI attempts to catch a possible case where someone added a kubebuilder annotation but did not re-generate the yaml files.
Usually yaml files can be patched with Kustomize (example: [Kustomize](https://github.com/NVIDIA/KAI-Scheduler/tree/238759247051ba0c7ff6450debb2b8b74ddecfcc/deployments/kustomization)) but I am not sure if there is a way to wrap the entire file with a condition using this method.

itayvallach:
@slaupster is there an update here?

slaupster:
sorry @itayvallach - just noticed this in my GH app 🤦 - I think this is now right ?

PR #226, opened 2025-06-10
pod labels for pre and post install hooks

slaupster:
The `Job` for these have metadata labels but the spec is missing the same.  It is useful to have the same labels on the pods so you can do things like, for example, attach cilium policy predictably to the pods that are created by the job.


enoodle:
I am sorry, I didn't mean to move the contents of the files on you.
It moved to `helpers.tmpl` file.

romanbaron:
@slaupster - can you please resolve conflicts so it will get merged?

davidLif:
The way the pre and post hooks are managed have changed. Closing PR.

PR #240, opened 2025-06-16
Fix typo, formatting and grammar tweaks in MPS docs

jacobtomlinson:
I was reading the docs and noticed a few small formatting problems. Thanks for your hard work here.


enoodle:
Thank you!

PR #269, opened 2025-06-26
bind request mutations design

enoodle:


docs/developer/designs/bindrequest-mutations/README.md
```diff
@@ -0,0 +1,118 @@
...
+## Design Details
+
+### Extension Point Definition
+
+Following the scheduler's plugin extension conventions, we introduce a function type for mutating BindRequest annotations and labels:
+
+```go
+// In pkg/scheduler/api/types.go
+// BindRequestMutateFn allows plugins to mutate annotations and labels before BindRequest creation.
+type BindRequestMutateFn func(pod *pod_info.PodInfo, nodeName string, annotations, labels map[string]string) (map[string]string, map[string]string)
```
davidLif: 1- What do you think about the name `BindRequestAnnotators`?
2- Do you see cases in which we would like to spesificly use bindrequest label, and not just an annotation?
3- Are the `annotations, labels map[string]string` parameters the pod metadata or previous plugins BindRequest metadata? In the second case, why would a plugin relay on the changes of a different plugin for it's logic?

enoodle: 1. I prefer to stick with the current name structure of the rest of the plugin function types
2. Lets do only annotations for now, We can add labels in the future if necessary, but I currently don't have a user story for that.
3. These should be the annotations of the bind request (the pod is available through the pod_info.PodInfo typed variable ) - I will remove them as input to the function.

davidLif (approved):

PR #297, opened 2025-07-03
Remove representative jobs from scenario info

itsomri:
Remove unused RepresentativeJobs from scenario info


SiorMeir (approved):
looks ok

PR #311, opened 2025-07-08
[WIP]: Added time aware fairness proposal

itsomri:
This PR proposes a way to consider historical resource consumption for fairness calculations, in order to achieve fair allocation of resources over time. The purpose of this PR is to provide a platform for discussion of different use cases and design considerations. Feedback is welcome.


docs/fairness/time-aware-fairness.md
```diff
@@ -0,0 +1,159 @@
...
+Users can rely on guaranteed resources for non-preemptible interactive & inference workloads, while queueing train jobs. Time-aware fairness will ensure fair access to resources, roughly proportional to the weight assigned to the different queues. Interactive and inference workloads will not be interrupted due to the usage of in-quota resources & non-preemptible properties.
+
+#### Story 4: No over-usage (open for debate)
+In a cluster with no deserved quotas, N queues are weighted arbitrarily (W_i for i in N is queue `i`'s weight, Ẃ_i = W_i/∑W_j for i,j in N is the normalized weight). If no queue ever goes above it's relative share - should they be penalized for their usage? On one hand, it's not intuitively fair to be penalized for usage of one's relative share. On the other, this could cause any queue with a pending job that requires more than it's relative share to be starved. For example, SDRF doesn't count in-share usage as "Commitment".
```
keinanDaniel: Whats SDRF?
What do you suggest? should we or should we not penalize for in share usage?

docs/fairness/time-aware-fairness.md
```diff
@@ -0,0 +1,159 @@
...
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation. What is the desired outcome and how do we measure success?.
+The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+### User Stories
```
keinanDaniel: We should maybe have slightly more requirments here
I think another requirement here is that a user should be able to:  Not use this feature at all (opt-in feature) - and all scheduler behavior should stay the same

docs/fairness/time-aware-fairness.md
```diff
@@ -0,0 +1,159 @@
...
+<!--
+ToDo
+-->
+
+## Design Details
+
+> **⚠️ Work in Progress**  
+> This design is currently under development. The mathematical formulation and implementation details are an initial proposal for further discussion. Feedback and suggestions are welcome.
+
+Accounting service will provide the scheduler with data for each cycle about the usage of resources per queue per resource. A vector will represent the usage of each queue per resource type.
```
keinanDaniel: Maybe it is just me. but It is not clear from wording -> is it a matrix? or a vector per queue?
Suggestion:
For each queue there will be a vector containing usage for every resource type.

docs/fairness/time-aware-fairness.md
```diff
@@ -0,0 +1,159 @@
...
+
+> **⚠️ Work in Progress**  
+> This design is currently under development. The mathematical formulation and implementation details are an initial proposal for further discussion. Feedback and suggestions are welcome.
+
+Accounting service will provide the scheduler with data for each cycle about the usage of resources per queue per resource. A vector will represent the usage of each queue per resource type.  
+The resource division formula for over-quota resource can be found in proportion plugin, resource_division.go:
+```Go
+fairShare := amountToGiveInCurrentRound * (overQuotaWeight / totalWeights)
+```
+This calculation is run for each resource type.
```
keinanDaniel: is done (instead of run)

docs/fairness/time-aware-fairness.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Option 2: SDRF-Inspired Approach
+
+This option adapts the Start-time Dominant Resource Fairness (SDRF) algorithm to account for different queue weights and guaranteed quotas, which the original SDRF does not consider.  
+
+#### Overusage Definition
+
+We define "overusage" as the cumulative excess resource consumption beyond a queue's fair share over time. For each resource type, the overusage for queue $i$ at time $t$ is calculated as:
+
+$$\text{O}_i(t) = \int_0^t \left[ \text{U}_i(\tau) - \text{D}_i - \text{NC}(\tau) \cdot \frac{w_i}{\sum_{j} w_j} \right] d\tau$$
```
keinanDaniel: Maybe i'm the only one but cannot read this expression

docs/fairness/time-aware-fairness.md
```diff
@@ -0,0 +1,159 @@
...
+- **U<sub>i</sub>(τ)** is the actual resource usage by queue $i$ at time $\tau$
+- **D<sub>i</sub>** is the deserved quota for queue $i$
+- **NC(τ)** is the non-committed capacity at time $\tau$ (total capacity minus sum of all deserved quotas)
+- **w<sub>i</sub>** is the weight of queue $i$
+- **∑w<sub>i</sub>** is the sum of weights across all competing queues
+
+This formulation ensures that:
+1. Queues are only penalized for usage beyond their guaranteed quota and fair share
+2. The penalty is proportional to the queue's weight in the over-quota allocation
+3. The historical usage is accumulated over time with appropriate decay
```
keinanDaniel: I do think a summary of pros and cons will be helpful at the end between two approaches.
I honestly say that i did not understand the 2nd option yet so i might missed something here

docs/fairness/time-aware-fairness.md
itsomri: It's a question up for debate - I don't know the answer yet.
SDRF is an algorithm proposed in a paper called "Stateful DRF: Considering the Past in a Multi-Resource Allocation", I will add a citation

docs/fairness/time-aware-fairness.md
itsomri: It's not clear because it's not determined yet - the exact API is out of scope for this doc

docs/developer/designs/time-aware-fairness/time-aware-fairness.md
```diff
@@ -0,0 +1,175 @@
...
+$$F_i = C \cdot (w'_i - \~U_i)$$
+
+Which will give us the following:
+* When usage is zero for all queues, we revert back to the current calculation
+* Decreased fair share for queues that are using more than average ($U_i > 0$)
+* Queues that their relative **usage** is bigger than their relative **share** will get a negative value
+
+Since we don't want to assign negative fair share, we will max this expression with 0:
+
+$$F_i = \max{\{C \cdot (w'_i - \~U_i), 0\}}$$
```
enoodle: I am afraid that this means we can have more fair share than the total available resources:

The sum of all normalized weights is one: $$\sum w'_i = 1 $$
The sum of all normalized usages  is also one: $$\sum U'_i = 1$$
So the sum of all fair shares before this $$\max$$ is $$\sum F_i = C \cdot \sum(w'_i - \~U_i) = C \cdot \sum{w'_i - U_i + m} = C \cdot (1 - 1 + 1)  = C$$
This is great, but if we use `max` on some of the values we will sometimes get a total that is more than C and we will have issues distributing it.

itsomri: Hmm, right.
Maybe we don't have to mean-center the usages, I'll try to play around with it

docs/developer/designs/time-aware-fairness/time-aware-fairness.md
```diff
@@ -0,0 +1,219 @@
...
+$$W'_i = \frac{W_i}{\sum{W}}$$
+
+
+### Vacant-adjusted normalized usage
+
+First, we'll define a normalized "usage" score for each queue and resource:
+
+$$U'_i = \frac{U_i}{\sum{U}}$$
+
+We can furthermore consider the **unallocated** resources in the cluster. If we go back to the definition of $U'_i$, we can add:
```
romanbaron: Do you also plan to keep historical data for the available resources in the cluster to keep track of what was available at specific time in scalable cluster?

docs/developer/designs/time-aware-fairness/time-aware-fairness.md
```diff
@@ -0,0 +1,219 @@
...
+
+![fig 1: job allocation with strict DRF](strict-drf-fig-1.svg)
+
+However, a time-aware scheduling algorithm will decrease the fair share for the queue of the running job, and increase it for the starved queue, resulting in an oscillating allocation of jobs between them:
+
+![fig 2: job allocation with time aware fairness](time-aware-fairness-fig-2.svg)
+
+This, potentially coupled with min-runtime protections for added guarantees, can ensure a non-disruptive oscillation between the queues' access to resources.
+
+#### Story 2: Burst access to resources
```
Hagay-RunAI, line 71: This depends on the time decay window size and the weights of the other queues, i.e. it is not always guaranteed that such a burst can take place. If the other queues initial weight is very high and the time window is short then I would imagine that the very low share queue does not get to have a fairshare that is ever larger than the high share queues.

docs/developer/designs/time-aware-fairness/time-aware-fairness.md
```diff
@@ -0,0 +1,219 @@
...
+
+#### Story 3: Time aware fairness with guaranteed resources
+In a heterogeneous cluster combining interactive, critical inference, and training use cases, different users with different queues require reliable access to deserved resources, while time-sharing over-quota resources for training jobs. Admin assigns deserved quota for each queue based on the users' needs, allowing them to run inference and some interactive workloads for use cases such as data exploration and small-scale experiments with jupyter notebooks and similar tools.  
+Users can rely on guaranteed resources for non-preemptible interactive & inference workloads, while queueing train jobs. Time-aware fairness will ensure fair access to resources, roughly proportional to the weight assigned to the different queues. Interactive and inference workloads will not be interrupted due to the usage of in-quota resources & non-preemptible properties.
+
+#### Story 4: No over-usage (open for debate)
```
Hagay-RunAI, line 78: I suggest to "keep it simple" and not penalize for "over-usage", we should penalize simply for any usage above the Deserved quota, that way we maintain a relatively simple to explain over-quota time-based sharing. We can leave this for future study or other future plugins/behaviors we want to explore and implement.

docs/developer/designs/time-aware-fairness/time-aware-fairness.md
```diff
@@ -0,0 +1,219 @@
...
+
+#### Intuitive understanding of fairness over time
+Several parameters and configurations can effect the outcome of the scheduling algorithm: good documentation, and potentially tools, will need to be provided in order to help the admin configure the cluster properly. An intuitive configuration should result in behavior that is predictable in simple scenarios: for example, "Using X GPUs for T hours will result in a project with W over-quota weight getting penalized by a factor of 50%". Several variables can go into achieving this result - we need to map these in a clear way.
```
Hagay-RunAI, line 97: Great point Omri, I think this is one of the major challenges for us - to assist Admins to make the right "weighting" and Decay time and rate configurations.

docs/developer/designs/time-aware-fairness/time-aware-fairness.md
```diff
@@ -0,0 +1,196 @@
...
+
+$$P_i = \max{\{W'_i - k \cdot (W'_i - U'_i), 0\}}$$
+
+$$P'_i = \frac{P_i}{\sum{P}}$$
+
+Where:
+- $W'_i$ is the over-quota weight of the queue, normalized to the sum of non-satisfied queues' over-quota weight for the round
+- $\sum{P}$ considers only queues that participate in the current resource division round
+- $U'_i$ is the historical usage for the queue, normalized to the capacity of the cluster
+- $k$ is set by the admin and controls the "significance" of historical usage
```
enoodle, line 191: Is k always $0 <= k <= 1$ ? If so it means the max might not be really needed.

itsomri: No, K defaults to 1 but can be much higher to give more significance to the usage-adjusted part of the equation
I'm thinking of maybe limiting K to [0, 1] and changing the formula to:

$$P_i = \max{\{(1 - k) \cdot W'_i + k \cdot (W'_i - U'_i), 0\}}$$

This way, the meaning of K might make sense to humans more: k=0 is like the old formula, k=1 is fully usage-adjusted, k=0.5 is the default

itsomri: Also you can't assume that u < 2w

PR #320, opened 2025-07-15
Natasha/add queue metrics

natasharomm:
* added some metrics to queuecontroller: queue_info, queue_deserved_gpus, queue_quota_cpu_cores, queue_quota_memory_bytes.
* having a configuration for queue labels converted to metric labels

### DIFF
--- a/cmd/queuecontroller/app/options.go
+++ b/cmd/queuecontroller/app/options.go
@@ -5,14 +5,31 @@ package app
 
 import (
 	"flag"
+
+	kaiflags "github.com/NVIDIA/KAI-scheduler/pkg/common/flags"
+)
+
+const (
+	defaultSchedulingQueueLabelKey = "kai.scheduler/queue"
+	defaultMetricsNamespace        = "kai"
+	defaultMetricsAddress          = ":8080"
 )
 
 type Options struct {
 	EnableLeaderElection    bool
 	SchedulingQueueLabelKey string
+
+	MetricsAddress                 string
+	MetricsNamespace               string
+	QueueLabelToMetricLabel        kaiflags.StringMapFlag
+	QueueLabelToDefaultMetricValue kaiflags.StringMapFlag
 }
 
 func (o *Options) AddFlags(fs *flag.FlagSet) {
 	fs.BoolVar(&o.EnableLeaderElection, "leader-elect", false, "Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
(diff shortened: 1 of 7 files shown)


cmd/queuecontroller/app/options.go
```diff
@@ -5,14 +5,31 @@ package app
 
 import (
 	"flag"
+
+	kaiflags "github.com/NVIDIA/KAI-scheduler/pkg/common/flags"
+)
+
+const (
+	defaultSchedulingQueueLabelKey = "kai.scheduler/queue"
+	defaultMetricsNamespace        = "kai"
```
davidLif: The default namespace should be "kai-scheduler"

davidLif: Please validate against the current default kai namespace

pkg/queuecontroller/controllers/queue_controller.go
```diff
@@ -54,7 +55,12 @@ func (r *QueueReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 	queue := &v2.Queue{}
 	err := r.Get(ctx, req.NamespacedName, queue)
 	if err != nil {
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		ignoreNotFoundErr := client.IgnoreNotFound(err)
+		if ignoreNotFoundErr == nil {
+			// If the queue is not found, reset its metrics
+			metrics.ResetQueueMetrics(req.Name)
```
davidLif, line 61: Even if the get fails only once? occasional error can occur if the API server is under pressure

cmd/queuecontroller/app/options.go
natasharomm: it's not really a k8s namespace, it's actually the prefix for the metric names.
I copied the behavior from the existing metrics in "scheduler" - https://github.com/NVIDIA/KAI-Scheduler/blob/main/cmd/scheduler/app/options/options.go#L34 

I think it should be the same for all metrics in the repo.
perhaps I can move this const to a common dir and use it in both places instead.

pkg/queuecontroller/controllers/queue_controller.go
natasharomm: I do it only for "NotFound" error, not on every error. If the API server has some kind of issue it would return a different error, and not "NotFound" error

cmd/queuecontroller/app/options.go
natasharomm: I pushed a fix for that - DefaultMetricsNamespace as a const in common

PR #324, opened 2025-07-17
Design doc for hierarchical PodGroup structure

romanbaron:
Grove workload architecture comprises multiple PodCliques, decoupling prefill from decode operations and leader from workers pods, thereby enabling each PodClique to utilize a distinct pod template specification and to scale independently. Upon scheduling a Grove workload, all constituent PodCliques are scheduled as a gang to ensure collective orchestration. Certain PodCliques may prefer or require close location within the cluster to benefit from reduced network latency and faster communication. A simple example is the leaders and workers pods, where, despite belonging to separate PodCliques, these components maintain tight coupling and frequent interactions.

This PR suggest an extension to PodGroup CRD that supports this requirement.

### DIFF
--- a/docs/developer/designs/hierarchical-podgroup/README.md
+++ b/docs/developer/designs/hierarchical-podgroup/README.md
@@ -1,17 +1,17 @@
 # Hierarchical PodGroup structure
 
 ## Overview
-This document outlines an enhancement to the PodGroup architecture by introducing SubGroups within a PodGroup, enabling fine-grained control over gang scheduling and topology-aware placement for subsets of pods within a workload. While the PodGroup continues to enforce gang scheduling for the workload as a whole, this extension allows workloads to specify minMember and topology constraints on individual SubGroups, providing greater resolution over how smaller groups of pods are placed and scheduled. For example, a SubGroup can specify that only a subset of its pods must be scheduled together (minMember) and define specific network topology constraints aligned with that subset’s requirements. This design preserves atomic workload scheduling semantics while allowing advanced, granular placement policies that improve alignment with topology-aware scheduling strategies.
+This document outlines an enhancement to the PodGroup architecture by introducing SubGroups within a PodGroup, enabling fine-grained control over gang scheduling and topology-aware placement for subsets of pods within a workload. While the PodGroup continues to enforce gang scheduling for the workload as a whole, this extension allows workloads to specify minMember and topology constraints on individual SubGroups, providing greater resolution over how smaller groups of pods are placed and scheduled. For example, a SubGroup can specify that only a subset of its pods must be scheduled together (minMember) and define specific network topology constraints aligned with that subset’s requirements. This design preserves atomic workload scheduling semantics while enabling advanced, granular placement policies that align closely with topology-aware scheduling strategies.
 
 ## Motivation
-Modern inference workloads consist of components with varying network sensitivity and resource requirements, necessitating the ability to express fine-grained scheduling and placement preferences within the context of gang-scheduled workloads. Enforcing uniform constraints at the PodGroup level can lead to overly restrictive placement requirements that do not reflect the nuanced needs of different workload components. By introducing SubGroups within a PodGroup, workloads can specify per-SubGroup minMember requirements and topology constraints while maintaining overall gang scheduling semantics. This enables advanced placement strategies, improves scheduling flexibility and efficiency, and ensures that differentiated placement and partial gang constraints can be expressed without relaxing the workload’s atomic scheduling guarantees.
+Modern inference workloads consist of components with varying network sensitivity and resource requirements, necessitating the ability to express fine-grained scheduling and placement preferences within the context of gang-scheduled workloads. Uniform constraints at the PodGroup level may impose overly restrictive placement requirements, failing to capture the nuanced needs of distinct workload components. By introducing SubGroups within a PodGroup, workloads can specify per-SubGroup minMember requirements and topology constraints while maintaining overall gang scheduling semantics. This enables advanced placement strategies, improves scheduling flexibility and efficiency, and ensures differentiated placement and subgroup-level gang constraints can be expressed without compromising the workload’s atomic scheduling guarantees.
 
 ## Goals
-- Fine-Grained Scheduling Specification: Allow PodGroups to define how smaller SubGroups of pods should be scheduled, specifying minMember and topology constraints per SubGroup.
-- Preserve Atomicity with Flexibility: Maintain gang scheduling semantics at the workload level while enabling per-SubGroup constraints for granular control over subsets of pods.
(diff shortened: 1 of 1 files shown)


docs/developer/designs/hierarchical-podgroup/README.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Motivation
+Modern inference workloads consist of components with varying network sensitivity and resource requirements, necessitating the ability to express fine-grained scheduling and placement preferences within the context of gang-scheduled workloads. Enforcing uniform constraints at the PodGroup level can lead to overly restrictive placement requirements that do not reflect the nuanced needs of different workload components. By introducing SubGroups within a PodGroup, workloads can specify per-SubGroup minMember requirements and topology constraints while maintaining overall gang scheduling semantics. This enables advanced placement strategies, improves scheduling flexibility and efficiency, and ensures that differentiated placement and partial gang constraints can be expressed without relaxing the workload’s atomic scheduling guarantees.
+
+## Goals
```
itsomri, line 9: IMO, an important goal is to not have a regression in the *experience* for existing, simpler use cases. Backwards compatibility is mentioned below, so no features are expected to be lost, but it's also important to keep the experience good as well - I mainly think of errors, conditions and logs, not becoming too complicated for simple 1-subgroup podgroups

docs/developer/designs/hierarchical-podgroup/README.md
```diff
@@ -0,0 +1,236 @@
...
+	SubGroups []SubGroup `json:"subGroups,omitempty"`
+}
+
+type SubGroup struct {
+    // Name uniquely identifies the SubGroup within the parent PodGroup.
+	Name string `json:"name"`
+
+	// Parent optionally specifies the name of a parent SubGroup, enabling hierarchical grouping.
+	// +optional
+	Parent string `json:"parent,omitempty"`
```
itsomri: An N-level hierarchy can quickly become incredibly complex - to request as a user, to schedule as a scheduler, and to debug and explain errors on. IMO, if we don't see a very important, valid use case, or get more validation from the community for this - we should keep to 2-levels.

docs/developer/designs/hierarchical-podgroup/README.md
```diff
@@ -0,0 +1,236 @@
...
+      podSelector:
+        matchLabels:
+          role: prefill-leader
+      minMember: 1
+      topologyConstraints:
+        requiredTopologyLevel: rack
+```
+
+## Development
+To ensure a controlled and backward-compatible rollout of PodGroup SubGroups with fine-grained scheduling, the following phased development plan is proposed:
```
itsomri, line 240: I don't see here an explanation on how we will keep BC - do we need a conversion webhook? Or some other way to convert old podgroups to the new format?

romanbaron: There is no need to convert old PodGroups, the API still allows creating PodGroups without SubGroups. These can specify minMember and TopologyConstraints at the global PodGroup scope without specifying SubGroups.

docs/developer/designs/hierarchical-podgroup/README.md
romanbaron: Right, and it is natively accomplished because we keep using a single PodGroup instance like we do today. Unlike the option of creating a separate CRD for each SubGroup, here the additional requirements compliment the existing structure while maintaining the same representation at the workload level. Conditions and status does not need to be changed.

docs/developer/designs/hierarchical-podgroup/README.md
romanbaron: Right, at the moment there is no such use case and I think it will not happen naturally so I am not sure if we want to block it.

docs/developer/designs/hierarchical-podgroup/README.md
```diff
@@ -0,0 +1,263 @@
...
+
+### User Story 2: SubGroup-Specific Topology Constraints
+I want one SubGroup to be placed within the same rack for low latency, while allowing another SubGroup pods to be placed flexibly across the cluster, aligning network-sensitive components with workload requirements.
+
+### User Story 3: Multi-SubGroup Workload with Co-Located Leaders and Workers
+I want to define decode-leaders, decode-workers, prefill-leaders, and prefill-workers SubGroups where leaders and workers are co-located within each group, but prefill and decode groups may be scheduled in different locations to optimize resource placement.
+
+### User Story 4: Backward Compatibility
+I want my existing PodGroups without SubGroups to continue working without modification, ensuring a seamless migration path for advanced scheduling capabilities.
+
```
davidLif, line 34: I would also add a user story for "Elastic subgroups" - To allocate the Podgroup, we need to be able to schdule some of the subgroups, but not all. For example, We would like to schedule a leaderWorkerSet object only if we can schedule at least 3 out of 5 of its replicas.

romanbaron: I think this can be achieved by introducing `minSubGroupMember` (or something similar) - do we want to add it now or just to discuss possible options for supporting it in the future?

docs/developer/designs/hierarchical-podgroup/README.md
```diff
@@ -0,0 +1,259 @@
...
+}
+
+// SubGroupSetTopologyConstraint defines a topology constraint for a set of subgroup names.
+type SubGroupSetTopologyConstraint struct {
+    // SubGroups is the list of subgroup names that this constraint applies to collectively.
+    SubGroups []string `json:"subGroups"`
+    
+    // Constraint defines the topology constraint to apply to the listed subgroups.
+    Constraint TopologyConstraint `json:"constraint"`
+}
```
enoodle, line 119: What would happen if I fill the same groups in several intersecting SubGroupSets ? Is that still a valid pod group  ?

```
  topologyConstraints:
    subGroups:
      decode-workers:
        requiredTopologyLevel: rack
      decode-leaders:
        requiredTopologyLevel: rack
      prefill-workers:
        requiredTopologyLevel: rack
      prefill-leaders:
        requiredTopologyLevel: rack

    subGroupSets:
      - subGroups:
          - decode-workers
          - decode-leaders
        constraint:
          requiredTopologyLevel: block
      - subGroups:
          - prefill-workers
          - prefill-leaders
        constraint:
          requiredTopologyLevel: block
      - subGroups:
          - decode-leaders
          - prefill-leaders
        constraint:
          requiredTopologyLevel: block
```

romanbaron: It is valid, each subGroup is constrained to a single rack. And there are SubGroup tuples that together as a set need to run on the same block. I imagined the constraints in `subGroupSets` to be more relaxed (i.e. preferred and not required) but this is possible as well.

enoodle (approved):

enoodle (approved):

PR #326, opened 2025-07-17
fair share by priority implementation

enoodle:
Implementing the priority based fair share distribution: https://github.com/NVIDIA/KAI-Scheduler/blob/main/docs/developer/designs/priority-based-fair-share.md


pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -59,14 +59,23 @@ type proportionPlugin struct {
...
 
 func New(arguments map[string]string) framework.Plugin {
-	return &proportionPlugin{
+	pp := &proportionPlugin{
 		totalResource:   rs.EmptyResourceQuantities(),
 		queues:          map[common_info.QueueID]*rs.QueueAttributes{},
 		pluginArguments: arguments,
 	}
+
+	if val, ok := arguments["priorityBasedFairShare"]; ok {
```
itsomri, line 73: Use `strconv.ParseBool()`: http://gosamples.dev/string-to-bool/

pkg/scheduler/plugins/proportion/resource_division/priority_resource_division.go
```diff
@@ -0,0 +1,81 @@
...
+// SetResourcesSharePriority distributes resources between queues according to the
+// three-phase algorithm described in docs/developer/designs/priority-based-fair-share.md.
+//
+//  1. Queues are grouped by priority (higher priority buckets processed first).
+//  2. Within a bucket we run:
+//     a) Phase-1: allocate Deserved quota (subject to remaining capacity).
+//     b) Phase-2: allocate over-quota resources proportionally to OverQuotaWeight.
+//     c) Phase-3: allocate any still-remaining capacity equally between the queues
+//     that still have unmet demand (treating all weights as 1).
+func SetResourcesSharePriority(totalResource rs.ResourceQuantities, queues map[common_info.QueueID]*rs.QueueAttributes) {
```
itsomri, line 23: I think this should be the only behavior. If anyone doesn't like priority-based fair share, just don't use priorities.

pkg/scheduler/plugins/proportion/resource_division/priority_resource_division.go
```diff
@@ -0,0 +1,81 @@
...
+				break
+			}
+			bucket := queuesByPriority[priority]
+
+			// ---------- Phase 1: Deserved quota ----------
+			for _, q := range bucket {
+				if remaining <= 0 {
+					break
+				}
+				rsShare := q.ResourceShare(resource)
```
itsomri: This is too deep! Let's keep the indentations low - this loop could be offloaded to a function

pkg/scheduler/plugins/proportion/resource_division/priority_resource_division.go
```diff
@@ -0,0 +1,81 @@
...
+
+			// ---------- Phase 2: OverQuotaWeight ----------
+			remaining = divideOverQuotaResource(remaining, bucket, resource)
+			if remaining <= 0 {
+				continue
+			}
+
+			// ---------- Phase 3: Weight-0 compensation ----------
+			// save original OverQuotaWeight
+			originalOverQuotaWeight := make(map[common_info.QueueID]float64)
```
itsomri: This is kinda awkward, I suggest another way: we can modify `divideUpToFairShare` to use a normalized weight, so that if `getTotalWeightsForUnsatisfied` is 0, instead of exiting, we can use a value of 1 automatically. Hard to explain but we can discuss it

pkg/scheduler/plugins/proportion/resource_division/priority_resource_division.go
enoodle: I want to preserve backwards compatibility with the current behavior.

pkg/scheduler/plugins/proportion/resource_division/priority_resource_division.go
enoodle: I am not sure it is very awkward. I think it is straight forward. After dividing over quota we do it once again for all (remaining) queues assuming their OQW is 1. 
You are suggesting a change in the function `divideOverQuotaResource` implementation that will effect the way over quota is distributed in all cases, even if the priority based flag is not in effect. This will either change current behavior which is not something I want to do or make us trickle down a flag into the `divideOverQuotaResource`   and `divideUpToFairShare` functions which might also be awkward.

pkg/scheduler/plugins/proportion/resource_division/priority_resource_division.go
enoodle: :+1: I will refactor it into several functions.

pkg/env-tests/scheduler_test.go
```diff
@@ -338,4 +341,128 @@ var _ = Describe("Scheduler", Ordered, func() {
 			Expect(podMap).To(HaveLen(deviceNum), "Expected exactly %d pods to be scheduled", deviceNum)
 		})
 	})
+
+	Context("priority based fair share", func() {
+		var priorityBasedFairShareSchedulerConf = &conf.SchedulerConfiguration{
```
itsomri, line 346: Can we outsource this to some kind of `DefaultSchedulerConfiguration`, that you will use and only override `priorityBasedFairShare`?
The reason it's important is that when we make later changes like add/modify plugins, we will not remember to update it in all the places, and each test will check different configurations.
I started working on something like this but got distracted: https://github.com/NVIDIA/KAI-Scheduler/pull/228/files

If it's too much then LMK and I'll create a dedicated ticket

pkg/scheduler/plugins/proportion/resource_division/priority_resource_division.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	// Phase 1: Distribute Deserved quota
+	remaining = divideDeservedQuota(remaining, bucket, resource, totalResource)
+	if remaining <= 0 {
+		return remaining
+	}
+
+	// Phase 2: Distribute over quota between queues
+	remaining = divideOverQuotaResource(remaining, bucket, resource)
```
itsomri, line 58: Does this mean that the higher priority will get it's over quota resources **before** lower priority gets its deserved resources? This is a major change

PR #330, opened 2025-07-21
Natasha/validate priority exists for podgroup

natasharomm:
* in podgrouper's "CalcPodGroupPriorityClass", validate the calculated PriorityClass object exists in the k8s cluster. If not - fallback to default from configmap, or the default priorityClass for type.
* had to move "kubeClient" param to the "NewDefaultGrouper" constructor - it is now required always to validate the priorityClass exists.
* fixed and added tests.

### DIFF
--- a/pkg/podgrouper/podgrouper/hub/hub.go
+++ b/pkg/podgrouper/podgrouper/hub/hub.go
@@ -76,8 +76,8 @@ func (ph *PluginsHub) GetPodGrouperPlugin(gvk metav1.GroupVersionKind) grouper.G
 func NewPluginsHub(kubeClient client.Client, searchForLegacyPodGroups,
 	gangScheduleKnative bool, queueLabelKey, nodePoolLabelKey string,
 	defaultPrioritiesConfigMapName, defaultPrioritiesConfigMapNamespace string) *PluginsHub {
-	defaultGrouper := defaultgrouper.NewDefaultGrouper(queueLabelKey, nodePoolLabelKey)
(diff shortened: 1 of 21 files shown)


pkg/podgrouper/podgrouper/hub/hub.go
```diff
@@ -76,8 +76,8 @@ func (ph *PluginsHub) GetPodGrouperPlugin(gvk metav1.GroupVersionKind) grouper.G
 func NewPluginsHub(kubeClient client.Client, searchForLegacyPodGroups,
 	gangScheduleKnative bool, queueLabelKey, nodePoolLabelKey string,
 	defaultPrioritiesConfigMapName, defaultPrioritiesConfigMapNamespace string) *PluginsHub {
-	defaultGrouper := defaultgrouper.NewDefaultGrouper(queueLabelKey, nodePoolLabelKey)
-	defaultGrouper.SetDefaultPrioritiesConfigMapParams(defaultPrioritiesConfigMapName, defaultPrioritiesConfigMapNamespace, kubeClient)
```
davidLif, line 80: Why did you move the kubeClient parameter to the ctor?

natasharomm: had to move "kubeClient" param to the "NewDefaultGrouper" constructor - it is now required always to validate the priorityClass exists, so it has to be in the constructor since it's not optional any more.

davidLif (approved):

PR #331, opened 2025-07-22
feat: copy webhooks to seperate files

SiorMeir:
## Goal & Motivation
Removing some tech debt and improving code hygiene by separating `PodValidator` and `PodMutator` to a dedicated file structure.

## What is done here

* adding a new file directory for admission webhooks
* copying existing hooks
* moving plugin interface to common directory
* adding unit tests

## What is next

* switching KAI-scheduler to use hooks in the new file directories
* removing old hooks


pkg/common/interfaces/plugin.go
SiorMeir, line 1: move interface into admission

pkg/admission/plugins/plugins.go
```diff
@@ -0,0 +1,78 @@
...
+			logger := log.FromContext(context.Background())
+			logger.Error(err, "pod mutation failed for pod",
+				"namespace", pod.Namespace, "name", pod.Name, "plugin", p.Name())
+			return err
+		}
+	}
+	return nil
+}
+
+func (bp *KaiAdmissionPlugins) PreBind(ctx context.Context, pod *v1.Pod, host *v1.Node, bindRequest *v1alpha2.BindRequest,
```
SiorMeir: remove PreBind and PostBind from the plugin as it is not relevant for this hook

PR #334, opened 2025-07-22
[WIP] filter job by topology - step 1

davidLif:
Base code fro prePredicate with node allocation checks instead of domain sum


pkg/scheduler/plugins/scores/scores.go
```diff
@@ -9,5 +9,6 @@ const (
 	Availability   = 100
 	GpuSharing     = 1000
 	K8sPlugins     = 10000
+	Topology       = 10000
```
itsomri, line 11: Having it share weight with K8sPlugins could be problematic. We should decide if it's higher or lower

pkg/scheduler/plugins/topology/topology_plugin.go
```diff
@@ -16,8 +19,11 @@ const (
 )
 
 type topologyPlugin struct {
-	enabled       bool
-	TopologyTrees map[string]*TopologyInfo
+	enabled            bool
+	taskOrderFunc      common_info.LessFn
+	sessionStateGetter k8s_internal.SessionStateProvider
+	nodesInfoMap       map[string]*node_info.NodeInfo
```
itsomri:
```suggestion
	nodesInfos       map[string]*node_info.NodeInfo
```

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
```diff
@@ -0,0 +1,328 @@
...
+
+func (t *topologyPlugin) calcTreeAlocationData(job *podgroup_info.PodGroupInfo, topologyTree *TopologyInfo) {
+	taskToAllocateCount, leafDomains := t.calcAllocationsForLeafDomains(job, topologyTree)
+
+	// Run from leafs to root updating distances and allocated pods
+	latestUpdatedDomains := leafDomains
+	currentlyUpdatedDomains := map[TopologyDomainID]*TopologyDomainInfo{}
+	for len(latestUpdatedDomains) > 0 {
+		for _, domain := range latestUpdatedDomains {
+			if domain.Parent != nil {
```
itsomri:
```suggestion
			if domain.Parent == nil {
				continue
			}
```
Happy flow to the left

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
```diff
@@ -0,0 +1,328 @@
...
+	return topologyTree, nil
+}
+
+func (t *topologyPlugin) calcTreeAlocationData(job *podgroup_info.PodGroupInfo, topologyTree *TopologyInfo) {
+	taskToAllocateCount, leafDomains := t.calcAllocationsForLeafDomains(job, topologyTree)
+
+	// Run from leafs to root updating distances and allocated pods
+	latestUpdatedDomains := leafDomains
+	currentlyUpdatedDomains := map[TopologyDomainID]*TopologyDomainInfo{}
+	for len(latestUpdatedDomains) > 0 {
```
itsomri, line 100: Is this meant to be for or if? Anyways, revert the condition

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+
+	// Calc tree job allocation data
+	t.calcTreeAlocationData(job, topologyTree)
+
+	// Get best domains for the job
+	jobAllocateableDomain, err := t.getBestjobAllocateableDomains(job, topologyTree)
+	if err != nil {
+		return err
+	}
```
itsomri: Is there any case where these two are not called together? If not, just call `calcTreeAlocationData` from inside `getBestjobAllocateableDomains`

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
```diff
@@ -0,0 +1,328 @@
...
+func (t *topologyPlugin) calcTreeAlocationData(job *podgroup_info.PodGroupInfo, topologyTree *TopologyInfo) {
+	taskToAllocateCount, leafDomains := t.calcAllocationsForLeafDomains(job, topologyTree)
+
+	// Run from leafs to root updating distances and allocated pods
+	latestUpdatedDomains := leafDomains
+	currentlyUpdatedDomains := map[TopologyDomainID]*TopologyDomainInfo{}
+	for len(latestUpdatedDomains) > 0 {
+		for _, domain := range latestUpdatedDomains {
+			if domain.Parent != nil {
+				// Known issue: shold be replaced with the real distance logic
```
itsomri:
```suggestion
				// Known issue: should be replaced with the real distance logic
```
This is a draft, So just make sure that you clean typos before the real PR, I saw some more

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
davidLif: I think that it makes sense to separate them as 2 different steps in the proccess:
1- Get matching topology
2- See if we have the result in cache
3- Calculate allocation data for the given job on all tree nodes
4- Get the domains which satisfies all the requirments
5- Save to cache

pkg/scheduler/plugins/scores/scores.go
enoodle: I agree, there this should have different weight (higher IMO)

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
```diff
@@ -0,0 +1,341 @@
...
+	cycleJobState.Write(
+		k8sframework.StateKey(topologyPluginName),
+		&topologyStateData{relevantDomains: jobAllocateableDomain},
+	)
+
+	return nil
+}
+
+func (t *topologyPlugin) getJobTopology(job *podgroup_info.PodGroupInfo) (*TopologyInfo, error) {
+	jobTopologyName := job.PodGroup.Annotations[topologyAnnotationKey]
```
enoodle: https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go#L75-L76

https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/podgrouper/podgroup/handler.go#L107-L111


the topology values are set in the pod group spec.

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
```diff
@@ -0,0 +1,341 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	k8sframework "k8s.io/kubernetes/pkg/scheduler/framework"
+)
+
+const (
+	topologyAnnotationKey                   = "kai.scheduler/topology"
+	topologyRequiredPlacementAnnotationKey  = "kai.scheduler/topology-required-placement"
+	topologyPreferredPlacementAnnotationKey = "kai.scheduler/topology-preferred-placement"
+)
+
```
enoodle: We don't use these annotations here, it is done already in the pod grouper (see comment about the usage)

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+	topologyTree := t.TopologyTrees[jobTopologyName]
+	if topologyTree == nil {
+		return nil, fmt.Errorf("matching topology tree haven't been found for job %s, workload topology name: %s",
+			job.PodGroup.Name, jobTopologyName)
+	}
+	return topologyTree, nil
+}
+
+func (t *topologyPlugin) calcTreeAlocationData(job *podgroup_info.PodGroupInfo, topologyTree *TopologyInfo) error {
```
enoodle: I suggest to rewrite it as post-order pass on the topology tree. I think it would be much clearer that way.

pkg/scheduler/plugins/topology/topology_structs.go
```diff
@@ -58,6 +58,9 @@ type TopologyDomainInfo struct {
 	// Number of pods that can be allocated in this domain for the job
 	AllocatablePods int
 
+	// Pods allocated distance from each other in the topology
+	Distance int
```
enoodle: I am not sure what is the meaning of this field. Is it the total distance between all pods that will be allocated on this specific domain?

enoodle: The distance that was meant to be used in the node ordering part was the distance of each node from the topology tree node that was selected for the job.

itayvallach:
This PR has been open for a long time. Yes it's in Draft - but is it still needed?

PR #337, opened 2025-07-23
Siormeir/feat-create-admission-webhooks-service-phase-2

SiorMeir:


pkg/binder/plugins/gpusharing/gpu_sharing.go
```diff
@@ -26,62 +24,25 @@ const (
 	CdiDeviceNameBase      = "k8s.device-plugin.nvidia.com/gpu=%s"
 )
 
-type GPUSharing struct {
+type BinderGPUSharing struct {
```
enoodle: I don't think that we need to change the name of the struct. because it is in this package `pkg/binder/plugins/gpusharing` it is already clear that it is part of the binder.

davidLif: I agree with @enoodle

SiorMeir: Accepted, will revert

SiorMeir: I'm applying this to the new `AdmissionGPUsharing` as well.

pkg/admission/webhook/v1alpha2/podhooks/pod_mutator.go
```diff
@@ -49,7 +49,6 @@ func (cpm *podMutator) Default(ctx context.Context, obj runtime.Object) error {
 	}
 
 	namespace, err := extractMutatingPodTargetNamespace(ctx, pod)
-	pod.Namespace = namespace
```
davidLif, line 52: Why did you remove this line?

SiorMeir: @davidLif the line is a duplicate. It appears again in after the error check

PR #339, opened 2025-07-23
Fair share simulator

itsomri:
Fair share calculations can be complicated, and are about to get even more complicated with time aware fair share.
It will be useful to have an interactive tool which will help admins, users, product, and CS people to get an intuitive feel for how fair share works. 

The simulator proposed here calls the `SetResourcesShare` function directly, which ensures that it will be consistent with the scheduler even when changes to the logic will be made.

The simulator runs as a stateless REST server, which will make it easy to use in different tools, as demonstrated in the attached examples (HTTP call, jupyter notebook)


.gitignore
```diff
@@ -5,4 +5,6 @@ cover.out
 .DS_Store
 coverage/
 *.test
-launch.json
\ No newline at end of file
+launch.json
+
+cmd/fairshare-simulator/fairshare-simulator
```
davidLif: How does it know to ignore the snapshot-tool binary but not this one?

itsomri: Looks like it doesn't - if you `go build . ` in `cmd/snapshot-tool`, it will not be ignored - I can add it to this .gitignore

davidLif: Please do

davidLif (dismissed):

PR #341, opened 2025-07-24
Small refactoring in job_info, preparing for SubGroups

romanbaron:


pkg/scheduler/cache/cluster_info/cluster_info.go
```diff
@@ -306,10 +306,6 @@ func (c *ClusterInfo) snapshotPodGroups(
 			continue
 		}
 
-		if podGroupInfo.MinAvailable == 0 {
-			podGroupInfo.MinAvailable = 1
-		}
```
romanbaron, line 311: I realized this part has no affect since `MinAvailable` is overwritten in `setPodGroupWithIndex` that is called in line 331 so I moved it there.

pkg/scheduler/api/podgroup_info/job_info.go
```diff
@@ -126,182 +126,182 @@ func NewPodGroupInfo(uid common_info.PodGroupID, tasks ...*pod_info.PodInfo) *Po
...
-	podGroupInfo.MinAvailable = pg.Spec.MinMember
-	podGroupInfo.Queue = common_info.QueueID(pg.Spec.Queue)
-	podGroupInfo.CreationTimestamp = pg.GetCreationTimestamp()
-	podGroupInfo.PodGroup = pg
-	podGroupInfo.PodGroupUID = pg.UID
+func (pgi *PodGroupInfo) SetPodGroup(pg *enginev2alpha2.PodGroup) {
+	pgi.Name = pg.Name
+	pgi.Namespace = pg.Namespace
+	pgi.NamespacedName = fmt.Sprintf("%s/%s", pgi.Namespace, pgi.Name)
+	pgi.MinAvailable = max(pg.Spec.MinMember, 1)
```
romanbaron, line 143: Commented about it in `cluster_info.go` - I am not sure it is really needed but we had some code that attempted to make sure `MinAvailable` is not zero so I moved it here. Maybe if the PodGroup is not created by the PodGrouper then it makes sense to have it here.

PR #342, opened 2025-07-24
Added SubGroups to PodGroup CRD and SubGroupInfo to PodGroupInfo

romanbaron:
Currently not being used anywhere, just instantiation and references from related objects.


pkg/scheduler/api/podgroup_info/subgroup_info.go
```diff
@@ -0,0 +1,31 @@
...
+		MinAvailable: minAvailable,
+		PodInfos:     pod_info.PodsMap{},
+	}
+}
+
+func FromSubGroup(subGroup *v2alpha2.SubGroup) *SubGroupInfo {
+	return NewSubGroupInfo(subGroup.Name, subGroup.MinMember)
+}
+
+func (sgi *SubGroupInfo) AddTaskInfo(ti *pod_info.PodInfo) {
```
davidLif: What about "removeTaskInfo"?

davidLif: Shouldn't this method be used by the jobInfo?

romanbaron: You are right, it probably will be more aligned this way.

While looking into it I noticed that the only use case for deleting a task is when task status is updated, old task is removed and new task, with the new status is added. And when it happens we don't delete the old task from PodInfos map - the old task is cloned and `Deleted` status is set.
https://github.com/NVIDIA/KAI-Scheduler/blob/daf2d15cf3c5266efe231108fa47c53ade426590/pkg/scheduler/api/podgroup_info/job_info.go#L265
Right after the task is cloned it is overwritten with the new PodInfo that has the same UID:
https://github.com/NVIDIA/KAI-Scheduler/blob/daf2d15cf3c5266efe231108fa47c53ade426590/pkg/scheduler/api/podgroup_info/job_info.go#L189

Isn't it redundant?

romanbaron: Unless there is a good reason for that, I can change the code so deleted task will not clone the task and instead - it will be removed from PodInfos map.

davidLif: I think that in some cases there are some pointers to the unupdated task. You can try and edit this, but please make a full e2e run with real gpus and see if something breaks

romanbaron: After speaking about it, we decided to do a small refactoring in this area:
https://github.com/NVIDIA/KAI-Scheduler/pull/345

davidLif (dismissed):

PR #347, opened 2025-07-28
Take SubGroups into consideration when checking if job is ready for s…

romanbaron:
…cheduling


pkg/scheduler/api/podgroup_info/subgroup_info.go
```diff
@@ -29,3 +30,28 @@ func fromSubGroup(subGroup *v2alpha2.SubGroup) *SubGroupInfo {
...
+	numTasks := 0
+	for _, task := range sgi.PodInfos {
+		if pod_status.IsAliveStatus(task.Status) {
+			numTasks += 1
+		}
+	}
+	return numTasks
+}
+
+func (sgi *SubGroupInfo) GetNumGatedTasks() int {
```
enoodle, line 49: What is your opinion on adding `PodStatusIndex`  field for each subgroup like the pod group has for faster calculation of these values?

romanbaron: Will be covered in a future PR

PR #351, opened 2025-07-31
Added SubGroups support in allocation_info.go and jobSolver

romanbaron:


pkg/scheduler/api/pod_info/pod_info.go
```diff
@@ -205,6 +205,7 @@ func (pi *PodInfo) Clone() *PodInfo {
 		Job:                  pi.Job,
 		Name:                 pi.Name,
 		Namespace:            pi.Namespace,
+		SubGroupName:         pi.SubGroupName,
```
omer-dayan, line 208: Do you think its correct to have this 2 direction pointing?

You have PodGroupInfo.SubGroup.PodInfos that points to PodInfo, and now you add PodInfo.SubGroupName that points to Subgroup

It feels like the data structure may not follow good practice.

omer-dayan: I see the only place you use it is in jobSolver `getPartialJobRepresentative`.

I feel like it could have been done in loop sort of like that:

```
func getPartialJobRepresentative() {
    podInfoToSubGroup = map[PodInfoID]SubGroupName{}

    for subGroup := range pgi.subGroups {
        for podInfo := range subGroup.podInfos {
            podInfoToSubGroup[podInfo.ID] = subGroup.Name
        }
    }

    subGroupToMinMember = map[SubGroupName]int{}
    for pendingPodInfo := range pendingPodInfos {
        subGroupToMinMember[podInfoToSubGroup[pendingPodInfo.ID]] += 1
    }
   ...
}
```

It will be o(n) when N is the number of PodInfos in PogGroup

omer-dayan: And if you really want to keep it o(1) you can store a map between PodInfo.ID to SubGroup in PodGroupInfo.

I will encapsulate all relations between PGI and PI inside PGI implementation

romanbaron: It can be implemented in different ways. Since pod references the relevant subgroup in a label, it seems intuitive to have an attribute in PodInfo with the subgroup name. Unlike SubGroupInfo/PodGroupInfo structs where tasks can be added or removed, this place remains permanent.
Let's discuss this one as well.

pkg/scheduler/api/podgroup_info/subgroup_info.go
```diff
@@ -55,3 +80,13 @@ func (sgi *SubGroupInfo) GetNumGatedTasks() int {
 	}
 	return numTasks
 }
+
+func (sgi *SubGroupInfo) GetNumPendingTasks() int {
```
itsomri, line 84: I think that we keep an index in podGroupInfo for indexing pods status - is there a specific reason not to do the same for subgroups?

romanbaron: It is something we will need to decide on, it is on my list. But not in this PR.

pkg/scheduler/api/pod_info/pod_info.go
romanbaron: @omer-dayan - let's continue this discussion and if we decide to change it I will do it across the repo in a future PR.

PR #356, opened 2025-08-02
predicate and nodeOrder for the topology plugin + topology result caching -  topology job filter PRs part 4

davidLif:


pkg/scheduler/framework/session_plugins.go
```diff
@@ -375,6 +375,20 @@ func (ssn *Session) NodeOrderFn(task *pod_info.PodInfo, node *node_info.NodeInfo
...
+func (ssn *Session) AddCleanAllocationAttemptCacheFn(fn api.CleanAllocationAttemptCacheFn) {
+	ssn.CleanAllocationAttemptCacheFns = append(ssn.CleanAllocationAttemptCacheFns, fn)
+}
+
+func (ssn *Session) CleanAllocationAttemptCache(job *podgroup_info.PodGroupInfo) {
+	for _, cleaner := range ssn.CleanAllocationAttemptCacheFns {
+		err := cleaner(job)
+		if err != nil {
+			log.InfraLogger.V(6).Infof(
+				"Failed to run CleanAllocationAttemptCache on job %s", job.Name)
```
SiorMeir, line 387: @davidLif Why only an INFO log? how severe is it to fail to clean the cache?

pkg/scheduler/plugins/scores/scores.go
```diff
@@ -8,6 +8,7 @@ const (
 	ResourceType   = 10
 	Availability   = 100
 	GpuSharing     = 1000
-	K8sPlugins     = 10000
-	NominatedNode  = 100000
+	Topology       = 10000
+	K8sPlugins     = 100000
+	NominatedNode  = 1000000
```
SiorMeir, line 13: @davidLif `NominatedNode` is increased here X10. 
Is this by design?

davidLif: Yes. The important think is to give a different values to each of the plugins according to the priroty between  them.

pkg/scheduler/framework/session_plugins.go
davidLif: You are right, this should be logged as a warning at the very least.

SiorMeir (approved):
Comments addressed, approving

PR #357, opened 2025-08-03
Added SubGroupOrder plugin

romanbaron:


pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -79,7 +79,7 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 func attemptToAllocateJob(ssn *framework.Session, stmt *framework.Statement, job *podgroup_info.PodGroupInfo) (allocated, pipelined bool) {
 	queue := ssn.Queues[job.Queue]
 
-	resReq := podgroup_info.GetTasksToAllocateInitResource(job, ssn.TaskOrderFn, true)
+	resReq := podgroup_info.GetTasksToAllocateInitResource(job, ssn.SubGroupOrderFn, ssn.TaskOrderFn, true)
```
itsomri, line 82: It's starting to look kinda weird that we inject those order functions to each podgroup in each invocation of this fn.. These are global to the session and are the same for each podgroup. Maybe there's a better way? Inject it at object creation? Or store it globally? let's think

pkg/scheduler/framework/session.go
```diff
@@ -63,6 +63,7 @@ type Session struct {
 	NodePreOrderFns                       []api.NodePreOrderFn
 	NodeOrderFns                          []api.NodeOrderFn
 	TaskOrderFns                          []common_info.CompareFn
```
itsomri: Lets revert the order of these 3.. Job, subgroup, task.. If we're already here

pkg/scheduler/plugins/subgrouporder/subgroup_order.go
```diff
@@ -0,0 +1,61 @@
...
+	rv := r.(*podgroup_info.SubGroupInfo)
+
+	lNumActiveTasks := lv.GetNumActiveAllocatedTasks()
+	rNumActiveTasks := rv.GetNumActiveAllocatedTasks()
+
+	// Prioritize SubGroup below minAvailable
+	lGangSatisfied := lNumActiveTasks >= int(lv.MinAvailable)
+	rGangSatisfied := rNumActiveTasks >= int(rv.MinAvailable)
+	if !lGangSatisfied && !rGangSatisfied {
+		return 0
```
itsomri: Let's use consts of lSubGroupPrioritized/rSubGroupPrioritized like we do for queues:
```Go
const (
	lQueuePrioritized   = -1
	rQueuePrioritized   = 1
	equalPrioritization = 0
)
```

pkg/scheduler/actions/allocate/allocate.go
romanbaron: We will get to it after subgroups are merged

Issue #340, opened 2025-07-23
NVIDIA runtime class in cluster


romanbaron:
### What you would like to be added?

Following this thread:
https://cloud-native.slack.com/archives/C08RKDPQPQU/p1753132066271239

KAI should explicitly state that KAI depends on the default runtime class to be `nvidia` for GPU sharing features.
This dependency affects the ability to run resource reservation pods where NVML library is used.

### Why is this needed?

1. Explain it in GPU sharing docs.
2. Explicitly set nvidia as the runtimeClass in resource reservation pod spec.

singh1203:
Hey @romanbaron, I would love to work on it, and after reading the thread, I understood that we need to document about default runtime class `nvidia`. Apart from that, what else are we looking for?
Thank you!

ManishaYogeshWaghmare:
resource reservation pod is not spinning up with Error while running the app: unable to initialize NVML: ERROR_LIBRARY_NOT_FOUND

romanbaron:
Thanks @singh1203 , your help is much appreciated 👏 

Resource reservation pods use NVML library in order to discover the device that will be used for GPU sharing.
The pod is created here:
https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/binder/binding/resourcereservation/resource_reservation.go#L438

I think that all we need to do is to explicitly set `nvidia` as runtime class in the pod spec.

The code that actually uses NVML library is located here:
https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/resourcereservation/discovery/discovery.go
And it is called from here:
https://github.com/NVIDIA/KAI-Scheduler/blob/main/cmd/resourcereservation/app/app.go#L61

In order to test it all you need to do is to run a GPU sharing pod (either fraction or gpu-memory) and make sure the reservation pod doesn't fail when the default runtime class is not set to `nvidia`.

Thanks for the assistance 🌟

romanbaron:
@singh1203 do you still plan to work on it?

singh1203:
@romanbaron Yes, I'm working on it. KubeCon India is around the corner, which is why I've been busy.

romanbaron:
Thanks a lot @singh1203 - your PR was merged 🚀

singh1203:
[...]

PR #360, opened 2025-08-03
Explicitly set runtimeClassName to `nvidia` for GPU reservation pods

singh1203:
Fixes - #340 

Reservation pods use the NVML library for GPU discovery, which requires the NVIDIA runtime.
Without explicitly setting RuntimeClassName, pods may fail if the cluster default is not nvidia.

This PR ensures reliable GPU sharing behavior by:

- [x] Explicitly setting runtimeClassName: "nvidia" in the resource reservation pod spec.
- [x] Updating GPU sharing documentation to include the requirement for the nvidia RuntimeClass.

### DIFF
--- a/docs/gpu-sharing/README.md
+++ b/docs/gpu-sharing/README.md
@@ -18,6 +18,11 @@ GPU sharing is disabled by default. To enable it, add the following flag to the 
 --set "global.gpuSharing=true"
 ```
 
+### RuntimeClass Requirement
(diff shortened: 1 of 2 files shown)


singh1203:
@romanbaron, I have tried to test this locally, on a non-GPU machine, here is the screenshot for reference 
Screenshot From 2025-08-03 23-05-52

Thank you!

PR #366, opened 2025-08-05
Few renames in topology_plugin_job_filtering.go

romanbaron:


pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
```diff
@@ -24,7 +24,8 @@ func (t *topologyPlugin) prePredicateFn(_ *pod_info.PodInfo, job *podgroup_info.
 		return err
 	}
 	if topologyTree == nil {
-		return nil
+		return fmt.Errorf("matching topology tree was not found for job %s, workload topology name: %s",
+			job.PodGroup.Name, job.PodGroup.Spec.TopologyConstraint.Topology)
```
davidLif: No. getJobTopology can return nill if no topology has been spesified on the job.

romanbaron: Reverted it.

itsomri (approved):

PR #372, opened 2025-08-06
Refactor subg

omer-dayan:


pkg/scheduler/api/cluster_info.go
```diff
@@ -94,15 +94,15 @@ func (ci ClusterInfo) String() string {
 		str = str + "PodGroupInfos:\n"
 		for _, job := range ci.PodGroupInfos {
 			str = str + fmt.Sprintf("\t Job(%s) name(%s) minAvailable(%v)\n",
-				job.UID, job.Name, job.MinAvailable)
+				job.UID, job.Name, job.GetDefaultMinAvailable())
 
-			for _, subGroup := range job.SubGroups {
+			for _, subGroup := range job.GetRealSubGroupInfo() {
```
romanbaron: `GetRealSubGroupInfo` might sound like there is a single SubGroup that is returned from this method.

pkg/scheduler/plugins/proportion/capacity_policy/capacity_policy_test.go
```diff
@@ -726,15 +744,17 @@ var _ = Describe("Capacity Policy Check", func() {
...
-								UID:       "task-a",
-								Job:       "job-a",
-								Name:      "task-a",
-								Namespace: "team-a",
-								Status:    pod_status.Pending,
-								ResReq:    resource_info.NewResourceRequirements(0, 1000, 0),
-							},
+						SubGroups: map[string]*podgroup_info.SubGroupInfo{
+							podgroup_info.DefaultSubGroup: podgroup_info.NewSubGroupInfo(podgroup_info.DefaultSubGroup, 0).WithPodInfos(map[common_info.PodID]*pod_info.PodInfo{
+								"task-a": {
```
romanbaron: I noticed that in many places you created a new instance of the default sub group with minAvailable of 0. Shouldn't it be equal to the number of tasks in the sub group?
I understand that also in the previous implementation it was 0 because it is the default value of int when not set. But, I would expect it to be greater than 0. wdyt?

omer-dayan: Yeah thats why I had it 0.

Changed it accordingly, in every test, to be equal to the amount of pods

omer-dayan: Except `job_order_by_queue_test.go` in which changing that make the tests fail and throw exception (In my opinion these tests are not doing what we think they do)

pkg/scheduler/api/cluster_info.go
omer-dayan: Changed to `GetActiveSubGroupInfos`.
However notice that we wont need such function in next PR when Default subGroup would be regular subGroup

PR #374, opened 2025-08-07
Allow reclaiming with lower utilization ratio

enoodle:
Currently it is possible for a reclaim to fail between two departments due to a third department requesting quota that it can't receive. To help remediate this issue this PR will allow queues to reclaim from other queues if their utilization ratio (allocation / fair share) is better for all resources.
The original intention for checking that the fair share is not over 1 was to prevent reclaim cycles where the reclaimed project will be able to reclaim the resources back and to create an infinite loop of reclaims and this test is a more relaxed version of this which will also achieve the same goal by always lowering the utilization ratio.

There is an action test for reclaim that show this issue and also unit test in reclaimable_test.go 

This change will allow the scheduler to find more reclaim scenarios at minimal performance cost.

### DIFF
--- a/docs/fairness/README.md
+++ b/docs/fairness/README.md
@@ -27,3 +27,16 @@ There are two main reclaim strategies:
 
 In both strategies, the scheduler ensures that the relative ordering is preserved: a queue that had the lowest utilisation ratio in its level before reclamation will still have the lowest ratio afterwards. Likewise, a queue that was below its quota will remain below its quota.
 The scheduler will prioritize the first strategy.
+
+### Reclaim Ratio Adjustment
+The utilization ratio comparison can be adjusted using the `reclaimerUtilizationMultiplier` plugin argument. This multiplier is applied to the reclaimer's utilization ratio before comparison:
+- Values > 1.0 make it harder for jobs to reclaim resources (more conservative)
+- Minimum value is 1.0 (standard comparison, default)
+- Values < 1.0 are not allowed and will be set to 1.0 - These values could cause infinite reclaim cycles that we want to avoid.
+
+Example configuration:
+```yaml
+pluginArguments:
+  proportion:
+    reclaimerUtilizationMultiplier: "1.2"  # Makes reclamation 20% more conservative
+```
--- a/docs/fairness/README.md
+++ b/docs/fairness/README.md
@@ -18,7 +18,7 @@ These two steps are repeated across all hierarchy levels until every leaf queue 
 ## Fair Share
 Once the fair share for each queue is calculated, it serves two primary purposes:
 1. Queue Order - Queues with a fair share further below their allocation will be prioritized for scheduling.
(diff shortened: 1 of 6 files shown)


docs/fairness/README.md
```diff
@@ -18,12 +18,25 @@ These two steps are repeated across all hierarchy levels until every leaf queue
 ## Fair Share
 Once the fair share for each queue is calculated, it serves two primary purposes:
 1. Queue Order - Queues with a fair share further below their allocation will be prioritized for scheduling.
-2. Reclaim action - If scheduling cannot be performed due to limited resources in the cluster, the scheduler will evict workloads from queues that have exceeded their fair share, giving priority to queues that are below their fair share. For more details, refer to the reclaim strategies.
+2. Reclaim action – When reclamation is required, the scheduler compares the **utilisation ratio** (`Allocated / FairShare`) of queues that share the same parent. A queue can only reclaim resources if, **after** the transfer, its utilisation ratio remains lower than that of every sibling queue. For more details see the reclaim strategies.
```
itsomri: I think "utilization" can be confusing. We need a term for the opposite of starvation - maybe "Saturation"?

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -54,19 +55,32 @@ type proportionPlugin struct {
...
+	reclaimablePlugin              *rec.Reclaimable
+	isInferencePreemptible         bool
+	allowConsolidatingReclaim      bool
+	reclaimerUtilizationMultiplier float64
 }
 
 func New(arguments map[string]string) framework.Plugin {
+	multiplier := 1.0
+	if val, exists := arguments["reclaimerUtilizationMultiplier"]; exists {
+		if m, err := strconv.ParseFloat(val, 64); err == nil {
```
itsomri, line 70: Let's log the error if we get it - it probably means that the admin made a typo
Plus - it would be easier if this argument extraction would be in a separate method - you could make much cleaner

pkg/scheduler/plugins/proportion/reclaimable/reclaimable.go
```diff
@@ -158,14 +188,47 @@ func (r *Reclaimable) reclaimingQueuesRemainWithinBoundaries(
 	return true
 }
 
-func (r *Reclaimable) getTopQueues(queues map[common_info.QueueID]*rs.QueueAttributes) map[common_info.QueueID]bool {
-	topQueues := make(map[common_info.QueueID]bool)
-	for queueID, queue := range queues {
-		if queue.IsTopQueue() {
-			topQueues[queueID] = true
+// isFairShareUtilizationLowerPerResource returns true if for every resource the
```
itsomri: Let's explain in the docstring that the ration should be lower when considering the `utilizationMultiplier`

itsomri (approved):

docs/fairness/README.md
enoodle: :+1: I will use that

pkg/scheduler/plugins/proportion/proportion.go
enoodle: :+1:

pkg/scheduler/plugins/proportion/reclaimable/reclaimable.go
enoodle: :+1:

itsomri (approved):

PR #377, opened 2025-08-07
Adjusted grove podGrouper plugin to use subgroups

romanbaron:


pkg/podgrouper/podgroup/handler.go
```diff
@@ -14,6 +16,10 @@ import (
 	schedulingv2alpha2 "github.com/NVIDIA/KAI-scheduler/pkg/apis/scheduling/v2alpha2"
 )
 
+const (
+	subGroupLabelKey = "kai.scheduler/subgroup-name"
```
enoodle: Maybe we want to set this constant somewhere we can also import it in e2e/env tests and also use the same one from the scheduler?
maybe it should be in `pkg/common/constants/constants.go` and also `pkg/scheduler/api/pod_info/pod_info.go` should point to it.

pkg/podgrouper/podgroup/handler.go
```diff
@@ -112,3 +131,24 @@ func (h *Handler) createPodGroupForMetadata(podGroupMetadata Metadata) *scheduli
...
+			pod := &v1.Pod{}
+			err := h.client.Get(ctx, *podRef, pod)
+			if err != nil {
+				return err
+			}
+
+			labeledPod := pod.DeepCopy()
+			labeledPod.Labels[subGroupLabelKey] = subGroup.Name
+
+			err = h.client.Patch(ctx, labeledPod, client.MergeFrom(pod))
```
enoodle, line 144: In large jobs this will make the pod grouper patch many objects which might take some time. Maybe we will want to run it concurrently and handle errors with retries . You don't have to do it in this PR but I feel like this will be a point of failure with large jobs.

pkg/podgrouper/podgroup/handler.go
romanbaron: Sure. moved to `pkg/common/constants`

pkg/podgrouper/podgroup/handler.go
romanbaron: Well, maybe but not in this PR.

enoodle (approved):

PR #381, opened 2025-08-10
Added subgroups support to stale gang eviction action

romanbaron:


pkg/scheduler/api/podgroup_info/job_info.go
```diff
@@ -372,7 +372,17 @@ func (pgi *PodGroupInfo) IsStale() bool {
 	}
 
 	activeUsedTasks := int32(pgi.GetNumActiveUsedTasks())
-	return activeUsedTasks > 0 && activeUsedTasks < pgi.MinAvailable
+	if activeUsedTasks > 0 {
```
enoodle, line 375: Maybe you can also add tests to this function?

romanbaron: Done

PR #383, opened 2025-08-11
fix: humanize message and add tests

SiorMeir:
A small bug that was discovered in messages outputted by KAI-scheduler in the scenario where a non-preemptible job requests more resources than available.

### DIFF
--- a/pkg/scheduler/api/unschedule_info_test.go
+++ b/pkg/scheduler/api/unschedule_info_test.go
@@ -6,9 +6,7 @@ package api
 import (
 	"testing"
 
-	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/pod_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/podgroup_info"
-	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/resource_info"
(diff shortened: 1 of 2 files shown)


pkg/scheduler/api/unschedule_info_test.go
```diff
@@ -0,0 +1,109 @@
...
+			used := 50000000000.0      // 50 GB in bytes
+			requiredResources := &podgroup_info.JobRequirement{
+				GPU:      0,
+				MilliCPU: 0,
+				Memory:   200000000000.0, // 200 GB in bytes
+			}
+
+			message := GetBuildOverCapacityMessageForQueue(queueName, resourceName, deserved, used, requiredResources)
+
+			expectedPrefix := "Non-preemptible workload is over quota. "
```
itsomri: I don't think that it's necessary to enforce the entire message - just validate that you see "299 GB memory", "quota is 100 GB", "50 GB are already allocated". The rest of the message is not what we test for, correct me if I'm wrong.

SiorMeir: I agree and will fix.
However, this will not affect the fact that if we change the warning message we'll need to revisit this test as well

enoodle (approved):

PR #384, opened 2025-08-11
Set PodGroup before adding tasks to job info

romanbaron:


SiorMeir (approved):
synced with @romanbaron on this. Seems ok

PR #386, opened 2025-08-11
Time aware fairness client plugins

itsomri:


pkg/scheduler/cache/usagedb/hub.go
```diff
@@ -0,0 +1,38 @@
...
+func GetClient(config *api.UsageDBConfig) (api.Interface, error) {
+	if config == nil {
+		return nil, nil
+	}
+
+	if config.ClientType == "" {
+		return nil, fmt.Errorf("client type cannot be empty")
+	}
+
+	clientMap := map[string]GetClientFn{
```
enoodle: This can be a global that is not re-created on every function call. It will help it stand out more and be clear that it is a "static" map of names to classes.

pkg/scheduler/cache/usagedb/api/interface.go
```diff
@@ -0,0 +1,14 @@
...
+package api
+
+import "github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/queue_info"
+
+type Interface interface {
+	GetResourceUsage() (*queue_info.ClusterUsage, error)
+}
+type UsageDBConfig struct {
+	ClientType       string `yaml:"clientType" json:"clientType"`
+	ConnectionString string `yaml:"connectionString" json:"connectionString"`
```
enoodle: Maybe the connection string will have to include a password, so we might want to pass it in a secret or something and not in the configmap.

enoodle (approved):

PR #389, opened 2025-08-13
Refactor Default SubGroup to contain only non SubGrouped Pods

omer-dayan:


pkg/scheduler/actions/allocate/allocate_subgroups_test.go
romanbaron, line 1: Something is not right here, the original order is the one we follow in our other go files

PR #396, opened 2025-08-14
Bug fix: Add the child domains properly to the parent when calculatin…

davidLif:
…g the topology tree in the scheduler


pkg/scheduler/actions/common/allocate.go
```diff
@@ -26,6 +26,8 @@ func AllocateJob(ssn *framework.Session, stmt *framework.Statement, nodes []*nod
 		}
 		return false
 	}
+
```
SiorMeir, line 29: Just a question for understanding - defer will make this execute last, so the line on which this statement is present shouldn't matter, no?

itsomri: It does matter in a few ways:
1. Most important IMO, readability - you want it to make sense in context
2. If there are a few defer statements, they will be called in LIFO order
3. If the function panics or return before the defer statement in the code, it will have no effect

davidLif: As @itsomri explained, it does matter. In this context, if the function returns because the job is over capacity (IsJobOverQueueCapacityFn), we don't need to call the CleanAllocationAttemptCache function.

SiorMeir: Got it.
Feels like a delicate point, glad I asked 😄

PR #403, opened 2025-08-17
Fixed getNumOfTasksToAllocatePerSubGroup so it will account tasks wit…

romanbaron:
…h virtualStatus


pkg/scheduler/api/podgroup_info/allocation_info.go
```diff
@@ -51,7 +51,7 @@ func GetTasksToAllocate(
 
 	} else {
 		taskPriorityQueue := getTasksPriorityQueue(podGroupInfo, taskOrderFn, isRealAllocation)
-		maxNumOfTasksToAllocate := getNumOfTasksToAllocate(podGroupInfo)
```
omer-dayan, line 54: This has no meaning removing the Of right?

I see now we have 
GetNumTasksToAllocate
GetNumOfTasksToAllocatePerSubgroup

I think either both would have "Of" or none

romanbaron: Good point, I will remove since it doesn't add any meaning and makes the name shorter

PR #404, opened 2025-08-17
Added leader election in all deployments and a global helm value that…

romanbaron:
… controls it during installaiton

### DIFF
--- a/cmd/admission/app/app.go
+++ b/cmd/admission/app/app.go
@@ -85,7 +85,7 @@ func New() (*App, error) {
 		}),
 		HealthProbeBindAddress: options.ProbeAddr,
 		LeaderElection:         options.EnableLeaderElection,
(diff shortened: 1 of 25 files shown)


cmd/admission/app/app.go
```diff
@@ -85,7 +85,7 @@ func New() (*App, error) {
 		}),
 		HealthProbeBindAddress: options.ProbeAddr,
 		LeaderElection:         options.EnableLeaderElection,
-		LeaderElectionID:       "2ad35f9c.kai.scheduler",
+		LeaderElectionID:       "ojptxr84.kai.scheduler",
```
enoodle, line 88: how did you set these?

romanbaron: It is just a random string as a prefix

PR #412, opened 2025-08-20
Basic topology e2e test

davidLif:


test/e2e/suites/allocate/topology/topology_test.go
```diff
@@ -0,0 +1,223 @@
...
+			scheduledRacks := map[string][]string{}
+			for _, pod := range podList.Items {
+				podRack := testTopologyData.TopologyNodes[pod.Spec.NodeName].Labels[rd.TestRackLabelKey]
+				scheduledRacks[podRack] = append(scheduledRacks[podRack], pod.Name)
+			}
+
+			Expect(len(scheduledRacks)).To(Equal(1), "Expected all pods scheduled to one rack, got %v", scheduledRacks)
+		})
+
+		It("preferred only - rack level", func(ctx context.Context) {
```
SiorMeir, line 101: I can see two consecutive tests with the same name.
Maybe add a bit more context in the name?

SiorMeir: In addition, I'm having trouble understanding what we're testing here in a glance, but it might be just me.
Maybe add a docstring or something similar to explain the tests?

davidLif: These are actually 2 different names "preferred only - rack level" and "required only - rack level".
I don't really like to give long names to tests, but these are setting required/prefferd topology constraint to a workload and we wait to see that both nodes have been scheduled to the same rack.

PR #417, opened 2025-08-21
fix: GPU memory calculation using incorrect unit conversion

### LINKED ISSUE #416: GPU Memory Validation Bug: Pods can exceed node's nvidia.com/gpu.memory limit
### What happened?

KAI-Scheduler allows pods to be scheduled even when their total GPU memory requests exceed the node's actual GPU memory capacity as reported by `nvidia.com/gpu.memory` label.
## Steps to Reproduce

1. Deploy the following configuration:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: gpu-sharing-7
  namespace: default
spec:
  replicas: 24
  selector:
    matchLabels:
      app: gpu-sharing
  template:
    metadata:
      labels:
        app: gpu-sharing
        kai.scheduler/queue: test
      annotations:
        gpu-memory: "2000" # in MiB
    spec:
      schedulerName: kai-scheduler
      containers:
        - name: ubuntu
          image: nginx
          args: ["sleep", "infinity"]
```

2. Apply the deployment: `kubectl apply -f gpu-sharing-deployment.yaml`

## Expected Behavior
- Total GPU memory requested: 24 replicas × 2000 MiB = 48000 MiB (48GB)
- Node GPU memory capacity: 46068 MiB (46GB)
- Expected: Only 23 pods should be scheduled (23 × 2000 = 46000 MiB), 1 pod should remain pending
- Expected: Scheduler should reject pods that would cause memory overcommitment

## Actual Behavior
- All 24 pods are scheduled successfully
- No pods remain in pending state
- Total allocated GPU memory (48GB) exceeds node capacity (46GB)

### Environment

- **KAI-Scheduler Version**: v0.8.1
- **GPU Hardware**: NVIDIA L40S / bare-metal
- **Node GPU Memory**: `nvidia.com/gpu.memory=46068` (46GB)

eveningcafe:
Fixes #416

The GPU memory validation was incorrectly converting MiB to MB and back, causing the scheduler to use inflated memory values (48300 MB instead of 46068 MiB) when checking shared GPU allocations. This allowed pods to be scheduled beyond the actual GPU memory capacity.

Changes:
- Use original MiB value from nvidia.com/gpu.memory node label directly
- Remove unnecessary MiB->MB->MiB conversion that caused precision loss
- Maintain the flooring logic for memory alignment but in MiB units

### DIFF
--- a/pkg/scheduler/api/node_info/node_info.go
+++ b/pkg/scheduler/api/node_info/node_info.go
@@ -27,7 +27,7 @@ import (
 
 	"go.uber.org/multierr"
 	"golang.org/x/exp/maps"
-	v1 "k8s.io/api/core/v1"	
+	v1 "k8s.io/api/core/v1"
 
 	commonconstants "github.com/NVIDIA/KAI-scheduler/pkg/common/constants"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/common_info"
@@ -646,7 +646,6 @@ func checkGpuMemoryIsInMib(gpuMemoryValue int64) bool {
 func convertBytesToMib(gpuMemoryValue int64) int64 {
 	return gpuMemoryValue / BitToMib
 }
-
 
 func (ni *NodeInfo) IsCPUOnlyNode() bool {
 	if ni.IsMIGEnabled() {
--- a/pkg/scheduler/api/node_info/node_info_test.go
+++ b/pkg/scheduler/api/node_info/node_info_test.go
@@ -452,7 +452,6 @@ func TestAddRemovePods(t *testing.T) {
 	}
 }
 
-
 type allocatableTestData struct {
 	node                    *v1.Node
 	podsResources           []v1.ResourceList


pkg/scheduler/api/node_info/node_info.go
```diff
@@ -637,8 +637,9 @@ func getNodeGpuMemory(node *v1.Node) (int64, bool) {
 		gpuMemoryLabelValue = convertBytesToMib(gpuMemoryLabelValue)
 	}
 
-	gpuMemoryInMb := convertMibToMb(gpuMemoryLabelValue)
```
enoodle, line 640: `convertMibToMb` is not used anymore, delete it as well (probably linter will fail anyway)

pkg/scheduler/api/node_info/node_info.go
```diff
@@ -637,8 +637,9 @@ func getNodeGpuMemory(node *v1.Node) (int64, bool) {
...
-	return gpuMemoryInMb - (gpuMemoryInMb % 100), true // Floor the memory count to make sure its divided by 100 so there will not be 2 jobs that get same bytes
+	// Use the original MiB value from the node label, don't convert to MB
+	// The flooring should be done in MiB, not MB to preserve accuracy
```
enoodle: :+1:
This comment only makes sense in the PR context, you can remove it from the final version.

enoodle (commented):
Amazing, thank you!

pkg/scheduler/api/node_info/node_info.go
eveningcafe: I just create a commit base on your review

enoodle:
@eveningcafe There are some errors when tyring to build, and then a few tests are failing after the change. Can you please look into that?

enoodle:
@eveningcafe This is still failing. Maybe you can tell Claude to run `make validate && make test` and make sure it passes without issues. I think there is also one test that will fail because it has hard coded numbers that will fail this.

enoodle:
I think you just need to run `go fmt` on your code now

eveningcafe:
> ### Merging this branch will **decrease** overall coverage
> Impacted Packages	Coverage Δ	🤖
> github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/node_info	66.74% (**-0.31%**)	👎
> Coverage by file
> ### Changed files (no unit tests)
> Changed File	Coverage Δ	Total	Covered	Missed	🤖
> github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/node_info/node_info.go	63.01% (**-0.50%**)	292 (-4)	184 (-4)	108	👎
> _Please note that the "Total", "Covered", and "Missed" counts above refer to _**code statements**_ instead of lines of code. The value in brackets refers to the test coverage of that file in the old version of the code._
> 
> ### Changed unit test files
> * github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/node_info/node_info_test.go

The convertMibToMb function was previously being used and covered by tests (contributing to the 67% coverage). My PR changes made this function unused by fixing the GPU  memory calculation bug, so the covered statements reduced by 4 lines where it was called, dropping coverage to 66.74%. This is not a concern.

enoodle:
Its OK, the coverage is just acting as FYI message to help notice big gaps.

enoodle (approved):

PR #422, opened 2025-08-22
Fix: incorrect scheduling decision and calculation when using MIG

hello2mao:
This PR tries to fix two problems we found:

1. If a MIG pod is not scheduled by Kai, its `ResourceRequestType` is not set to `MigInstance`. This results in incorrect idle resource calculation for the node, which in turn leads to inaccurate scheduling decisions.
2. When calculating resource proportions, the MIG resource of a node is counted twice.

### DIFF
--- a/pkg/scheduler/cache/cluster_info/cluster_info_test.go
+++ b/pkg/scheduler/cache/cluster_info/cluster_info_test.go
@@ -278,51 +278,15 @@ func TestSnapshotNodes(t *testing.T) {
 			Phase: v1core.PodRunning,
 		},
 	}
-	exampleMIGPod := &v1core.Pod{
-		ObjectMeta: v1.ObjectMeta{
-			Name: "mig-pod-1",
-		},
-		Spec: v1core.PodSpec{
-			NodeName: "node-1",
-			Containers: []v1core.Container{
-				{
-					Resources: v1core.ResourceRequirements{
-						Requests: v1core.ResourceList{
-							"cpu":                   resource.MustParse("2"),
-							"nvidia.com/mig-1g.5gb": resource.MustParse("2"),
-						},
-					},
-				},
-			},
-		},
-		Status: v1core.PodStatus{
-			Phase: v1core.PodRunning,
-		},
+	exampleMIGPod := examplePod.DeepCopy()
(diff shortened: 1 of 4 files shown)


pkg/scheduler/cache/cluster_info/cluster_info_test.go
```diff
@@ -278,6 +278,51 @@ func TestSnapshotNodes(t *testing.T) {
...
+						},
+					},
+				},
+			},
+		},
+		Status: v1core.PodStatus{
+			Phase: v1core.PodRunning,
+		},
+	}
+	exampleMIGPodWithPG := &v1core.Pod{
```
enoodle: Nit: Can this just DeepCopythe previous pod definition  and add the annotation  instead of re-writing all of it?
Also the 1st `exampleMIGPod` can do that from the pod before it.

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -245,7 +245,7 @@ func getNodeResources(ssn *framework.Session, node *node_info.NodeInfo) rs.Resou
 	if shouldIgnoreGPUs {
 		nodeResource.Add(rs.NewResourceQuantities(node.Allocatable.Cpu(), node.Allocatable.Memory(), 0))
 	} else {
-		nodeResource.Add(utils.QuantifyResource(node.Allocatable))
+		nodeResource.Add(rs.NewResourceQuantities(node.Allocatable.Cpu(), node.Allocatable.Memory(), node.Allocatable.GPUs()))
```
enoodle: Good catch.
I think it is better to remove lines 249-260 here instead of this change. This calculation should only be done in one place, and it is not the proportion plugin IMO

pkg/scheduler/cache/cluster_info/cluster_info_test.go
hello2mao: @enoodle Appreciate the suggestions — I’ve made the changes accordingly.

pkg/scheduler/plugins/proportion/proportion.go
hello2mao: Yes, you're right.

enoodle (approved):

PR #435, opened 2025-08-25
Time aware fairness usage: prometheus client

itsomri:


pkg/scheduler/cache/usagedb/api/defaults.go
```diff
@@ -0,0 +1,87 @@
...
+
+package api
+
+import "time"
+
+func (up *UsageParams) SetDefaults() {
+	if up.HalfLifePeriod == nil {
+		// noop: disabled by default
+	}
+	if up.WindowSize == nil {
```
enoodle, line 12: Use the "GetXOrDefault" functions here. Otherwise this logic is duplicated

pkg/scheduler/cache/usagedb/prometheus/prometheus.go
```diff
@@ -0,0 +1,152 @@
...
+		queryResolution:      queryResolution,
+		allocationMetricsMap: allocationMetricsMap,
+	}, nil
+}
+
+func (p *PrometheusClient) GetResourceUsage() (*queue_info.ClusterUsage, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), p.usageQueryTimeout)
+	defer cancel()
+
+	usage := queue_info.NewClusterUsage()
```
enoodle, line 77: Maybe we can add a `SetResourceUsage(resourceType, usage)` method the Queues field and then this function will loop on the resource types. This will be helpful in the future when more resource types are added and also now this will be less repetetive.

pkg/scheduler/cache/usagedb/api/defaults.go
itsomri: I removed the GetWindowTypeOrDefault: I ended up going for the singe `SetDefaults` fn for the regular usagedb params (window type, size, half life), using the same pattern we use in engine config. It keeps the defaults of the common params concentrated and not dependent on the plugins..
For plugin-specific params (ExtraParams) I did the GetXOrDefault because it's unstructured and to keep it unopinionated - let the plugin decide for itself what the defaults are

pkg/scheduler/cache/usagedb/prometheus/prometheus.go
itsomri: I changed QueueUsage to a map[ResourceName]float64

enoodle (approved):

PR #443, opened 2025-08-26
Not using summerized MinAvailable for PodGroup, but look per SubGroup

omer-dayan:


pkg/scheduler/plugins/elastic/elastic.go
```diff
@@ -53,4 +53,24 @@ func JobOrderFn(l, r interface{}) int {
...
+			allSubgroupsAboveMinAvailable = false
+		}
+		if allocatedCount == subGroup.GetMinAvailable() {
+			allSubgroupsAboveMinAvailable = false
+		} else {
+			allSubgroupsExactlyMinAvailable = false
+		}
+	}
+	return atLeastOneSubgroupUnderMinAvailable, allSubgroupsAboveMinAvailable, allSubgroupsExactlyMinAvailable
+}
```
omer-dayan, line 68: In this commit we generalize the use case of elastic comparison, but not handled all cases.

All previous cases are handled the same, meaning not breaking changes compared to the days when SubGroup were not exists - As the behavior of no SubGroups is equal to single default SubGroup.

The unhandled behavior is when we have multiple SubGroups in both jobs.
In each one subGroup is above MinMember and second is under MinMember.
Which job is first?
Not handled

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -177,19 +177,34 @@ func (pp *proportionPlugin) getVictimResources(victim *api.VictimInfo) []*resour
...
+
+		// Handle case where minAvailable is greater than or equal to the number of tasks
+		if subGroup.GetMinAvailable() >= int32(len(subGroupTasks)) {
+			// All tasks are considered core tasks, no elastic tasks
+			coreTasks = append(coreTasks, subGroupTasks...)
+			continue
+		}
+
+		coreTasks = append(coreTasks, subGroupTasks[:subGroup.GetMinAvailable()]...)
+		elasticTasks = append(elasticTasks, subGroupTasks[subGroup.GetMinAvailable():]...)
```
omer-dayan, line 206: In here we generalize the case of no SubGroups so no breaking changes when not using SubGroups.

But when multiple SubGroups we return all elastic tasks unioned accross all subgroups and all core tasks unioned accross all subgroups.

Alternativly we could return list of list for elastic tasks and list of list for core tasks,

pkg/scheduler/api/podgroup_info/subgroup_info.go
```diff
@@ -62,6 +62,10 @@ func (sgi *SubGroupInfo) IsGangSatisfied() bool {
 	return numActiveTasks >= int(sgi.minAvailable)
 }
 
+func (sgi *SubGroupInfo) IsElastic() bool {
+	return sgi.GetMinAvailable() < int32(len(sgi.GetPodInfos()))
+}
+
```
romanbaron, line 123: Please add tests for this method.

pkg/scheduler/plugins/elastic/elastic.go
```diff
@@ -29,22 +29,22 @@ func JobOrderFn(l, r interface{}) int {
 	lv := l.(*podgroup_info.PodGroupInfo)
 	rv := r.(*podgroup_info.PodGroupInfo)
 
-	lvAllocatedCount := int32(lv.GetActiveAllocatedTasksCount())
-	rvAllocatedCount := int32(rv.GetActiveAllocatedTasksCount())
+	lvAtLeastOneSubgroupUnderMinAvailable, lvAllSubgroupsAboveMinAvailable, lvAllSubgroupsExactlyMinAvailable := minAvailableState(lv)
```
romanbaron, line 32: maybe `lvGangSatisfied`?

pkg/scheduler/actions/common/solvers/job_solver.go
```diff
@@ -122,11 +122,12 @@ func (s *JobSolver) solvePartialJob(ssn *framework.Session, state *solvingState,
 func getPartialJobRepresentative(
 	job *podgroup_info.PodGroupInfo, pendingTasks []*pod_info.PodInfo) *podgroup_info.PodGroupInfo {
 	jobRepresentative := job.CloneWithTasks(pendingTasks)
-	jobRepresentative.SetDefaultMinAvailable(int32(len(pendingTasks)))
 	subGroupsMinAvailable := map[string]int{}
 	for _, pendingTask := range pendingTasks {
 		if _, found := jobRepresentative.SubGroups[pendingTask.SubGroupName]; found {
 			subGroupsMinAvailable[pendingTask.SubGroupName] += 1
+		} else {
+			subGroupsMinAvailable[podgroup_info.DefaultSubGroup] += 1
```
romanbaron: This part should probably be adjusted to our latest design where default subgroup is used when no explicit subgroups are defined.

pkg/scheduler/plugins/elastic/elastic.go
romanbaron: What do you think about making this part more general by reducing subgroup related terminology.
For example:
```lvBelowMinAvailable, lvAboveMinAvailable, lvExactlyAtMinAvailable := minAvailableState(lv)```
I think that subgroups is an implementation details in the scope of this part of the code

File: pkg/scheduler/plugins/elastic/elastic.go [hunk 6]
```diff
@@ -53,4 +53,24 @@ func JobOrderFn(l, r interface{}) int {
 	return 0
 }
 
+func minAvailableState(pgi *podgroup_info.PodGroupInfo) (bool, bool, bool) {
+	atLeastOneSubgroupUnderMinAvailable := false
+	allSubgroupsAboveMinAvailable := true
+	allSubgroupsExactlyMinAvailable := true
+
+	for _, subGroup := range pgi.SubGroups {
+		allocatedCount := int32(subGroup.GetNumActiveAllocatedTasks())
```
romanbaron:
```suggestion
		numAllocatedTasks := int32(subGroup.GetNumActiveAllocatedTasks())
```

pkg/scheduler/plugins/elastic/elastic.go
```diff
@@ -53,4 +53,24 @@ func JobOrderFn(l, r interface{}) int {
 	return 0
 }
 
+func minAvailableState(pgi *podgroup_info.PodGroupInfo) (bool, bool, bool) {
+	atLeastOneSubgroupUnderMinAvailable := false
+	allSubgroupsAboveMinAvailable := true
```
romanbaron: These two variables do the same thing, they are defined together and their value is changed together. I think you don't really need them (especially not both). You can also return in line 64 when you find the first subgroup that is below minAvailable instead of iterating through all of them.

pkg/scheduler/api/podgroup_info/subgroup_info_test.go
```diff
@@ -332,6 +332,25 @@ func TestGetNumPendingTasks(t *testing.T) {
...
+		{UID: "5", Status: pod_status.Pending},
+	}
+
+	for _, pod := range pods {
+		sgi.AssignTask(pod)
+	}
+
+	if got := sgi.IsElastic(); !got {
+		t.Errorf("IsElastic() = %v, want %v", got, true)
+	}
```
romanbaron: Let's also add at least one case where the returned value is false.

pkg/scheduler/plugins/elastic/elastic.go
```diff
@@ -53,4 +53,24 @@ func JobOrderFn(l, r interface{}) int {
...
+			atLeastOneSubgroupUnderMinAvailable = true
+			allSubgroupsAboveMinAvailable = false
+		}
+		if numAllocatedTasks == subGroup.GetMinAvailable() {
+			allSubgroupsAboveMinAvailable = false
+		} else {
+			allSubgroupsExactlyMinAvailable = false
+		}
+	}
+	return atLeastOneSubgroupUnderMinAvailable, allSubgroupsAboveMinAvailable, allSubgroupsExactlyMinAvailable
```
romanbaron:
```suggestion
	underMinAvailable := false
	aboveMinAvailable := true
	exactlyAtMinAvailable := true

	for _, subGroup := range pgi.SubGroups {
		numAllocatedTasks := int32(subGroup.GetNumActiveAllocatedTasks())
		if numAllocatedTasks < subGroup.GetMinAvailable() {
			underMinAvailable = true
			aboveMinAvailable = false
		}
		if numAllocatedTasks == subGroup.GetMinAvailable() {
			aboveMinAvailable = false
		} else {
			exactlyAtMinAvailable = false
		}
	}
	return underMinAvailable, aboveMinAvailable, exactlyAtMinAvailable
}
```

romanbaron (approved):

romanbaron (approved):

PR #449, opened 2025-08-28
Add scheduler topology envtest

davidLif:
1- Add scheduler topology envtest
2- Bug fix: filter nodes that doesn't have the required topology labels

### DIFF
--- a/pkg/env-tests/scheduler_test.go
+++ b/pkg/env-tests/scheduler_test.go
@@ -449,12 +449,24 @@ var _ = Describe("Scheduler", Ordered, func() {
 		})
 
 		It("Schedule pods with rack topology constraints", func(ctx context.Context) {
+
+			// schedule a single gpu pod outside of the topology to try and "pull" the topology constraint workload pods outside of a valid rack
+			binPackingPullPod := CreatePodObject(testNamespace.Name, "bin-packing-pull-pod", corev1.ResourceRequirements{
+				Limits: corev1.ResourceList{constants.GpuResource: resource.MustParse("1")},
+			})
+			binPackingPullPod.Spec.NodeSelector = map[string]string{
+				hostnameLabelKey: "test-node",
+			}
+			Expect(ctrlClient.Create(ctx, binPackingPullPod)).To(Succeed(), "Failed to create bin-packing-pull-pod")
+			err := GroupPods(ctx, ctrlClient, podGroupConfig{queueName: testQueue.Name, podgroupName: "bin-packing-pull-podgroup", minMember: 1}, []*corev1.Pod{binPackingPullPod})
+			Expect(err).NotTo(HaveOccurred(), "Failed tocreate a pod group for bin-packing-pull-pod")
+
 			singlePodResourceRequirements := corev1.ResourceRequirements{
 				Limits: corev1.ResourceList{
-					constants.GpuResource: resource.MustParse(fmt.Sprintf("%d", gpusPerNode)),
+					constants.GpuResource: resource.MustParse(fmt.Sprintf("%d", gpusPerNode-1)),
 				},
 				Requests: corev1.ResourceList{
-					constants.GpuResource: resource.MustParse(fmt.Sprintf("%d", gpusPerNode)),
+					constants.GpuResource: resource.MustParse(fmt.Sprintf("%d", gpusPerNode-1)),
 				},
 			}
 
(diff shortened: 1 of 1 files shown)


pkg/env-tests/scheduler_test.go
```diff
@@ -338,4 +369,132 @@ var _ = Describe("Scheduler", Ordered, func() {
...
+
+			nodeToRack := map[int]string{}
+			for i := range numNodes {
+				nodeToRack[i] = fmt.Sprintf("%d", i%numRacks)
+			}
+			rand.Shuffle(len(nodeToRack), func(i, j int) {
+				nodeToRack[i], nodeToRack[j] = nodeToRack[j], nodeToRack[i]
+			})
+
+			for i := 0; i < numNodes; i++ {
```
itsomri:
```suggestion
			for i := range numNodes {
```

pkg/env-tests/scheduler_test.go
```diff
@@ -338,4 +369,132 @@ var _ = Describe("Scheduler", Ordered, func() {
...
+				topologyNode.Labels[rackLabelKey] = fmt.Sprintf("clusterUuid.%s", nodeToRack[i])
+				topologyNode.Labels[hostnameLabelKey] = topologyNode.Name
+				nodeObj := CreateNodeObject(ctx, ctrlClient, topologyNode)
+				Expect(ctrlClient.Create(ctx, nodeObj)).To(Succeed(), "Failed to create topology test node")
+				topologyNodes = append(topologyNodes, nodeObj)
+			}
+
+			topologyTree = &kueuev1alpha1.Topology{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: topologyName,
```
itsomri, line 413: Add label `app=engine-e2e`. This helps when the tests' cleanup fails for any reason.

pkg/env-tests/scheduler_test.go
```diff
@@ -338,4 +369,132 @@ var _ = Describe("Scheduler", Ordered, func() {
...
+					constants.GpuResource: resource.MustParse(fmt.Sprintf("%d", gpusPerNode)),
+				},
+				Requests: corev1.ResourceList{
+					constants.GpuResource: resource.MustParse(fmt.Sprintf("%d", gpusPerNode)),
+				},
+			}
+
+			numWorkloadPods := 4
+			workloadPods := []*corev1.Pod{}
+			for i := 0; i < numWorkloadPods; i++ {
```
itsomri:
```suggestion
			for i := range numWorkloadPods {
```

pkg/env-tests/scheduler_test.go
```diff
@@ -338,4 +369,132 @@ var _ = Describe("Scheduler", Ordered, func() {
...
+				To(Succeed(), "Failed to list pods")
+
+			Expect(len(pods.Items)).To(Equal(numWorkloadPods), "Expected %d pods to be created in the test", numWorkloadPods)
+
+			scheduledRacks := map[string]int{}
+			for _, pod := range pods.Items {
+				scheduledRacks[pod.Labels[rackLabelKey]]++
+			}
+
+			Expect(scheduledRacks).To(HaveLen(1), "Expected all pods to be scheduled on the same rack. Got %v", scheduledRacks)
```
itsomri, line 509: But is there any reason they won't? Even without topology?
Feels like the tests should interfere, like somehow mess with the scoring of nodes, so the same scenario will fail without the topology.

enoodle: Maybe also a test with preferred constraints and one with both on different levels?

pkg/env-tests/scheduler_test.go
enoodle: But this is not e2e, so it will not stay anywhere anyway.

davidLif: You can see that we don't add this label or similar one on the other objects in the envtests

pkg/env-tests/scheduler_test.go
davidLif: The rack nodes are chosen randomly. This means that without the topology requirement, there is a good chance that at least one of the pods will be scheduled on a node that belongs to the other rack.

To increase the chance for a failure in case of a bug, I added a ginkgo parameter for the e2e to re-run the test (with different racks and node labels) several times. Do you think I should do the same with this test?

pkg/env-tests/scheduler_test.go
itsomri: Oh right, sorry

pkg/env-tests/scheduler_test.go
itsomri: First, I think the tests should always be deterministic if possible - why not swap specific nodes?
Second, re-running the test just seems like a waste
What if you put 1 gpu-job on a node outside the rack so it will "pull" the pods due to bin packing?

davidLif: I implemented @itsomri suggestion

PR #451, opened 2025-08-30
move crds into helm chart

enoodle:
To simplify creating objects from those CRDs by the helm chart, we will add them to the chart which means that they will be created when the chart is first installed.
The crd-upgrader will still run to upgrade the CRDs .


SiorMeir (dismissed):

.github/workflows/on-pr.yaml
```diff
@@ -152,6 +152,7 @@ jobs:
         env:
           PACKAGE_VERSION: ${{ needs.build.outputs.package_version }}
         run: |
+          kubectl apply -f ./deployments/external-crds
```
romanbaron: Why is it required here? Won't it be deployed by crd-upgrader?

enoodle: You are right, I will remove it

romanbaron (dismissed):

PR #452, opened 2025-08-30
remove unused variable and volumes from binder

enoodle:


cmd/binder/app/app.go
```diff
@@ -190,15 +186,6 @@ func (app *App) Run() error {
 	}
 	// +kubebuilder:scaffold:builder
 
-	if err = app.manager.AddHealthzCheck("healthz", app.manager.GetWebhookServer().StartedChecker()); err != nil {
```
SiorMeir, line 193: Don't we need a ready/healthy endpoints on the binder?

SiorMeir (dismissed):

cmd/binder/app/app.go
enoodle: Now the binder is only a controller - handling the BindRequest CRD. These were endpoints for the webhook that is not functional anymore.
The readiness endpoint is used to determine if a specific container of a server is ready to receive incoming traffic, but the binder is not receiving any calls now (except for metrics, which is ok that will just fail)
The health check is used to determine if a pod is in internal failure status and restart the pod, but it was not doing that anyway (always returning true after startup) and this pod will just crush anyway. The controller itself is also serving a health endpoint, which still exists.
So they are just not needed. I do think that the tests are failing because I didn't remove those probes from the deployment.

SiorMeir (approved):

PR #454, opened 2025-08-31
PGInfo Clean SubGroup messages + Remove GetDefaultMinAvailable

omer-dayan:


pkg/scheduler/api/unschedule_info.go
```diff
@@ -82,12 +82,28 @@ func GetJobOverMaxAllowedMessageForQueue(
...
+
+func getGangEvictionForDefaultSubgroupMessage(taskNamespace, taskName string, minimum int32) string {
 	return fmt.Sprintf(
 		"Workload doesn't have the minimum required number of pods (%d), evicting remaining pod: %s/%s",
 		minimum, taskNamespace, taskName)
 }
 
+func getGangEvictionWithSubGroupsMessage(taskNamespace, taskName, subGroup string, minimum int32) string {
+	return fmt.Sprintf(
+		"Workload doesn't have the minimum required number of pods (%d) on sub-group %s, evicting remaining pod: %s/%s",
```
romanbaron: I have a small suggestion here:
"Workload doesn't meet the minimum required number of pods (%d) in sub-group %s. Evicting remaining pod: %s/%s"

If you agree, please also align the message for the default subgroup.

pkg/scheduler/actions/common/allocate.go
```diff
@@ -125,13 +125,34 @@ func handleFailedTaskAllocation(job *podgroup_info.PodGroupInfo, unschedulableTa
 		allocationError.SetError(common_info.DefaultPodError)
 	}
 
-	numRunningTasks := job.GetActivelyRunningTasksCount()
-	if job.GetDefaultMinAvailable() > 1 && numRunningTasks < job.GetDefaultMinAvailable() {
+	isGangScheduling := false
+	var taskSubGroup *podgroup_info.SubGroupInfo
+	for subGroupName, subGroup := range job.GetSubGroups() {
```
romanbaron: you can get the subGroup without iterating through all the subgroups, I am curious why not getting it from the map directly.

pkg/scheduler/api/podgroup_info/job_info.go
```diff
@@ -139,7 +139,7 @@ func (pgi *PodGroupInfo) GetSubGroups() map[string]*SubGroupInfo {
 	return pgi.SubGroups
 }
 
-func (pgi *PodGroupInfo) SetDefaultMinAvailable(minAvailable int32) {
+func SetDefaultMinAvailable(pgi *PodGroupInfo, minAvailable int32) {
```
romanbaron: Why it is no longer a method?

romanbaron: Got it.

pkg/scheduler/api/unschedule_info.go
```diff
@@ -82,12 +82,28 @@ func GetJobOverMaxAllowedMessageForQueue(
...
 }
 
-func GetGangEvictionMessage(taskNamespace, taskName string, minimum int32) string {
+func GetGangEvictionMessage(task *pod_info.PodInfo, job *podgroup_info.PodGroupInfo) string {
+	if len(job.GetSubGroups()) == 1 {
+		if defaultSubgroup, found := job.GetSubGroups()[podgroup_info.DefaultSubGroup]; found {
+			return getGangEvictionForDefaultSubgroupMessage(task.Namespace, task.Name, defaultSubgroup.GetMinAvailable())
+		}
+	}
+	subGroup := job.GetSubGroups()[task.SubGroupName]
```
romanbaron: Maybe worth checking if subGroup was found so it won't panic in case there is a problem?

pkg/scheduler/actions/common/allocate.go
```diff
@@ -125,13 +125,34 @@ func handleFailedTaskAllocation(job *podgroup_info.PodGroupInfo, unschedulableTa
...
+				return
+			}
+		}
 		job.SetJobFitError(
 			podgroup_info.PodSchedulingErrors,
-			fmt.Sprintf("Resources were found for %d pods while %d are required for gang scheduling. "+
-				"Additional pods cannot be scheduled due to: %s",
-				numSchedulableTasks, job.GetDefaultMinAvailable(), allocationError.Error()),
+			fmt.Sprintf("Resources were found for %d pods from all sub-groups while sub-group %s minimum %d are required for gang scheduling. "+
+				"Additional pods cannot be scheduled in this sub-group due to: %s",
```
romanbaron: It is also possible that there will be more than one subgroup that doesn't meet gang requirement

pkg/scheduler/actions/common/allocate.go
```diff
@@ -125,13 +125,34 @@ func handleFailedTaskAllocation(job *podgroup_info.PodGroupInfo, unschedulableTa
...
+			if taskSubGroup.GetName() == podgroup_info.DefaultSubGroup {
+				job.SetJobFitError(
+					podgroup_info.PodSchedulingErrors,
+					fmt.Sprintf("Resources were found for %d pods while %d are required for gang scheduling. "+
+						"Additional pods cannot be scheduled due to: %s",
+						numSchedulableTasks, taskSubGroup.GetMinAvailable(), allocationError.Error()),
+					nil)
+				return
+			}
+		}
```
romanbaron: Any chance to simplify this part?

pkg/scheduler/api/unschedule_info.go
omer-dayan: I have changed it to have a message 
`Workload doesn't have the minimum required number of pods on undefined sub-group [SUBGROUP], evicting remaining pod: [NS]/[NAME]`

It is a case when we evict the entire workload. Do you think its fine to tell that to the user that the subgroup is not found, or should it be only logged in the scheduler?

pkg/scheduler/actions/common/allocate.go
omer-dayan: Yes but this function is in the context of single task.
Its called for each unschedulableTask.
So other SubGroups that doesnt meet requirement will have their own pods

pkg/scheduler/actions/common/allocate.go
omer-dayan: Tried. Tell me what you think

pkg/scheduler/api/unschedule_info.go
```diff
@@ -82,12 +82,38 @@ func GetJobOverMaxAllowedMessageForQueue(
...
+func getGangEvictionForDefaultSubgroupMessage(taskNamespace, taskName string, minimum int32) string {
 	return fmt.Sprintf(
-		"Workload doesn't have the minimum required number of pods (%d), evicting remaining pod: %s/%s",
+		"Workload doesn't meet the minimum required number of pods (%d), evicting remaining pod: %s/%s",
 		minimum, taskNamespace, taskName)
 }
 
+func getGangEvictionWithSubGroupsMessage(taskNamespace, taskName, subGroup string, minimum int32) string {
+	return fmt.Sprintf(
+		"Workload doesn't meet the minimum required number of pods (%d) on sub-group %s, evicting remaining pod: %s/%s",
```
romanbaron:
```suggestion
		"Workload doesn't meet the minimum required number of pods (%d) in sub-group %s, evicting remaining pod: %s/%s",
```

pkg/scheduler/api/unschedule_info.go
```diff
@@ -82,12 +82,38 @@ func GetJobOverMaxAllowedMessageForQueue(
...
+		minimum, subGroup, taskNamespace, taskName)
+}
+
+func getGangEvictionWithUndefinedSubGroupMessage(taskNamespace, taskName, subGroup string) string {
+	return fmt.Sprintf(
+		"Workload doesn't meet the minimum required number of pods on undefined sub-group %s, evicting remaining pod: %s/%s",
```
romanbaron:
```suggestion
		"Workload doesn't meet the minimum required number of pods in sub-group %s, evicting remaining pod: %s/%s",
```

pkg/scheduler/actions/common/allocate.go
```diff
@@ -125,19 +125,36 @@ func handleFailedTaskAllocation(job *podgroup_info.PodGroupInfo, unschedulableTa
...
 
-	numRunningTasks := job.GetActivelyRunningTasksCount()
-	if job.GetDefaultMinAvailable() > 1 && numRunningTasks < job.GetDefaultMinAvailable() {
+	isGangScheduling := false
+	taskSubGroup, found := job.GetSubGroups()[unschedulableTask.SubGroupName]
+	for _, subGroup := range job.GetSubGroups() {
+		if subGroup.GetMinAvailable() > 1 {
+			isGangScheduling = true
+		}
+	}
```
romanbaron: let's extract it to a separate function

pkg/scheduler/actions/common/allocate.go
```diff
@@ -125,19 +125,36 @@ func handleFailedTaskAllocation(job *podgroup_info.PodGroupInfo, unschedulableTa
...
-				unschedulableTask.Namespace, unschedulableTask.Name, allocationError.Error()),
+			fmt.Sprintf("Resources were found for %d pods while %d are required for gang scheduling. "+
+				"Additional pods cannot be scheduled due to: %s",
+				numSchedulableTasks, taskSubGroup.GetMinAvailable(), allocationError.Error()),
 			nil)
+		return
 	}
+	job.SetJobFitError(
+		podgroup_info.PodSchedulingErrors,
+		fmt.Sprintf("Resources were found for %d pods from all sub-groups while sub-group %s minimum %d are required for gang scheduling. "+
```
romanbaron:
```suggestion
		fmt.Sprintf("Resources were found for %d pods from all sub-groups while sub-group %s requires %d pods for gang scheduling. "+
```

pkg/scheduler/api/unschedule_info.go
romanbaron: I think the user will not know what it means. I suggested a small enhancement.

pkg/scheduler/actions/common/allocate.go
romanbaron: Yes, looks better.

pkg/scheduler/api/unschedule_info.go
omer-dayan: Changed to "in" but kept the "undefined"

pkg/scheduler/api/unschedule_info.go
```diff
@@ -82,12 +82,38 @@ func GetJobOverMaxAllowedMessageForQueue(
...
 
+func getGangEvictionWithSubGroupsMessage(taskNamespace, taskName, subGroup string, minimum int32) string {
+	return fmt.Sprintf(
+		"Workload doesn't meet the minimum required number of pods (%d) in sub-group %s, evicting remaining pod: %s/%s",
+		minimum, subGroup, taskNamespace, taskName)
+}
+
+func getGangEvictionWithUndefinedSubGroupMessage(taskNamespace, taskName, subGroup string) string {
+	return fmt.Sprintf(
+		"Workload doesn't meet the minimum required number of pods in undefined sub-group %s, evicting remaining pod: %s/%s",
```
romanbaron: consider removing the word "undefined"

romanbaron (approved):

PR #457, opened 2025-08-31
Fixed a bug where workload with subgroups would not consider addition…

romanbaron:
…al tasks above minAvailable

### DIFF
--- a/pkg/scheduler/api/podgroup_info/allocation_info.go
+++ b/pkg/scheduler/api/podgroup_info/allocation_info.go
@@ -32,32 +32,23 @@ func GetTasksToAllocate(
 	}
 
 	var tasksToAllocate []*pod_info.PodInfo
-	priorityQueueMap := getTasksPriorityQueuePerSubGroup(podGroupInfo, taskOrderFn, isRealAllocation)
-	maxNumOfTasksToAllocateMap := getNumTasksToAllocatePerSubGroup(podGroupInfo, isRealAllocation)
-
 	subGroupPriorityQueue := getSubGroupsPriorityQueue(podGroupInfo.GetSubGroups(), subGroupOrderFn)
 	maxNumOfSubGroups := getNumOfSubGroupsToAllocate(podGroupInfo)
-	numAllocatedSubGroups := 0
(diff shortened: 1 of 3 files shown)


pkg/scheduler/api/podgroup_info/allocation_info.go
```diff
@@ -32,35 +32,26 @@ func GetTasksToAllocate(
 	}
 
 	var tasksToAllocate []*pod_info.PodInfo
-	priorityQueueMap := getTasksPriorityQueuePerSubGroup(podGroupInfo, taskOrderFn, isRealAllocation)
-	maxNumOfTasksToAllocateMap := getNumTasksToAllocatePerSubGroup(podGroupInfo, isRealAllocation)
-
 	subGroupPriorityQueue := getSubGroupsPriorityQueue(podGroupInfo.GetSubGroups(), subGroupOrderFn)
 	maxNumOfSubGroups := getNumOfSubGroupsToAllocate(podGroupInfo)
```
omer-dayan: How come `getNumOfSubGroupsToAllocate` return only 1 or `len(SubGroups)`?
Isnt there a case when you need to allocate only from part of the subgroups?

pkg/scheduler/api/podgroup_info/allocation_info.go
```diff
@@ -32,35 +32,26 @@ func GetTasksToAllocate(
...
-	numAllocatedSubGroups := 0
+	numSubGroupsToAllocate := 0
```
omer-dayan, line 37: Because its a counter doesnt the previous name `numAllocatedSubGroups` is better?

omer-dayan (approved):

pkg/scheduler/api/podgroup_info/allocation_info.go
romanbaron: I thought it might be confusing because these subgroups are not actually allocated at this point, they are in a process to be allocated. I don't mind changing it back if you think the previous name was better.

omer-dayan: No that doesnt really matter

PR #459, opened 2025-09-01
adding initial KAI-operator

enoodle:
This operator is not yet deployed, but it will replace in the future all of the other deployments in `deployments/kai-scheduler/templates/services`

### DIFF
--- a/.github/workflows/on-pr.yaml
+++ b/.github/workflows/on-pr.yaml
@@ -163,6 +163,13 @@ jobs:
         run: |
           go install github.com/onsi/ginkgo/v2/ginkgo@v2.19.0
 
+      - name: Delete images from disk
+        env:
+          PACKAGE_VERSION: ${{ needs.build.outputs.package_version }}
+        run: |
+          docker images --format '{{.Repository}}:{{.Tag}}' | grep $PACKAGE_VERSION | xargs docker rmi -f
+          rm -rf images
+
       - name: Run e2e tests
         run: |
           ginkgo -r --keep-going --randomize-all --randomize-suites --trace -vv ./test/e2e/suites --label-filter '!autoscale', '!scale'


cmd/operator/config/custom_options.go
```diff
@@ -0,0 +1,22 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package config
+
+import (
+	"regexp"
+	"strings"
+)
+
+type ArrayFlags []string
```
itsomri: What is this? Why is it neede?

cmd/operator/config/options.go
```diff
@@ -0,0 +1,87 @@
...
+	if err := opts.parseCommandLineArgs(flagSet, options); err != nil {
+		return nil, err
+	}
+	if err := opts.loadConfigFromFile(); err != nil {
+		return nil, err
+	}
+	return opts, nil
+}
+
+func (opts *Options) loadConfigFromFile() error {
```
itsomri: Why aren't command line args enough?

pkg/apis/kai/v1/pod_group_controller/pod_group_controller.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+type PodGroupController struct {
+	// Enabled defines whether the pod-group-controller should be deployed
+	// +kubebuilder:validation:Optional
+	Enabled *bool `json:"enabled,omitempty"`
+
+	// Image is the configuration of the pod-group-controller image
+	// +kubebuilder:validation:Optional
+	Image *common.Image `json:"image,omitempty"`
```
itsomri: An idea - maybe Image, Resources, and others could be a part of "ControllerCommon" struct - all other controllers will use this struct, and will be able to use the same logic to generate an initial deployment from, and only modify specific things if needed (most will probably only need to override the image and args)

itsomri: So a typical operand will be something like:
```Go
deployment := config.podGroupController.controllerCommon.GetDeployment()
deployment.Spec.Template.Spec.Containers[0].Args = getArgs(config.podGroupController.args)
```

Something like that

cmd/operator/config/custom_options.go
enoodle: It is implementation of variable type. it is used for the AdditionalImagePullSecrets.

pkg/apis/kai/v1/pod_group_controller/pod_group_controller.go
enoodle: its possible to just create this struct and embed it without name so the usage will be the same but it will save a lot of duplications in definition. :+1:

cmd/operator/config/options.go
enoodle: its very awkward to pass tolerations and affinity in command line.

itsomri (approved):

cmd/operator/config/custom_options.go
enoodle: I understood that the argument is not used, so I will remove it and this variable type too.

cmd/operator/config/options.go
enoodle: I removed that as well

pkg/apis/kai/v1/pod_group_controller/pod_group_controller.go
```diff
@@ -0,0 +1,66 @@
...
+
+	// Replicas specifies the number of replicas of the pod-group controller
+	// +kubebuilder:validation:Optional
+	Replicas *int32 `json:"replicas,omitempty"`
+}
+
+// Args defines command line arguments for the pod-group-controller
+type Args struct {
+	// InferencePreemptible should inference priority class be counted as preemptibile
+	InferencePreemptible *bool `json:"inferencePreemptible,omitempty"`
```
itsomri: I think we deprecated this, @romanbaron might know

pkg/operator/config/feature-gates.go
```diff
@@ -0,0 +1,24 @@
+// Copyright 2025 NVIDIA CORPORATION
```
itsomri: Let's use underscore instead of dash (-) in the file name, this seems to be the convention

pkg/apis/kai/v1/pod_group_controller/pod_group_controller.go
enoodle: :+1:

pkg/operator/config/feature-gates.go
enoodle: :+1:

itsomri (approved):

PR #460, opened 2025-09-01
If a podgroup has scheduled pod and a required topology constraint, t…

davidLif:
…ake the active pods location into account when calculating domains for any pending pods of the same podgroup


pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
```diff
@@ -199,11 +200,20 @@ func (t *topologyPlugin) getBestJobAllocatableDomains(job *podgroup_info.PodGrou
 	if err != nil {
 		return nil, err
 	}
+
+	// Validate that the domains do not crush with the active pods of the job
```
itsomri:
```suggestion
	// Validate that the domains match the active pods of the job
```

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
```diff
@@ -231,6 +241,47 @@ func (t *topologyPlugin) getBestJobAllocatableDomains(job *podgroup_info.PodGrou
 	return []*TopologyDomainInfo{maxDepthDomains[0]}, nil
 }
 
+func getRelevantDomainsWithAllocatedPods(job *podgroup_info.PodGroupInfo, topologyTree *TopologyInfo, topLevel string) map[string]map[TopologyDomainID]*TopologyDomainInfo {
+	relevantDomainsByLevel := map[string]map[TopologyDomainID]*TopologyDomainInfo{}
+	for _, domainAtRequiredLevel := range topologyTree.DomainsByLevel[topLevel] {
```
itsomri: Why are we looking only at topLevel?

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
```diff
@@ -231,6 +241,47 @@ func (t *topologyPlugin) getBestJobAllocatableDomains(job *podgroup_info.PodGrou
...
+		}
+	}
+	return activePodsInDomain
+}
+
+func addSubTreeToDomainMap(domain *TopologyDomainInfo, domainsMap map[string]map[TopologyDomainID]*TopologyDomainInfo) {
+	if domainsMap[domain.Level] == nil {
+		domainsMap[domain.Level] = map[TopologyDomainID]*TopologyDomainInfo{}
+	}
+	if domain.Children != nil {
```
itsomri: I think this if is redundant

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
davidLif: I think the better message might be:
"Validate that the domains do not clash with the chosen domain for active pods of the job"

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
```diff
@@ -14,6 +14,7 @@ import (
 
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/node_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/pod_info"
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/pod_status"
```
enoodle, line 17: I just noticed the name of the file, maybe we can just simplify it to `job_filtering.go` because `topology_plugin` is kind of redundant here.

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
davidLif: We are trying to calculate all the domains that satisfy the following requirment:
Any domain returned by this function has a common parent with the active pods of this job at the required level. 

In other words, all the pods (old and new) set under the same "domain" at the required level.

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
davidLif: It's not - you need to set a value for new keys in a map instead. You can see that if you disable 270-272, some unitests fail.

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
```diff
@@ -231,6 +241,47 @@ func (t *topologyPlugin) getBestJobAllocatableDomains(job *podgroup_info.PodGrou
 	return []*TopologyDomainInfo{maxDepthDomains[0]}, nil
 }
 
+func getRelevantDomainsWithAllocatedPods(job *podgroup_info.PodGroupInfo, topologyTree *TopologyInfo, topLevel string) map[string]map[TopologyDomainID]*TopologyDomainInfo {
+	relevantDomainsByLevel := map[string]map[TopologyDomainID]*TopologyDomainInfo{}
```
enoodle: This complicated type comes up a lot, maybe it is worth making it a named type.

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
enoodle: @itsomri I think that this `topLevel` is a bit of misleading name - it is the level that is required by the job (see the function call) - so it is the top level that is "interesting" for the job, and everything under it is added by `addSubTreeToDomainMap`

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
enoodle: I think that omri means that only the test `if domain.Childern != nil` is redundant, because if it is nil, then the for loop will just finish immediately.

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
itsomri: ok

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
itsomri: Yes, like Erez said - the if statement itself is redundant

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
davidLif: I renamed it to "requiredLevel"

pkg/scheduler/plugins/topology/topology_plugin_job_filtering.go
davidLif: Renamed the file

PR #462, opened 2025-09-02
Time aware fairness: prometheus time decay

itsomri:
Implement exponential decay in prometheus using the half life config.


pkg/scheduler/cache/usagedb/prometheus/prometheus.go
```diff
@@ -108,7 +113,7 @@ func (p *PrometheusClient) queryResourceUsage(ctx context.Context, allocationMet
 
 	// Log warnings if exist
 	for _, w := range warnings {
-		log.InfraLogger.V(3).Warnf("Warning querying cluster usage metric %s: %s", allocationMetric, w)
+		log.InfraLogger.V(3).Warnf("Warning querying cluster usage metric %s: %s", decayedAllocationMetric, w)
```
SiorMeir, line 116: just a minor note - I don't think we need the word Warning when the log level is already at warning level

pkg/scheduler/cache/usagedb/prometheus/prometheus.go
```diff
@@ -129,3 +134,14 @@ func (p *PrometheusClient) queryResourceUsage(ctx context.Context, allocationMet
 
 	return queueUsage, nil
 }
+
+func getExponentialDecayQuery(halfLifePeriod *time.Duration) string {
+	if halfLifePeriod == nil {
+		return ""
+	}
```
singh1203, line 141: It’s up to you - would you prefer returning `1` (no-op) or `nil` when `halfLifePeriod == nil`? That way, the query makes it clearer that no decay is applied.

singh1203 (approved):
Just for my knowledge, want to confirm, if `halfLifePeriod` isn’t set, does it just use the raw metric without decay?

Otherwise, all LGTM. 👍

pkg/scheduler/cache/usagedb/prometheus/prometheus.go
itsomri: I get what you're saying but I want to explain to the reader that this warning was received when trying to query the prometheus - what do you suggest?

itsomri:
> Just for my knowledge, want to confirm, if `halfLifePeriod` isn’t set, does it just use the raw metric without decay?
> 
> Otherwise, all LGTM. 👍 Thank you!

Yes, it sums the usage without applying half life

pkg/scheduler/cache/usagedb/prometheus/prometheus.go
itsomri: You mean a vector(1) that will multiply the original metric? If so, then no - I prefer to not add the half life to the expression at all - when things go wrong with querying prometheus, you sometimes need to debug the query that ran - I prefer to not make it more complex then it has to

pkg/scheduler/cache/usagedb/prometheus/prometheus.go
SiorMeir: @itsomri 
So I think we should specify here explicitly that this comes from the interaction with Prometheus.

Maybe something along the lines of `Failure to query cluster usage. Check Prometheus and TSDB`
?

In addition - I think Prometheus / metrics is a pretty important infra in k8s. Shouldn't this be upgraded to an Error?

itsomri: But it's not a failure, we just got some warnings (I don't really know what kind of warnings, maybe usage of deprecated functions/ unoptimized queries?)
I don't think we should consider this an error. Think about it: let's say you get a warning because your query is suboptimal - would you want to fail the query? This means the scheduler will operate without historical usage data, which means unfair evictions/allocations

SiorMeir: Got your point.
In this case, I'd stay with the current warning string.

SiorMeir (approved):

pkg/scheduler/cache/usagedb/prometheus/prometheus.go
singh1203: Got it!

PR #463, opened 2025-09-02
New plugin endpoint - for partition

omer-dayan:


pkg/scheduler/framework/session_plugins.go
```diff
@@ -341,6 +345,25 @@ func (ssn *Session) PredicateFn(task *pod_info.PodInfo, job *podgroup_info.PodGr
 	return nil
 }
 
+func (ssn *Session) PartitionFn(job *podgroup_info.PodGroupInfo, nodes api.Partition) ([]api.Partition, error) {
+	var err error
```
omer-dayan, line 349: In this function we take a partition (all cluster) and partition it to sub-partitions.
P1 -> P1s1, P1s2, P1s3.

And then each sub partition is partitioned so we sould have

P1s1p1, P1s1p2, P1s1p3, P1s2p1, P1s2p2, P1s2p3, P1s3p1, P1s3p2, P1s3p3.

That means that the order of partition functions is matter.
What do you think about it?

pkg/scheduler/api/types.go
```diff
@@ -28,9 +28,13 @@ import (
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/resource_info"
 )
 
+type Partition []*node_info.NodeInfo
```
omer-dayan: What do you think about the term "partition"?

Can you think of more appropriate term?

pkg/scheduler/actions/common/allocate.go
```diff
@@ -27,6 +27,22 @@ func AllocateJob(ssn *framework.Session, stmt *framework.Statement, nodes []*nod
...
+		if allocateJobOnPartition(ssn, stmt, partition, job, isPipelineOnly, tasksToAllocate) {
+			return true
+		}
+	}
+	return false
+}
+
+func allocateJobOnPartition(ssn *framework.Session, stmt *framework.Statement, nodes []*node_info.NodeInfo,
+	job *podgroup_info.PodGroupInfo, isPipelineOnly bool, tasksToAllocate []*pod_info.PodInfo) bool {
 	defer ssn.CleanAllocationAttemptCache(job)
```
omer-dayan, line 46: Notice the `defer ssn.CleanAllocationAttemptCache(job)` in here that would now happen on each partition try, multiple times for a single job, as opposed to before that it was happen once per job

pkg/scheduler/api/types.go
romanbaron: Just in case we will have other partitions in the future, maybe it should be `NodePartitions`?
Other possible names can be: Segments, Pools, Cohorts

pkg/scheduler/api/types.go
```diff
@@ -28,9 +28,13 @@ import (
...
+
 // PredicateFn is used to predicate node for task.
 type PredicateFn func(*pod_info.PodInfo, *podgroup_info.PodGroupInfo, *node_info.NodeInfo) error
 
+type PartitionFn func(*podgroup_info.PodGroupInfo, Partition) ([]Partition, error)
```
romanbaron: I understand why the first parameter is the job since it dictates how nodes are partitioned, why the second parameter is a Partition? Maybe it should simply be a slice of all cluster nodes?

omer-dayan: So we can have multiple plugins implement this function.
If we have only one plugin so yes the nodes could be infered from `ssn.Nodes`.
But when you have multiple plugins what is the expected behaviour?

In my implementation its partitining any previous partition

pkg/scheduler/framework/session_plugins.go
romanbaron: Can you maybe explain a bit more about this function?

I can imagine a pluggable function that divides the nodes into partitions, but I don't understand why it is invoked on every partitions again

omer-dayan: In our current plugin mechanism, most of the endpoints allow for multiple plugin implementations.
For example: `NodeScore` endpoint is implemented in multiple plugins: `nodeplacement`, `podaffinity`, etc.

I wanted to make this new endpoint the same. Allowing multiple plugins implement this new `Partition` endpoint.
For that, a question need to be asked - what is the expected behavior when you have 2 or more plugins?

What I decided to implement is that all plugins work in order.
Given you have `NodePoolPlugin` and `TopologyPlugin` both implement `Partition` endpoint - what would happen is that the first plugin, `NodePoolPlugin` will partition the cluster first:
AllCluster -> `NodePool1`, `NodePool2`, `NodePool3` ...

And then `TopologyPlugin` will partition each `NodePoolX`:
`NodePoolX` -> `NodePoolXDomain1`, `NodePoolXDomain2` ...

So we end up with:
[`NodePool1Domain1`, `NodePool1Domain2`, `NodePool2Domain1`, `NodePool2Domain2`, `NodePool2Domain3`, `NodePool3Domain1`]

pkg/scheduler/framework/session_plugins.go
```diff
@@ -341,6 +345,25 @@ func (ssn *Session) PredicateFn(task *pod_info.PodInfo, job *podgroup_info.PodGr
...
+func (ssn *Session) NodePartitionFn(job *podgroup_info.PodGroupInfo, nodes api.NodePartition) ([]api.NodePartition, error) {
+	var err error
+	partitions := []api.NodePartition{nodes}
+	for _, pfn := range ssn.NodePartitionFns {
+		newPartitions := []api.NodePartition{}
+		for _, partition := range partitions {
+			partitions, err = pfn(job, partition)
+			if err != nil {
+				log.InfraLogger.V(6).Infof(
+					"Failed to run Partition on job %s", job.Name)
```
romanbaron:
```suggestion
					"Failed to run NodePartitionFn on job %s", job.Name)
```

pkg/scheduler/actions/common/allocate.go
```diff
@@ -27,6 +27,22 @@ func AllocateJob(ssn *framework.Session, stmt *framework.Statement, nodes []*nod
 		return false
 	}
 
+	nodePartitions, err := ssn.NodePartitionFn(job, tasksToAllocate, nodes)
+	if err != nil {
+		log.InfraLogger.Errorf("Failed to partition nodes for job <%s/%s> in session %v, err: %v", job.Namespace, job.Name, ssn.UID, err)
+		return false
+	}
+
+	for _, partition := range nodePartitions {
```
itsomri, line 36: Please add a high-level log (maybe 5 or 6) that will give us details on the partitions (how many, attempt x/N, names, etc)

pkg/scheduler/actions/common/allocate.go
```diff
@@ -27,6 +27,22 @@ func AllocateJob(ssn *framework.Session, stmt *framework.Statement, nodes []*nod
 		return false
 	}
 
+	nodePartitions, err := ssn.NodePartitionFn(job, tasksToAllocate, nodes)
+	if err != nil {
+		log.InfraLogger.Errorf("Failed to partition nodes for job <%s/%s> in session %v, err: %v", job.Namespace, job.Name, ssn.UID, err)
```
itsomri, line 32:
```suggestion
		log.InfraLogger.V(1).Errorf("Failed to partition nodes for job <%s/%s> in session %v, err: %v", job.Namespace, job.Name, ssn.UID, err)
```

pkg/scheduler/api/types.go
```diff
@@ -28,9 +28,13 @@ import (
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/resource_info"
 )
 
+type NodePartition []*node_info.NodeInfo
```
itsomri, line 31: kinda feels like it belongs in nodeinfo

pkg/scheduler/framework/session_plugins.go
```diff
@@ -341,6 +345,25 @@ func (ssn *Session) PredicateFn(task *pod_info.PodInfo, job *podgroup_info.PodGr
 	return nil
 }
 
+func (ssn *Session) NodePartitionFn(job *podgroup_info.PodGroupInfo, tasks []*pod_info.PodInfo, nodes api.NodePartition) ([]api.NodePartition, error) {
+	var err error
+	partitions := []api.NodePartition{nodes}
```
enoodle, line 350: This always adds a partition of all the nodes, maybe it should be empty and some default plugin will fill it up? This way we can also remove it as an option if needed.

pkg/scheduler/api/types.go
itsomri: I think partitions should be named. Otherwise it will be hard to convey scheduling errors

PR #467, opened 2025-09-03
deploy operator in chart with pod group controller

enoodle:


deployments/kai-scheduler/templates/rbac/podgroupcontroller-binding.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
   name: kai-podgroup-controller
 subjects:
   - kind: ServiceAccount
-    name: podgroup-controller
+    name: pod-group-controller
```
SiorMeir: why the name change?

enoodle: It should match this: https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/operator/operands/pod_group_controller/resources.go#L20
I don't really care which name takes.. It seems more logical that pod group should be two words :shrug:

SiorMeir (approved):

deployments/kai-scheduler/templates/rbac/podgroupcontroller-binding.yaml
enoodle: I changed it in the other place eventually :)

PR #469, opened 2025-09-03
add queue controller operand

enoodle:


pkg/apis/kai/v1/queue_controller/queue_controller.go
```diff
@@ -0,0 +1,165 @@
...
+	// QueueLabelToMetricLabel maps queue label keys to metric label keys for metrics exposure
+	// +kubebuilder:validation:Optional
+	QueueLabelToMetricLabel *string `json:"queueLabelToMetricLabel,omitempty"`
+
+	// QueueLabelToDefaultMetricValue maps queue label keys to default metric values when the label is absent
+	// +kubebuilder:validation:Optional
+	QueueLabelToDefaultMetricValue *string `json:"queueLabelToDefaultMetricValue,omitempty"`
+}
+
+func (q *QueueController) SetDefaultsWhereNeeded(replicaCount *int32) {
```
SiorMeir, line 46: Can / Should we extract the pattern of == nil then set to a method? might look cleaner
Saw that pattern elsewhere, wondering if this would be an anti-pattern in Go

SiorMeir (approved):

pkg/apis/kai/v1/queue_controller/queue_controller.go
enoodle: :+1: Thanks, great idea

SiorMeir (approved):

PR #476, opened 2025-09-07
Topology scheduling scheduler tests

omer-dayan:


pkg/scheduler/actions/integration_tests/allocate/allocate_test.go
```diff
@@ -4,18 +4,20 @@
...
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/pod_status"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/constants"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/test_utils"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/test_utils/jobs_fake"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/test_utils/nodes_fake"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/test_utils/tasks_fake"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+	kueuev1alpha1 "sigs.k8s.io/kueue/apis/kueue/v1alpha1"
+	"testing"
```
romanbaron: It could be nice if all our imports were sorted so the import blocks would make more sense. Goland has an automatic way to do it if you configure it here:
Screenshot 2025-09-07 at 10 00 31

pkg/scheduler/actions/integration_tests/allocate/allocate_test.go
```diff
@@ -565,10 +567,10 @@ func getAllocateTestsMetadata() []integration_tests_utils.TestTopologyMetadata {
 						RequiredGPUsPerTask: 1,
 						Priority:            constants.PriorityTrainNumber,
 						QueueName:           "queue0", Tasks: []*tasks_fake.TestTaskBasic{
-							{
-								State: pod_status.Pending,
-							},
+						{
+							State: pod_status.Pending,
 						},
+					},
```
romanbaron: I think it was properly indented before

pkg/scheduler/actions/integration_tests/allocate/allocate_test.go
```diff
@@ -1079,5 +1081,123 @@ func getAllocateTestsMetadata() []integration_tests_utils.TestTopologyMetadata {
...
+					"pending_job0-1": {
+						NodeName:             "node3",
+						GPUsRequired:         1,
+						Status:               pod_status.Running,
+						DontValidateGPUGroup: true,
+					},
+				},
+				Mocks: &test_utils.TestMock{
+					CacheRequirements: &test_utils.CacheMocking{
+						NumberOfCacheBinds: 1,
```
romanbaron: Two pods get scheduled, how can it be only single bind?

omer-dayan: Actually the test pass no matter what number I put in here...
Any idea why?

romanbaron: It used:
```
cacheMock.EXPECT().Bind(Any(), Any(), Any()).Return(nil).MaxTimes(cacheRequirements.NumberOfCacheBinds)
```
So any number below the number of bound pods should fail, higher numbers should pass.

romanbaron (approved):

PR #477, opened 2025-09-07
feat: add prometheus operand

SiorMeir:


pkg/apis/kai/v1/global.go
```diff
@@ -61,6 +61,10 @@ type GlobalConfig struct {
 	// PodLabelSelector filters pods for webhooks and pod grouper
 	// +kubebuilder:validation:Optional
 	PodLabelSelector map[string]string `json:"podLabelSelector,omitempty"`
+
+	// PrometheusEnabled defines whether Prometheus should be enabled
```
itsomri:
```suggestion
	// PrometheusEnabled defines whether a Prometheus instance should be created for kai
```

itsomri: Wait why is this in global?

pkg/apis/kai/v1/prometheus/prometheus.go
```diff
@@ -0,0 +1,23 @@
...
+
+// +kubebuilder:object:generate:=true
+package prometheus
+
+import (
+	"k8s.io/utils/ptr"
+)
+
+type Prometheus struct {
+	// Enabled defines whether Prometheus should be deployed
```
itsomri:
```suggestion
	// Enabled defines whether a Prometheus instance should be deployed
```

pkg/operator/controller/status_reconciler/status_reconciler.go
```diff
@@ -165,3 +168,47 @@ func (r *StatusReconciler) getDependenciesFulfilledCondition(gen int64) metav1.C
...
+		Name: "prometheuses.monitoring.coreos.com",
+	}, crd)
+
+	if err != nil {
+		if client.IgnoreNotFound(err) != nil {
+			return metav1.Condition{
+				Type:               "PrometheusOperatorInstalled",
+				Status:             metav1.ConditionFalse,
+				Reason:             "PrometheusOperatorCheckFailed",
+				Message:            err.Error(),
```
itsomri: The message will not be the most clear, it will probably say something like "failed to find CRD ...", maybe wrap it with "`Failed to detect prometheus operator installation: %s", err.Error()`"

itsomri: Also, I think you're not supposed to ignore not found in this case: you're looking for the CRD, not an instance of the CRD, so not found == the CRD is not installed in the cluster. Unless I misunderstood something

pkg/operator/operands/prometheus/resources.go
```diff
@@ -0,0 +1,122 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	kaiv1 "github.com/NVIDIA/KAI-scheduler/pkg/apis/kai/v1"
+)
+
+const (
+	mainResourceName = "prometheus"
```
itsomri:
```suggestion
	mainResourceName = "kai"
```
This will result in "prometheus-kai" (or kai-prometheus, I don't remember)
if you leave it as "prometheus", it will be "prometheus-prometheus"

pkg/apis/kai/v1/global.go
SiorMeir: leftover. Will be removed entirely

pkg/apis/kai/v1/prometheus/prometheus.go
```diff
@@ -0,0 +1,23 @@
...
+	// +kubebuilder:validation:Optional
+	Enabled *bool `json:"enabled,omitempty"`
+}
+
+func (p *Prometheus) SetDefaultsWhereNeeded() {
+
+	if p.Enabled == nil {
+		p.Enabled = ptr.To(false)
```
SiorMeir, line 20: @itsomri Note that I've set the default to False - made sense to me..
Do you agree?

pkg/operator/controller/status_reconciler/status_reconciler.go
```diff
@@ -165,3 +168,47 @@ func (r *StatusReconciler) getDependenciesFulfilledCondition(gen int64) metav1.C
 		LastTransitionTime: metav1.Now(),
 	}
 }
+
+func (r *StatusReconciler) getPrometheusOperatorCondition(ctx context.Context, gen int64) metav1.Condition {
```
enoodle: This should be implemented in the prometheus operand, probably in the DesiredState method. Then the error message can be returnred in IsDeployed  / IsAvailable functions.

pkg/operator/operands/prometheus/prometheus.go
```diff
@@ -0,0 +1,46 @@
...
+	p.lastDesiredState = objects
+	return objects, nil
+}
+
+func (b *Prometheus) IsDeployed(ctx context.Context, readerClient client.Reader) (bool, error) {
+	return common.AllObjectsExists(ctx, readerClient, b.lastDesiredState)
+}
+
+func (b *Prometheus) IsAvailable(ctx context.Context, readerClient client.Reader) (bool, error) {
+	return common.AllControllersAvailable(ctx, readerClient, b.lastDesiredState)
```
enoodle: This operand is not creating any controllers, here should be a test for the avialability of the prometheus instance.

pkg/operator/controller/status_reconciler/status_reconciler.go
SiorMeir: enriched the message.
Regarding the second message - you are correct. The logic here grabs and checks for the CRD existance beforet this line. This condition will be true if there's another issue other than that the CRD is not found

pkg/apis/kai/v1/prometheus/prometheus.go
itsomri: Yes

pkg/operator/operands/prometheus/resources.go
```diff
@@ -0,0 +1,122 @@
...
+		return nil, err
+	}
+
+	// If Prometheus Operator is not installed, we can't create a Prometheus CR
+	if !hasPrometheusOperator {
+		logger.Info("Prometheus Operator not found - Prometheus CRD is not available")
+		return []client.Object{}, nil
+	}
+
+	logger.Info("Prometheus Operator found, proceeding with Prometheus instance creation")
```
itsomri: This might be confusing, since this log will be printed every reconcile loop, but we will only create the prometheus once (if everything works right)
You're not "creating" you're "reconciling" (trying to move the prometheus to the desired state)

pkg/apis/kai/v1/prometheus/prometheus.go
```diff
@@ -0,0 +1,23 @@
...
+	"k8s.io/utils/ptr"
+)
+
+type Prometheus struct {
+	// Enabled defines whether a Prometheus instance should be deployed
+	// +kubebuilder:validation:Optional
+	Enabled *bool `json:"enabled,omitempty"`
+}
+
+func (p *Prometheus) SetDefaultsWhereNeeded() {
```
itsomri, line 17: Need to add properties to configure the persistency of the prometheus: a boolean for persistency (defaults to false), an optional StorageClass (if not using the cluster default), and a volume size
It should configure the following fields in the prometheus instance: https://github.com/NVIDIA/KAI-Scheduler/blob/80d98917c0c9119a0f1a99e44c727dfc9445819f/docs/metrics/prometheus.yaml#L52

SiorMeir: As discussed - will be add at a later PR

pkg/operator/operands/prometheus/resources.go
SiorMeir: I think we will remove it alltogether as we have a log message when creating the status

pkg/operator/operands/prometheus/prometheus.go
SiorMeir: done, now checking for a  `monitoringv1.Prometheus{}` instance

pkg/operator/operands/prometheus/prometheus.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// Only check for Prometheus operator if we actually have objects to deploy
+	if len(objects) > 0 {
+		err = p.GetPrometheusOperatorCondition(ctx, runtimeClient, 0)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	p.lastDesiredState = objects
```
enoodle: Set this before making the operator check so that the IsDeployed method will reflect that it is missing.

enoodle (changes requested):

pkg/operator/operands/prometheus/prometheus.go
SiorMeir: done, moved before check

PR #482, opened 2025-09-08
introduce scheduling shards

enoodle:
Scheduling shards are meant to enable partitioning of the cluster nodes in order to allow separate scheduling on each partition. This can be a solution to heterogeneous clusters with different types of accelerators which allow defining different quota for each partition and schedule jobs only in their intended partition.


pkg/apis/kai/v1/schedulingshard_types.go
```diff
@@ -0,0 +1,123 @@
...
+	// ReclaimMinRuntime specifies the minimum runtime of a job in queue before it can be reclaimed
+	// +kubebuilder:validation:Optional
+	ReclaimMinRuntime *string `json:"reclaimMinRuntime,omitempty"`
+}
+
+// PlacementStrategy defines the scheduling strategy of NodePool
+type PlacementStrategy struct {
+	// Gpu scheduling strategy (binpack/spread)
+	// +kubebuilder:validation:Optional
+	Gpu *string `json:"gpu,omitempty"`
```
SiorMeir: GPU / CPU instead of Gpu / Cpu ?

SiorMeir (approved):

pkg/apis/kai/v1/schedulingshard_types.go
enoodle: :+1:

SiorMeir (approved):

PR #484, opened 2025-09-09
kai operator docs

enoodle:


docs/operator/README.md
```diff
@@ -0,0 +1,72 @@
...
+Each KAI service is managed by a dedicated operand:
+- **Admission Webhooks** - Pod validation and mutation
+- **Scheduler** - Pod scheduling decisions
+- **Queue Controller** - Queue management
+- **Pod Group Controller** - Pod grouping functionality
+- **Node Scale Adjuster** - Node scaling operations
+
+### CRDs
+
+#### Config CRD (`config.kai.scheduler`)
```
itsomri, line 28: Maybe not here - but we should have a doc of all possible configurations for the config. I think it can be generated including descriptions with kube-builder or something like that. We can also generate a yaml example with all the default configs and make sure it stays up-to-date by calling SetDefaultsWhereNeeded and marshalling to YAML

enoodle: Lets implement this in a different PR - It seems important but we need to understand what format and contents we want to have there.

PR #486, opened 2025-09-09
adding binder operand

enoodle:


pkg/apis/kai/v1/binder/binder.go
```diff
@@ -0,0 +1,129 @@
...
+	imageName                           = "binder"
+	defaultResourceReservationImageName = "resource-reservation"
+)
+
+type Binder struct {
+	Service *common.Service `json:"service,omitempty"`
+
+	// Webhook defines configuration for the binder service
+	// +kubebuilder:validation:Optional
+	Webhook *Webhook `json:"webhook,omitempty"`
```
davidLif: Does the binder still needs a webhook config after the kai-admission service?

enoodle: :+1: right it doesn't need it!

PR #490, opened 2025-09-09
Time fairshare - support tumbling window config

davidLif:
Add support for Prometheus client tumbling window configuration - allow the usage data used for fairshare calculations to be reset each X time instead of a sliding window

### DIFF
--- a/pkg/scheduler/cache/usagedb/api/interface.go
+++ b/pkg/scheduler/cache/usagedb/api/interface.go
@@ -37,6 +37,9 @@ type UsageParams struct {
 	WindowSize *time.Duration `yaml:"windowSize" json:"windowSize"`
 	// Window type for time-series aggregation. If not set, defaults to sliding.
 	WindowType *WindowType `yaml:"windowType" json:"windowType"`
+	// Cron string for the usage. If not set, the usage will not be aggregated.
+	CronString string `yaml:"cronString" json:"cronString"`
+
 	// ExtraParams are extra parameters for the usage db client, which are client specific.
 	ExtraParams map[string]string `yaml:"extraParams" json:"extraParams"`
 }
--- a/pkg/scheduler/cache/usagedb/api/interface.go
+++ b/pkg/scheduler/cache/usagedb/api/interface.go
@@ -37,8 +37,8 @@ type UsageParams struct {
 	WindowSize *time.Duration `yaml:"windowSize" json:"windowSize"`
 	// Window type for time-series aggregation. If not set, defaults to sliding.
 	WindowType *WindowType `yaml:"windowType" json:"windowType"`
-	// Cron string for the usage. If not set, the usage will not be aggregated.
-	CronString string `yaml:"cronString" json:"cronString"`
+	// A cron string used to determine when to reset resource usage for all queues.
+	TumblingWindowCronString string `yaml:"tumblingWindowCronString" json:"tumblingWindowCronString"`
 
 	// ExtraParams are extra parameters for the usage db client, which are client specific.
 	ExtraParams map[string]string `yaml:"extraParams" json:"extraParams"`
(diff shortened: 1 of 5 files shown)


pkg/scheduler/cache/usagedb/api/interface.go
```diff
@@ -37,6 +37,9 @@ type UsageParams struct {
 	WindowSize *time.Duration `yaml:"windowSize" json:"windowSize"`
 	// Window type for time-series aggregation. If not set, defaults to sliding.
 	WindowType *WindowType `yaml:"windowType" json:"windowType"`
+	// Cron string for the usage. If not set, the usage will not be aggregated.
+	CronString string `yaml:"cronString" json:"cronString"`
```
itsomri: "CronString" is the "type", not the "what" this param does.. Needs a better name, maybe "ResetTime", "ResetCron", "TumblingWindowCron" etc.
Also, the docstring doesn't make sense at all.

pkg/scheduler/cache/usagedb/prometheus/prometheus.go
```diff
@@ -59,15 +62,33 @@ func NewPrometheusClient(address string, params *api.UsageParams) (api.Interface
...
+	}
+
+	if params.WindowType == nil {
+		return nil, fmt.Errorf("window type is not set in usage params")
+	}
+	switch *params.WindowType {
+	case api.TumblingWindow:
+		clientObj.queryUsageFunction = clientObj.queryTumblingWindowUsage
+
+		cronExpression, err := cronexpr.Parse(params.CronString)
```
itsomri: Cron validation should probably be validated globally, in `pkg/scheduler/cache/usagedb/hub.go`, since it's not a client-specific thing

pkg/scheduler/cache/usagedb/prometheus/prometheus.go
```diff
@@ -145,3 +166,59 @@ func getExponentialDecayQuery(halfLifePeriod *time.Duration) string {
...
+func (p *PrometheusClient) queryTumblingWindowUsage(ctx context.Context, allocationMetric string) (map[common_info.QueueID]float64, error) {
+	queuesUsage := make(map[common_info.QueueID]float64)
+
+	usageQuery := fmt.Sprintf("(%s)[%s:%s]",
+		allocationMetric,
+		p.usageParams.WindowSize.String(),
+		p.queryResolution.String(),
+	)
+
+	usageResult, warnings, err := p.client.Query(ctx, usageQuery, time.Now())
```
itsomri, line 179: Looks like a lot of code is duplicated - can you not just change the query string? All the error handling and data processing should look the same

pkg/scheduler/cache/usagedb/prometheus/prometheus.go
```diff
@@ -145,3 +166,59 @@ func getExponentialDecayQuery(halfLifePeriod *time.Duration) string {
...
+	}
+
+	usageMatrix := usageResult.(model.Matrix)
+	if len(usageMatrix) == 0 {
+		return nil, fmt.Errorf("no data returned for cluster usage metric %s", allocationMetric)
+	}
+
+	lastUsageReset := p.getLatestUsageResetTime()
+
+	for _, usageSample := range usageMatrix {
```
itsomri: Why do it in code? Why not in prometheus?

pkg/scheduler/cache/usagedb/api/interface.go
davidLif: Renamed and changed the description

pkg/scheduler/cache/usagedb/prometheus/prometheus.go
davidLif: Changed

pkg/scheduler/cache/usagedb/prometheus/prometheus.go
davidLif: Yes, but I don't think that for this single line it's worth it to add this as a very spesific parameter to the client ctor or extend the interface.

itsomri (approved):

PR #493, opened 2025-09-10
adding scheduler operand

enoodle:


pkg/operator/operands/scheduler/scheduler.go
```diff
@@ -0,0 +1,131 @@
...
+	serviceAccount, err := serviceAccountForKAIConfig(ctx, readerClient, kaiConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	s.lastDesiredState = []client.Object{serviceAccount}
+	return s.lastDesiredState, nil
+}
+
+func (s *SchedulerForConfig) IsAvailable(ctx context.Context, readerClient client.Reader) (bool, error) {
```
SiorMeir, line 121: Suggestion - Maybe should the SchedulerForConfig funcs be in a separate file? 
Feels a bit confusing to have similarly named funcs in the same file

SiorMeir (approved):

PR #496, opened 2025-09-10
NodeSet plugin infra

omer-dayan:


pkg/scheduler/framework/session_plugins.go
```diff
@@ -317,6 +321,41 @@ func (ssn *Session) IsTaskAllocationOnNodeOverCapacityFn(task *pod_info.PodInfo,
...
+	for _, subsetNodesFn := range ssn.SubsetNodesFns {
+		log.InfraLogger.V(7).Infof(
+			"Running plugin func <%v> on podGroup <%s/%s>", subsetNodesFn, podGroup.Namespace, podGroup.Namespace)
+		newNodeSets := []node_info.NodeSet{}
+		for _, nodeSet := range nodeSets {
+			nodeSubsets, err := subsetNodesFn(podGroup, isRealAllocation, nodeSet)
+			if err != nil {
+				log.InfraLogger.V(6).Infof(
+					"Failed to run SubsetNodes function <%v> on job <%s/%s>", subsetNodesFn, podGroup.Namespace, podGroup.Namespace)
+				return nil, err
```
romanbaron: As an infra function, I don't think it should log here - I suggest to return a meaningful error and let the caller log it instead.

pkg/scheduler/actions/common/allocate.go
```diff
@@ -27,10 +27,26 @@ func AllocateJob(ssn *framework.Session, stmt *framework.Statement, nodes []*nod
 		return false
 	}
 
+	nodeSets, err := ssn.SubsetNodesFn(job, !isPipelineOnly, nodes)
+	if err != nil {
+		log.InfraLogger.V(6).Infof(
+			"Failed to run SubsetNodes on job <%s/%s>", job.Namespace, job.Namespace)
+		return false
```
romanbaron: Since it is going to reject workload scheduling, it should be clearly logged with error logger so it will be visible regardless of currently set log level:
```
log.InfraLogger.Errorf
```

pkg/scheduler/framework/session_plugins.go
omer-dayan: I see that `PrePredicate` + `Predicate` are returning the error + writing logs.
Also they log the error in Infof.

What do you think we should do? Change it in them as well?

pkg/scheduler/framework/session_plugins.go
```diff
@@ -317,6 +321,47 @@ func (ssn *Session) IsTaskAllocationOnNodeOverCapacityFn(task *pod_info.PodInfo,
...
 
+func (ssn *Session) SubsetNodesFn(podGroup *podgroup_info.PodGroupInfo, tasks []*pod_info.PodInfo, initNodeSet node_info.NodeSet) ([]node_info.NodeSet, error) {
+	nodeSets := []node_info.NodeSet{initNodeSet}
+	for _, subsetNodesFn := range ssn.SubsetNodesFns {
+		log.InfraLogger.V(7).Infof(
+			"Running plugin func <%v> on podGroup <%s/%s>", subsetNodesFn, podGroup.Namespace, podGroup.Namespace)
+		newNodeSets := []node_info.NodeSet{}
+		shouldSkipPlugin := false
+		for _, nodeSet := range nodeSets {
+			isRelevant, nodeSubsets, err := subsetNodesFn(podGroup, tasks, nodeSet)
```
romanbaron: The flow here potentially allows a plugin to partition several nodesets and then skip on the next one - it is weird, but possible. Maybe we should introduce two extension points:
1 - one that will check if it is relevant
2- partitions the nodes (only if relevant)

There are cons in this method, because it is an option

omer-dayan: I think its over-engineering the problem.

I see the following options:
1. Revert the new changes - it is the internal plugin responsibility to decide weather it returns new list or the original nodes, pros: Simple, cons: require internal plugin responsibility
2. Keep the current implementation - Check for relevance for each sub-set, pros: not complex, cons: open to unexpected behavior
3. Divide to 2 distinguished hooks, pros: clear well defined API, cons: complex, no other hook works like that

I have to say I start to prefer option 1

PR #511, opened 2025-09-16
feat: add TSDB PVC

SiorMeir:
# Prometheus Operand - Adding a PVC
As we need a long term storage for the metrics gathered by Prometheus, we'll need to add a PersistentVolumeClaim and allow for configuration of the way the data is stored. 

Luckily, the Promtheus Operator comes with a PVC out-of-the-box, so all we need to do is enriching it with our attributes.

We're exposing 2 configurable attributes: `retentionPeriod` and `sampleFrequency` that will allow the user to determine the amount of data gathered. 

From these attributes we're estimating the amount of storage needed to be claimed (rounded to the nearest `Gi`) 

## What is done here
* Adding configurable attributes to a TSDB 
* Setting PVC storage claim based on an estimation formula
* Adding some unit tests


hack/run-e2e-kind.sh
```diff
@@ -44,11 +44,11 @@ while [[ $# -gt 0 ]]; do
   esac
 done
 
-kind create cluster --config ${KIND_CONFIG} --name $CLUSTER_NAME
+# kind create cluster --config ${KIND_CONFIG} --name $CLUSTER_NAME
```
itsomri: ?

pkg/apis/kai/v1/prometheus/prometheus.go
```diff
@@ -5,19 +5,255 @@
...
 	// +kubebuilder:validation:Optional
 	Enabled *bool `json:"enabled,omitempty"`
+
+	// TSDB defines the configuration for Prometheus TSDB storage
+	// +kubebuilder:validation:Optional
+	TSDB *TSDB `json:"tsdb,omitempty"`
+
+	// Status defines the observed state of the Prometheus TSDB
+	// +kubebuilder:validation:Optional
+	Status *TSDBStatus `json:"status,omitempty"`
```
itsomri: This is park of the kai config spec; status does not belong here

pkg/apis/kai/v1/prometheus/prometheus.go
```diff
@@ -5,19 +5,255 @@
...
+		t.SampleFrequency = ptr.To("1m")
+	}
+
+	if t.Connection == nil {
+		t.Connection = &Connection{}
+	}
+}
+
+// CalculateStorageSize estimates the required storage size based on TSDB parameters according to design
+func (t *TSDB) CalculateStorageSize(ctx context.Context, client client.Reader) (string, error) {
```
itsomri: This should be overridable in case our calculations are wrong - maybe let the user override it? @enoodle do you agree?

pkg/apis/kai/v1/prometheus/prometheus.go
```diff
@@ -5,19 +5,255 @@
...
+
+	minutes := int(durationValue.Minutes())
+	if minutes == 0 {
+		minutes = 1 // Ensure minimum 1 minute
+	}
+	return minutes, nil
+}
+
+// parseCustomDuration parses custom duration formats like "2w", "1d", "30m"
+func (t *TSDB) parseCustomDuration(durationStr string) (int, error) {
```
itsomri: I don't get this, what's different between this type of duration and the time.ParseDuration?

pkg/operator/operands/prometheus/resources.go
```diff
@@ -73,18 +75,46 @@ func prometheusForKAIConfig(
...
-	prometheus.Spec = monitoringv1.PrometheusSpec{
-		// Basic configuration - let the Prometheus Operator handle defaults
+	prometheusSpec := monitoringv1.PrometheusSpec{
+		// Basic configuration required for Prometheus Operator to create pods
+		// Using minimal spec to avoid field name issues
 	}
-	logger.Info("Using default Prometheus configuration")
+
+	// Configure TSDB storage if TSDB is configured
+	if config.TSDB != nil {
```
itsomri: Keep your happy flow to the left
```suggestion
	if config.TSDB == nil {
		return []client.Object{}, nil
	}
```

pkg/operator/operands/prometheus/resources.go
```diff
@@ -73,18 +75,46 @@ func prometheusForKAIConfig(
...
+		storageSize, err := config.TSDB.CalculateStorageSize(ctx, runtimeClient)
+		if err != nil {
+			logger.Error(err, "Failed to calculate storage size")
+			return nil, err
+		}
+		prometheusSpec.Storage = &monitoringv1.StorageSpec{
```
itsomri, line 90: Somewhere here you need to allow the admin to configure storageclass

pkg/operator/operands/prometheus/resources.go
```diff
@@ -73,18 +75,46 @@ func prometheusForKAIConfig(
 		return nil, err
 	}
 
-	// If it exists, use the existing spec or merge with our configuration
+	// If it exists, we'll update it with our configuration
 	if err == nil {
-		logger.Info("Prometheus instance already exists", "name", mainResourceName, "namespace", kaiConfig.Spec.Namespace)
-		return []client.Object{prometheus}, nil
+		logger.Info("Prometheus instance already exists, will update with new configuration", "name", mainResourceName, "namespace", kaiConfig.Spec.Namespace)
```
itsomri: I don't see a reason to log this, this is the expected behavior in reconciler pattern

itsomri: In fact, I think there's no reason to even get the current object - the kai-operator framework just needs the objects and will handle the reconcile itself - @enoodle right?

pkg/operator/operands/prometheus/resources.go
itsomri: And it should be wayy up probably - no reason to do anything else if TSDB is disabled

itsomri: Actually why is the TSDB sub-struct necessary? Why not have these properties directly in prometheus?

pkg/operator/operands/prometheus/resources.go
```diff
@@ -73,18 +75,46 @@ func prometheusForKAIConfig(
...
+		}
+
+		// Set retention period if specified
+		if config.TSDB.RetentionPeriod != nil {
+			prometheusSpec.Retention = monitoringv1.Duration(*config.TSDB.RetentionPeriod)
+		}
+	}
+
+	prometheus.Spec = prometheusSpec
+	logger.Info("Using Prometheus configuration with TSDB settings")
```
itsomri: No need of this log really, adds no new info vs the log right after

pkg/operator/operands/prometheus/resources.go
enoodle: :+1: There really is no need for many info logs here because it is wrapped by the operator logic.

enoodle: The reason to get the current object is if we want to allow the user to change other fields that we are not setting (add annotations etc..)
You should use this function to simplify getting the objects:
```
	sa, err := common.ObjectForKAIConfig(ctx, runtimeClient, &v1.ServiceAccount{}, mainResourceName,
		kaiConfig.Spec.Namespace)
```
like here: https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/operator/operands/admission/resources.go#L93

pkg/apis/kai/v1/prometheus/prometheus.go
```diff
@@ -5,19 +5,255 @@
...
+	// Status defines the observed state of the Prometheus TSDB
+	// +kubebuilder:validation:Optional
+	Status *TSDBStatus `json:"status,omitempty"`
 }
 
-func (p *Prometheus) SetDefaultsWhereNeeded() {
+type TSDB struct {
+	// Connection defines the connection configuration for TSDB
+	// +kubebuilder:validation:Optional
+	Connection *Connection `json:"connection,omitempty"`
```
gshaibi: Where is that used?

pkg/apis/kai/v1/prometheus/prometheus.go
enoodle: :+1: Yes, we can add an optional size field.

pkg/operator/operands/prometheus/resources.go
enoodle: I think TSDB struct should be part of global / directly in the spec of kai config.
The prometheus sub struct should be used to create a prometheus only if time decay feature is enabled and no other TSDB is provided.
It could have other fields to help configure the prometheus like storage size, retention etc..

pkg/apis/kai/v1/prometheus/prometheus.go
```diff
@@ -5,19 +5,255 @@
...
+
+	// RetentionPeriod defines how long to retain data (e.g., "2w", "1d", "30d")
+	// +kubebuilder:validation:Optional
+	RetentionPeriod *string `json:"retentionPeriod,omitempty"`
+
+	// SampleFrequency defines the frequency of sampling (e.g., "1m", "30s", "5m")
+	// +kubebuilder:validation:Optional
+	SampleFrequency *string `json:"sampleFrequency,omitempty"`
```
gshaibi: I find `SampleFrequency` a bit confusing, as frequency is associated in my mind with "times per second".
Consider renaming to `SamplingInterval` or something alike if you feel the same.

pkg/operator/operands/prometheus/resources.go
SiorMeir: @itsomri I think a sub-struct is cleaner, no? 
we can flatten this out but it is a logical unit AFAIK

pkg/apis/kai/v1/prometheus/prometheus.go
```diff
@@ -5,19 +5,255 @@
...
 )
 
 type Prometheus struct {
 	// Enabled defines whether a Prometheus instance should be deployed
 	// +kubebuilder:validation:Optional
 	Enabled *bool `json:"enabled,omitempty"`
+
+	// TSDB defines the configuration for Prometheus TSDB storage
+	// +kubebuilder:validation:Optional
+	TSDB *TSDB `json:"tsdb,omitempty"`
```
gshaibi: What are your thoughts on allowing users to configure the entire Prometheus specification, while still providing sensible defaults, instead of limiting them to partial configuration?

hack/run-e2e-kind.sh
SiorMeir: leftover, will be corrected.

pkg/apis/kai/v1/prometheus/prometheus.go
```diff
@@ -5,19 +5,255 @@
...
+	sampleFrequencyMinutes, err := t.parseDurationToMinutes(t.SampleFrequency)
+	if err != nil {
+		logger.Error(err, "Failed to parse sample frequency")
+		return defaultStorageSize, err // Fallback to default
+	}
+
+	// Calculate storage size using the formula
+	sampleSize := 2.0           // [bytes]
+	recordedResourcesLen := 5.0 //
+	storageSizeGi := ((sampleSize * recordedResourcesLen * float64(nodePools) * float64(numQueues) * float64(retentionMinutes)) / float64(sampleFrequencyMinutes)) / (1024 * 1024 * 1024)
```
gshaibi: Does this mean that whenever queue/nodepool count is changing, the Prometheus will be edited to claim a new volume?
If so, will it retain data? Will it restart the Prometheus instance?

pkg/apis/kai/v1/prometheus/prometheus.go
SiorMeir: @itsomri @enoodle cool, will add

pkg/apis/kai/v1/prometheus/prometheus.go
SiorMeir: @itsomri Apperantly the built in `ParseDuration` doesn't support longer time period when parsing (anything longer than hours)
We can use a third-party package instead of this logic, but it seems a bit of an overkill as we're only parsing once, here.

pkg/apis/kai/v1/prometheus/prometheus.go
SiorMeir: @gshaibi Accepted, will change

pkg/operator/operands/prometheus/resources.go
SiorMeir: done, defaulting to the `standard` storageclass if not provided

pkg/operator/operands/prometheus/resources.go
SiorMeir: done, simlified the flow using `ObjectForKAIConfig`

pkg/apis/kai/v1/prometheus/prometheus.go
SiorMeir: This is a provision for external Prometheus instance (not managed by KAI)
To be implemented in the next step

pkg/operator/operands/prometheus/resources.go
SiorMeir: Thought about this - followed @itsomri's suggestion and flatten out the struct to include all the TSDB stuff

pkg/apis/kai/v1/prometheus/prometheus.go
SiorMeir: No. I've tested this and the behavior is sticky - the PVC will not be replaced until the prom deployment is deleted.

pkg/apis/kai/v1/prometheus/prometheus.go
enoodle: We want this to be an opinionated deployment, but also allow users to bring their own prometheus / promQL supporting DB with the connection string.

pkg/apis/kai/v1/prometheus/prometheus.go
```diff
@@ -5,19 +5,259 @@
...
 )
 
 type Prometheus struct {
 	// Enabled defines whether a Prometheus instance should be deployed
 	// +kubebuilder:validation:Optional
 	Enabled *bool `json:"enabled,omitempty"`
+
+	// TSDB defines the configuration for Prometheus TSDB storage
+	// +kubebuilder:validation:Optional
+	TSDB *TSDB `json:"tsdb,omitempty"`
```
enoodle: I think that the TSDB sub struct doesn't make much sense here. 
It holds fields that can be directly under the main struct. I think that we should move the connection fields that are about connecting to external DB to a different location (probably the global or in the main spec of  kai config.) - The reason is that they are not about this Prometheus deployment but will replace it and they make setting all the rest of the fields irrelevant.

SiorMeir: Done. 
`TSDB` struct was flattened & merged into `Prometheus` struct
`Connection` moved to the global config & got a `nil` default

PR #514, opened 2025-09-16
Refactor topology IDE warnings

omer-dayan:
A list of commits

Remove unused topologyPlugin.enabled
Unused arg pluginArgs map[string]string -> _ map[string]string
Unused arg ssn *framework.Session -> _ *framework.Session
Remove unused topologyPlugin field: nodesInfos
Change initializeTopologyTree to not accept twice same parameter (first parameter included in second parameter)
Remove unneeded cast (IDE warning)
Extract "root" to const rootLevel
Extract "root" to const rootDomainId
Remove unused string as root domain name "cluster"
Remove unused field of domainTopologyInfo - depth
Remove unused field of domainTopologyInfo - parent
Remove unused field of domainTopologyInfo - name
Not holding root domain twice
Reduce initializeTopologyTree parameters data to exactly what it needs instead of whole ssn
Rename singleTopology -> topology
Remove prefix Topology from types as the package is called topology (IDE warning)
Remove duplicate validations (All validation is done at the begining isNodePartOfTopology
Remove 1liner function connectDomainToParent
Not using nodeInfo.node.name but nodeInfo.name

### DIFF
--- a/pkg/scheduler/plugins/topology/topology_plugin.go
+++ b/pkg/scheduler/plugins/topology/topology_plugin.go
@@ -40,7 +40,7 @@ func (t *topologyPlugin) initializeTopologyTree(topologies []*kueuev1alpha1.Topo
 	for _, topology := range topologies {
 		topologyTree := &Info{
 			Name: topology.Name,
-			DomainsByLevel: map[DomainLevel]map[DomainID]*DomainInfo{
+			DomainsByLevel: map[DomainLevel]LevelDomainInfos{
 				rootLevel: {
 					rootDomainId: NewDomainInfo(rootDomainId, rootLevel),
 				},
--- a/pkg/scheduler/plugins/topology/topology_structs.go
+++ b/pkg/scheduler/plugins/topology/topology_structs.go
@@ -16,7 +16,9 @@ type DomainID string
 
 type DomainLevel string
 
-type domainsByLevel map[DomainLevel]map[DomainID]*DomainInfo
+type LevelDomainInfos map[DomainID]*DomainInfo
+
+type domainsByLevel map[DomainLevel]LevelDomainInfos
 
 // Info represents a topology tree for the cluster
 type Info struct {
--- a/pkg/scheduler/plugins/topology/job_filtering_test.go
+++ b/pkg/scheduler/plugins/topology/job_filtering_test.go
@@ -111,7 +111,7 @@ func TestTopologyPlugin_subsetNodesFn(t *testing.T) {
 							},
 						},
 					},
-					DomainsByLevel: map[DomainLevel]map[DomainID]*DomainInfo{
+					DomainsByLevel: map[DomainLevel]LevelDomainInfos{
 						"rack": {
 							"rack1.zone1": {
 								ID:    "rack1.zone1",
@@ -306,7 +306,7 @@ func TestTopologyPlugin_subsetNodesFn(t *testing.T) {
 							},
 						},
 					},
-					DomainsByLevel: map[DomainLevel]map[DomainID]*DomainInfo{
+					DomainsByLevel: map[DomainLevel]LevelDomainInfos{
 						"zone": {
 							"zone1": {
 								ID:    "zone1",
@@ -360,7 +360,7 @@ func TestTopologyPlugin_subsetNodesFn(t *testing.T) {
 							},
 						},
 					},
-					DomainsByLevel: map[DomainLevel]map[DomainID]*DomainInfo{
+					DomainsByLevel: map[DomainLevel]LevelDomainInfos{
 						"rack": {
 							"rack1.zone1": {
 								ID:    "rack1.zone1",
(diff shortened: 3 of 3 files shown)


pkg/scheduler/plugins/topology/job_filtering.go
```diff
@@ -17,10 +17,10 @@ import (
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/log"
 )
 
-type domainsByLevel map[string]map[TopologyDomainID]*TopologyDomainInfo
+type domainsByLevel map[string]map[DomainID]*DomainInfo
```
romanbaron: Maybe it will be nice to create a dedicated type for the domain level as well?
Something like:
```
type domainsByLevel map[DomainLevel]map[DomainID]*DomainInfo
```

pkg/scheduler/plugins/topology/topology_plugin.go
```diff
@@ -32,70 +31,64 @@ func (t *topologyPlugin) Name() string {
...
-	for levelIndex := len(singleTopology.Spec.Levels) - 1; levelIndex >= 0; levelIndex-- {
-		level := singleTopology.Spec.Levels[levelIndex]
+	var nodeContainingChildDomain *DomainInfo
+	for levelIndex := len(topology.Spec.Levels) - 1; levelIndex >= 0; levelIndex-- {
+		level := topology.Spec.Levels[levelIndex]
 
-		domainName, foundLevelLabel := nodeInfo.Node.Labels[level.NodeLabel]
-		if !foundLevelLabel {
-			continue // Skip if the node is not part of this level
-		}
```
romanbaron, line 75: not needed?

pkg/scheduler/plugins/topology/topology_structs.go
```diff
@@ -28,63 +26,42 @@ type TopologyInfo struct {
...
-
-	// Name of this domain
-	Name string
+	ID DomainID
 
 	// Level in the hierarchy (e.g., "datacenter", "zone", "rack", "node")
 	Level string
 
-	// Parent domain, nil for root
-	Parent *TopologyDomainInfo
```
romanbaron, line 43: What about the option of storing parent domain id?

pkg/scheduler/plugins/topology/topology_structs.go
```diff
@@ -28,63 +26,42 @@ type TopologyInfo struct {
...
 
-func calcDomainId(leafLevelIndex int, levels []kueuev1alpha1.TopologyLevel, nodeLabels map[string]string) TopologyDomainID {
+func calcDomainId(leafLevelIndex int, levels []kueuev1alpha1.TopologyLevel, nodeLabels map[string]string) DomainID {
 	domainsNames := make([]string, leafLevelIndex+1)
 	for levelIndex := leafLevelIndex; levelIndex >= 0; levelIndex-- {
 		levelLabel := levels[levelIndex].NodeLabel
-		levelDomainName, foundLevelOnNode := nodeLabels[levelLabel]
-		if !foundLevelOnNode {
-			levelDomainName = "missing"
-		}
```
romanbaron, line 81: I am not following this part

pkg/scheduler/plugins/topology/topology_plugin.go
omer-dayan: This code is testing if a label exist on a node, and if not, continue to next node.

If you look at the begining of this function you see it first call `isNodePartOfTopology`, which internally validates that each label of the topology exist on the node.
https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/scheduler/plugins/topology/topology_plugin.go#L102

This code perform the test again, unnecesserily

pkg/scheduler/plugins/topology/topology_structs.go
omer-dayan: `DomainInfo` struct has a field of `Children`.
Having a parent causing 2 sided pointing.

Generally, I am not against such code, if you have a good reason.
But in this case, the `Parent` was exist, but it was never used.

So why not keeping it single point of direction? Easier to manage/change/etc, and no harm in code at all...

pkg/scheduler/plugins/topology/topology_structs.go
omer-dayan: Like in this comment: https://github.com/NVIDIA/KAI-Scheduler/pull/514#discussion_r2357954907

This code check if a label exist on a node, but it actually runs after we call `isNodePartOfTopology` which already doing so

pkg/scheduler/plugins/topology/job_filtering.go
omer-dayan: Done

omer-dayan: I actually a bit uncomfortable with the map of maps. It looks weird.

Do you think its worth having `map[DomainID]*DomainInfo` as some type X and then
`domainsByLevel map[DomainLevel]X`?

romanbaron: Ok, looks good to me

pkg/scheduler/plugins/topology/topology_plugin.go
romanbaron: Is it the same? in `isNodePartOfTopology` it checks that node is part of some topology. here it skips specific levels.

pkg/scheduler/plugins/topology/topology_structs.go
romanbaron: Ok, if not in use that it is not worth discussing

romanbaron (dismissed):

PR #517, opened 2025-09-17
Binder env tests

itsomri:
This PR adds the infra for running the binder in env tests, and adds one simple sanity test.

### DIFF
--- a/cmd/binder/app/app.go
+++ b/cmd/binder/app/app.go
@@ -132,7 +132,7 @@ func (app *App) RegisterPlugins(plugins *plugins.BinderPlugins) {
 // +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get;list;watch;create;update;patch;delete
 
-func (app *App) Run(stopCh chan struct{}) error {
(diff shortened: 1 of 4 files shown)


cmd/binder/app/app.go
```diff
@@ -186,16 +168,23 @@ func (app *App) Run() error {
 	}
 	// +kubebuilder:scaffold:builder
 
+	// Create a context that cancels on either OS signals or when stopCh is closed
+	ctx, cancel := context.WithCancel(ctrl.SetupSignalHandler())
+	defer cancel()
+	go func() {
+		<-stopCh
+		cancel()
+	}()
```
gshaibi: IMHO this might belong to the caller who creates the stopCh.
Consider passing a ctx from the caller and using ctx.Done() instead of the stopCh to trigger the stop in case the context is cancelled (by a signal).

gshaibi: But this existed before your PR so feel free to ignore 😅

enoodle: I think it makes more sense than the original where we had a `stopCh` and never stopped it. ;)
Here we listen on the `stopCh` and act if it says to stop.

itsomri: The stopCh is used by the informers so I still need it

itsomri: I removed the redundant `defer cancel()` though

PR #518, opened 2025-09-17
Hierarchical subgroup structure

romanbaron:
This PR extends the subgroup structure in podgroup CRD by adding a reference to parent subgroup.
It will allow us adding topology constraint at the subgroup set level in the future.
As part of this change podgroup validating webhook logic was extended to detect possible cycles in the subgroup structure.


pkg/apis/scheduling/v2alpha2/podgroup_types.go
```diff
@@ -77,6 +77,9 @@ type SubGroup struct {
 	// MinMember defines the minimal number of members to run this SubGroup;
 	// if there are not enough resources to start all required members, the scheduler will not start anyone.
 	MinMember int32 `json:"minMember,omitempty"`
+
+	// Parent is an optional attribute that specifies the name of the parent SubGroup
+	Parent string `json:"parent,omitempty"`
```
omer-dayan: I remeber there is an issue with pointer in go crd struct.

I think that if you have it `Parent *string` it means that the field is optional, and may not exist in a cr.
And if you have it `Parent string` it will alway be in the cr, but as empty string.

If so, dont we prefer the first option?

pkg/apis/scheduling/v2alpha2/podgroup_webhook.go
```diff
@@ -5,50 +5,148 @@ package v2alpha2
...
+	}
+
+	if err := validateParent(subGroupMap); err != nil {
+		return err
+	}
+
+	if detectCycle(subGroupMap) {
+		return errors.New("cycle detected in subgroups")
+	}
+	return nil
```
omer-dayan, line 83: Do we allow for multiple "root"s? (Empty parent)

If so, what is the expected behavior in term of scheduling? All the roots must be satisfied?

pkg/apis/scheduling/v2alpha2/podgroup_webhook.go
```diff
@@ -5,50 +5,148 @@ package v2alpha2
...
+}
+
+func validateSubGroupFields(subGroup *SubGroup) error {
+	if subGroup.Name == "" {
+		return fmt.Errorf("subgroup name cannot be empty")
```
omer-dayan: Are you sure there is no markup validation that can check emptyness?
https://book.kubebuilder.io/reference/markers/crd-validation

And minMember > 0 as well

pkg/apis/scheduling/v2alpha2/podgroup_webhook.go
```diff
@@ -5,50 +5,148 @@ package v2alpha2
...
+	graph := map[string][]string{}
+	for _, subGroup := range subGroupMap {
+		graph[subGroup.Parent] = append(graph[subGroup.Parent], subGroup.Name)
+	}
+
+	visited := map[string]bool{}
+	recStack := map[string]bool{}
+
+	for name := range subGroupMap {
+		if !visited[name] {
```
omer-dayan: You have such check in line 137 I would remove it from here, as both function are tightly coupled anyway

pkg/apis/scheduling/v2alpha2/podgroup_types.go
romanbaron: I will change it to pointer for clarity.

pkg/apis/scheduling/v2alpha2/podgroup_webhook.go
romanbaron: I was thinking that the root is the PodGroup and subgroups without a parent ref are it's direct children.
If we agree on that, PodGroup's minMember specifies gang semantics for these subgroups.

As a beginning, for topology constraints, we can start with "all required" approach.

pkg/apis/scheduling/v2alpha2/podgroup_types.go
enoodle: `// +kubebuilder:validation:Optional` -  specifies that this field is optional.

pkg/apis/scheduling/v2alpha2/podgroup_webhook.go
```diff
@@ -5,50 +5,137 @@ package v2alpha2
...
+	subGroupMap := map[string]*SubGroup{}
+	for i := range subGroups {
+		name := subGroups[i].Name
+		subGroup := &subGroups[i]
+
+		if subGroupMap[name] != nil {
+			return fmt.Errorf("duplicate subgroup name %s", name)
+		}
+		subGroupMap[name] = subGroup
+	}
```
itsomri:
```suggestion
for _, sg := range subGroups {
		if subGroupMap[sg.Name] != nil {
			return fmt.Errorf("duplicate subgroup name %s", sg.Name)
		}
		subGroupMap[sg.Name] = &sg
	}
```

itsomri (approved):

PR #521, opened 2025-09-17
Priority-Preemptibility Separation Design

gshaibi:


docs/developer/designs/priority-preemptibility-separation/README.md
```diff
@@ -0,0 +1,192 @@
...
+   * [Example 2: Low-Priority Non-Preemptible PodGroup](#example-2-low-priority-non-preemptible-podgroup)
+   * [Example 3: External Workload with Explicit Preemptibility](#example-3-external-workload-with-explicit-preemptibility)
+   * [Example 4: Default Behavior PodGroup (Backward Compatible)](#example-4-default-behavior-podgroup-backward-compatible)
+
+
+---
+
+## Background
+
+Currently, Run:ai users can submit workloads with associated priority classes that apply to all subordinate pods. The scheduler implicitly assumes that all workloads using priority classes lower than 100 are preemptible, while workloads using priority classes higher than or equal to 100 are non-preemptible.
```
enoodle:
```suggestion
Currently, KAI-scheduler users can submit workloads with associated priority classes that apply to all subordinate pods. The scheduler implicitly assumes that all workloads using priority classes lower than 100 are preemptible, while workloads using priority classes higher than or equal to 100 are non-preemptible.
```

docs/developer/designs/priority-preemptibility-separation/README.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+```go
+type PodGroupSpec struct {
+    // ... existing fields ...
+    
+    // Preemptibility defines whether this PodGroup can be preempted
+    // +kubebuilder:validation:Enum=preemptible;non-preemptible;semi-preemptible
+    // +optional
+    Preemptibility string `json:"preemptibility,omitempty"`
```
enoodle, line 90: Maybe it should be `*string` if it is optional

gshaibi: I also thought about it but then noticed that `PriorityClassName` is also optional and not a pointer, and checks zero value against an empty string, so I kept the convention.
What do you think?

docs/developer/designs/priority-preemptibility-separation/README.md
gshaibi: oops... thanks ^_^

docs/developer/designs/priority-preemptibility-separation/README.md
enoodle: If it works well with the enum validation then its OK because then it will have 4 possible values, with empty string meaning that we go for the old option. :+1:

enoodle (approved):

enoodle (approved):

PR #525, opened 2025-09-18
fix: impove docs and READMEs

SiorMeir:


CONTRIBUTING.md
```diff
@@ -37,7 +37,7 @@ Help us keep the docs clear and useful by fixing typos, updating outdated inform
 ### Pull Request Checklist
 Before introducing major changes, we strongly recommend opening a PR that outlines your proposed design.
 Each pull request should meet the following requirements:
-- All tests pass – Run the full test suite locally with: `make build validate test`
+- All tests pass – Run the full test suite locally with: `make build && validate test`
```
itsomri: Why? Doesn't `make` handle this itself?

README.md
```diff
@@ -82,6 +80,10 @@ To start scheduling workloads with KAI Scheduler, please continue to [Quick Star
 
 We’d love to hear from you! Here are the best ways to connect:
 
+### Contributing
+Contributions are encouraged and appreciated! 
+Please have a look at KAI-scheduler's [contribution guide](https://github.com/NVIDIA/KAI-Scheduler/blob/main/CHANGELOG.md) before submitting PRs.
```
itsomri: How is the CHANGELOG a contribution guide?

docs/fairness/README.md
```diff
@@ -1,42 +1,56 @@
...
 
-## Fair Share
-Once the fair share for each queue is calculated, it serves two primary purposes:
-1. Queue Order - Queues with a fair share further below their allocation will be prioritized for scheduling.
-2. Reclaim action – When reclamation is required, the scheduler compares the **Saturation Ratio** (`Allocated / FairShare`) of queues that share the same parent. A queue can only reclaim resources if, **after** the transfer, its utilisation ratio remains lower than that of every sibling queue. For more details see the reclaim strategies.
+> Resource Allocation is done on each scheduling cycle
+
+Resources are allocated hierarchically across queue levels:
+
+1. **Quota allocation**: Guaranteed resources distributed first
```
itsomri:
```suggestion
1. **Quota allocation**: Deserved resources are assigned first
```

docs/fairness/README.md
```diff
@@ -1,42 +1,56 @@
 # Fairness
```
itsomri, line 1: I'm missing some explanation about how the "fair share" is used in the queue: try to explain that for each cycle, each queue is assigned a fair share of the cluster's resources, and the scheduler attempts to allocate jobs so that the actual allocation matches as closely as possible the fair share allocation.

docs/fairness/README.md
```diff
@@ -1,42 +1,56 @@
...
+2. **Over-quota distribution**: Remaining resources allocated by priority and weight
+3. **Hierarchical propagation**: Process repeated at each queue level
+
+## Fair Share Calculation
+
+Fair share determines queue scheduling priority and reclaim eligibility:
+
+- **Scheduling Priority**: Queues below fair share are prioritized
+- **Saturation Ratio**: `Allocated / FairShare` used for reclaim decisions
+- **Reclaim Eligibility**: Queues can only reclaim if their saturation ratio remains lowest among siblings
```
itsomri: ?? this is not true at all

itsomri: It's true that we **attempt** this, but it's not guaranteed

docs/fairness/README.md
```diff
@@ -1,42 +1,56 @@
...
-There are two main reclaim strategies:
-1. Workloads from queues with resources below their fair share can evict workloads from queues that have exceeded their fair share.
-2. Workloads from queues under their quota can evict workloads from queues that have exceeded their quota.
+KAI scheduler uses two main reclaim strategies:
+1. **Fair Share Reclaim** - Workloads from queues with resources below their fair share can evict workloads from queues that have exceeded their fair share.
+2. **Quota Reclaim** - Workloads from queues under their quota can evict workloads from queues that have exceeded their quota.
 
-In both strategies, the scheduler ensures that the relative ordering is preserved: a queue that had the lowest utilisation ratio in its level before reclamation will still have the lowest ratio afterwards. Likewise, a queue that was below its quota will remain below its quota.
+In both strategies, the scheduler ensures that the **relative ordering is preserved**: a queue that had the lowest utilisation ratio in its level before reclamation will still have the lowest ratio afterwards. Likewise, a queue that was below its quota will remain below its quota.
```
itsomri: This is not true (wasn't true before your changes). There are cases where a queue will be more starved then it's siblings, but will fail to reclaim, while one of it's siblings succeeds. Need to fix this. Also, utilisation is not good in this context because it might invoke "gpu utilisation", while the intent is queue saturation (or starvation). Also utilisation is a typo.

docs/fairness/README.md
```diff
@@ -1,42 +1,56 @@
...
-2. Reclaim action – When reclamation is required, the scheduler compares the **Saturation Ratio** (`Allocated / FairShare`) of queues that share the same parent. A queue can only reclaim resources if, **after** the transfer, its utilisation ratio remains lower than that of every sibling queue. For more details see the reclaim strategies.
+> Resource Allocation is done on each scheduling cycle
+
+Resources are allocated hierarchically across queue levels:
+
+1. **Quota allocation**: Guaranteed resources distributed first
+2. **Over-quota distribution**: Remaining resources allocated by priority and weight
+3. **Hierarchical propagation**: Process repeated at each queue level
+
+## Fair Share Calculation
```
itsomri: There's no actual explanation about how we calculate fair share, I'm missing a part were we explain the actual algorithm (divide deserved resources, then go by priorities, in each priority divide by normalized weight for unsatisfied queues, leftover resources for each priority go to the next...)

docs/fairness/README.md
itsomri: Maybe we can also refer to the fairness simulator?

README.md
SiorMeir: woops.
This should point to the contribution guide, this is a typo

docs/fairness/README.md
SiorMeir: @itsomri I had a feeling this is a bit lacking. I'll try add a bit more.

This definitly should point to the simulator. In fact I think that the sim should be the main focus on this (when ready)

docs/fairness/README.md
SiorMeir: will fix

CONTRIBUTING.md
SiorMeir: it is, thought otherwise.
reverting

docs/fairness/README.md
SiorMeir: @itsomri added a note on this.
AFAIK this is a part of our algorithmic solution so this is just an FYI for the user

docs/fairness/README.md
SiorMeir: @itsomri Corrected - please have a look to see if it makes more sense

docs/fairness/README.md
SiorMeir: Added psudeo-code from one of our knowledge-share sessions.r

docs/fairness/README.md
```diff
@@ -1,42 +1,77 @@
...
+    remainingResources -= min(q.deserved, requested)
+```
+> Fair share calculation:
+```python
+while remainingResources > 0:
+    totalWeights = sum(q.OverQuotaWeight for q in queues)
+    for q in queues:
+        resourcesToAssign = min(remainingResources * q.OverQuotaWeight / totalWeights, q.RemainingRequest)
+        q.fairShare += resourcesToAssign
+        remainingResources -= resourcesToAssign
```
davidLif, line 56: What about fair-share over time? Maybe we should add it as a seperate example?

docs/fairness/README.md
itsomri: Let's add a more human-readable introduction. Keep in mind that the reader might not be very familiar with the concept of fair share, or might be familiar with other implementations.

Also, describing the algorithm is good, but it might not be enough to convey intuitive understanding of the expected result.

Some things that are important to convey:
1. Fair share is a "nominal" resource allocation. It's an amount of resources that the scheduler assigns each queue with the intent of dividing the resources that way, but it's not guaranteed to be the exact result of the scheduling cycle.
2. The fair share algorithm allows us to lend resources from queues that deserve them but don't need them at the moment, to queues that might not deserve them but do need them

Order of Guarantees:
1. If a queue has deserved quota, and requests resources, then the fair share is guarantee to fill it's request, up to it's deserved quota. In other words, deserved quota gets assigned first.
2. For over quota resources, queues with higher priority will get their over quota resources before queues with lower priority.
3. For queues in the same priority, quota will be divided proportionally to the weights of the queues
4. For each step of the algorithm, once a queue's resource request is satisfied, it's no longer considered, and the rest of it's deserved resources (if it has any) can be assigned to the rest of the queues


You don't have to stick to this structure - these are just important points to explain, and they should be conveyed even to the reader that will not dig deeper into the algorithm

itsomri: (You covered some of these well in the algorithm description, I'm just not convinced that it's easy to understand for someone who has no background)

docs/fairness/README.md
SiorMeir: In my opinion this is really hard to put into an understandable text.
We'll have the sim & the charts that appear in the design doc

docs/fairness/README.md
SiorMeir: Broke the structure a bit.
Added a `philosophy` header for quick-to-understand points & added the guarantees as part of the flow.

enoodle (approved):

PR #526, opened 2025-09-18
shaibi/priority preemptability separation p0 implementation

gshaibi:
- **TODOs**
- **PodGroup API change**
- **API Doc**
- **Handle PodGrouper**
- **Handle preemptability in PodGroupController and Scheduler**

### DIFF
--- a/pkg/podgroupcontroller/utilities/pod-group/preemptible.go
+++ b/pkg/podgroupcontroller/utilities/pod-group/preemptible.go
@@ -20,6 +20,7 @@ const (
 )
 
 func IsPreemptible(ctx context.Context, podGroup *v2alpha2.PodGroup, kubeClient client.Client) (bool, error) {
+	// GuyTODO: Change to use the new Preemptible field, defaults this old behavior
 	priority, err := getPodGroupPriority(ctx, podGroup, kubeClient)
 	if errors.IsNotFound(err) {
 		logger := log.FromContext(ctx)
--- a/pkg/podgroupcontroller/utilities/pod-group/preemptible.go
+++ b/pkg/podgroupcontroller/utilities/pod-group/preemptible.go
@@ -5,14 +5,15 @@ package pod_group
 
 import (
 	"context"
+	"fmt"
 
 	schedulingv1 "k8s.io/api/scheduling/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/NVIDIA/KAI-scheduler/pkg/apis/scheduling/v2alpha2"
+
+	pg "github.com/NVIDIA/KAI-scheduler/pkg/common/resources/podgroup"
 )
 
 const (
@@ -20,23 +21,14 @@ const (
 )
 
 func IsPreemptible(ctx context.Context, podGroup *v2alpha2.PodGroup, kubeClient client.Client) (bool, error) {
-	// GuyTODO: Change to use the new Preemptible field, defaults this old behavior
-	priority, err := getPodGroupPriority(ctx, podGroup, kubeClient)
-	if errors.IsNotFound(err) {
-		logger := log.FromContext(ctx)
-		logger.Info(
-			"Priority Class not found",
-			"podGroup", podGroup.Name,
-			"namespace", podGroup.Namespace,
-			"priorityClass", podGroup.Spec.PriorityClassName,
-		)
-		return true, nil
-	}
+	isPreemptible, err := pg.IsPreemptible(podGroup.Spec.Preemptibility, func() (int32, error) {
+		return getPodGroupPriority(ctx, podGroup, kubeClient)
+	})
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("failed to determine podgroup's preemptibility: %w", err)
 	}
 
-	return priority < PriorityBuildNumber, nil
+	return isPreemptible, nil
 }
 
 func getPodGroupPriority(ctx context.Context, podGroup *v2alpha2.PodGroup, kubeClient client.Client) (int32, error) {
--- a/pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
+++ b/pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
@@ -69,6 +69,7 @@ func (dg *DefaultGrouper) GetPodGroupMetadata(topOwner *unstructured.Unstructure
 		Queue:             dg.CalcPodGroupQueue(topOwner, pod),
 		PriorityClassName: dg.CalcPodGroupPriorityClass(topOwner, pod, constants.TrainPriorityClass),
 		MinAvailable:      1,
+		// GuyTodo: Add Preemptibility field
 	}
 
 	annotations := topOwner.GetAnnotations()
--- a/pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
+++ b/pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
@@ -17,6 +17,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
+	"github.com/NVIDIA/KAI-scheduler/pkg/apis/scheduling/v2alpha2"
 	commonconsts "github.com/NVIDIA/KAI-scheduler/pkg/common/constants"
 
 	"github.com/NVIDIA/KAI-scheduler/pkg/podgrouper/podgroup"
@@ -68,8 +69,8 @@ func (dg *DefaultGrouper) GetPodGroupMetadata(topOwner *unstructured.Unstructure
 		Labels:            dg.CalcPodGroupLabels(topOwner, pod),
 		Queue:             dg.CalcPodGroupQueue(topOwner, pod),
 		PriorityClassName: dg.CalcPodGroupPriorityClass(topOwner, pod, constants.TrainPriorityClass),
+		Preemptibility:    dg.CalcPodGroupPreemptibility(topOwner, pod),
 		MinAvailable:      1,
-		// GuyTodo: Add Preemptibility field
 	}
 
 	annotations := topOwner.GetAnnotations()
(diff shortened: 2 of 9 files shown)


pkg/common/podgroup/preemptible.go
```diff
@@ -0,0 +1,32 @@
+// Copyright 2025 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+package podgroup
+
+import (
+	"fmt"
+
+	"github.com/NVIDIA/KAI-scheduler/pkg/apis/scheduling/v2alpha2"
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/constants"
```
enoodle: Maybe not as part of this PR but the contents of pkg/scheudler/constants/priorities.go should move into pkg/common/constants

test/e2e/suites/preempt/preempt_test.go
```diff
@@ -25,7 +28,7 @@ import (
 	"github.com/NVIDIA/KAI-scheduler/test/e2e/modules/wait"
 )
 
-var _ = Describe("Priority Preemption", Ordered, func() {
+var _ = FDescribe("Priority Preemption", Ordered, func() {
```
enoodle: I think this change is by mistake.

gshaibi: You're right, thanks

enoodle (approved):

pkg/common/constants/constants.go
```diff
@@ -10,6 +10,7 @@ const (
 	UnlimitedResourceQuantity = float64(-1)
 
 	DefaultQueuePriority                  = 100
+	DefaultPodGroupPriority               = 50
```
itsomri: This is usually determined by the default priority class in the cluster - why change it?

pkg/podgroupcontroller/utilities/pod-group/preemptible.go
```diff
@@ -5,48 +5,84 @@ package pod_group
...
+
+	// Try to get the specific priority class
+	if priority, err := getSpecificPriorityClass(ctx, podGroup.Spec.PriorityClassName, kubeClient); err == nil {
+		return priority, nil
+	} else if !errors.IsNotFound(err) {
+		return -1, fmt.Errorf("failed to get priority class %s: %w", podGroup.Spec.PriorityClassName, err)
+	}
+
+	// Priority class not found, try to get global default
+	logger.Info("Priority Class not found, trying to get global default priority class")
```
itsomri, line 48: specify the name of the missing priority class, and for which podgroup

pkg/podgroupcontroller/utilities/pod-group/preemptible.go
```diff
@@ -5,48 +5,84 @@ package pod_group
...
 func getPodGroupPriority(ctx context.Context, podGroup *v2alpha2.PodGroup, kubeClient client.Client) (int32, error) {
+	logger := log.FromContext(ctx,
+		"podGroup", podGroup.Name,
+		"namespace", podGroup.Namespace,
+		"priorityClassName", podGroup.Spec.PriorityClassName)
+
+	// Try to get the specific priority class
+	if priority, err := getSpecificPriorityClass(ctx, podGroup.Spec.PriorityClassName, kubeClient); err == nil {
+		return priority, nil
+	} else if !errors.IsNotFound(err) {
```
itsomri: no need for `else if` because of the return directive

pkg/podgroupcontroller/utilities/pod-group/preemptible.go
```diff
@@ -5,48 +5,84 @@ package pod_group
...
+		return -1, fmt.Errorf("failed to get global default priority class: %w", err)
+	}
+
+	// No global default found, use system default
+	logger.Info("Global default priority class not found, using default priority")
+	return constants.DefaultPodGroupPriority, nil
+}
+
+func getSpecificPriorityClass(ctx context.Context, priorityClassName string, kubeClient client.Client) (int32, error) {
```
itsomri, line 62: Honestly I don't really think we need a function here, no? It's just calls kubeClient.Get

pkg/podgroupcontroller/utilities/pod-group/preemptible.go
```diff
@@ -5,48 +5,84 @@ package pod_group
...
+	}
+
+	for _, priorityClass := range priorityClasses.Items {
+		if priorityClass.GlobalDefault {
+			return priorityClass.Value, nil
+		}
+	}
+
+	// No global default found
+	return -1, errors.NewNotFound(schedulingv1.Resource("priorityclass"), "")
```
itsomri, line 90: It's ok to not find a global priority class, it's not guaranteed to exist, we can just use 50 instead

pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
```diff
@@ -176,6 +178,24 @@ func (dg *DefaultGrouper) CalcPodGroupPriorityClass(topOwner *unstructured.Unstr
...
 }
 
+func (dg *DefaultGrouper) calcPodGroupPreemptibility(topOwner *unstructured.Unstructured, pod *v1.Pod) v2alpha2.Preemptibility {
+	if preemptibility, found := topOwner.GetLabels()[constants.PreemptibilityLabelKey]; found {
+		if preemptibility, err := v2alpha2.ParsePreemptibility(preemptibility); err == nil {
+			return preemptibility
+		} else {
+			logger.Error(err, "Invalid preemptibility label found on top owner", "topOwner", topOwner.GetName())
+		}
+	} else if preemptibility, found = pod.GetLabels()[constants.PreemptibilityLabelKey]; found {
```
itsomri, line 188: Why lookup twice in the map?

pkg/scheduler/cache/cluster_info/cluster_info.go
```diff
@@ -328,6 +330,10 @@ func (c *ClusterInfo) snapshotPodGroups(
 		log.InfraLogger.V(7).Infof("The priority of job <%s/%s> is <%s/%d>", podGroup.Namespace, podGroup.Name,
 			podGroup.Spec.PriorityClassName, podGroupInfo.Priority)
 
+		podGroupInfo.Preemptibility = getPodGroupPreemptibility(podGroup.Spec.Preemptibility, podGroupInfo.Priority)
+		log.InfraLogger.V(7).Infof("The preemptibility of job <%s/%s> is <%s>", podGroup.Namespace, podGroup.Name,
```
itsomri, line 334: Let's print the priority & preemptibility in the same line

itsomri: Maybe it makes more sense to print "Job X is preemptible" or "Job X is non-preemptible"
Will look better than "The preemptibility of job X is false"

pkg/scheduler/cache/cluster_info/cluster_info.go
```diff
@@ -414,7 +420,7 @@ func (c *ClusterInfo) snapshotTopologies() ([]*kueue.Topology, error) {
 }
 
 func getDefaultPriority(dataLister data_lister.DataLister) (int32, error) {
```
itsomri, line 422: I have deja vu from https://github.com/NVIDIA/KAI-Scheduler/pull/526/files#diff-a4d20671885379d4aaca001082a6e14c91251db3bdc4e4c41de7da0a0a2d74ecR73 - please consolidate these two

pkg/scheduler/cache/cluster_info/cluster_info.go
```diff
@@ -449,6 +455,15 @@ func getPodGroupPriority(
 	return chosenPriorityClass.Value
 }
 
+func getPodGroupPreemptibility(preemptibility enginev2alpha2.Preemptibility, priority int32) enginev2alpha2.Preemptibility {
```
itsomri: This function feels redundant - it just ignores the error, you can do that in `snapshotPodGroups` as well

pkg/common/constants/constants.go
gshaibi: You're right, it is, but if such default doesn't exist in the cluster we used to fallback to 50, I just declared it as a constant for clarity.

pkg/podgroupcontroller/utilities/pod-group/preemptible.go
gshaibi: This should already be contained in the context - https://github.com/NVIDIA/KAI-Scheduler/blob/907d09e83422c84c6b9a07a9393cca2095120e3d/pkg/podgroupcontroller/utilities/pod-group/preemptible.go#L34 no?

pkg/podgroupcontroller/utilities/pod-group/preemptible.go
gshaibi: It is, but imho this makes `getPodGroupPriority` function more readable.

pkg/podgroupcontroller/utilities/pod-group/preemptible.go
gshaibi: You're right, that is the current behaviour

pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
gshaibi: I'm can't find the other place we lookup in the map, can you please help me with a reference?

pkg/scheduler/cache/cluster_info/cluster_info.go
gshaibi: The value is not true/false, it is an enum string that may be "preemptible"/"non-preemptible"/"semi-preemptible".

Do you think we should print a similar log to the privious priority log or change it to what you suggested?

pkg/scheduler/cache/cluster_info/cluster_info.go
gshaibi: Yes, it is very much alike.
I thought about unifying these two but this one uses dataLister and the other place use a kubeClient, and half of this function logic is to list the priority classes using it.
Do we have a common interface for both?

pkg/scheduler/cache/cluster_info/cluster_info.go
gshaibi: You're right, not much meat here - its just to make the `snapshotPodGroups` function more readable

pkg/common/constants/constants.go
gshaibi: I've added a comment to clarify that

pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
itsomri: Sorry, didn't realize that one was for pod and one for topOwner

pkg/scheduler/cache/cluster_info/cluster_info.go
itsomri: It will not be needed if `pg.CalculatePreemptibility` has no error return value

PR #527, opened 2025-09-18
Topology scheduling - on require single relevant level

omer-dayan:


pkg/scheduler/actions/integration_tests/allocate/allocate_topology_test.go
```diff
@@ -488,7 +488,7 @@ func getTopologyTestsMetadata() []integration_tests_utils.TestTopologyMetadata {
 						DontValidateGPUGroup: true,
 					},
 					"pending_job0-0": {
-						NodeName:             "node1",
```
omer-dayan, line 491: Because we dont choose smaller level than what is required, we will now cheese exactly what is required.

The tree is rack, and under that spines.

Until now even though the require was rack, we choose a spine, thats why both pending pods got to node1.
From now, we choose rack, and within that rack we perform node bin-packing so first pod gets to node0 and second to node1.

romanbaron (approved):

PR #531, opened 2025-09-18
Topology plugin - Filter out worse case domains

omer-dayan:
In this PR we change the mechanism that calculates allocatable pods.

Instead of using `IsTaskAllocatable` we use `IsTaskAllocatableOnReleasingOrIdle`.

This mechanism goal is to filter out domains that can have the job in terms of resources.
If `IsTaskAllocatableOnReleasingOrIdle` is false, it means for sure that `IsTaskAllocatable` is false.
But not the other way arround.

That means we filter out less domains with that function.
Which is good, as this mechanism is only optimization, but it filters out domains, so we prefer false-positive then false-negative


romanbaron (approved):

romanbaron:
Can you also add some tests for this specific scenario?

omer-dayan (commented):
Added unit test inside the package.

As I think the "IsAllocatable" or "OnReleasingOrIdle" for choosing optimization is internal logic and no behavior

romanbaron (approved):

PR #534, opened 2025-09-25
Prepare infra for time aware env tests

itsomri:
* Init default actions & plugins outside the scheduler, to allow injecting scheduler config externally
* Changed binder metrics port in env tests
* Allow configuration of informer params from UsageDB params (fetch interval, staleness period, timeout)

### DIFF
--- a/pkg/env-tests/binder/binder.go
+++ b/pkg/env-tests/binder/binder.go
@@ -20,7 +20,7 @@ import (
 func RunBinder(cfg *rest.Config, ctx context.Context) error {
 	options := app.InitOptions(pflag.NewFlagSet("binder-test", pflag.ContinueOnError))
 
-	options.MetricsAddr = ":8084"
+	options.MetricsAddr = ":8085"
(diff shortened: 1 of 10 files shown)


pkg/env-tests/binder/binder.go
```diff
@@ -20,7 +20,7 @@ import (
 func RunBinder(cfg *rest.Config, ctx context.Context) error {
 	options := app.InitOptions(pflag.NewFlagSet("binder-test", pflag.ContinueOnError))
 
-	options.MetricsAddr = ":8084"
+	options.MetricsAddr = ":8085"
```
SiorMeir, line 23: why?

SiorMeir (approved):

pkg/env-tests/binder/binder.go
itsomri: To avoid collision with the other controllers, you can't run more than one controller which tries to use the same port

PR #538, opened 2025-09-28
Moved SubGroupInfo into a separate package

romanbaron:


omer-dayan (dismissed):
My only comment, which is fine, is that we have now `subgroup_info.SubGroupInfo`
Means, package name in the beggining of struct and functions. I guess its OK though, all infos are like that

omer-dayan (approved):

PR #545, opened 2025-09-29
Set job fit error for topology job misconfiguration

omer-dayan:


pkg/scheduler/actions/common/allocate.go
```diff
@@ -30,7 +30,7 @@ func AllocateJob(ssn *framework.Session, stmt *framework.Statement, nodes []*nod
 	nodeSets, err := ssn.SubsetNodesFn(job, tasksToAllocate, nodes)
 	if err != nil {
 		log.InfraLogger.Errorf(
```
romanbaron, line 32: I think we said it is not a scheduler error and it should be info:
```
log.InfraLogger.V(6).Infof(...)
```

pkg/scheduler/plugins/topology/job_filtering.go
```diff
@@ -33,10 +32,13 @@ type jobAllocationMetaData struct {
 }
 
 func (t *topologyPlugin) subSetNodesFn(job *podgroup_info.PodGroupInfo, tasks []*pod_info.PodInfo, nodeSet node_info.NodeSet) ([]node_info.NodeSet, error) {
-	topologyTree, err := t.getJobTopology(job)
+	topologyTree, found, err := t.getJobTopology(job)
 	if err != nil {
 		return nil, err
 	}
+	if !found {
+		return []node_info.NodeSet{}, nil
```
romanbaron, line 41: Maybe the fit error should be set here instead of in `getJobTopology`

pkg/scheduler/actions/common/allocate.go
omer-dayan: We said that if a topology object not found is not a scheduler error, but a job error. I agree.

But now if its job error (Topology typo / Not found domain).

If SubsetNode fails now, it is a scheduler error

pkg/scheduler/plugins/topology/job_filtering.go
```diff
@@ -48,8 +51,10 @@ func (t *topologyPlugin) subSetNodesFn(job *podgroup_info.PodGroupInfo, tasks []
 	}
 
 	if maxAllocatablePods < len(tasks) {
-		log.InfraLogger.V(6).Infof("no relevant domains found for job %s, workload topology name: %s",
-			job.PodGroup.Name, topologyTree.Name)
+		job.SetJobFitError(
+			podgroup_info.PodSchedulingErrors,
+			fmt.Sprintf("No relevant domains found for job %s, workload topology name: %s", job.Name, topologyTree.Name),
```
romanbaron: It will be recorded on the workload, I don't the name of the workload is not really needed. It will be part of the metadata anyway.

Let's just say:
```
No relevant domains found for workload in topology tree: %s
```

romanbaron (approved):

PR #548, opened 2025-09-29
Moved global topology constraint from PodGroupInfo to SubGroupSet

romanbaron:


pkg/scheduler/api/podgroup_info/job_info.go
```diff
@@ -209,7 +199,14 @@ func (pgi *PodGroupInfo) setSubGroups(podGroup *enginev2alpha2.PodGroup) error {
 		return err
 	}
 	pgi.RootSubGroupSet = rootSubGroupSet
-	pgi.PodSets = rootSubGroupSet.GetPodSets()
+	podSets := rootSubGroupSet.GetPodSets()
+	if len(podSets) > 0 {
+		pgi.PodSets = podSets
+	} else {
+		for _, podSet := range pgi.PodSets {
+			rootSubGroupSet.AddPodSet(podSet)
```
omer-dayan, line 207: I dont understand how in the "SetSubGroup", which again happen directly after "NewPodGroup" the `pgi.PodSets` has some value

omer-dayan: OK Got it, when someone creates PodGroupInfo and not running "SetSubGroup" it will have the default SubGroup.
Happens in tests only.
Again its code that exists for tests only in prod structs

romanbaron:
Will be merged as part of https://github.com/NVIDIA/KAI-Scheduler/pull/556

PR #552, opened 2025-09-30
feat: set up service monitor

SiorMeir:
# Prometheus Operand - Adding ServiceMonitores

We're adding `ServiceMonitor` CR for each service. We'll allow customization of scraping rules through our `promtheus` config.

## What is done here
* adding ServiceMonitors for each KAI service
* adding configuration for ServiceMonitors 
* adding relevant RBAC permissions
* adjusting unit tests


pkg/operator/controller/config_controller.go
```diff
@@ -73,13 +73,13 @@ func (r *ConfigReconciler) SetOperands(ops []operands.Operand) {
 // +kubebuilder:rbac:groups=kai.scheduler,resources=configs/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=kai.scheduler,resources=configs/finalizers,verbs=update
 // +kubebuilder:rbac:groups=apps,resources=deployments;daemonsets,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups="",resources=services;secrets;serviceaccounts;configmaps;persistentvolumeclaims,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=services;secrets;serviceaccounts;configmaps;persistentvolumeclaims;nodes;nodes/proxy;endpoints;pods,verbs=get;list;watch;create;update;patch;delete
```
enoodle: I dont think we want this change - we are not editing nodes.

pkg/operator/operands/common/common.go
```diff
@@ -193,3 +209,48 @@ func AddK8sClientConfigToArgs(k8sClientConfig *kaiv1common.K8sClientConfig, args
...
+
+	err := runtimeClient.Get(ctx, types.NamespacedName{
+		Name: "servicemonitors.monitoring.coreos.com",
+	}, crd)
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
```
itsomri: Add a little bit more context
```suggestion
		return false, fmt.Errorf("failed to check service monitor CRD availability, error: %v", err)
```

pkg/operator/operands/common/common.go
```diff
@@ -193,3 +209,48 @@ func AddK8sClientConfigToArgs(k8sClientConfig *kaiv1common.K8sClientConfig, args
...
+
+	err := client.Get(ctx, types.NamespacedName{
+		Name: "prometheuses.monitoring.coreos.com",
+	}, crd)
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
```
itsomri: Same here

pkg/operator/operands/known_types/prometheus.go
```diff
@@ -104,3 +116,55 @@ func getCurrentPrometheusState(ctx context.Context, runtimeClient client.Client,
...
+
+	serviceMonitorList := &monitoringv1.ServiceMonitorList{}
+	reconcilerKey := getReconcilerKey(reconciler)
+
+	// Try to list with field selector first, but fall back to listing all if field indexer is not available
+	err = runtimeClient.List(ctx, serviceMonitorList, client.MatchingFields{CollectableOwnerKey: reconcilerKey})
+	if err != nil {
+		// If field indexer is not available, fall back to listing all ServiceMonitor resources
+		// and filter by owner reference manually
+		log.FromContext(ctx).Info("Field indexer not available, falling back to manual filtering")
```
itsomri: You should check that the failure was actually due to the field indexer missing, and not some other reason

pkg/operator/operands/prometheus/resources.go
```diff
@@ -72,44 +72,82 @@ func prometheusForKAIConfig(
 		logger.Error(err, "Failed to check for existing Prometheus instance")
 		return nil, err
 	}
-	prometheus = prom.(*monitoringv1.Prometheus)
+	var ok bool
+	prometheus, ok = prom.(*monitoringv1.Prometheus)
+	if !ok {
+		logger.Error(nil, "Failed to cast object to Prometheus type")
+		return nil, err
```
itsomri: You are actually returning the wrong err object here - it wasn't populated
```suggestion
		return nil, fmt.Errorf("failed to cast object to Prometheus type")
```

pkg/operator/operands/prometheus/resources.go
```diff
@@ -72,44 +72,82 @@ func prometheusForKAIConfig(
...
-			VolumeClaimTemplate: monitoringv1.EmbeddedPersistentVolumeClaim{
-				Spec: v1.PersistentVolumeClaimSpec{
-					StorageClassName: config.StorageClassName,
-					AccessModes:      []v1.PersistentVolumeAccessMode{v1.ReadWriteOnce},
-					Resources: v1.VolumeResourceRequirements{
-						Requests: v1.ResourceList{
-							v1.ResourceStorage: resource.MustParse(storageSize),
-						},
+	// Configure TSDB storage
+	storageSize, err := config.CalculateStorageSize(ctx, runtimeClient)
```
itsomri, line 74: Please remind me if this is overridable by the user

pkg/operator/operands/common/common.go
```diff
@@ -193,3 +209,48 @@ func AddK8sClientConfigToArgs(k8sClientConfig *kaiv1common.K8sClientConfig, args
...
+	}
+
+	return true, nil
+}
+
+func CheckPrometheusCRDAvailable(ctx context.Context, client client.Client) (bool, error) {
```
enoodle: These two functions are very similar, also they are always called together, maybe it can be unified?

pkg/operator/operands/prometheus/resources.go
```diff
@@ -139,3 +177,109 @@ func CheckPrometheusOperatorInstalled(ctx context.Context, runtimeClient client.
...
+			Name:      serviceAccountName,
+			Namespace: kaiConfig.Spec.Namespace,
+			Labels: map[string]string{
+				"app": mainResourceName,
+			},
+		},
+	}
+
+	// Check if ServiceAccount already exists
+	saObj, err := common.ObjectForKAIConfig(ctx, runtimeClient, serviceAccount, serviceAccountName, kaiConfig.Spec.Namespace)
```
enoodle: Should first get the object and then set the fields. It is true for all resources as well. Otherwise changes will not be overwritten.

pkg/operator/operands/prometheus/resources.go
```diff
@@ -72,44 +72,82 @@ func prometheusForKAIConfig(
...
+		prometheusSpec.ServiceMonitorNamespaceSelector = &metav1.LabelSelector{}
 	}
 
 	prometheus.Spec = prometheusSpec
 
-	return []client.Object{prometheus}, nil
+	var objects []client.Object
+
+	// Create ServiceAccount for Prometheus
+	serviceAccount, err := createPrometheusServiceAccount(ctx, runtimeClient, kaiConfig)
```
enoodle: I think it is better to have the same structure as the other operands where each of these functions is called individually. It is easier to debug and see what is included. Also this function will be shorter.

pkg/operator/operands/prometheus/resources.go
```diff
@@ -139,3 +177,109 @@ func CheckPrometheusOperatorInstalled(ctx context.Context, runtimeClient client.
...
+				Namespace: kaiConfig.Spec.Namespace,
+				Labels: map[string]string{
+					"app":        mainResourceName,
+					"accounting": mainResourceName,
+				},
+			},
+		}
+		serviceMonitorObj, err := common.ObjectForKAIConfig(ctx, runtimeClient, serviceMonitor, kaiService.Name, kaiConfig.Spec.Namespace)
+		if err != nil {
+			logger.Error(err, "Failed to check for existing ServiceMonitor instance", "service", kaiService.Name)
```
itsomri, line 139: You should probably ignoreNotFound here, no?

pkg/operator/operands/common/common.go
SiorMeir: We can.
The reasoning here was that we might only need the ServiceMonitors when we're using an external connection.
I'll try leave the option to test only one, defaulting to checking both CRDs

enoodle: You are right! then maybe use a basic function to check for CRD existence and send different names?

SiorMeir: @enoodle sounds good, will implement

pkg/operator/operands/prometheus/resources.go
SiorMeir: @itsomri yup, we've added a short-circuit if the user inputs a fixed size

pkg/operator/operands/prometheus/resources.go
SiorMeir: @itsomri `ObjectForKAIConfig` handles this for us here

pkg/operator/operands/prometheus/resources.go
SiorMeir: done

pkg/operator/operands/prometheus/resources.go
SiorMeir: accepted.

pkg/operator/operands/known_types/prometheus.go
SiorMeir: this is implicit because we're using the MatchingFields attribute..

itsomri: It's one possible source of failure, yes, but you can also fail due to permissions, or network issues, or context timeout, or any other reason

pkg/operator/operands/common/common.go
```diff
@@ -24,6 +25,21 @@ import (
...
 
+// KAI services that should be monitored via ServiceMonitor
+var KaiServicesForServiceMonitor = []struct {
+	Name     string
+	Port     string
+	JobLabel string
+}{
+	{"binder", "http-metrics", "binder"},
+	{"scheduler", "http-metrics", "scheduler"},
+	{"queuecontroller", "metrics", "queuecontroller"},
```
enoodle, line 35: I think that we only need this one (queue controller) to support the usage based scheduling feature. The other ones will just fill the DB with data that it less relevant for that. I suggest that for now we will comment the others out and we will add them later under a feature flag in another PR.

pkg/operator/operands/prometheus/resources.go
```diff
@@ -69,73 +69,168 @@ func prometheusForKAIConfig(
 	// Check if Prometheus already exists
 	prom, err := common.ObjectForKAIConfig(ctx, runtimeClient, prometheus, mainResourceName, kaiConfig.Spec.Namespace)
```
enoodle:
```suggestion
prom, err := common.ObjectForKAIConfig(ctx, runtimeClient, &monitoringv1.Prometheus{}, mainResourceName, kaiConfig.Spec.Namespace)
```

enoodle: replace 55-70 with this one line

pkg/operator/operands/prometheus/resources.go
```diff
@@ -69,73 +69,168 @@ func prometheusForKAIConfig(
...
+	}
+
+	// Configure ServiceMonitor selector to match KAI ServiceMonitors
+	if config.ServiceMonitor != nil && *config.ServiceMonitor.Enabled {
+		prometheusSpec.ServiceMonitorSelector = &metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				"accounting": mainResourceName,
+			},
 		}
+		prometheusSpec.ServiceMonitorNamespaceSelector = &metav1.LabelSelector{}
```
enoodle, line 105: Maybe use the namespace selector to select all ServiceMonitors in the KAI-scheduler namespace? it will be simpler.

pkg/operator/operands/prometheus/resources.go
```diff
@@ -69,73 +69,168 @@ func prometheusForKAIConfig(
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      kaiService.Name,
+				Namespace: kaiConfig.Spec.Namespace,
+				Labels: map[string]string{
+					"app":        mainResourceName,
+					"accounting": mainResourceName,
+				},
+			},
+		}
+		serviceMonitorObj, err := common.ObjectForKAIConfig(ctx, runtimeClient, serviceMonitor, kaiService.Name, kaiConfig.Spec.Namespace)
```
enoodle:
```suggestion
		serviceMonitorObj, err := common.ObjectForKAIConfig(ctx, runtimeClient, &monitoringv1.ServiceMonitor{}, kaiService.Name, kaiConfig.Spec.Namespace)
```

pkg/operator/operands/known_types/prometheus.go
SiorMeir: generalized the log message

pkg/operator/operands/prometheus/resources.go
SiorMeir: @enoodle I agree, but apprerantly the `CommonPrometheusFields` we're using here in the spec only supports LabelSelector, so I suggest we will leave it like that for now.

pkg/operator/operands/common/common.go
SiorMeir: commented out and left a comment

pkg/operator/operands/prometheus/resources.go
SiorMeir: @enoodle Makes total sense. `ObjectForKAIConfig` takes care of this for us.

itsomri (approved):

PR #554, opened 2025-10-01
Time aware fairness simulation env tests

itsomri:
This PR adds a simulator to test time-aware-fairness logic over time.
A simulation has the following inputs:
* Queues: a list of queues, with optional quota, priority and weight
* Jobs: a list of jobs, with pod & gpu count
* Nodes: a list of nodes with GPU counts
* The number of cycles the simulation should run for

### DIFF
--- a/pkg/env-tests/time_aware_fairness_test.go
+++ b/pkg/env-tests/time_aware_fairness_test.go
@@ -41,7 +41,7 @@ import (
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/plugins"
 )
 
-const interval = time.Millisecond * 10
+const simulationCycleInterval = time.Millisecond * 10
 const defaultTimeout = time.Second * 10
 
 type TestQueue struct {
@@ -70,7 +70,7 @@ type TimeAwareSimulation struct {
 	Cycles *int //default is 100
 }
 
-func setupControllers(backgroundCtx context.Context, cfg *rest.Config) (chan struct{}, context.CancelFunc, *fake.FakeWithHistoryClient, error) {
+func setupControllers(backgroundCtx context.Context, cfg *rest.Config) (chan struct{}, context.CancelFunc, *fake.FakeUsageDBClient, error) {
 	ctx, cancel := context.WithCancel(backgroundCtx)
 
 	err := queuecontroller.RunQueueController(cfg, ctx)
@@ -115,7 +115,7 @@ func setupControllers(backgroundCtx context.Context, cfg *rest.Config) (chan str
 	if err != nil {
 		return nil, cancel, nil, fmt.Errorf("failed to create fake usage client: %w", err)
 	}
-	usageClient := fakeClient.(*fake.FakeWithHistoryClient)
+	usageClient := fakeClient.(*fake.FakeUsageDBClient)
 
 	cfg.QPS = -1
 	cfg.Burst = -1
@@ -202,7 +202,7 @@ func RunSimulation(ctx context.Context, simulation TimeAwareSimulation) (fake.Al
 		simulation.Cycles = ptr.To(100)
 	}
 	for range *simulation.Cycles {
-		time.Sleep(interval * 10)
+		time.Sleep(simulationCycleInterval * 10)
 		usageClient.AppendQueuedAllocation(getAllocations(ctx, ctrlClient), getClusterResources(ctx, ctrlClient, true))
 	}
 
@@ -235,7 +235,7 @@ func cleanupSimulation(ctx context.Context, ctrlClient client.Client, testNamesp
 		cleanupErrors = append(cleanupErrors, fmt.Errorf("failed to delete simulation pods and podgroups during cleanup: %w", err))
 	}
 
-	if err := WaitForNoObjectsInNamespace(ctx, ctrlClient, testNamespace.Name, defaultTimeout, interval,
+	if err := WaitForNoObjectsInNamespace(ctx, ctrlClient, testNamespace.Name, defaultTimeout, simulationCycleInterval,
 		&corev1.PodList{},
 		&schedulingv2alpha2.PodGroupList{},
 		&resourceapi.ResourceClaimList{},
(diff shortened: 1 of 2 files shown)


pkg/env-tests/time_aware_fairness_test.go
```diff
@@ -0,0 +1,353 @@
...
+	stopCh, cancel, usageClient, err := setupControllers(ctx, cfg)
+	if err != nil {
+		return nil, err
+	}
+	defer close(stopCh)
+	defer cancel()
+
+	testNamespace := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "test-" + randomstring.HumanFriendlyEnglishString(10),
```
SiorMeir: maybe use the `test-{simulationName}` or a derivative of it for easy lookup?

pkg/env-tests/time_aware_fairness_test.go
```diff
@@ -0,0 +1,353 @@
...
+		nodeObject := CreateNodeObject(ctx, ctrlClient, DefaultNodeConfig(fmt.Sprintf("test-node-%d", node.GPUs)))
+		nodeObject.ObjectMeta.Labels = map[string]string{
+			"simulation": simulationName,
+		}
+		if err := ctrlClient.Create(ctx, nodeObject); err != nil {
+			return nil, err
+		}
+		nodes = append(nodes, nodeObject)
+	}
+	defer func() {
```
SiorMeir: maybe worth exporting cleanup logic to a non-anonymous func?
this might be a source of bugs (incomplete cleanup)

SiorMeir: And adding a cleanup validation?

pkg/env-tests/time_aware_fairness_test.go
```diff
@@ -0,0 +1,353 @@
...
+		}
+
+		err := ctrlClient.Create(ctx, queueObject)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create queue object: %w", err)
+		}
+		queues = append(queues, queueObject)
+	}
+
+	defer func() {
```
SiorMeir: cleanup note here as well

pkg/env-tests/time_aware_fairness_test.go
itsomri: Actually I will make it constant, no reason to randomize it, will help debugging

pkg/env-tests/time_aware_fairness_test.go
itsomri: I refactored the cleanup errors a bit.
There's not much to do in the code if the cleanups fail, but now they will return as a separate return value, so the caller will know not to attempt another simulation.
Since these are env tests, it's not really a big deal, as the environment will get re-created when the process is done, gracefully or not

itsomri: The anonymous function doesn't really matter as the call stack will not printed anyway, but know it will be very clear when there are cleanup errors

pkg/queuecontroller/metrics/metrics.go
```diff
@@ -54,6 +56,11 @@ var (
 // queueLabelToMetricLabelMap        = map[string]string{"priority": "queue_priority"}
 // queueLabelToDefaultMetricValueMap = map[string]string{"priority": "normal"}
 func InitMetrics(namespace string, queueLabelToMetricLabelMap, queueLabelToDefaultMetricValueMap map[string]string) {
+	if initiated {
+		return
+	}
+	initiated = true
```
gshaibi, line 62: I don't know if this should be thread safe, but if it is, consider using `sync.Once`

pkg/env-tests/time_aware_fairness_test.go
```diff
@@ -0,0 +1,357 @@
...
+	if err != nil {
+		return nil, cancel, nil, fmt.Errorf("failed to run podgroupcontroller: %w", err)
+	}
+
+	err = binder.RunBinder(cfg, ctx)
+	if err != nil {
+		return nil, cancel, nil, fmt.Errorf("failed to run binder: %w", err)
+	}
+
+	fakeClient, err := fake.NewFakeWithHistoryClient("fake-connection", nil)
```
enoodle, line 114: rename to `fakeUsageDBClient`

pkg/env-tests/time_aware_fairness_test.go
enoodle: Maybe you can have just one cleanup function that will delete all nodes, queues , pods with specific label, and just call it at the beginning ?

pkg/env-tests/time_aware_fairness_test.go
```diff
@@ -0,0 +1,357 @@
...
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/actions"
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/common_info"
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/queue_info"
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/cache/usagedb/api"
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/cache/usagedb/fake"
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/conf_util"
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/plugins"
+)
+
+const interval = time.Millisecond * 10
```
enoodle: Maybe rename to 'simulationCycleInterval` ?

enoodle (dismissed):

gshaibi (approved):

enoodle (dismissed):

pkg/queuecontroller/metrics/metrics.go
itsomri: It's not called concurrently, but it needs to work if called more than once

enoodle (approved):

PR #557, opened 2025-10-05
E2E Flakiness fix

gshaibi:
If e2e tests are re-run quickly, workload namespaces may remain in a  terminating  state and cause the test to fail.
This aims to wait for them to finish terminating before proceeding.


test/e2e/modules/wait/namespace.go
```diff
@@ -0,0 +1,52 @@
...
+	condition := func(event watch.Event) bool {
+		namespaceListObj, ok := event.Object.(*v1.NamespaceList)
+		if !ok {
+			return false
+		}
+
+		// Check if any of the specified namespaces still exist
+		for _, ns := range namespaceListObj.Items {
+			if _, exists := namespaceSet[ns.Name]; exists {
+				return false
```
itsomri, line 40: Instead of immediately returning, iterate over all namespace names, so you can print all remaining namespaces below in case it fails

gshaibi: I want to align with the rest of the wait functions that do not print the remaining undeleted resources (pods for example)

davidLif (approved):

PR #558, opened 2025-10-05
shaibi/topology plugin node sorting

gshaibi:
- **Implement topology plugin enhancements: add node scoring and pre-order functions, introduce node set ID generation, and improve tree sorting for job allocation. Update session management to include new functions for better task placement based on topology.**
- **Topology plugin node sorting**
- **Topology scoring up to the preferred level + bugfix**

### DIFF
--- a/pkg/scheduler/plugins/topology/job_filtering.go
+++ b/pkg/scheduler/plugins/topology/job_filtering.go
@@ -4,14 +4,16 @@
 package topology
 
 import (
+	"cmp"
 	"fmt"
-	"sort"
+	"slices"
 
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/node_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/pod_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/pod_status"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/podgroup_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/resource_info"
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/log"
 )
 
 type jobAllocationMetaData struct {
@@ -41,7 +43,10 @@ func (t *topologyPlugin) subSetNodesFn(job *podgroup_info.PodGroupInfo, tasks []
 		return nil, err
 	}
 
+	// GuyTodo: Maybe we should sort afterwards for needed domains instead of sorting the whole tree.
+	// Sorting so we can traverse the tree in-order later on for node scoring.
 	sortTree(topologyTree.DomainsByLevel[rootLevel][rootDomainId])
+	log.InfraLogger.V(7).Infof("Tree sorted: %+v", topologyTree.DomainsByLevel[rootLevel][rootDomainId])
 
 	if maxAllocatablePods < len(tasks) {
 		job.SetJobFitError(
@@ -314,10 +319,13 @@ func (*topologyPlugin) treeAllocatableCleanup(topologyTree *Info) {
 	}
 }
 
-// Only sort children, not nodes.
 func sortTree(root *DomainInfo) {
-	sort.Slice(root.Children, func(i, j int) bool {
-		return root.Children[i].AllocatablePods > root.Children[j].AllocatablePods
+	if root == nil {
+		return
+	}
+
+	slices.SortStableFunc(root.Children, func(i, j *DomainInfo) int {
+		return cmp.Compare(i.AllocatablePods, j.AllocatablePods)
 	})
 
 	for _, child := range root.Children {
--- a/pkg/scheduler/plugins/topology/job_filtering.go
+++ b/pkg/scheduler/plugins/topology/job_filtering.go
@@ -5,6 +5,7 @@ package topology
 
 import (
 	"fmt"
+	"sort"
 
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/node_info"
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/pod_info"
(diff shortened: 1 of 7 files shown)


pkg/scheduler/plugins/topology/topology_plugin.go
```diff
@@ -16,13 +18,30 @@ const (
...
 )
 
+type topologyName = string
+
 type topologyPlugin struct {
-	TopologyTrees map[string]*Info
+	TopologyTrees map[topologyName]*Info
+
+	// Cache for storing node set to domain by topology name
+	nodeSetToDomain map[topologyName]map[nodeSetID]*DomainInfo
```
davidLif: Usually we a different mechanizem for plugins cache (see stateProvider.GetSessionStateForResource)

pkg/scheduler/actions/allocate/allocateTopology_test.go
```diff
@@ -995,13 +995,13 @@ func getTopologyTestsMetadata() []integration_tests_utils.TestTopologyMetadata {
...
 					"pending_job0-0": {
-						NodeName:             "node0",
+						NodeName:             "node1",
 						GPUsRequired:         1,
 						Status:               pod_status.Binding,
 						DontValidateGPUGroup: true,
 					},
 					"pending_job0-1": {
-						NodeName:             "node1",
+						NodeName:             "node0",
```
davidLif, line 1021: Why the change?

pkg/scheduler/actions/allocate/allocateTopology_test.go
```diff
@@ -1161,13 +1161,13 @@ func getTopologyTestsMetadata() []integration_tests_utils.TestTopologyMetadata {
...
 					"pending_job0-0": {
-						NodeName:             "node2",
+						NodeName:             "node3",
 						GPUsRequired:         1,
 						Status:               pod_status.Binding,
 						DontValidateGPUGroup: true,
 					},
 					"pending_job0-1": {
-						NodeName:             "node3",
+						NodeName:             "node2",
```
davidLif, line 1189: Why the change?

pkg/scheduler/actions/allocate/allocateTopology_test.go
gshaibi: Since we now sort the nodes differently

pkg/scheduler/actions/allocate/allocateTopology_test.go
gshaibi: As above, since we now sort the nodes differently

pkg/scheduler/plugins/topology/topology_plugin.go
gshaibi: I took an example from [another plugin](https://github.com/NVIDIA/KAI-Scheduler/blob/1b801e9791233077b94d9c7845dcdbb452c52d46/pkg/scheduler/plugins/nodeplacement/nodeplacement.go#L29) that holds a local cache that is not bound to the Session.

Can you please refer me to a plugin that uses this SessionState so I can see how to use it?

pkg/scheduler/plugins/topology/topology_plugin.go
```diff
@@ -32,8 +42,16 @@ func (t *topologyPlugin) Name() string {
...
 	t.initializeTopologyTree(ssn.Topologies, ssn.Nodes)
+	t.session = ssn
 
 	ssn.AddSubsetNodesFn(t.subSetNodesFn)
+	ssn.AddNodeOrderFn(t.nodeOrderFn)
+	ssn.AddPreJobAllocationFn(t.preJobAllocationFn)
+}
+
+func (t *topologyPlugin) preJobAllocationFn(job *podgroup_info.PodGroupInfo) {
+	// Invalidate the sub-group node scores
```
davidLif, line 53: I would call it "initiallize". What do you think would be more readable?

pkg/scheduler/plugins/topology/job_filtering.go
```diff
@@ -333,18 +346,40 @@ func (*topologyPlugin) treeAllocatableCleanup(topologyTree *Info) {
 	}
 }
 
-func sortDomainInfos(domainInfos []*DomainInfo) []*DomainInfo {
-	sort.SliceStable(domainInfos, func(i, j int) bool {
-		if domainInfos[i].Level != domainInfos[j].Level {
-			return false
-		}
+func sortTree(root *DomainInfo, maxDepthLevel DomainLevel) {
```
davidLif, line 353: I recommend adding the function's purpose to the name. Something like sortBYDomainScore

pkg/scheduler/plugins/topology/job_filtering.go
```diff
@@ -333,18 +346,40 @@ func (*topologyPlugin) treeAllocatableCleanup(topologyTree *Info) {
...
+		return
+	}
+
+	for _, child := range root.Children {
+		sortTree(child, maxDepthLevel)
+	}
+}
+
+// Assumes the topology tree is already sorted
+func sortDomainInfos(topologyTree *Info, domainInfos []*DomainInfo) []*DomainInfo {
```
davidLif, line 376: What's the difference between this sort and the previous one?

pkg/scheduler/plugins/topology/node_scoring.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func calculateNodeScores(domain *DomainInfo, preferredLevel DomainLevel) map[string]float64 {
+	orderedPreferredLevelDomains := getLevelDomains(domain, preferredLevel)
+
+	nodeScores := make(map[string]float64)
+	for i, leafDomain := range orderedPreferredLevelDomains {
+		for _, node := range leafDomain.Nodes {
+			// Score nodes by their domain's order
+			score := (float64(i+1) / float64(len(orderedPreferredLevelDomains))) * 10
```
davidLif, line 44: Please use const for the 10

pkg/scheduler/plugins/topology/job_filtering.go
```diff
@@ -52,12 +53,24 @@ func (t *topologyPlugin) subSetNodesFn(
...
+	// Sorting the tree for both packing and closest preferred level domain scoring
+	preferredLevel := DomainLevel(subGroup.GetTopologyConstraint().PreferredLevel)
+	requiredLevel := DomainLevel(subGroup.GetTopologyConstraint().RequiredLevel)
+	maxDepthLevel := preferredLevel
+	if maxDepthLevel == "" {
+		maxDepthLevel = requiredLevel
+	}
+	sortTree(topologyTree.DomainsByLevel[rootLevel][rootDomainId], maxDepthLevel)
+	if preferredLevel != "" {
+		t.subGroupNodeScores[subGroup.GetName()] = calculateNodeScores(topologyTree.DomainsByLevel[rootLevel][rootDomainId], preferredLevel)
```
enoodle, line 65: This will run on all nodes before we have a chance to maybe filter out some of them. Maybe it can run some how lazily? If it is too complicated we can do it in another PR later.

gshaibi: I can't think of such a way at the moment. Do you have something in mind?

enoodle: I can't think of something currently. Lets remember it for future improvements.

PR #560, opened 2025-10-09
Topology constraint at any subgroup hierarchy level

romanbaron:
Topology constraint at any subgroup hierarchy level

### DIFF
--- a/pkg/scheduler/actions/allocate/allocateTopology_test.go
+++ b/pkg/scheduler/actions/allocate/allocateTopology_test.go
@@ -1717,8 +1717,8 @@ func getTopologyTestsMetadata() []integration_tests_utils.TestTopologyMetadata {
 						RootSubGroupSet: func() *subgroup_info.SubGroupSet {
 							replica1 := subgroup_info.NewSubGroupSet("replica1",
 								&topology_info.TopologyConstraintInfo{
-									RequiredLevel: "k8s.io/rack",
-									Topology:      "cluster-topology",
+									PreferredLevel: "k8s.io/rack",
+									Topology:       "cluster-topology",
 								},
 							)
 							replica1.AddPodSet(subgroup_info.NewPodSet(
@@ -1734,8 +1734,8 @@ func getTopologyTestsMetadata() []integration_tests_utils.TestTopologyMetadata {
 
 							replica2 := subgroup_info.NewSubGroupSet("replica2",
 								&topology_info.TopologyConstraintInfo{
-									Topology:      "cluster-topology",
-									RequiredLevel: "k8s.io/rack",
+									Topology:       "cluster-topology",
+									PreferredLevel: "k8s.io/rack",
 								},
 							)
 							replica2.AddPodSet(subgroup_info.NewPodSet(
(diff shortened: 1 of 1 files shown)


pkg/scheduler/actions/common/allocate.go
```diff
@@ -26,32 +27,93 @@ func AllocateJob(ssn *framework.Session, stmt *framework.Statement, nodes []*nod
...
-	if err != nil {
-		log.InfraLogger.Errorf(
-			"Failed to run SubsetNodes on job <%s/%s>: %v", job.Namespace, job.Namespace, err)
-		return false
+func allocateSubGroupSet(ssn *framework.Session, stmt *framework.Statement, nodes []*node_info.NodeInfo,
+	job *podgroup_info.PodGroupInfo, subGroupSet *subgroup_info.SubGroupSet, tasksToAllocate []*pod_info.PodInfo,
+	isPipelineOnly bool,
+) bool {
+	nodeSets := []node_info.NodeSet{nodes}
+	if subGroupSet.GetTopologyConstraint() != nil {
```
itsomri: This if feels very specific, I thought node subset plugin was supposed to make this generic?

itsomri: Feels to me like we can give up this IF completely, and give `ssn.SubsetNodesFn` a default behavior that will return a single subset of all nodes by default

pkg/scheduler/actions/common/allocate.go
```diff
@@ -26,32 +27,93 @@ func AllocateJob(ssn *framework.Session, stmt *framework.Statement, nodes []*nod
...
+	}
+	return true
+}
+
+func allocatePodSet(ssn *framework.Session, stmt *framework.Statement, nodes node_info.NodeSet,
+	job *podgroup_info.PodGroupInfo, podSet *subgroup_info.PodSet, tasksToAllocate []*pod_info.PodInfo,
+	isPipelineOnly bool,
+) bool {
+	nodeSets := []node_info.NodeSet{nodes}
+	if podSet.GetTopologyConstraint() != nil {
```
itsomri: Again, this feels to specific, I feel like we can contain Topology to it's own plugin without messing with the core allocate logic

pkg/scheduler/actions/common/allocate.go
romanbaron: You are correct, will do

pkg/scheduler/actions/allocate/allocateTopology_test.go
```diff
@@ -1180,5 +1181,689 @@ func getTopologyTestsMetadata() []integration_tests_utils.TestTopologyMetadata {
...
+					CacheRequirements: &test_utils.CacheMocking{
+						NumberOfCacheBinds: 2,
+					},
+				},
+			},
+			RoundsUntilMatch: 1,
+		},
+		{
+			TestTopologyBasic: test_utils.TestTopologyBasic{
+				Name: "Preferred topology at multiple subgroup hierarchy levels - rack and zone",
```
enoodle, line 1681: I am not sure that this test is setting preferred topology

romanbaron: Thanks, fixed and added one more test that combines preferred and required constraints.

PR #563, opened 2025-10-09
feat: add externalURL config

SiorMeir:
# Prometheus Operand - Add support for non-managed Prometheus instance

As prometheus is a common service in many k8s cluster, we want to allow for an external, non managed instance to monitor and collect metrics from KAI-scheduler services. 

we're adding an optional `externalPrometheusUrl` to be provided, creating only ServiceMonitors per tracked service instead of creating an instance managed by the Operator.

## What is done here
* adding an `externalPrometheusUrl` spec
* setting up ServiceMontiors for tracked service
* updating config status with successful / unsuccessful connetion


pkg/operator/controller/config_controller.go
```diff
@@ -101,8 +106,8 @@ func (r *ConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 	defer func() {
-		reconcileStatusErr := r.ReconcileStatus(
-			ctx, &status_reconciler.KAIConfigWithStatusWrapper{Config: kaiConfig},
+		reconcileStatusErr := r.ReconcileStatusWithConfig(
```
SiorMeir: This feels a bit weird - I think about changing this to be simply `Reconcile`

pkg/apis/kai/v1/prometheus/prometheus.go
```diff
@@ -253,3 +263,114 @@ func (s *ServiceMonitor) SetDefaultsWhereNeeded() {
...
+	if err != nil || !ok {
+		return fmt.Errorf("failed to ping external Prometheus: %w", err)
+	}
+	return nil
+}
+
+// StartPrometheusMonitoring starts a background goroutine to monitor external Prometheus connectivity
+// and update the status periodically. The goroutine will stop when ctx is cancelled or when
+// ExternalPrometheusUrl is set to nil or empty string.
+func (p *Prometheus) StartPrometheusMonitoring(ctx context.Context, statusUpdater func(ctx context.Context, condition metav1.Condition) error) {
```
enoodle: This should move to be part of the operator code.

pkg/operator/controller/status_reconciler/status_reconciler.go
```diff
@@ -48,7 +48,17 @@ func (r *StatusReconciler) ReconcileStatus(ctx context.Context, object objectWit
...
+		return err
+	}
+	return nil
+}
+
+func (r *StatusReconciler) ReconcileStatusWithConfig(ctx context.Context, object objectWithConditions, kaiConfig *kaiv1.Config) error {
+	if err := r.ReconcileStatus(ctx, object); err != nil {
+		return err
+	}
+	return r.reconcileCondition(ctx, object, r.getPrometheusConnectivityCondition(ctx, object.GetGeneration(), kaiConfig))
```
enoodle: Lets find a way to call this from the operand's code. If you need to you can add to the interface.

pkg/operator/controller/config_controller.go
enoodle: It will cause naming issues because this function comes from a type that is embedded in this type, which already have a function named `Reconcile`

pkg/apis/kai/v1/prometheus/prometheus.go
```diff
@@ -253,3 +263,114 @@ func (s *ServiceMonitor) SetDefaultsWhereNeeded() {
...
+
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+}
+
+// validateExternalPrometheusConnection validates connectivity to an external Prometheus instance
+func pingExternalPrometheus(ctx context.Context, prometheusURL string) (bool, error) {
```
enoodle: This seems to be implemented twice, the other place is here: https://github.com/NVIDIA/KAI-Scheduler/pull/563/files#diff-988dc692576ef8dc42a4e8c08a006e6bc8fe82b3df418cbe8526a40c33458d71R228

SiorMeir: removed - this was part of an older implementation

pkg/operator/controller/config_controller.go
```diff
@@ -161,3 +170,68 @@ func enqueueWatched(_ context.Context, _ client.Object) []ctrl.Request {
...
 }
+
+// managePrometheusMonitoring manages the lifecycle of the Prometheus monitoring goroutine
+func (r *ConfigReconciler) managePrometheusMonitoring(ctx context.Context, kaiConfig *kaiv1.Config) {
+	// Check if external Prometheus is configured
+	hasExternalPrometheus := kaiConfig.Spec.Prometheus != nil &&
+		kaiConfig.Spec.Prometheus.ExternalPrometheusUrl != nil &&
+		*kaiConfig.Spec.Prometheus.ExternalPrometheusUrl != ""
+
+	if hasExternalPrometheus {
```
itsomri: In the context of this function, the happy flow is "hasExternalPrometheus". So just invert this condition, something like:
```Go
if !hasExternalPrometheus {
	if r.monitoringCancel != nil {
			r.monitoringCancel()
			r.monitoringCtx = nil
			r.monitoringCancel = nil
		}
	return
}

This will keep the rest of the code less indented, and make the function more readable.
In general: keep happy flow to the left (un-nested) as much as possible

File: pkg/operator/controller/config_controller.go [hunk 6]
```diff
@@ -161,3 +170,68 @@ func enqueueWatched(_ context.Context, _ client.Object) []ctrl.Request {
...
+		// Start monitoring if not already running
+		if r.monitoringCtx == nil || r.monitoringCtx.Err() != nil {
```
itsomri: Again: if running, return. Else, continue with the monitoring logic, un-nested

pkg/operator/controller/config_controller.go
```diff
@@ -161,3 +170,68 @@ func enqueueWatched(_ context.Context, _ client.Object) []ctrl.Request {
...
+				if !found {
+					configToUpdate.Status.Conditions = append(configToUpdate.Status.Conditions, condition)
+				}
+
+				// Update the status using the controller's client
+				return r.Client.Status().Patch(ctx, configToUpdate, client.MergeFrom(currentConfig))
+			}
+
+			// Start the monitoring goroutine
+			prometheus.StartPrometheusMonitoring(r.monitoringCtx, kaiConfig.Spec.Prometheus, statusUpdater)
```
itsomri: Nitpicking, but you can drop the "Prometheus" in "StartPrometheusMonitoring", it's cleaner:
```suggestion
			prometheus.StartMonitoring(r.monitoringCtx, kaiConfig.Spec.Prometheus, statusUpdater)
```

pkg/operator/controller/config_controller.go
```diff
@@ -161,3 +170,68 @@ func enqueueWatched(_ context.Context, _ client.Object) []ctrl.Request {
...
+		kaiConfig.Spec.Prometheus.ExternalPrometheusUrl != nil &&
+		*kaiConfig.Spec.Prometheus.ExternalPrometheusUrl != ""
+
+	if hasExternalPrometheus {
+		// Start monitoring if not already running
+		if r.monitoringCtx == nil || r.monitoringCtx.Err() != nil {
+			r.monitoringCtx, r.monitoringCancel = context.WithCancel(ctx)
+
+			// Create status updater function that uses the controller's client
+			statusUpdater := func(ctx context.Context, condition metav1.Condition) error {
```
itsomri: Is there a specific reason why we create this function here, instead of having it as a named method/function (or, having a "statusUpdater" object which can also keep state)?
(Generally it's a bit less convenient to debug these kinds of functions, but not a deal breaker if there's a good reason)

pkg/operator/operands/prometheus/prometheus.go
```diff
@@ -81,3 +89,115 @@ func (b *Prometheus) IsAvailable(ctx context.Context, readerClient client.Reader
...
+// StartPrometheusMonitoring starts a background goroutine to monitor external Prometheus connectivity
+// and update the status periodically. The goroutine will stop when ctx is cancelled or when
+// ExternalPrometheusUrl is set to nil or empty string.
+func StartPrometheusMonitoring(ctx context.Context, prometheusConfig *kaiprometheus.Prometheus, statusUpdater func(ctx context.Context, condition metav1.Condition) error) {
+	if prometheusConfig == nil || prometheusConfig.ExternalPrometheusUrl == nil || *prometheusConfig.ExternalPrometheusUrl == "" {
+		return
+	}
+
+	go func() {
+		ticker := time.NewTicker(30 * time.Second) // Check every 30 seconds
```
itsomri: Let's make this configurable

pkg/operator/operands/prometheus/prometheus.go
```diff
@@ -81,3 +89,115 @@ func (b *Prometheus) IsAvailable(ctx context.Context, readerClient client.Reader
...
+
+	// Ensure the URL has a scheme
+	if !strings.Contains(prometheusURL, "://") {
+		prometheusURL = "http://" + prometheusURL
+	}
+
+	// Parse the URL to ensure it's valid
+	_, err := url.Parse(prometheusURL)
+	if err != nil {
+		return false, fmt.Errorf("invalid Prometheus URL: %w", err)
```
itsomri: Does this print the URL itself? if so, let's add it to the error

pkg/operator/operands/prometheus/prometheus.go
```diff
@@ -81,3 +89,115 @@ func (b *Prometheus) IsAvailable(ctx context.Context, readerClient client.Reader
...
+	}
+
+	// Create HTTP client with timeout
+	client := &http.Client{
+		Timeout: 10 * time.Second,
```
itsomri: Should probably be configurable

pkg/operator/operands/prometheus/prometheus.go
```diff
@@ -81,3 +89,115 @@ func (b *Prometheus) IsAvailable(ctx context.Context, readerClient client.Reader
...
+	}
+
+	// Try to connect to the Prometheus /api/v1/status/config endpoint
+	statusURL := prometheusURL + "/api/v1/status/config"
+	logger.Info("Validating external Prometheus connection", "url", statusURL)
+
+	req, err := http.NewRequestWithContext(ctx, "GET", statusURL, nil)
+	if err != nil {
+		return false, fmt.Errorf("failed to create request: %w", err)
```
itsomri: Here as well - let's verify that the URL we tried is printed in the error

pkg/operator/operands/prometheus/prometheus.go
```diff
@@ -81,3 +89,115 @@ func (b *Prometheus) IsAvailable(ctx context.Context, readerClient client.Reader
...
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return false, fmt.Errorf("failed to connect to external Prometheus: %w", err)
+	}
+	defer resp.Body.Close()
+
+	// Check if we got a successful response
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return false, fmt.Errorf("external Prometheus returned status code %d", resp.StatusCode)
```
itsomri: Add the URL we tried to the error. Also, if there is more info we can get from the response body, we should add it as well

pkg/operator/operands/prometheus/prometheus.go
```diff
@@ -81,3 +89,115 @@ func (b *Prometheus) IsAvailable(ctx context.Context, readerClient client.Reader
...
+
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+}
+
+// pingExternalPrometheus validates connectivity to an external Prometheus instance
+func pingExternalPrometheus(ctx context.Context, prometheusURL string) (bool, error) {
```
itsomri: Is there a reason to return both a bool and an err? From what I see, the bool is basically `err == nil`, which makes it redundant

pkg/operator/controller/config_controller.go
SiorMeir: cool, will implement

pkg/operator/controller/config_controller.go
SiorMeir: yup, agree and will implement

pkg/operator/controller/config_controller.go
SiorMeir: no good reason other than the nesting. 
will try to refactor this out

pkg/operator/operands/prometheus/prometheus.go
SiorMeir: should. will validate

pkg/operator/operands/prometheus/prometheus.go
SiorMeir: @itsomri Thought it is clearer as this basically a boolean operation we don't really care why we couldn't ping prometheus - we just couldn't 

We can remove this and return an err but this didn't feel like an err cause nothing was wrong on our side

pkg/apis/kai/v1/config_types.go
```diff
@@ -24,19 +24,22 @@ import (
...
 const (
-	ConditionTypeReconciling       ConditionType = "Reconciling"
-	ConditionTypeDeployed          ConditionType = "Deployed"
-	ConditionTypeAvailable         ConditionType = "Available"
-	ConditionDependenciesFulfilled ConditionType = "DependenciesFulfilled"
+	ConditionTypeReconciling            ConditionType = "Reconciling"
+	ConditionTypeDeployed               ConditionType = "Deployed"
+	ConditionTypeAvailable              ConditionType = "Available"
+	ConditionDependenciesFulfilled      ConditionType = "DependenciesFulfilled"
+	ConditionTypePrometheusConnectivity ConditionType = "PrometheusConnectivity"
```
enoodle: This shouldn't be be a condition, but part of the message. When Prometheus can't connect we should be "unavailable"

pkg/operator/operands/prometheus/prometheus.go
SiorMeir: done

pkg/operator/controller/status_reconciler/status_reconciler.go
```diff
@@ -48,7 +48,10 @@ func (r *StatusReconciler) ReconcileStatus(ctx context.Context, object objectWit
 	if err := r.reconcileCondition(ctx, object, r.getAvailableCondition(ctx, object.GetGeneration())); err != nil {
 		return err
 	}
-	return r.reconcileCondition(ctx, object, r.getDependenciesFulfilledCondition(object.GetGeneration()))
+	if err := r.reconcileCondition(ctx, object, r.getDependenciesFulfilledCondition(object.GetGeneration())); err != nil {
+		return err
+	}
+	return nil
```
davidLif: Whats the difference between this code and the prevoius version (the direct return)?

pkg/operator/operands/prometheus/prometheus.go
davidLif: If I understand the code correctly, a single failure to reach the promethues will update the kai config status to "PrometheusConnectionFailed". Do we want to add some kind of hysteresis?

SiorMeir: this is a really good point... I think we need a retry before resorting to failure
I'll add a configurable param with default value of 5

SiorMeir: done. added a backoff as well

pkg/operator/controller/status_reconciler/status_reconciler.go
SiorMeir: This is redundant. Will revert

pkg/operator/controller/config_controller.go
```diff
@@ -161,3 +170,68 @@ func enqueueWatched(_ context.Context, _ client.Object) []ctrl.Request {
 		},
 	}
 }
+
+// managePrometheusMonitoring manages the lifecycle of the Prometheus monitoring goroutine
+func (r *ConfigReconciler) managePrometheusMonitoring(ctx context.Context, kaiConfig *kaiv1.Config) {
```
enoodle: Please extend the operands interface (https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/operator/operands/interface.go) to add a function for monitoring resources and implement it in the prometheus operand. Then here add something to call this function from all operands.

This will break the strong connection between the operator code and the prometheus operand - allowing us to remove it in the future, change it or even add similar ones.

pkg/apis/kai/v1/prometheus/prometheus.go
```diff
@@ -43,6 +43,30 @@ type Prometheus struct {
...
+	// for the external instance and validate connectivity
+	// +kubebuilder:validation:Optional
+	ExternalPrometheusUrl *string `json:"externalPrometheusUrl,omitempty"`
+
+	// ExternalPrometheusPingConfig defines the configuration for external Prometheus connectivity validation, with defaults.
+	// +kubebuilder:validation:Optional
+	ExternalPrometheusPingConfig *ExternalPrometheusPingConfig `json:"externalPrometheusPingConfig,omitempty"`
+}
+
+type ExternalPrometheusPingConfig struct {
```
itsomri: This struct is a good idea, let's just name it `ExternalPrometheusHealthProbe` or something similar, it works better with [known kubernetes constructs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)

pkg/apis/kai/v1/prometheus/prometheus.go
```diff
@@ -43,6 +43,30 @@ type Prometheus struct {
...
+	// PingsInterval defines the interval for external Prometheus connectivity validation (in seconds)
+	// +kubebuilder:validation:Optional
+	PingsInterval *int `json:"pingsInterval,omitempty"`
```
itsomri: Just "Interval", it's cleaner. You're already in the context of "ping" (or "probe"). In general, try to avoid repeating the context if you can.

A good tip is to look at the usage:
`ExternalPrometheusPingConfig.PingsInterval` vs `ExternalPrometheusPingConfig.Interval`

pkg/apis/kai/v1/prometheus/prometheus.go
SiorMeir: Accepted `ExternalPrometheusHealthProbe` even though this is not a native-k8s API in the sense of liveness/healthy endpoints.

PR #566, opened 2025-10-12
fix: Convert Ephemeral-Storage in MaxNodePoolResources from Bytes to GB

lakshyaj02:
Added test case for ephemeral storage testing. Made changes in reading and converting memory from Bytes to GB.


pkg/scheduler/api/resource_info/resource_info.go
```diff
@@ -128,7 +130,7 @@ func (r *Resource) DetailedString() string {
 	messageBuilder.WriteString(r.String())
 
 	for rName, rQuant := range r.scalarResources {
-		messageBuilder.WriteString(fmt.Sprintf(", %s: %v", rName, rQuant))
+            messageBuilder.WriteString(fmt.Sprintf(", %s: %v", rName, rQuant))
```
enoodle: This will cause the linter to file

pkg/scheduler/api/resource_info/resource_requirment.go
```diff
@@ -151,7 +153,10 @@ func (r *ResourceRequirements) DetailedString() string {
 	messageBuilder.WriteString(r.String())
 
 	for rName, rQuant := range r.scalarResources {
-		messageBuilder.WriteString(fmt.Sprintf(", %s: %v", rName, rQuant))
+		if rName == v1.ResourceEphemeralStorage || rName == v1.ResourceStorage {
+			rQuant = rQuant / (1000 * 1000 * 1000) // convert from milli-bytes to GB
```
enoodle:
```suggestion
			rQuant = rQuant / (MemoryToGB) // convert from milli-bytes to GB
```

enoodle (changes requested):
Thanks, I have one final request

pkg/scheduler/api/resource_info/resource_requirment.go
lakshyaj02: Added the change with the appropriate unit conversion.

enoodle (approved):

enoodle:
Thank you @lakshyaj02

PR #581, opened 2025-10-19
chore(ci,docs): add conventional PR title guidelines and validation + pull request template

gshaibi:
## Description

- Add conventional PR title guidelines and validation
- Add a Pull Request template

## Related Issues

Fixes #

## Checklist

- [x] Self-reviewed
- [x] Updated [CHANGELOG.md](/CHANGELOG.md) (if needed)

## Additional Notes

- I've looked for a way to enforce conventional commits on release branches, but the only option I found involved adding a merge bot, which might not be worthwhile. The current proposal instead enforces conventional commit formatting on PR titles, which then serve as the default commit titles.
- Release Notes section will be added on a separate PR

### DIFF
--- a/pkg/env-tests/env_tests_suite_test.go
+++ b/pkg/env-tests/env_tests_suite_test.go
@@ -11,7 +11,6 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	resourceapi "k8s.io/api/resource/v1"
-	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


.github/workflows/on-pr.yaml
```diff
@@ -181,3 +179,31 @@ jobs:
...
+            queue-controller
+            pod-group-controller
+            resource-reservation
+            chart
+            api
+            node-scale-adjuster
+            ci
+            release
+            docs
+          requireScope: false
```
SiorMeir, line 206: cool, this is important and gives us flexibility

CONTRIBUTING.md
```diff
@@ -34,13 +34,74 @@ Help us keep the docs clear and useful by fixing typos, updating outdated inform
...
+- 'chart'
+- 'api'
+- 'node-scale-adjuster'
+- 'ci'
+- 'release'
+- 'docs'
+
+#### Breaking Changes
+
+Breaking changes MUST be indicated by adding `!` after the type/scope: `feat(api)!: remove deprecated field`
```
SiorMeir, line 79: 2 things here:

1. BREAKING CHANGE is correlated to a major version.
I think we should add that this is not in use right now or should at least should receive a green light from the maintaining team

2. according to the standard BREAKING CHANGE can also be specified in the description of the commit. This is rarer but legit.

gshaibi: 1. We aren’t using commit types to strictly follow semantic versioning; we just adopt the conventional commit style for better readability, which is why I didn’t include versioning here.
2. You’re right - but this section refers to the PR title, which can’t include a description. Since we don’t enforce any rules on commit messages, I figured we should include that information in the title instead. What do you think?

SiorMeir: yeah I think that's ok - as you said, this is mostly a style guide. I personally don't like descriptive commit bodies.

PR #582, opened 2025-10-20
fix: update default scaling pod image name

avi-airis:
It seems the name of the scaling pod image changed recently. The default configuration in the operator would still try to use the old name causing ImagePullBackoffs unless set explicitly in the Config.

This PR updates the variable to the new name.


enoodle:
Thank you @avi-airis

PR #583, opened 2025-10-20
feat: Configure kValue for time aware fairness

itsomri:
This PR:
1. Adds the option to configure kValue in proportion plugin, which effects the importance of historical usage in time aware fairness
2. Refactors plugin arguments to use getters with defaults, which simplifies the code whenever reading plugin configs

### DIFF
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -78,8 +78,8 @@ func New(arguments framework.PluginArguments) framework.Plugin {
 		log.InfraLogger.Warningf("Failed to parse kValue: %v. Using default value of 1.0", err)
 	}
 	if kValue <= 0.0 {
-		log.InfraLogger.Warningf("kValue must be > 0.0, got %v. Using default value of 1.0", kValue)
-		kValue = 1.0
+		log.InfraLogger.Warningf("kValue must be > 0.0, got %v. Setting as 0", kValue)
+		kValue = 0.0
 	}
 
 	return &proportionPlugin{


pkg/scheduler/framework/interface.go
```diff
@@ -45,3 +53,86 @@ type Plugin interface {
...
 	OnSessionClose(ssn *Session)
 }
+
+type PluginArguments map[string]string
+
+// GetFloat64 parses a float64 value from plugin arguments.
+// Returns the parsed value and nil error on success.
+// Returns defaultValue and nil if the key doesn't exist.
+// Returns defaultValue and an error if parsing fails.
+func (pa PluginArguments) GetFloat64(key string, defaultValue float64) (float64, error) {
```
SiorMeir, line 63: This parsing methods feel useful in other places
Should they be here or in something like `common`?

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -55,33 +54,40 @@ type proportionPlugin struct {
...
+		log.InfraLogger.Warningf("relcaimerSaturationMultiplier must be >= 1.0, got %v. Using default value of 1.0", multiplier)
+		multiplier = 1.0
+	}
+
+	kValue, err := arguments.GetFloat64("kValue", 1.0)
+	if err != nil {
+		log.InfraLogger.Warningf("Failed to parse kValue: %v. Using default value of 1.0", err)
+	}
+	if kValue <= 0.0 {
+		log.InfraLogger.Warningf("kValue must be > 0.0, got %v. Using default value of 1.0", kValue)
```
SiorMeir: This is an error, no?
We can't do anything with negative values

SiorMeir (approved):
Looks real good.
Really liked `PluginArguments` - makes stuff more standardized & readable

Approving, but left 2 suggestions / questions

pkg/scheduler/plugins/proportion/proportion.go
itsomri: We can: we reset it to 1.
Actually I will reset it to 0, so -1 will be like disabling it.
We can pretty easily handle it so it's a warning only in my opinion.

pkg/scheduler/framework/interface.go
itsomri: If it will get used in other places then maybe, right now I think it's fine here

PR #588, opened 2025-10-21
Topology plugin node sorting performance improvement

gshaibi:


pkg/scheduler/plugins/topology/common.go
```diff
@@ -0,0 +1,29 @@
...
+package topology
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"slices"
+	"strings"
+
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/node_info"
+	"github.com/samber/lo"
```
davidLif: Why add this as a dependency instead of just iterating in a loop?

enoodle (approved):

PR #598, opened 2025-10-25
test: Time aware fairness burst simulation

itsomri:
This PR adds a test that validates a simple "burst" scenario for time aware fairness.

### DIFF
--- a/pkg/env-tests/timeaware/timeaware.go
+++ b/pkg/env-tests/timeaware/timeaware.go
@@ -47,6 +47,18 @@ type TestQueue struct {
 	Weight       *float64 // default is 1
 }
 
+func (q *TestQueue) SetDefaults() {
+	if q.Priority == nil {
+		q.Priority = ptr.To(100)
+	}
+	if q.DeservedGPUs == nil {
+		q.DeservedGPUs = ptr.To(0.0)
+	}
+	if q.Weight == nil {
+		q.Weight = ptr.To(1.0)
+	}
+}
+
 type TestJobs struct {
 	GPUs    int
 	NumPods int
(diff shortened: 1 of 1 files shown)


pkg/env-tests/time_aware_fairness_test.go
```diff
@@ -64,13 +63,15 @@ type TestNodes struct {
...
 type TimeAwareSimulation struct {
-	Queues []TestQueue
-	Jobs   map[string]TestJobs // key is the queue name
-	Nodes  []TestNodes
-	Cycles *int //default is 100
+	Queues     []TestQueue
+	Jobs       map[string]TestJobs // key is the queue name
+	Nodes      []TestNodes
+	Cycles     *int     // default is 100
+	WindowSize *int     // default is 5
```
gshaibi: Consider documenting it is in seconds

pkg/env-tests/time_aware_fairness_test.go
```diff
@@ -123,10 +141,10 @@ func setupControllers(backgroundCtx context.Context, cfg *rest.Config) (chan str
 	return stopCh, cancel, usageClient, nil
 }
 
-func RunSimulation(ctx context.Context, simulation TimeAwareSimulation) (fake.AllocationHistory, error, error) {
+func RunSimulation(ctx context.Context, simulation TimeAwareSimulation) (allocationHistory fake.AllocationHistory, err error, cleanupErr error) {
```
gshaibi: Consider setting all the `TimeAwareSimulation` defaults together at the start of this function

gshaibi (approved):

pkg/env-tests/time_aware_fairness_test.go
itsomri: It's not really in seconds though, it's "number of cycles", the seconds is just to conform to the API

pkg/env-tests/time_aware_fairness_test.go
itsomri: Moved setting defaults to a method which is called at the beginning

PR #601, opened 2025-10-26
Topology domain sort based on resource ratio

davidLif:
For the topology plugin, sort domain for "bin packing" calculation based on tasksRequest/FreeResourcesForDomain


gshaibi:
@davidLif Do we still need the `AllocatablePods`? 
Can/should we filter out irrelevant domains based on resource count only?

davidLif:
> @davidLif Do we still need the `AllocatablePods`? Can/should we filter out irrelevant domains based on resource count only?

Yes we do - Currently we use it to filter domais that cann't allocate the job.

pkg/scheduler/plugins/topology/topology_structs.go
```diff
@@ -49,19 +49,27 @@ type DomainInfo struct {
...
+		Nodes:                    map[string]*node_info.NodeInfo{},
+		AllocatablePods:          0,
+		IdleOrReleasingResources: resource_info.EmptyResource(),
 	}
 }
 
 func (di *DomainInfo) AddNode(nodeInfo *node_info.NodeInfo) {
 	di.Nodes[nodeInfo.Name] = nodeInfo
+	di.IdleOrReleasingResources.Add(nodeInfo.Idle)
+	di.IdleOrReleasingResources.Add(nodeInfo.Releasing)
```
gshaibi, line 71: As we discussed, we should take into account that this happens on the beginning of the session and not updated again during it, so the data may be slightly incorrect.

PR #602, opened 2025-10-27
feat(admission): Explicitly apply 'nvidia' runtimeClass to GPU pods

omeryahud:
In the upcoming release, gpu-operator no longer sets the `nvidia` runtime class as the default runtimeClass in the cluster, hence the need to set it explicitly for pod that request GPUs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 - Fixed a bug where the scheduler would not re-try updating podgroup status after failure
+- Fixed a bug where ray workloads gang scheduling would ignore `minReplicas` if autoscaling was not set
 
 ## [v0.9.1] - 20250-09-15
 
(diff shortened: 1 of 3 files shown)


pkg/binder/common/gpu_access.go
```diff
@@ -155,3 +157,7 @@ func UpdateConfigMapEnvironmentVariable(
 
 	return nil
 }
+
+func SetNVIDIARuntimeClass(pod *v1.Pod) {
```
enoodle: Thsi should be defined somewhere in the admission package because it is only used there and uses nothing from the binder package.

pkg/admission/webhook/v1alpha2/runtimeenforcement/runtime_enforcement.go
```diff
@@ -0,0 +1,74 @@
...
+	return "runtimeenforcement"
+}
+
+func (p *RuntimeEnforcement) Validate(pod *v1.Pod) error {
+	return nil
+}
+
+func (p *RuntimeEnforcement) Mutate(pod *v1.Pod) error {
+	// nvidia runtimeClass is not supported on openshift
+	isOpenshift, err := k8s_utils.IsOpenshift(context.Background(), p.kubeClient)
```
enoodle: This should be moved to the operator so that it happens only once every time the admission service is updated and not for every pod.
https://github.com/NVIDIA/KAI-Scheduler/pull/602/files#diff-393e26df056afd757588a4cbd58d05324043c8f90b4048984f3d0ff6d657a186R347

README.md
```diff
@@ -48,6 +48,13 @@ helm upgrade -i kai-scheduler oci://ghcr.io/nvidia/kai-scheduler/kai-scheduler -
 #### Build from Source
 Follow the instructions [here](docs/developer/building-from-source.md)
 
+## Flavor Specific Instructions
+### Openshift
+When `gpu-operator` <v25.10.0 is installed, the following flag should be added to the installation command:
+```
+--set admission.gpuPodRuntimeClassName=null
+```
+
```
enoodle, line 57: :+1:

pkg/common/resources/gpu_sharing.go
```diff
@@ -24,6 +24,22 @@ func RequestsGPUFraction(pod *v1.Pod) bool {
 	return foundFraction || foundGPUMemory
 }
 
+func RequestsWholeGPU(pod *v1.Pod) bool {
```
itsomri, line 27: Are you sure we don't have implementation for those already?

omeryahud: I didn't find any function that does this specifically, can you point me to it if it exists?

enoodle (dismissed):

itsomri (approved):

PR #604, opened 2025-10-27
feat: Better topology allocation for non homogeneous jobs 

davidLif:
If the topology plugin detects a non homogeneous job (different pods require different resource types) the topology plugin will check if a domain can allocate the job based on the job requested resources to domain free resources ration, instead of checking the amount of representor pods each node in the domain.

Small additional improvments:
1- Calculate AllocatablePods only for the domains relevant to the nodes subset
2- Cache ratio results of topology tree sorting by job requested resources to free domain resources 

Known issue: Does not count fractional gpus as part of the free resources

### DIFF
--- a/pkg/scheduler/plugins/topology/job_filtering.go
+++ b/pkg/scheduler/plugins/topology/job_filtering.go
@@ -493,6 +493,10 @@ func sortDomainInfos(topologyTree *Info, domainInfos []*DomainInfo) []*DomainInf
 	return sortedDomainInfos
 }
 
+// useRepresentorPodsAccounting checks if the tasks are using representor pods accounting.
+// If all the tasks are homogeneous, i.e. all the tasks have the same type of resource requirements, then use representor pods accounting (AllocatablePods).
+// If the tasks are heterogeneous, i.e. some of the tasks require resources that other tasks do not require,
+// then use the job resources sum to see if a domain can allocate the job.
 func useRepresentorPodsAccounting(tasks []*pod_info.PodInfo) bool {
 	extendedResources := map[v1.ResourceName]int{}
 	podsUsingGpu := 0


pkg/scheduler/plugins/topology/job_filtering.go
```diff
@@ -447,3 +492,25 @@ func sortDomainInfos(topologyTree *Info, domainInfos []*DomainInfo) []*DomainInf
 
 	return sortedDomainInfos
 }
+
+func useRepresentorPodsAccounting(tasks []*pod_info.PodInfo) bool {
```
gshaibi, line 500: Consider adding documentation to the function

pkg/scheduler/plugins/topology/job_filtering.go
```diff
@@ -94,6 +100,15 @@ func (t *topologyPlugin) subSetNodesFn(
 	return domainNodeSets, nil
 }
 
+func getTasksAllocationMetadata(tasks []*pod_info.PodInfo) (*resource_info.Resource, int) {
```
gshaibi, line 103: Consider only calculating the aggregated resource request, and use len(tasks) wherever you need `tasksCount` for clarity

pkg/scheduler/plugins/topology/job_filtering.go
```diff
@@ -109,20 +124,17 @@ func (t *topologyPlugin) getJobTopology(subGroup *subgroup_info.SubGroupInfo) (*
 	return topologyTree, true
 }
 
-func (t *topologyPlugin) calcTreeAllocatable(tasks []*pod_info.PodInfo, topologyTree *Info, nodeSet node_info.NodeSet) (int, error) {
-	jobAllocationData, err := initJobAllocationMetadataStruct(tasks)
+func (t *topologyPlugin) calcTreeAllocatable(tasks []*pod_info.PodInfo, domain *DomainInfo) error {
```
gshaibi, line 127: Nice optimization!

pkg/scheduler/plugins/topology/job_filtering.go
```diff
@@ -378,9 +418,14 @@ func sortTree(tasksResources *resource_info.Resource, root *DomainInfo, maxDepth
 		return
 	}
 
+	domainRatiosCache := make(map[DomainID]float64, len(root.Children))
+	for _, child := range root.Children {
+		domainRatiosCache[child.ID] = getJobRatioToFreeResources(tasksResources, child)
+	}
```
gshaibi, line 424: Thats great!

gshaibi (dismissed):

pkg/scheduler/plugins/topology/job_filtering.go
davidLif: many times I prefer to pass the metadata instead of the actual tasks list

pkg/scheduler/plugins/topology/job_filtering.go
davidLif: Added comment

PR #639, opened 2025-11-12
docs: Add time aware docs

itsomri:
## Description

This PR documents the time aware fairness feature in KAI.

## Checklist

- [x] Self-reviewed
- [x] Updated [CHANGELOG.md](/CHANGELOG.md) (if needed)

## Breaking Changes

N/A

## Additional Notes

N/A

### DIFF
--- a/docs/timeaware/README.md
+++ b/docs/timeaware/README.md
@@ -2,7 +2,7 @@
 
 Time aware fairness is a feature in KAI-Scheduler which makes use of historical resource usage by queues for making allocation and reclaim decisions. Key features are:
 
-1. Consider past usage for order of allocation: all else being equal, queues with higher past usage will get to run jobs after queues with lower usage
+1. All else being equal, queues with higher past usage will get to run jobs after queues with lower usage
 2. Reclaim based on usage: queues which are starved over time will reclaim resources from queues which used a lot of resources.
     1. Note: this does not effect in-quota allocation: deserved quota still takes precedence over time-aware fairness
 
@@ -34,11 +34,11 @@ Where:
 
 #### Normalization to cluster capacity
 
-The aggregated usage for each queue is then normalized to the **cluster capacity** at the relevant time period: the scheduler looks at the available resources in the cluster for that time period, and normalizes all resource usage to it. For example, in a cluster with 10 GPUs, and considering a time period of 10 hours, a queue which consumed 24 GPU hours (wether it's 8 GPUs for 3 hours, or 12 GPUs for 2 hours), will get a normalized usage score of 0.24 (used 24 GPU hours out of a potential 100). This normalization ensures that a small amount of resource usage in a vacant cluster will not result in a heavy penalty.
+The aggregated usage for each queue is then normalized to the **cluster capacity** at the relevant time period: the scheduler looks at the available resources in the cluster for that time period, and normalizes all resource usage to it. For example, in a cluster with 10 GPUs, and considering a time period of 10 hours, a queue which consumed 24 GPU hours (wether it's 8 GPUs for 3 hours, or 12 GPUs for 2 hours), will get a normalized usage score of 0.24 (used 24 GPU hours out of a potential 100). This normalization ensures that a small amount of resource usage relative to the cluster size will not result in a heavy penalty.
 
 ### Effect on fair share
 
-Usually, over quota resources is divided to each queue proportionally to it's Over Quota Weight. With time-aware fairness, queues with historical usage will get relatively less resources in over-quota. The significance of the resource usage in this calculation can be controlled with a parameter called "kValue": the bigger it is, the more significant the historical usage be.
+Usually, over quota resources are assigned to each queue proportionally to it's Over Quota Weight. With time-aware fairness, queues with historical usage will get relatively less resources in over-quota. The significance of the resource usage in this calculation can be controlled with a parameter called "kValue": the bigger it is, the more impact (or weight) the historical usage has on the calculated fairshare, i.e. it will decrease the fairshare of that queue.
 
 Check out the [time aware simulator](../../cmd/time-aware-simulator/README.md) to understand scheduling behavior over time better.
 
@@ -91,12 +91,12 @@ Add the following section under `spec`:
     usageParams:
       windowSize: 1w # The time period considered for fairness calculations. One week is the default
       windowType: sliding # Change to the desired value (sliding/tumbling). Sliding is the default
-      halfLifePeriod: 10m # Leave empty to not use time decay
+      halfLifePeriod: 10m # Leave empty to not use time decay. Off by default
 ```
 
 #### kValue
 
-KValue is a parameter used by the proportion plugin to determine the significance of historical usage in fairness calculations - higher values mean more aggressive effects on fairness. To set it, add it to the scheduling shard spec:
+KValue is a parameter used by the proportion plugin to determine the impact of historical usage in fairness calculations - higher values mean more aggressive effects on fairness. To set it, add it to the scheduling shard spec:
 ```sh
 kubectl edit schedulingshard default
 ```


cmd/time-aware-simulator/examples/oscillating.csv
enoodle, line 1: If this file is generated from the python script, maybe you can omit it?

docs/timeaware/results.png
enoodle, line 1: Maybe you can use the image from `cmd/time-aware-simulator/example` ?

docs/timeaware/README.md
```diff
@@ -0,0 +1,190 @@
+# Time Aware fairness
+
+Time aware fairness is a feature in KAI-Scheduler which makes use of historical resource usage by queues for making allocation and reclaim decisions. Key features are:
+
+1. Consider past usage for order of allocation: all else being equal, queues with higher past usage will get to run jobs after queues with lower usage
```
Hagay-RunAI: I would change the order of the explanation: "all else being equal, queues with lower past usage will get to run jobs before queues with higher past usage.

docs/timeaware/README.md
```diff
@@ -0,0 +1,190 @@
...
+2. Reclaim based on usage: queues which are starved over time will reclaim resources from queues which used a lot of resources.
+    1. Note: this does not effect in-quota allocation: deserved quota still takes precedence over time-aware fairness
```
Hagay-RunAI, line 7: I would actually add this to the top of the explanation, right before talking about the main features. The scheduler first allocated workloads and reclaims resources from queue (while preempting workloads) to satisfythe deserve in-quota queues.

docs/timeaware/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+> **Prerequisites**: Familiarity with [fairness](../fairness/README.md)
+
+## How it works
+
+In high level: resource usage data in the cluster is collected and persisted in prometheus. It is then used by the scheduler to make resource fairness calculations: the more resources consumed by a queue, the less over-quota resources it will get compared to other queues. This will eventually result in the queues' over-quota resources being reclaimed by more starved queues, thus achieving a more fair allocation of resources over time.
+
+### Resource usage data
+
+Queue historical resource usage data is collected in a prometheus instance in the cluster *(external prometheus instances can be used - see [External prometheus](#external-prometheus))*. The scheduler configuration determines the time period that will be considered, as well as allows configuration for time-decay, which, if configured, gives more weight to recent usage than past usage.
```
Hagay-RunAI, line 18: "determines the time period that will be considered" - considered for what ?  is that the time window or the "micro" periods of time in which scheduler measures/records usage ?
Generally I think we need to either explain in a separate paragraph the terminology and it meaning.

docs/timeaware/README.md
```diff
@@ -0,0 +1,190 @@
...
+1. Consider past usage for order of allocation: all else being equal, queues with higher past usage will get to run jobs after queues with lower usage
+2. Reclaim based on usage: queues which are starved over time will reclaim resources from queues which used a lot of resources.
+    1. Note: this does not effect in-quota allocation: deserved quota still takes precedence over time-aware fairness
+
+
+> **Prerequisites**: Familiarity with [fairness](../fairness/README.md)
+
+## How it works
+
+In high level: resource usage data in the cluster is collected and persisted in prometheus. It is then used by the scheduler to make resource fairness calculations: the more resources consumed by a queue, the less over-quota resources it will get compared to other queues. This will eventually result in the queues' over-quota resources being reclaimed by more starved queues, thus achieving a more fair allocation of resources over time.
```
Hagay-RunAI, line 14: "the more resources consumed by a queue" - is it more correct to say "the more **over-quota** resources consumed by a queue" ?

docs/timeaware/results.png
itsomri: I'll leave it here and remove the other one, it's more relevant here

docs/timeaware/README.md
```diff
@@ -0,0 +1,190 @@
...
+Where:
+
+- $U$ is the usage
+- $t_{1/2}$ is the half life constant set by the user
+- $\Delta{t}$ is the time elapsed since that usage
+- $A$ is the allocated resource
+
+#### Normalization to cluster capacity
+
+The aggregated usage for each queue is then normalized to the **cluster capacity** at the relevant time period: the scheduler looks at the available resources in the cluster for that time period, and normalizes all resource usage to it. For example, in a cluster with 10 GPUs, and considering a time period of 10 hours, a queue which consumed 24 GPU hours (wether it's 8 GPUs for 3 hours, or 12 GPUs for 2 hours), will get a normalized usage score of 0.24 (used 24 GPU hours out of a potential 100). This normalization ensures that a small amount of resource usage in a vacant cluster will not result in a heavy penalty.
```
Hagay-RunAI: Need to explain what "time period" is, I already made similar comments earlier - we need a glossary to explain all those terms, or just explain what time period is before we start using it to explain things.

Hagay-RunAI: "This normalization ensures that a small amount of resource usage in a vacant cluster will not result in a heavy penalty." - not sure I understand this, if the cluster is vacant or fully occupied, how does it impact you calculation ? you always normalize against the total available GPU hours in this time period. I may be missing something here but Vacant Cluster means a lot of available resources, hence small amount of used resources is normalized as even smaller vs. the case an occupied cluster there are few free resources and the small amount is relatively large to the total available resources.

docs/timeaware/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Effect on fair share
+
+Usually, over quota resources is divided to each queue proportionally to it's Over Quota Weight. With time-aware fairness, queues with historical usage will get relatively less resources in over-quota. The significance of the resource usage in this calculation can be controlled with a parameter called "kValue": the bigger it is, the more significant the historical usage be.
```
Hagay-RunAI: 1. "Over quota resources are divided" instead of "is divided"
2. Maybe use "given" or "allocated" instead of "divided"
3. ""kValue": the bigger it is, the more significant the historical usage be." -> ""kValue": the bigger it is, the more impact (or weight) the historical usage has on the calculated fairshare, i.e. it will decrease the fairshare of that queue."

docs/timeaware/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+Add the following section under `spec`:
+```yaml
+  usageDBConfig:
+    clientType: prometheus
+    connectionString: http://prometheus-operated.kai-scheduler.svc.cluster.local:9090 # Optional: if not configured, the kai config will populate it 
+    usageParams:
+      windowSize: 1w # The time period considered for fairness calculations. One week is the default
+      windowType: sliding # Change to the desired value (sliding/tumbling). Sliding is the default
+      halfLifePeriod: 10m # Leave empty to not use time decay
```
Hagay-RunAI: Half life is 10 minutes by default ??
Wouldn't it make sense to make it at least 1 hour ? CS are using 24hours for 2 weeks window ...

docs/timeaware/README.md
```diff
@@ -0,0 +1,189 @@
...
+```
+
+#### kValue
+
+KValue is a parameter used by the proportion plugin to determine the significance of historical usage in fairness calculations - higher values mean more aggressive effects on fairness. To set it, add it to the scheduling shard spec:
```
Hagay-RunAI: - "the significance of historical usage" maybe change to "the impact of historical usage on fairshare calculations "
- "higher values mean more aggressive effects on fairness." maybe change to. "higher K-values means historical usage has more aggressive impact on fairshare, at K=1 ... at K=0 ...

docs/timeaware/README.md
```diff
@@ -0,0 +1,189 @@
...
+kubectl edit schedulingshard default
+```
+
+```yaml
+  usageDBConfig:
+    extraParams:
+      gpuAllocationMetric: kai_queue_allocated_gpus
+      cpuAllocationMetric: kai_queue_allocated_cpu_cores
+      memoryAllocationMetric: kai_queue_allocated_memory_bytes
+      gpuCapacityMetric: sum(kube_node_status_capacity{resource=\"nvidia_com_gpu\"})
```
Hagay-RunAI, line 128: I assume DRA won't affect this since we will make sure the same metrics is still updated, right ?

docs/timeaware/README.md
itsomri: I want the intro to be as straightforward as possible: readers might not know or care about deserved quota. This is still very high, it's the 4th line

docs/timeaware/README.md
itsomri: No, in-quota resources count as well

docs/timeaware/README.md
itsomri: The time period for which we look at consumed resources. Resource consumption outside this window does not matter. How do you suggest to write this?

docs/timeaware/README.md
itsomri: It's the time for which we count consumed resources, I don't know how to explain it better - do you have a suggestion?

itsomri: You're right, the vacancy actually doesn't matter, changed to "relative to the cluster size"

docs/timeaware/README.md
itsomri: 1. Change do "are assigned", allocated has a different connotation
2. Assigned
3. 👍

docs/timeaware/README.md
itsomri: No, it's turned off by default

docs/timeaware/README.md
itsomri: I don't know

docs/timeaware/README.md
itsomri: Fixed the first point
Regarding the second - I don't really want to give specific examples here, I don't want to create expectations for something that's really easily impacted by different factors. We can later on add examples of different simulation results

PR #628, opened 2025-11-10
feat(chart): Add flexible image tag configuration with priority-based overrides

lokielse:
## Description

This PR introduces a flexible image tag configuration system for the Helm chart, allowing users to override image tags at different levels with a clear priority order. Additionally, this PR fixes an incorrect reference where `.Chart.Version` was being used instead of `.Chart.AppVersion` for image tags.

### Changes Made

1. **Added `tag` field to all component image configurations** in `values.yaml`:
   - operator.image.tag
   - podgrouper.image.tag
   - podgroupcontroller.image.tag
   - binder.image.tag
   - binder.resourceReservationImage.tag
   - scheduler.image.tag
   - queuecontroller.image.tag
   - admission.image.tag
   - nodescaleadjuster.image.tag
   - nodescaleadjuster.scalingPodImage.tag
   - crdupgrader.image.tag

2. **Implemented priority-based tag resolution** using Helm's `default` function:
   ```
   Priority: <component>.image.tag > global.tag > .Chart.AppVersion
   ```

3. **Fixed incorrect Chart reference**: Changed all occurrences of `.Chart.Version` to `.Chart.AppVersion` for image tags
   - `.Chart.Version` is intended for the Helm chart version
   - `.Chart.AppVersion` is the correct field for application/image versions

### Updated Templates

- `deployments/kai-scheduler/templates/kai-config.yaml`
- `deployments/kai-scheduler/templates/services/operator.yaml`
- `deployments/kai-scheduler/templates/crd-upgrader.yaml`

### Benefits

- **Flexibility**: Users can now override image tags at three different levels
- **Global control**: Set `global.tag` to apply a tag to all components
- **Component-specific overrides**: Set individual component tags when needed (e.g., testing a new version of one component)
- **Backward compatibility**: Existing deployments continue to work without changes
- **Correctness**: Fixed the incorrect use of `.Chart.Version` for image tags

### Usage Examples

```yaml
# Use global.tag for all components
global:
  tag: "v1.2.0"

# Override specific component
operator:
  image:
    tag: "v1.3.0-rc1"  # operator will use v1.3.0-rc1, others use v1.2.0

# Fall back to Chart.AppVersion when nothing is set
# (default behavior, no configuration needed)
```

## Related Issues

N/A

## Checklist

- [x] Self-reviewed
- [x] Added/updated tests (tested with `helm template` command)
- [ ] Updated [CHANGELOG.md](/CHANGELOG.md) (if needed)
- [ ] Updated documentation (if needed)

[...]

PR #640, opened 2025-11-12
docs: update helm installation commands for image tag compatibility

lokielse:
## Description

This PR updates the helm installation commands in the documentation to include the `--set-string global.tag=<VERSION>` parameter, which is required after the changes introduced in PR #628.

**Background:**
PR #628 changed the image tag resolution mechanism from using `{{ .Chart.Version }}` directly to using a fallback chain: `component.image.tag → global.tag → Chart.AppVersion`. Since `global.tag` now defaults to an empty string in `values.yaml`, the `--version` flag alone no longer controls the image tags used by the deployment.

**Impact:**
Without this fix, users installing KAI Scheduler with `--version X.Y.Z` would get the correct chart version but incorrect image tags (using `Chart.AppVersion` instead of the specified version).

**Changes:**
1. Updated production installation command in `README.md` to include `--set-string global.tag=<VERSION>`
2. Updated local build installation command in `docs/developer/building-from-source.md` to include `--set-string global.tag=0.0.0`

This ensures that both the Helm chart version and container image tags align with the user's intended version.

## Related Issues

Related to #628

> **Note:** This PR updates the documentation to reflect the changes introduced in #628, which modified the image tag resolution mechanism from `{{ .Chart.Version }}` to a fallback chain of `component.image.tag → global.tag → Chart.AppVersion`.

## Checklist

- [x] Self-reviewed
- [ ] Added/updated tests (if needed) - N/A for documentation changes

## Breaking Changes

No breaking changes - this PR fixes documentation to match the current implementation.

## Additional Notes

**Files Modified:**
- `README.md` - Added `--set-string global.tag=<VERSION>` to the production installation command
- `docs/developer/building-from-source.md` - Added `--set-string global.tag=0.0.0` to the local build installation command

**Verification:**
Searched all markdown files in the repository to ensure all helm installation commands have been updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ KAI Scheduler can be installed:
 Locate the latest release version in [releases](https://github.com/NVIDIA/KAI-Scheduler/releases) page.
 Run the following command after replacing `<VERSION>` with the desired release version:
 ```sh
-helm upgrade -i kai-scheduler oci://ghcr.io/nvidia/kai-scheduler/kai-scheduler -n kai-scheduler --create-namespace --version <VERSION> --set-string global.tag=<VERSION>
+helm upgrade -i kai-scheduler oci://ghcr.io/nvidia/kai-scheduler/kai-scheduler -n kai-scheduler --create-namespace --set-string global.tag=<VERSION>
 ```
 #### Build from Source
 Follow the instructions [here](docs/developer/building-from-source.md)


README.md
```diff
@@ -43,7 +43,7 @@ KAI Scheduler can be installed:
 Locate the latest release version in [releases](https://github.com/NVIDIA/KAI-Scheduler/releases) page.
 Run the following command after replacing `<VERSION>` with the desired release version:
 ```sh
-helm upgrade -i kai-scheduler oci://ghcr.io/nvidia/kai-scheduler/kai-scheduler -n kai-scheduler --create-namespace --version <VERSION>
+helm upgrade -i kai-scheduler oci://ghcr.io/nvidia/kai-scheduler/kai-scheduler -n kai-scheduler --create-namespace --version <VERSION> --set-string global.tag=<VERSION>
```
enoodle: Why is this needed? when not set the helm version should be used.

lokielse: There are 2 reasons:
1. `chat.version` is a semantic version by its definition like `1.2.3`, so `v1.2.3` is not a standard version.
2. There is another `chat.appVersion` which is a right choice for image tag because usually we may publish 1 chart to registry with multiple app versions(images)

In conclusion
- `chat.version` is for the helm chart itself. We need to bump it if we make changes to helm chart. The chart need to be published to registry if its version is changed.
- `chat.appVersion` is for application/docker images. No need to publish a new version of helm chart.

image

enoodle: I didn't understand that when approving the other PR. Maybe we should change the chart so that it is easy to set up instead of adding more arguments for it to work well?

lokielse: sorry for misleading you by the previous commit.
It's not add an additional argument.
It's replacing the `--version <VERSION>` to `--set-string global.tag=<VERSION>`.

I've pushed the latest commit.

enoodle: But then which version of the chart will be installed?

enoodle: I think that we need to make changes to the chart so that it will support the current commands, because there are already many scripts that depend on them.

lokielse: Let me close this PR. 
I checked the details of building helm chart step.
It sets both `version` and `appVersion` so that it can select a correct chart by specify the `version` argument.
And images reference to `appVersion` which is the same value as `version`.

```yaml
      - name: Build helm chart
        run: |
          sed -i 's#registry/local/kai-scheduler#${{ env.DOCKER_REGISTRY }}#' deployments/kai-scheduler/values.yaml
          helm package ./deployments/kai-scheduler -d ./charts --app-version $PACKAGE_VERSION --version $PACKAGE_VERSION

      - name: Push Helm Chart
        run: |
          helm push ./charts/kai-scheduler-$PACKAGE_VERSION.tgz oci://${{ env.DOCKER_REGISTRY }}

```

PR #678, opened 2025-11-21
fix(chart): Helm uninstall leaves behind scheduler pods due to "kept" SchedulingShard

srujanreddya:
## Description
**What does this PR do?**

- This PR ensures that scheduler-related resources (SchedulingShard CRs and their deployments/pods) are properly cleaned up during helm uninstall.
- Previously, the uninstall workflow left behind scheduler pods because the SchedulingShard CR was marked as “kept”, which prevented Helm from removing the associated deployment.
- To resolve this, the PR introduces a post-delete cleanup Job (with `post-delete-cleanup` SA/RBAC) that explicitly deletes the all Deployments.
- All cleanup resources are registered as post-delete Helm hooks so they are created after the main uninstall and automatically removed once the cleanup completes.

**Why is this needed?**

1. Helm does not delete resources that are:
- owned by CRDs
- marked as "kept"
- created dynamically by an operator
2. The SchedulingShard resource was left behind, which caused:
- scheduler pods to remain running
- partial uninstall
- stale resources in the cluster
3. Users expect a clean uninstall of the operator and all scheduler components.

## Related Issues
https://github.com/NVIDIA/KAI-Scheduler/issues/661

## Checklist

- [x] Self-reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+## [v0.10.1] - 2025-11-23
+
 ### Fixed
+- Fixed scheduler pod group status update conflict [#676](https://github.com/NVIDIA/KAI-Scheduler/pull/676) [davidLif](https://github.com/davidLif) 
 - Fixed gpu request validations for pods [#660](https://github.com/NVIDIA/KAI-Scheduler/pull/660) [itsomri](https://github.com/itsomri)
 
-## [v0.10.0] - 20250-11-18
+### Changed
+- Dependabot configuration to update actions in workflows [#651](https://github.com/NVIDIA/KAI-Scheduler/pull/651) [ScottBrenner](https://github.com/ScottBrenner)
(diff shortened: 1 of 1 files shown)


enoodle:
Direction looks good, but I guess the changes to the RBAC brock the chart and the new files need license added to them (look at the test failures)

srujanreddya:
> Direction looks good, but I guess the changes to the RBAC brock the chart and the new files need license added to them (look at the test failures)

Thank you. I see instead of using the operator service account, created new rbacs explicitly for post delete clean up job since the controller gen is overwriting the clusterrole.

srujanreddya:
Also, QQ: I don't see in workflow the uninstall case? is that something we can add as validation?

enoodle:
> Also, QQ: I don't see in workflow the uninstall case? is that something we can add as validation?

Do you mean some kind of test to validate that this works well? We have the e2e tests that use helm to install the whole system and run tests but they don't do uninstall. If you would like to you can add a step to verify that no pods are left in the e2e workflow.

srujanreddya:
> > Also, QQ: I don't see in workflow the uninstall case? is that something we can add as validation?
> 
> Do you mean some kind of test to validate that this works well? We have the e2e tests that use helm to install the whole system and run tests but they don't do uninstall. If you would like to you can add a step to verify that no pods are left in the e2e workflow.

Added stage to uninstall kai scheduler and check if stale resources exists in `kai-scheduler` namespace.

PR #682, opened 2025-11-24
feat(ci): release notes automation

gshaibi:
## Release Notes

NONE

## Related Issues

Fixes #

## Checklist

- [ ] Self-reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+## [v0.10.2] - 2025-11-24
+
 ### Fixed
-- Removed the requirement to specify container type for init container gpu fractions [#684](https://github.com/NVIDIA/KAI-Scheduler/pull/684) (itsomri)(https://github.com/itsomri)
+- Removed the requirement to specify container type for init container gpu fractions [#684](https://github.com/NVIDIA/KAI-Scheduler/pull/684) [itsomri](https://github.com/itsomri)
+- When a status update for a podGroup in the scheduler is flushed due to update conflict, delete the update payload data as well [#691](https://github.com/NVIDIA/KAI-Scheduler/pull/691) [davidLif](https://github.com/davidLif)
 
 ## [v0.10.1] - 2025-11-23
 
--- a/pkg/scheduler/cache/status_updater/concurrency.go
+++ b/pkg/scheduler/cache/status_updater/concurrency.go
@@ -111,6 +111,7 @@ func (su *defaultStatusUpdater) updatePodGroup(
 				// Don't retry this update if the resource version is outdated - The status update cannot be updated with the given object.
 				// If a pod group status update is required (e.g. a scheduling condition) a new status update with an updated object
 				//  will be enqueued in the next scheduling cycle.
+				su.inFlightPodGroups.Delete(key)
 				return
 			}
 		}


.github/scripts/release-notes.py
```diff
@@ -0,0 +1,277 @@
...
+    categories = parse_release_notes(release_notes)
+    
+    # Validate that we have at least one category with entries
+    if not categories:
+        return False, "Release notes must contain at least one valid category (Added, Changed, Deprecated, Removed, Fixed, Security) with entries, or 'NONE' to opt out", None
+    
+    # Check that each category has at least one entry
+    empty_categories = [cat for cat, entries in categories.items() if not entries]
+    if empty_categories:
+        return False, f"Categories {', '.join(empty_categories)} have no entries", None
```
davidLif, line 158: Why each category needs at least one entry?

.github/scripts/release-notes.py
```diff
@@ -0,0 +1,277 @@
...
+    
+    for line in cleaned.split('\n'):
+        line = line.strip()
+        if not line:
+            continue
+            
+        # Check if this is a category heading
+        category_match = re.match(r'^###\s+(.+)$', line)
+        if category_match:
+            category = category_match.group(1).strip()
```
SiorMeir, line 106: should this be case sensitive / insensitive?

.github/scripts/release-notes.py
```diff
@@ -0,0 +1,277 @@
...
+                    formatted_entry = f"{entry} [#{pr_number}]({pr_url}) [{author}]({author_url})"
+                else:
+                    formatted_entry = entry
+                
+                # Avoid duplicates
+                if formatted_entry not in all_categories[category]:
+                    all_categories[category].append(formatted_entry)
+    
+    # Generate final changelog content
+    category_order = ["Added", "Changed", "Deprecated", "Removed", "Fixed", "Security"]
```
SiorMeir: A bit of nitpicking - you can just use `VALID_CATEGORIES` you already created

.github/scripts/release-notes.py
gshaibi: This is on the PR validation, it validates that such thing doesn't happen:
```
## Release Notes

### Fixed
```

which is invalid since the `Fixed` category is empty

.github/scripts/release-notes.py
gshaibi: You're right, thanks!

.github/scripts/release-notes.py
gshaibi: It's currently case-sensitive, but I’m not strongly attached to that behavior - let me know if you’d prefer it to be case-insensitive instead.

SiorMeir: either is ok - I just think it needs to be clear (had a hard time deducting that from the regex)

.github/scripts/release-notes.py
```diff
@@ -0,0 +1,275 @@
...
+    """
+    # Read PRs from file
+    with open(prs_json_path, 'r') as f:
+        prs = json.load(f)
+    
+    # Collect all entries by category across all PRs
+    all_categories = {}
+    
+    for pr in prs:
+        pr_number = pr['number']
```
SiorMeir, line 181: `pr_number = pr.get("number", None)` is safer (otherwise it might raise an exception)

.github/scripts/release-notes.py
```diff
@@ -0,0 +1,275 @@
...
+        Formatted changelog text with all entries organized by category
+    """
+    # Read PRs from file
+    with open(prs_json_path, 'r') as f:
+        prs = json.load(f)
+    
+    # Collect all entries by category across all PRs
+    all_categories = {}
+    
+    for pr in prs:
```
SiorMeir, line 180: by the way you might be able to unpack with `zip` builtin method

.github/workflows/generate-release.yaml
```diff
@@ -0,0 +1,298 @@
...
+        with:
+          fetch-depth: 0  # Fetch all history for tag comparison
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+      
+      - name: Install Python dependencies
+        run: pip install packaging
```
SiorMeir: I suggest setting explicit version

`pip install packaging==x.y.z` 

This may prevent some shenanigans :)

.github/workflows/generate-release.yaml
```diff
@@ -0,0 +1,298 @@
...
+          fi
+          echo "previous_tag=$PREVIOUS_TAG" >> $GITHUB_OUTPUT
+          echo "Previous tag: $PREVIOUS_TAG"
+
+      - name: Get merged PRs since last release
+        id: get_prs
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const previousTag = '${{ steps.previous_tag.outputs.previous_tag }}';
```
SiorMeir, line 108: python , js and bash :) 

Maybe extract this to a file? a bit hard to read

.github/workflows/generate-release.yaml
```diff
@@ -0,0 +1,298 @@
...
+          else
+            git commit -m "Update CHANGELOG.md for $VERSION"
+            git push origin main
+            echo "✅ CHANGELOG.md also committed and pushed to main"
+          fi
+
+      - name: Create GitHub Release
+        uses: actions/github-script@v7
+        with:
+          script: |
```
SiorMeir, line 251: here as well?

.github/scripts/release-notes.py
gshaibi: I prefer allowing the exception here, the code shouldn't proceed without a PR number.

.github/scripts/release-notes.py
gshaibi: Not sure what you mean, can you explain how exactly `zip` fits here?

.github/workflows/generate-release.yaml
gshaibi: While I generally prefer keeping all logic within the Python script for consistency, certain GitHub-specific operations are significantly easier to handle using the  `github-script`  action (JavaScript).
Because of this, I split the workflow:
	•	JavaScript ( `github-script` ): Handles interactions with the GitHub API, specifically the “Get PRs” and “Generate Release” steps.
	•	Python: Handles the majority of the remaining logic and processing.

.github/workflows/generate-release.yaml
gshaibi: True, thanks!

.github/scripts/release-notes.py
SiorMeir: sorry, I'm mistaken with `zip`

but you can use unpacking syntax:

```
def display_info(name, age, city):
    print(f"Name: {name}, Age: {age}, City: {city}")

person_data = {"name": "Alice", "age": 25, "city": "New York"}
display_info(**person_data)
```

PR #707, opened 2025-12-03
feat: decouple from kueue

gshaibi:
## Description

Currently, KAI relies on Kueue’s Topology CRD to define topology logic. 
This creates two main issues:
- Compatibility: It introduces potential versioning and compatibility friction between KAI and Kueue.
- User Experience: It causes confusion for customers who question why they must import/install Kueue dependencies when strictly using KAI.

This PR aims to decouple KAI from Kueue's definitions by introducing a native KAI Topology CRD. This involves creating the new definition and ensuring a smooth transition for existing deployments.

## Related Issues

Fixes #

## Checklist

- [x] Self-reviewed

### DIFF
--- a/docs/developer/designs/segmented-subgroups/README.md
+++ b/docs/developer/designs/segmented-subgroups/README.md
@@ -1,0 +1,206 @@
+# TAS - SubGroup Segment Support
+
+## Motivation
+
+Distributed workloads often use hierarchical communication patterns (e.g., 16 workers in 4 tensor-parallel groups of 4). Communication within groups is frequent and requires tight locality (e.g., NVLink/NVSwitch), while inter-group communication is less frequent.
+
+While SubGroups support hierarchical topology constraints, uniform grouping is currently not supported. We aim to support dividing a subgroup into **Segments**, where each segment has its pods co-located, without the user having to manually define $N$ identical subgroups in the spec.
+
+## Reference Implementations
+
+- Kueue Two-Level Topology: [PR #5449](https://github.com/kubernetes-sigs/kueue/pull/5449), [PR #5353](https://github.com/kubernetes-sigs/kueue/pull/5353), [PR #5596](https://github.com/kubernetes-sigs/kueue/pull/5596)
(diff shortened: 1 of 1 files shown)


davidLif (commented):
I think that we might need to add some ackloedgment in the NOTICE file and/or the crd file that this crd is based on the kueue topology crd

gshaibi:
> I think that we might need to add some ackloedgment in the NOTICE file and/or the crd file that this crd is based on the kueue topology crd

The NOTICE contains the following:
image

Do you think we should change it?

deployments/kai-scheduler/templates/hooks/pre/topology-migration-configmap.yaml
```diff
@@ -0,0 +1,75 @@
...
+kind: ConfigMap
+metadata:
+  name: topology-migration-script
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "3"
+    "helm.sh/hook-delete-policy": hook-succeeded,before-hook-creation
+data:
+  migrate-topologies.sh: |
```
SiorMeir: move to a separate file?

deployments/kai-scheduler/values.yaml
```diff
@@ -119,6 +119,18 @@ crdupgrader:
...
 
+topologyMigration:
+  # Enable migration of Kueue Topology CRs to KAI Topology CRs
+  # When enabled, a pre-install/pre-upgrade hook will:
+  # 1. Check if Kueue Topology CRD exists
+  # 2. Copy existing Kueue Topologies to KAI Topologies with the same spec
+  # The migrated topologies will have an annotation indicating they were migrated
+  enabled: true
+  image:
+    repository: bitnami/kubectl
```
itsomri: This will probably not pass security checks. You can use the same image as the crd upgrader https://github.com/NVIDIA/KAI-Scheduler/blob/main/deployments/crd-upgrader/Dockerfile

Issue #659, opened 2025-11-19
Bug: Is kai-scheduler required to have GPU Operator (ClusterPolicy CRD) installed?


googs1025:
### What happened?

Hi maintainers,

I'm trying to deploy `kai-scheduler` v0.10.0 in a Kubernetes cluster that does not have NVIDIA GPU Operator installed, and I noticed the following recurring error in the kai-operator logs:

```bash
➜  dynamo kubectl logs -f kai-operator-6c7598cd96-zqwp2 -nkai-scheduler
2025-11-19T11:21:30Z	INFO	Attempting to register Prometheus resource management
2025-11-19T11:21:30Z	INFO	Successfully registered Prometheus resource management
2025-11-19T11:21:30Z	INFO	Attempting to register ServiceMonitor resource management
2025-11-19T11:21:30Z	INFO	Successfully registered ServiceMonitor resource management
2025-11-19T11:21:30Z	INFO	setup	starting manager
2025-11-19T11:21:30Z	INFO	starting server	{"name": "health probe", "addr": "[::]:8081"}
2025-11-19T11:21:30Z	INFO	controller-runtime.metrics	Starting metrics server
2025-11-19T11:21:30Z	INFO	controller-runtime.metrics	Serving metrics server	{"bindAddress": ":8080", "secure": false}
2025-11-19T11:21:30Z	INFO	Starting EventSource	{"controller": "schedulingshard", "controllerGroup": "kai.scheduler", "controllerKind": "SchedulingShard", "source": "kind source: *v1.Config"}
2025-11-19T11:21:30Z	INFO	Starting EventSource	{"controller": "config", "controllerGroup": "kai.scheduler", "controllerKind": "Config", "source": "kind source: *v1.Secret"}
2025-11-19T11:21:30Z	INFO	Starting EventSource	{"controller": "schedulingshard", "controllerGroup": "kai.scheduler", "controllerKind": "SchedulingShard", "source": "kind source: *v1.ServiceAccount"}
2025-11-19T11:21:30Z	INFO	Starting EventSource	{"controller": "config", "controllerGroup": "kai.scheduler", "controllerKind": "Config", "source": "kind source: *v1.Deployment"}
2025-11-19T11:21:30Z	INFO	Starting EventSource	{"controller": "schedulingshard", "controllerGroup": "kai.scheduler", "controllerKind": "SchedulingShard", "source": "kind source: *v1.SchedulingShard"}
[...]

Issue #649, opened 2025-11-14
[Question] no matches for kind "ClusterPolicy" in version "nvidia.com/v1"


lokielse:
## What happened?

I tried to install kai-scheduler in our **GKE** cluster, however it encountered this error in the `kai-operator` pod.
It looks like we need to install [`gpu-operator`](https://github.com/NVIDIA/gpu-operator).
As we have already installed `nvidia driver` & `device-plugins`, so is it safe to install `gpu-operator`? will it cause compatible issues? What's the relationship between `gpu-operator` and `nvidia-driver` and `nvidia-gpu-device-plugin`?

## Error logs
```
2025-11-14T09:45:03Z	ERROR	controller-runtime.source.Kind	if kind is a CRD, it should be installed before calling Start	{"kind": "ClusterPolicy.nvidia.com", "error": "no matches for kind \"ClusterPolicy\" in version \"nvidia.com/v1\""}
sigs.k8s.io/controller-runtime/pkg/internal/source.(*Kind[...]).Start.func1.1
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/source/kind.go:75
k8s.io/apimachinery/pkg/util/wait.loopConditionUntilContext.func1
	/go/pkg/mod/k8s.io/apimachinery@v0.34.1/pkg/util/wait/loop.go:53
k8s.io/apimachinery/pkg/util/wait.loopConditionUntilContext
	/go/pkg/mod/k8s.io/apimachinery@v0.34.1/pkg/util/wait/loop.go:54
k8s.io/apimachinery/pkg/util/wait.PollUntilContextCancel
	/go/pkg/mod/k8s.io/apimachinery@v0.34.1/pkg/util/wait/poll.go:33
sigs.k8s.io/controller-runtime/pkg/internal/source.(*Kind[...]).Start.func1
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/source/kind.go:68
```

## What do we have?
```bash
kubectl get daemonset -n kube-system | grep nvidia
nvidia-driver-installer                  0         0         0       0            0           <none>                                                               2y154d
nvidia-gpu-device-plugin-large-cos       15        15        15      15           15          <none>                                                               654d
nvidia-gpu-device-plugin-large-ubuntu    0         0         0       0            0           <none>                                                               654d
[...]

PR #717, opened 2025-12-05
fix(operator): only watch cluster policy if it exists

enoodle:
## Description

This will relax the dependency on gpu-operator being installed on the same cluster with KAI-scheduler. ) by only watching ClusterPolicy  CRD if it exists in the cluster. This will also mean that if the cluster policy CRD with the GPU operator will be installed later than the operator pod will not watch it until  it is deleted and a new operator pod is created.

## Related Issues
referencing #649 and #659 

## Checklist

- [x] Self-reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 - Fixed a bug where the snapshot tool would not load topology objects [#720](https://github.com/NVIDIA/KAI-Scheduler/pull/720) [itsomri](https://github.com/itsomri)
+- Operator to conditionally watch ClusterPolicy based on its existence, preventing errors in its absence
 
 ## [v0.10.2] - 2025-11-24
 


pkg/operator/controller/config_controller.go
```diff
@@ -168,3 +172,22 @@ func enqueueWatched(_ context.Context, _ client.Object) []ctrl.Request {
...
 	}
 }
+
+func checkForClusterPolicy(mgr ctrl.Manager) bool {
+	logger := log.FromContext(context.Background())
+	tempClient, err := client.New(mgr.GetConfig(), client.Options{Scheme: mgr.GetScheme()})
+	if err != nil {
+		logger.Info("Failed to create temporary client to check for cluster policy", "error", err)
+		return false
+	}
```
romanbaron, line 182: I think that you can use `mgr.GetClient()` instead, or maybe there is a specific reason you did it that I am missing?

enoodle: Yes, this is called before the manager starts which means the cache is not started yet and the call always fails for that reason.

romanbaron: Ok, I suspected that it might be the case.

PR #735, opened 2025-12-08
feat: default preemtibility from configmap

natasharomm:
Description:

Default "Preemptibility" per workload-type is now (optionally) taken from the configmap of default configs (what used to have only the default priorities, now has default "Preemptibility" as well).
So if "preemptibility" per pod-group is not provided using the label, will use the default from the configmap.

### DIFF
--- a/pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
+++ b/pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
@@ -198,10 +198,6 @@ func (dg *DefaultGrouper) calcPriorityClassWithDefaults(topOwner *unstructured.U
 	return defaultPriorityClassForJob, defaultConfigs
 }
 
-// calcPodGroupPreemptibilityWithDefaults - resolves preemptibility using:
-// 1) explicit labels (top owner/pod)
-// 2) provided defaults (if any)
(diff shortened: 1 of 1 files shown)


pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
```diff
@@ -157,42 +162,84 @@ func (dg *DefaultGrouper) calculateQueueName(topOwner *unstructured.Unstructured
...
+	return defaultPriorityClassForJob, defaultConfigs
 }
 
-func (dg *DefaultGrouper) calcPodGroupPreemptibility(topOwner *unstructured.Unstructured, pod *v1.Pod) v2alpha2.Preemptibility {
-	if preemptibility, found := topOwner.GetLabels()[constants.PreemptibilityLabelKey]; found {
-		if preemptibility, err := v2alpha2.ParsePreemptibility(preemptibility); err == nil {
+// calcPodGroupPreemptibilityWithDefaults - resolves preemptibility using:
+// 1) explicit labels (top owner/pod)
+// 2) provided defaults (if any)
+// 3) defaults from ConfigMap
```
enoodle: I think the implementation has 2 and 3 swapped.
Maybe the comment is not a must have in this code.

PR #737, opened 2025-12-09
Make post-delete-cleanup resources configurable

dttung2905:
## Description

This PR aims to make `post-delete-cleanup` resources ( jobs , service account, role binding ) configurable. This is particularly useful when users do not use `helm install` or `helm delete` to manage the chart lifecycle. I'm not sure whether extra tests are needed so please advise me on this

## Checklist

- [x] Self-reviewed

## Breaking Changes

No this is backward-compatible with default value to be `true`

## Additional Notes

nil

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Added
 - Added support for having the default "preemptibility" per top-owner-type read from the default configs configmap in the pod-grouper
+- Added option to profile CPU when running the snapshot tool [#726](https://github.com/NVIDIA/KAI-Scheduler/pull/726) [itsomri](https://github.com/itsomri)
 
 ### Fixed
 - Fixed a bug where the snapshot tool would not load topology objects [#720](https://github.com/NVIDIA/KAI-Scheduler/pull/720) [itsomri](https://github.com/itsomri)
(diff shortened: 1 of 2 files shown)


enoodle:
I have a question about the usecase and the solution: If someone is not using helm, then this solution of adding helm `if`s will not necessarily help. Can you explain more about the usecase and how this change will help?

dttung2905:
> I have a question about the usecase and the solution: If someone is not using helm, then this solution of adding helm `if`s will not necessarily help. Can you explain more about the usecase and how this change will help?

In my company's use case, we use a script to pull helm chart into a single template file ( with values.yaml supplied as an input ) . Then we use kustomize to apply an additional overlay on top of it.

deployments/kai-scheduler/values.yaml
```diff
@@ -120,6 +120,10 @@ crdupgrader:
     # tag: ""  # Optional: Override global.tag or Chart.AppVersion
 
 postCleanup:
+  # enabled: Set to false to disable post-delete cleanup resources.
+  # This is useful when not using Helm for installation/uninstallation (e.g., when using kustomize).
+  # When disabled, the post-delete Job, ClusterRole, ClusterRoleBinding, and ServiceAccount will not be created.
```
enoodle: Can you please remove this comment?

enoodle (changes requested):
Its OK, I just have one comment.

deployments/kai-scheduler/values.yaml
dttung2905: Thank you for the quick review. I have removed it and also added a line in the changelog.md file

PR #743, opened 2025-12-11
docs: workload segmentation design

gshaibi:
## Related Issues

Fixes #

## Checklist

- [ ] Self-reviewed


docs/developer/designs/segmented-subgroups/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Assumptions
+
+1. _Distributed workloads frameworks are responsible for ordering their pods_, and advertising their index(also known as `rank` in PyTorch) as a label on the pod and informing the process of it (in any way they wish).
+2. The PodGrouper creates a SubGroup for each PodTemplate within a workload.
+3. Segmentation should only be supported for _homogeneous SubGroups_ within a workload.
+
+## Proposal
+
+- We will leverage the pod ordering established by the workload framework to assign each pod to its corresponding segment (e.g. when segment size is 4, pods with indices 0-3 belong to segment-0, pods with indices 4-7 belong to segment-1).
```
bugra-gedik-nv: Would like to understand this better. There seems to be an assumption that the ranks are known at pod creation time. Is that really the case for all frameworks? How would KAI know the index if the rank is determined later at runtime?

The Slurm approach is reverse here. Slurm provices a SLURM_PROCID, which can be used by the framework or the launchers to find the segment.

An alternative approach is that, KAI makes available to the POD an environment variable that indicates the segment index.

bugra-gedik-nv: I think it is typically the launcher's responsibility to assign ranks. I recall, for instance, training operator does not necessarily have pod index = rank.

I think this needs to be understood better.

gshaibi: You're right that pod index ≠ rank, and ranks may be assigned dynamically at runtime by the launcher. I should clarify:

From KAI's perspective, we need a deterministic way to group pods for co-location at scheduling time. We use the pod-index label (e.g., batch.kubernetes.io/job-completion-index) for this purpose, not the runtime rank. This index is typically available at pod creation for indexed jobs and similar workloads.

The relationship between pod-index and rank is framework-specific. For frameworks where this mapping is predictable (like indexed Jobs, PyTorchJob workers), users can configure segment size knowing how their framework assigns ranks, and infer their segment index from the pod index (e.g., segment_index = global_pod_index / segment_size). For frameworks with truly dynamic rank assignment at runtime, we wouldn't be able to co-schedule based on ranks - but the segment-based co-location would still work; users just need to be aware of how their ranks map to pod indices.

Good point about making the segment index available to the pod. I've added this as a future enhancement - KAI could inject a segment index (similar to SLURM_PROCID) so the process can determine its segment without needing to calculate it.

How does that sound?

bugra-gedik-nv: OK, as long as this works for the major frameworks.

docs/developer/designs/segmented-subgroups/README.md
```diff
@@ -28,79 +28,109 @@ I want to specify segment size and topology via annotations on my workload witho
...
+        replicas: 16
+        template:
+          spec:
+            containers:
+              - name: tensorflow
+                image: tensorflow/tensorflow:latest
+  ```
+
+  This creates segments with composite global indices:
+  - **Segment 0**: Chief-0 (idx 0), PS-0 (idx 1), PS-1 (idx 2), Worker-0 (idx 3)
```
bugra-gedik-nv: Treating the workers/ps/chief as one pool does not seem to be the desired behavior.

docs/developer/designs/segmented-subgroups/README.md
```diff
@@ -28,79 +28,109 @@ I want to specify segment size and topology via annotations on my workload witho
...
-                    - python
-                    - train.py
-    ```
-
-### Validation
-
-- **Divisibility**: `MinMembers` of a SubGroup should be divided by its `SegmentSize` without remainder. This is to ensure that the SubGroup can be satisfied by the segment requirements.
+  - `kai.scheduler/pod-index-label` - (Optional) The label that denotes the pod's global index.
+
+    > **⚠️ Important**: When `pod-index-label` is specified, the user asserts that the label values are **globally unique across all pods in the workload** (no overlapping indices between replica types). If not set, the PodGrouper computes a composite global index automatically based on the workload type (see [Global Index Computation](#global-index-computation)).
```
bugra-gedik-nv: Does it have to cover all pods? Can pod-index-lable be present only on pods that desire to participate in the segmentation?

gshaibi: Yes, it would have to cover all pods for now.
We might support partial segmentation in the future if we get a feedback that it is needed.

docs/developer/designs/segmented-subgroups/README.md
gshaibi: After taking it with the team we changed it to support segmentation per replicaset.

docs/developer/designs/segmented-subgroups/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+## User Stories
+
+### Story 1: Tensor Parallelism Placement
+
+I want my 16 pods distributed training job to be split into 4 groups of 4 pods, where each group is placed on the same rack for fast NVLink/NVSwitch communication.
+
+### Story 2: Multi-Level Constraints
+
+Same as story 1, but the groups should be within the same zone.
```
davidLif:
```suggestion
Same as story 1, but all 16 pods should be within the same zone.
```

docs/developer/designs/segmented-subgroups/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+### API
+
+- Users may define segment requirements via annotations on the **PodTemplate** of each replica type:
+  - `kai.scheduler/segment-size` - Pod count for each segment
+  - `kai.scheduler/segment-topology-required-placement` - Required Topology level
+  - `kai.scheduler/segment-topology-preferred-placement` - Preferred Topology level
+  - `kai.scheduler/pod-index-label` - (Optional) The label that denotes the pod's replica index. If not set, the PodGrouper uses the default index label for the workload type (see [Workload Index Labels](#workload-index-labels)).
+
+> **Note**: The `kai.scheduler/topology` annotation can be set on either the PodTemplate annotations or on the workload root annotations. When defined on both, the PodTemplate annotation takes precedence as the more specific configuration. If `kai.scheduler/topology` is not provided, segment annotations are ignored.
```
davidLif, line 53: Can kai.scheduler/segment and "regular" topology annotation be defined for the same level? what happens in that case?

docs/developer/designs/segmented-subgroups/README.md
```diff
@@ -0,0 +1,206 @@
...
+### PodGrouper
+
+When the PodGrouper detects segment annotations on a PodTemplate, it performs the following:
+
+1. **Calculate number of segments** for that replica type:
+$$\text{NumSegments} = \lceil \frac{\text{ReplicaCount}}{\text{SegmentSize}} \rceil$$
+2. **Create segment SubGroups**: Under the replica type's SubGroup, create `NumSegments` child SubGroups. Each SubGroup has:
+   - `MinMember` = `SegmentSize` for segments covering mandatory pods (up to the replica type's MinMember)
+   - `MinMember` = `0` for segments covering excess/elastic pods (beyond the replica type's MinMember)
+   - The last segment may have fewer pods if the count is not divisible by SegmentSize
```
davidLif: "Fewer pods" - a smaller minMember?

docs/developer/designs/segmented-subgroups/README.md
```diff
@@ -0,0 +1,206 @@
...
+- **Mandatory segments**: Segments containing pods within the replica type's `MinMember` threshold have `MinMember` = `SegmentSize`. These pods are required for a successful allocation.
+- **Elastic segments**: Segments containing pods beyond the replica type's `MinMember` threshold have `MinMember` = `0`. These pods are optional and their segment SubGroups do not block allocation if they cannot be satisfied.
+
+**Example**: A replica type with `MinMember=12`, `ReplicaCount=20`, and `SegmentSize=4`:
+- Segments 0-2 (pods 0-11): `MinMember=4` each — mandatory
+- Segments 3-4 (pods 12-19): `MinMember=0` each — elastic
+
+#### Segment requirements to SubGroup tree mapping
+
+```mermaid
```
davidLif, line 145: Some of text in this image is hard to read due to the text coloring (for example, "Chief SubGroup
(1 pod)"). Can you change all the text coloring to black?

docs/developer/designs/segmented-subgroups/README.md
gshaibi: Currently, we do not support "regular" topology constraints on the PodTemplate level, so such mutuality shouldn't happen, but theoretically, they can be defined together on the same level.
In such case, the level will be segmented where each segment is constrained by the `segment-topology-required/preferred-placement`, and all the segments together are constrained by the `topology-required/preferred-placement`

docs/developer/designs/segmented-subgroups/README.md
gshaibi: Exactly. I'll be more explicit.

docs/developer/designs/segmented-subgroups/README.md
gshaibi: Sure

PR #749, opened 2025-12-15
feat(scheduler): Add a tumbling window usage type option

davidLif:
## Description

1- Add a "tumbling window" usage configuration - calculate a tumbling window size based on a start time configuration and a duration config field.
2- Convert the current "tumbling window" scheduler usage window type to "cron window".
3- For the cron window, do not use the duration as part of the window calculation - use only the cron string to calculate the window size.

## Related Issues

Fixes #

## Checklist

- [x] Self-reviewed

## Breaking Changes

The scheduling shard doesn't have a "tumblingWindowCronString" field anymore. It has bee renamed to "cronWindowDurationString". 
Also, to use a tumbling window with the cron string configuration, the "windowType" value needs to be changed to "cron" instead of "tumbling".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Fixed confusing resource division log message [#733](https://github.com/NVIDIA/KAI-Scheduler/pull/733) [itsomri](https://github.com/itsomri)
 - Made post-delete-cleanup resources configurable [#737](https://github.com/NVIDIA/KAI-Scheduler/pull/737) [dttung2905](https://github.com/dttung2905)
 - GPU Memory pods are not reclaimed or consolidated correctly
+- Added missing leases permission for the operator [#753](https://github.com/NVIDIA/KAI-Scheduler/pull/753) [dttung2905](https://github.com/dttung2905)
+- Fixed reclaim/preempt/consolidate actions for topology workloads [#739](https://github.com/NVIDIA/KAI-Scheduler/pull/739)  [itsomri](https://github.com/itsomri)
 
 ### Changed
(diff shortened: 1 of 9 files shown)


pkg/scheduler/cache/usagedb/api/interface.go
```diff
@@ -47,8 +47,10 @@ type UsageParams struct {
 	WindowSize *metav1.Duration `yaml:"windowSize,omitempty" json:"windowSize,omitempty"`
 	// Window type for time-series aggregation. If not set, defaults to sliding.
 	WindowType *WindowType `yaml:"windowType,omitempty" json:"windowType,omitempty"`
-	// A cron string used to determine when to reset resource usage for all queues.
-	TumblingWindowCronString string `yaml:"tumblingWindowCronString,omitempty" json:"tumblingWindowCronString,omitempty"`
+	// The start timestamp of the tumbling window. If not set, defaults to the current time.
+	TumblingWindowStartTime *metav1.Time `yaml:"tumblingWindowStartTime,omitempty" json:"tumblingWindowStartTime,omitempty"`
+	// The cron string defining the behavior of the cron window.
+	CronWindowDurationString string `yaml:"cronWindowDurationString,omitempty" json:"cronWindowDurationString,omitempty"`
```
itsomri:
```suggestion
	CronString string `yaml:"cronString,omitempty" json:"cronWindowDurationString,omitempty"`
```
It's not a "duration", it's simply a cron string

davidLif: But this string determines the size of the window (duration).

itsomri: It determines the size, but it's not "duration" in the sense that duration usually hints at duration type (time.Duration or something similar). The cron duration is not a "duration", it's a reset time. For example, a cron for the first day of the month will have different durations in different months. "CronString" is a known industry term that should describe this well

PR #756, opened 2025-12-21
fix: Prometheus operand

itsomri:
## Description

This PR introduces 2 changes to the prometheus operand in kai operator:
1. Prometheus storage size will not be inferred anymore, but will be set to a default of 50Gi unless explicitly set by the user in KAI config. This change was made to avoid frequent changes to the prometheus storage size.
2. Prometheus instances will not be deleted right away if prometheus changes from enabled to disabled, to avoid data loss in cases of misconfiguration or tests by the user. It will be deleted only after the retention period passed.

## Checklist

- [x] Self-reviewed

### DIFF
--- a/pkg/operator/operands/prometheus/resources.go
+++ b/pkg/operator/operands/prometheus/resources.go
@@ -183,31 +183,31 @@ func deprecatePrometheusForKAIConfig(
 		return []client.Object{prometheusObj}, nil
 	}
 
-	// Calculate the grace period (use retention period from config, default to 30 days)
-	gracePeriod := 30 * 24 * time.Hour
+	// Use retention period from config, default to 30 days
+	retentionPeriod := 30 * 24 * time.Hour
 	if kaiConfig.Spec.Prometheus != nil && kaiConfig.Spec.Prometheus.RetentionPeriod != nil {
 		duration, err := time.ParseDuration(*kaiConfig.Spec.Prometheus.RetentionPeriod)
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse retention period %q: %w", *kaiConfig.Spec.Prometheus.RetentionPeriod, err)
 		}
-		gracePeriod = duration
+		retentionPeriod = duration
 	}
 
-	// Check if grace period has passed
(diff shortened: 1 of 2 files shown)


pkg/operator/operands/prometheus/resources.go
```diff
@@ -106,10 +107,111 @@ func prometheusForKAIConfig(
...
+		logger.Info("Grace period has passed, allowing Prometheus deletion",
+			"deprecationTime", deprecationTime,
+			"gracePeriod", gracePeriod,
+			"deletionTime", deletionTime)
+		return []client.Object{}, nil
+	}
+
+	// Grace period has not passed yet, keep the instance
+	remainingTime := time.Until(deletionTime)
+	logger.Info("Prometheus instance marked for deprecation, waiting for grace period",
```
SiorMeir: nitpicking: waiting for grace period to end? not clear

pkg/operator/operands/prometheus/resources.go
```diff
@@ -106,10 +107,111 @@ func prometheusForKAIConfig(
...
+	deprecationTime, err := time.Parse(time.RFC3339, deprecationTimeStr)
+	if err != nil {
+		logger.Error(err, "Failed to parse deprecation timestamp, re-setting it", "timestamp", deprecationTimeStr)
+		annotations[deprecationTimestampKey] = metav1.Now().Format(time.RFC3339)
+		prometheusObj.SetAnnotations(annotations)
+		return []client.Object{prometheusObj}, nil
+	}
+
+	// Calculate the grace period (use retention period from config, default to 30 days)
+	gracePeriod := 30 * 24 * time.Hour
```
SiorMeir: Can this also called RetentionPeriod? naming feels a bit confusing

pkg/operator/operands/prometheus/resources.go
itsomri: fixed

pkg/operator/operands/prometheus/resources.go
itsomri: fixed

Issue #763, opened 2025-12-22
Add JobSet Support to PodGrouper


rich7420:
### What you would like to be added?

We'd like to add a PodGrouper plugin for JobSet workloads so it can automatically create PodGroups for gang scheduling. The plugin will create a PodGroup for each Job replica in a JobSet, using the naming pattern `pg-<jobset-name>-<replicatedjob-name>-<job-index>-<jobset-uid>`. It sets `MinAvailable` based on `spec.replicatedJobs[].template.spec.parallelism`. The implementation follows the same pattern as our existing plugins and comes with unit tests.

### Why is this needed?

JobSet is pretty commonly used for distributed training workloads, but right now KAI Scheduler's PodGrouper doesn't support it. This means users have to manually create PodGroups or use the default grouper, which doesn't handle JobSet's gang scheduling needs well. With this feature, we can enable automatic gang scheduling for JobSet workloads and avoid sequencing deadlocks by creating separate PodGroups per Job replica. This way users can take advantage of KAI Scheduler's gang scheduling without having to manage PodGroups manually.

cc @romanbaron

davidLif:
Hi @rich7420 ,

I think that the JobSet definition have a few features we need to consider:
1- 
`startupPolicyOrder: InOrder`. In this case, each replicatedJob executes separatly. We might want to create a separate podgroup for each  replicatedJob in this case.

2- 
If you look at https://github.com/kubernetes-sigs/jobset?tab=readme-ov-file#conceptual-diagram and read about the replicas field in a replicatedJob object, you can sse that the `MinAvailable` calculation needs to take the replicas field too as a part of the calculation.

3- 
I think we should discuss the what should the `MinAvailable` value be when in cases where `parallelism != completions`. 
For example, if `completions = parallelism + 1` and  `parallelism > 1` and 2 pods reached a "complete" state, what should be the `MinAvailable` value? This is a valid state, but the number of running pods might be lower then parallelism. This will trigger a stale gang eviction.

rich7420:
@davidLif  Thanks for the suggestions!
I'll follow the clues and have a PR as soon as possible!

davidLif:
A podgrouper plugin for JobSets have been added to the podgrouper by @rich7420 .

PR #784, opened 2025-12-25
feat(podgrouper): Add JobSet Support to PodGrouper

rich7420:
## Description

1. This PR added PodGrouper plugin support for JobSet workloads so they can automatically create PodGroups for gang scheduling.
2. The plugin creates a separate PodGroup for each replicatedJob in a JobSet, which helps avoid sequencing deadlocks when using startupPolicyOrder: InOrder.
3. It calculates MinAvailable as replicas * min(parallelism, completions) to properly handle the replicas field and edge cases where completions differs from parallelism.
4. This enables users to take advantage of KAI Scheduler's gang scheduling for JobSet workloads without having to manually manage PodGroups.
If something wrong , plz correct me.

## Related Issues

Related #763 

## Checklist

- [x] Self-reviewed

### DIFF
--- a/.github/workflows/on-pr.yaml
+++ b/.github/workflows/on-pr.yaml
@@ -5,9 +5,11 @@ name: KAI Scheduler - Pull Request
 on:
   pull_request:
     types: [opened, reopened, synchronize]
+  merge_group:
+    types: [checks_requested]
 
 concurrency:
-  group: ${{ github.head_ref || github.ref || github.run_id }}
+  group: ${{ github.event_name == 'merge_group' && github.ref || github.head_ref || github.run_id }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/validate-changelog.yaml
+++ b/.github/workflows/validate-changelog.yaml
@@ -9,18 +9,26 @@ on:
     branches:
       - main
       - 'v*.*'
+  merge_group:
+    types: [checks_requested]
 
 jobs:
   validate-changelog:
     name: Validate Changelog
     runs-on: ubuntu-latest
-    if: ${{ !contains(github.event.pull_request.labels.*.name, 'skip-changelog') }}
     steps:
+      # For merge queue: changelog was already validated, just succeed
+      - name: Skip validation for merge queue
+        if: github.event_name == 'merge_group'
+        run: echo "Changelog was already validated before entering merge queue"
+
       - uses: actions/checkout@v4
+        if: github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'skip-changelog')
         with:
           fetch-depth: 0
 
       - name: Check CHANGELOG.md updated
+        if: github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'skip-changelog')
         run: |
           if git diff --name-only origin/${{ github.base_ref }}...HEAD | grep -q '^CHANGELOG.md$'; then
             echo "CHANGELOG.md has been updated."
@@ -28,4 +36,3 @@ jobs:
             echo "::error::CHANGELOG.md has not been updated. Please update the changelog or add the 'skip-changelog' label to opt out."
             exit 1
           fi
-
--- a/.github/workflows/validate-pr-title.yaml
+++ b/.github/workflows/validate-pr-title.yaml
@@ -6,13 +6,21 @@ name: Validate PR Title
 on:
   pull_request:
     types: [opened, edited, reopened]
+  merge_group:
+    types: [checks_requested]
 
 jobs:
   validate-pr-title:
     name: Validate PR Title
     runs-on: ubuntu-latest
     steps:
+      # For merge queue: PR title was already validated, just succeed
+      - name: Skip validation for merge queue
+        if: github.event_name == 'merge_group'
+        run: echo "PR title was already validated before entering merge queue"
+
       - uses: amannn/action-semantic-pull-request@48f256284bd46cdaab1048c3721360e808335d50 # v6
+        if: github.event_name == 'pull_request'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Added rule selector for resource accounting prometheus [#818](https://github.com/NVIDIA/KAI-Scheduler/pull/818) [itsomri](https://github.com/itsomri)
 - Made accounting labels configurable [#818](https://github.com/NVIDIA/KAI-Scheduler/pull/818) [itsomri](https://github.com/itsomri)
 - Added support for Grove hierarchical topology constraints in PodGroup subgroups
+- Added support for n-level queue hierarchies [#858](https://github.com/NVIDIA/KAI-Scheduler/pull/858) [gshaibi](https://github.com/gshaibi)
 
 ### Fixed
 - Fixed pod controller logging to use request namespace/name instead of empty pod object fields when pod is not found
--- a/docs/queues/README.md
+++ b/docs/queues/README.md
@@ -2,7 +2,7 @@
 
 Scheduling queues are the core resource management primitive in KAI Scheduler, providing hierarchical resource allocation with quota guarantees and priority-based distribution.
 
-Currently only queues that have parent queues can be scheduled to and top level queues cannot be used by jobs. This is a limitation that will be fixed in the future [#621](github.com/NVIDIA/KAI-scheduler/issues/621).
+Only leaf queues (queues with no children) can be used for scheduling jobs. Parent queues serve as organizational units for resource distribution among their child queues.
 
 ## Table of Contents
 - [Queue Attributes](#queue-attributes)
--- a/docs/quickstart/README.md
+++ b/docs/quickstart/README.md
@@ -3,11 +3,11 @@
 ## Scheduling queues
 A queue is an object which represents a job queue in the cluster. Queues are an essential scheduling primitive, and can reflect different scheduling guarantees, such as resource quota and priority. 
 Queues are typically assigned to different consumers in the cluster (users, groups, or initiatives). A workload must belong to a queue in order to be scheduled.
-KAI Scheduler operates with two levels of hierarchical scheduling queue system.
+KAI Scheduler supports multi-level hierarchical scheduling queues.
 
 ### Default Queue on Fresh Install
 
-After installing KAI Scheduler, a **two-level queue hierarchy** is automatically created:
+After installing KAI Scheduler, a default queue hierarchy is automatically created:
 * `default-parent-queue` – Top-level (parent) queue. By default, this queue has no reserved resource quotas, allowing governance of resource distribution for its leaf queues.
 * `default-queue` – Leaf (child) queue under the `default-parent-queue` top-level queue. Workloads should reference this queue.
 
--- a/examples/quickstart/README.md
+++ b/examples/quickstart/README.md
@@ -6,11 +6,11 @@ This directory contains basic examples to get you started with KAI Scheduler.
 
 A queue represents a job queue in the cluster. Queues are an essential scheduling primitive and can reflect different scheduling guarantees, such as resource quota and priority. Queues are typically assigned to different consumers in the cluster (users, groups, or initiatives). A workload must belong to a queue in order to be scheduled.
 
-KAI Scheduler operates with a two-level hierarchical scheduling queue system.
+KAI Scheduler supports multi-level hierarchical scheduling queues.
 
 ### Default Queues
 
-After installing KAI Scheduler, a two-level queue hierarchy is automatically created:
+After installing KAI Scheduler, a default queue hierarchy is automatically created:
 - `default-parent-queue` – Top-level (parent) queue. By default, this queue has no reserved resource quotas, allowing governance of resource distribution for its leaf queues.
 - `default-queue` – Leaf (child) queue under the `default-parent-queue` top-level queue. Workloads should reference this queue.
 
--- a/pkg/operator/operands/common/common.go
+++ b/pkg/operator/operands/common/common.go
@@ -47,14 +47,15 @@ func AllControllersAvailable(
 	errorMessages := []string{}
 
 	for _, obj := range objects {
+		objKind := obj.GetObjectKind().GroupVersionKind().Kind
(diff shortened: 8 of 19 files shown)


rich7420:
cc @romanbaron , @davidLif

pkg/podgrouper/podgrouper/plugins/jobset/jobset_grouper.go
```diff
@@ -0,0 +1,148 @@
...
+	if !ok || replicatedJobName == "" {
+		return nil, fmt.Errorf("pod %s/%s missing required label %q", pod.Namespace, pod.Name, jobSetLabelReplicatedJobName)
+	}
+
+	pgMeta.Name = fmt.Sprintf(
+		"%s-%s-%s-%s",
+		jobSetPodGroupNamePrefix,
+		jobSetName,
+		replicatedJobName,
+		string(jobSetUID),
```
davidLif: Maybe a switch between the replicatedJobName and uid would look better? Then it's easier to understand that the uid belong to the jobset and not the replicatedJob.

davidLif:
> 2. The plugin creates a separate PodGroup for each replicatedJob in a JobSet, which helps avoid sequencing deadlocks when using startupPolicyOrder: InOrder.

If the "Inorder" parameter isn't set, it would make more sence to schedule them as a single podgroup. You might want to test for the startupPolicyOrderspec and create the podgroups accordingly.


> 3. It calculates MinAvailable as replicas * min(parallelism, completions) to properly handle the replicas field and edge cases where completions differs from parallelism.

Please add an e2e test for the case of parallelism=2, completions=3, with the first pod completing 5 seconds earlier than the other 2.

davidLif (commented):
Please add a few e2e tests under `"test/e2e/suites/integrations/third_party"`

rich7420:
@davidLif thanks for the review!
I've updated it following your suggestion.

test_jobset.yaml
```diff
@@ -0,0 +1,109 @@
+# Copyright 2025 NVIDIA CORPORATION
+# SPDX-License-Identifier: Apache-2.0
+
+# Test Scenario 1: Single ReplicatedJob with high parallelism
```
davidLif: This test example file should be in the main folder. I recommend changing it into an e2e test

pkg/podgrouper/podgrouper/plugins/jobset/jobset_grouper.go
```diff
@@ -0,0 +1,224 @@
...
+	if !ok || replicatedJobName == "" {
+		return nil, fmt.Errorf("pod %s/%s missing required label %q", pod.Namespace, pod.Name, jobSetLabelReplicatedJobName)
+	}
+
+	startupPolicyOrder, err := getStartupPolicyOrder(topOwner)
+	if err != nil {
+		return nil, err
+	}
+
+	if startupPolicyOrder == "InOrder" {
```
davidLif: Please move the "InOrder" to const.

pkg/podgrouper/podgrouper/plugins/jobset/jobset_grouper.go
```diff
@@ -0,0 +1,224 @@
...
+			jobSet.GetNamespace(), jobSet.GetName(), err)
+	}
+	if !found {
+		return "InOrder", nil
+	}
+	return order, nil
+}
+
+// getJobSetTotalMinAvailable calculates total minAvailable for all replicatedJobs.
+func getJobSetTotalMinAvailable(jobSet *unstructured.Unstructured) (int32, error) {
```
davidLif: The name of the function is too similar to "getJobSetMinAvailable". Maybe change them to something like "getReplicatedJobMinAvaliable" and "getJobSetMinAvailable"?

pkg/podgrouper/podgrouper/plugins/jobset/jobset_grouper.go
```diff
@@ -0,0 +1,224 @@
...
+		if !ok {
+			continue
+		}
+
+		name, _, _ := unstructured.NestedString(rjMap, "name")
+		if name == "" {
+			continue
+		}
+
+		minAvailable, err := getJobSetMinAvailable(jobSet, name)
```
davidLif: Considering the fact that getJobSetMinAvailable iterates over all the replicatedJobs each time is called, maybe it would be better to extract the core per replicatedJob to a separate function and call it from getJobSetMinAvailable and getJobSetTotalMinAvailable

rich7420:
@davidLif thanks for the review and suggestions!

pkg/podgrouper/podgrouper/plugins/jobset/jobset_grouper.go
```diff
@@ -0,0 +1,230 @@
...
+
+		minAvailable, err := calculateReplicatedJobMinAvailable(rjMap, jobSet.GetNamespace(), jobSet.GetName(), name)
+		if err != nil {
+			return 0, err
+		}
+		totalMinAvailable += int64(minAvailable)
+	}
+
+	if totalMinAvailable <= 0 {
+		return 1, nil
```
davidLif, line 224: Why return 1 is the sum is 0? Maybe for some reason the user decided to "pause" the job and set the number of replicas to 0?

test/e2e/suites/integrations/third_party/jobset/jobset_test.go
```diff
@@ -0,0 +1,558 @@
...
+	}
+)
+
+var _ = Describe("JobSet integration", Ordered, func() {
+	var (
+		testCtx *testcontext.TestContext
+	)
+
+	BeforeAll(func(ctx context.Context) {
+		testCtx = testcontext.GetConnectivity(ctx, Gomega)
```
davidLif: Gomega cann't be passed as a parameter. Please look at other GetConnectivity calls.

test/e2e/suites/integrations/third_party/jobset/jobset_test.go
```diff
@@ -0,0 +1,558 @@
...
+		parentQueue := queue.CreateQueueObject(utils.GenerateRandomK8sName(10), "")
+		childQueue := queue.CreateQueueObject(utils.GenerateRandomK8sName(10), parentQueue.Name)
+		testCtx.InitQueues([]*v2.Queue{childQueue, parentQueue})
+
+		skipIfJobSetNotInstalled(ctx, testCtx)
```
davidLif: Please check if "JobSetNotInstalled" before creating the queues.

test/e2e/suites/integrations/third_party/jobset/jobset_test.go
```diff
@@ -0,0 +1,558 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+// createJobSetWithDefaultParallelism creates a JobSet with single replica and default parallelism.
+func createJobSetWithDefaultParallelism(name, namespace, queueName string) *unstructured.Unstructured {
```
davidLif: You can use a Unstructured like that, but importing the relevant struct would have also been fine.

test/e2e/suites/integrations/third_party/jobset/jobset_test.go
```diff
@@ -0,0 +1,558 @@
...
+
+			// Wait for pods to be created
+			pods := waitForJobSetPods(ctx, testCtx, jobSetName, testNamespace, 3)
+
+			// Wait for all pods to be scheduled
+			wait.ForPodsScheduled(ctx, testCtx.ControllerClient, testNamespace, pods)
+
+			// Wait for first pod to complete (should complete 5 seconds earlier)
+			firstPod := pods[0]
+			Eventually(func() v1.PodPhase {
```
davidLif: Please change `Eventually`"calls to the appropriate `wait` calls where possible. For example, here you can use `wait.ForPodSucceededOrError` .

davidLif: See test/e2e/modules/wait

test/e2e/suites/integrations/third_party/jobset/jobset_test.go
```diff
@@ -0,0 +1,558 @@
...
+			jobSetName := "inorder-jobset-" + utils.GenerateRandomK8sName(10)
+
+			jobSet := createJobSetWithStartupPolicy(jobSetName, testNamespace, testCtx.Queues[0].Name, "InOrder")
+			Expect(testCtx.ControllerClient.Create(ctx, jobSet)).To(Succeed())
+			defer func() {
+				Expect(testCtx.ControllerClient.Delete(ctx, jobSet)).To(Succeed())
+			}()
+
+			// Wait for pods and verify PodGroups are created separately
+			pods := waitForJobSetPods(ctx, testCtx, jobSetName, testNamespace, 2)
```
davidLif: How does this call validates that 2 separate podgroups have been created?

test/e2e/suites/integrations/third_party/jobset/jobset_test.go
```diff
@@ -0,0 +1,558 @@
...
+			jobSetName := "anyorder-jobset-" + utils.GenerateRandomK8sName(10)
+
+			jobSet := createJobSetWithStartupPolicy(jobSetName, testNamespace, testCtx.Queues[0].Name, "AnyOrder")
+			Expect(testCtx.ControllerClient.Create(ctx, jobSet)).To(Succeed())
+			defer func() {
+				Expect(testCtx.ControllerClient.Delete(ctx, jobSet)).To(Succeed())
+			}()
+
+			// Wait for pods and verify they all use the same PodGroup
+			pods := waitForJobSetPods(ctx, testCtx, jobSetName, testNamespace, 2)
```
davidLif: How does this call validates that only one podgroup have been created?

test/e2e/suites/integrations/third_party/jobset/jobset_test.go
```diff
@@ -0,0 +1,558 @@
...
+	AfterEach(func(ctx context.Context) {
+		testCtx.TestContextCleanup(ctx)
+	})
+
+	AfterAll(func(ctx context.Context) {
+		testCtx.ClusterCleanup(ctx)
+	})
+
+	Context("JobSet submission", func() {
+		It("should schedule pods with parallelism=2 and completions=3, first pod completes 5 seconds earlier", func(ctx context.Context) {
```
davidLif: This is a good test, please set default-staleness-grace-period to 1 second at the begining of the test and remove this config at the end. Please look at `test/e2e/modules/configurations/feature_flags` before implementing.

test/e2e/suites/integrations/third_party/jobset/jobset_test.go
```diff
@@ -0,0 +1,558 @@
...
+	crdList, err := testCtx.KubeClientset.ApiextensionsV1().CustomResourceDefinitions().List(ctx, metav1.ListOptions{
+		FieldSelector: "metadata.name=jobsets.jobset.x-k8s.io",
+	})
+	Expect(err).NotTo(HaveOccurred())
+	if len(crdList.Items) == 0 {
+		Skip("JobSet CRD not installed, skipping JobSet integration tests")
+	}
+}
+
+func createJobSet(name, namespace, queueName string, parallelism, completions int32) *unstructured.Unstructured {
```
davidLif: Please import the relevant struct from the JobSet repo instead of manipulating Unstructured for all JobSet creations.

test/e2e/suites/integrations/third_party/jobset/jobset_test.go
```diff
@@ -0,0 +1,558 @@
...
+				g.Expect(len(podGroups.Items)).To(Equal(1))
+				// MinMember should default to 1 (replicas=1, parallelism defaults to 1)
+				g.Expect(podGroups.Items[0].Spec.MinMember).To(Equal(int32(1)))
+			}, time.Minute).Should(Succeed())
+		})
+	})
+})
+
+func skipIfJobSetNotInstalled(ctx context.Context, testCtx *testcontext.TestContext) {
+	crdList, err := testCtx.KubeClientset.ApiextensionsV1().CustomResourceDefinitions().List(ctx, metav1.ListOptions{
```
davidLif: Please use `crd.SkipIfCrdIsNotInstalled` instead. See `test/e2e/suites/integrations/third_party/kubeflow.org/pytorch/pytorch_test.go` for example

davidLif (commented):
1- Please add a `deploy_jobset.sh` script under `hack/third_party_integrations` to enable installation for `run-e2e-kind.sh` (and add a call to this script appropriatly). 

2- After adding deploy_jobset.sh`, please validate that all Jobset e2e tests pass on a kinf cluster deployed by `run-e2e-kind.sh --test-third-party-integrations` (You can use --preserve-cluster if you need to debug some of the tests)

@rich7420  Thank you very much for you patience and effort!

pkg/podgrouper/podgrouper/plugins/jobset/jobset_grouper.go
rich7420: I set it to 1 because the PodGroup CRD validation requires minMember to be at least 1. However, you are right regarding the paused job scenario (replicas=0). Forcing it to 1 might cause the PodGroup to hang waiting for a pod that will never exist.
Perhaps we should skip creating the PodGroup entirely if totalMinAvailable calculates to 0? WDYT?

davidLif: The the PodGroup CRD validation requires minMember > 1, then I think the code you wrote (minMember 1) is the best option.
I wouldn't try to "skip creating the PodGroup". 
It would create a mess for cases like this:  if the podgroup existed, the use changed the JobSet, the podgroup would be deleted, the use changed the JobSet again, and now we create a new podgroup for the same entity again.

rich7420:
@davidLif thanks for the patient reviewing! I've updated.

hack/third_party_integrations/deploy_jobset.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+# Copyright 2025 NVIDIA CORPORATION
+# SPDX-License-Identifier: Apache-2.0
+set -e
+
+kubectl apply --server-side -k "github.com/kubernetes-sigs/jobset.git/config/default?ref=v0.2.0"
```
davidLif: When i tried to run it on my machine, the image pull failed. 
I recommend using `kubectl apply --server-side -f https://github.com/kubernetes-sigs/jobset/releases/download/v0.10.1/manifests.yaml` instead, as described by https://github.com/kubernetes-sigs/jobset?tab=readme-ov-file#installation .

rich7420:
@davidLif thanks! I've updated.

davidLif (approved):

rich7420:
thanks @davidLif , @gshaibi for the review and patience !
What’s the next step I can help with?

PR #817, opened 2025-12-29
fix: ToDo consider handling gracefully in storage.go

rich7420:
## Description
This PR changed error handling in snapshotStorageCapacities() to gracefully skip invalid CSIStorageCapacity objects with a warning log instead of failing the entire snapshot, allowing the scheduler to continue with valid capacities.
If i'm wrong plz correct me.

## Checklist

- [x] Self-reviewed

### DIFF
--- a/pkg/scheduler/cache/cluster_info/storage.go
+++ b/pkg/scheduler/cache/cluster_info/storage.go
@@ -106,8 +106,11 @@ func (c *ClusterInfo) snapshotStorageCapacities() (map[common_info.StorageCapaci
 	for _, capacity := range capacities {
 		capacityInfo, err := storagecapacity_info.NewStorageCapacityInfo(capacity)
 		if err != nil {
-			// ToDo: consider handling gracefully
-			return nil, err
+			// Skip invalid capacity and continue processing others
+			log.InfraLogger.V(2).Warnf(
+				"Failed to process CSIStorageCapacity %s/%s, skipping: %v",
+				capacity.Namespace, capacity.Name, err)
+			continue
 		}
 		result[capacityInfo.UID] = capacityInfo
 	}


pkg/scheduler/cache/cluster_info/storage.go
```diff
@@ -106,8 +106,11 @@ func (c *ClusterInfo) snapshotStorageCapacities() (map[common_info.StorageCapaci
 	for _, capacity := range capacities {
 		capacityInfo, err := storagecapacity_info.NewStorageCapacityInfo(capacity)
 		if err != nil {
-			// ToDo: consider handling gracefully
-			return nil, err
+			// Skip invalid capacity and continue processing others
+			log.InfraLogger.V(2).Warnf(
+				"Failed to process CSIStorageCapacity %s/%s, skipping: %v",
+				capacity.Namespace, capacity.Name, err)
+			continue
```
itsomri, line 113: We need to check what happens if this is indeed the case. What if there's a job that requests storage from this capacity? Will the scheduling error be clear enough for the user?

rich7420: If we skip it, the scheduler won't see this capacity. Any job requesting it will remain Pending due to insufficient resources/unschedulable nodes.
The user will see a standard scheduling error, but the Admin will see the specific validation error in the logs (via the Warnf).
In my opinion, keeping the job Pending is better than failing the entire snapshotStorageCapacities function, which could disrupt scheduling for all other valid jobs as well. WDYT?

itsomri: Okay, approving

itsomri: Please rebase and I'll re-run the CI

rich7420: OK, thanks!

PR #830, opened 2026-01-04
feat: record dra enabled resources in snapshot

SiorMeir:
## Description

Part of DRA support in KAI-Scheduler, we add recording of DRA enabled GPU resources in snapshot taking.

ResourceSlices as well as ResourceClaims will now be recorded & analyzed with `snapshot-tool`
## Related Issues

Fixes #

## Checklist

- [X] Self-reviewed
- [X] Added/updated tests (if needed)
- [X] Updated documentation (if needed)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,21 +18,43 @@ make vet-go                   # Run go vet
 ```
 
 ### Testing
+#### unit and integration tests
+
+- Test files MUST ALWAYS be in the same directory as code
+- Test files names MUST ALWAYS end in `_test.go`. Example: `resolver_test.go`
+
 ```bash
 make test                     # Run all tests (unit + helm chart tests)
 
 # Run a single test file
-go test -v ./pkg/scheduler/actions/allocate/...
+ginkgo -v ./pkg/scheduler/actions/allocate
 
 # Run a specific test function
-go test -v ./pkg/scheduler/actions/allocate/... -run TestHandleAllocation
+ginkgo -v --focus "TestHandleAllocation" ./pkg/scheduler/actions/allocate
 
 # Run tests with Ginkgo (for integration tests)
-go test -v ./pkg/binder/controllers/integration_tests/... -ginkgo.focus="test name pattern"
+ginkgo -v --focus "test name pattern" ./pkg/binder/controllers/integration_tests
 
 # Run tests with envtest (requires setup-envtest)
 make envtest
 KUBEBUILDER_ASSETS="$(bin/setup-envtest use 1.34.0 -p path --bin-dir bin)" go test ./pkg/... -timeout 30m
+```
+#### E2E tests
+
+E2E tests run against a real Kubernetes using [`kind`](https://kind.sigs.k8s.io/) cluster and are located in `test/e2e/suites/`.
+
+```bash
+# Run locally with Kind (recommended for development)
+./hack/run-e2e-kind.sh                          # Full e2e suite
+./hack/run-e2e-kind.sh --preserve-cluster       # Keep cluster after tests
+./hack/run-e2e-kind.sh --local-images-build     # Build images locally
+
+# Run specific test suites (requires cluster with KAI installed)
+ginkgo -r --randomize-all ./test/e2e/suites/allocate
+ginkgo -r --randomize-all --focus "quota" ./test/e2e/suites
+
+# Run with verbose output and trace
+ginkgo -r --randomize-all --trace -vv ./test/e2e/suites/preempt
 ```
 
 ### Code Generation
@@ -41,7 +63,7 @@ make generate                 # Generate DeepCopy methods
 make manifests                # Generate CRDs and RBAC
 make clients                  # Generate client code
 make generate-mocks           # Generate mock implementations
-make validate                 # Verify generated code is up to date
+make validate                 # Verify generated code is up to date. Also format and vet codebase
 ```
 
 ## Repository Structure
@@ -141,4 +163,16 @@ PRs trigger: `make validate` → `make test` → `make build` → E2E tests
 
 ## General Rules
 - Use `git mv` when moving files to preserve history
-- Don't add obvious comments that duplicate the code
+- DO NOT add obvious comments that duplicate the code.
+    - Write comments ONLY when ABSOLUTELY nessessary
+    - Keep Comments short, explicit and concise
+    - DO NOT use `I`, `we` or any other pronoun
+- When performing a major change, run `make validate` after changes
+- 
+## Philosophy & Design
+
+- Documentation can be found in [`docs`](docs/) folder
+
+- Design documents for major features are in [`docs/developer/designs/`](docs/developer/designs/):
+
+- Usage Examples are in [`examples`](/examples/)


pkg/scheduler/api/node_info/node_info.go
```diff
@@ -600,7 +606,8 @@ func (ni *NodeInfo) GetNumberOfGPUsInNode() int64 {
 	numberOfGPUs, err := ni.getNodeGpuCountLabelValue()
 	if err != nil {
 		log.InfraLogger.V(6).Infof("Node: <%v> had no annotations of nvidia.com/gpu.count", ni.Name)
-		return int64(ni.Allocatable.GPUs())
+		// Check both extended resources and DRA GPUs
```
enoodle: NIT: I don't think that we need this comment everywhere - lets just assume that after this PR it should check for both, everywhere. :)

pkg/scheduler/api/node_info/node_info.go
```diff
@@ -79,6 +79,10 @@ type NodeInfo struct {
 
 	Allocatable *resource_info.Resource
 
+	// DRAGPUsByResource tracks GPU capacity from DRA ResourceSlices per resource type.
+	// Key is the resource name (e.g., "nvidia.com/gpu", "amd.com/gpu").
+	DRAGPUsByResource map[v1.ResourceName]float64
```
enoodle: Isn't it a bit over complicated? Is it a real use case to have different GPU types on the same node?

pkg/scheduler/cache/cluster_info/cluster_info.go
```diff
@@ -240,9 +241,51 @@ func (c *ClusterInfo) snapshotNodes(
 		}
 	}
 
+	// Populate DRA GPU counts from ResourceSlices
```
enoodle: please remove comment, function name is clear.

pkg/common/resources/dra_node_resources.go
```diff
@@ -0,0 +1,72 @@
...
+	return totalGPUs
+}
+
+func isSliceForNode(slice *resourceapi.ResourceSlice, nodeName string) bool {
+	if slice.Spec.NodeName != nil && *slice.Spec.NodeName != "" {
+		return *slice.Spec.NodeName == nodeName
+	}
+	if slice.Spec.AllNodes != nil && *slice.Spec.AllNodes {
+		return true
+	}
```
enoodle: Can it be a GPU device if it is for all nodes?

pkg/common/resources/dra_node_resources.go
```diff
@@ -0,0 +1,72 @@
...
+	for _, slice := range slices {
+		if slice == nil {
+			continue
+		}
+
+		if !isSliceForNode(slice, nodeName) {
+			continue
+		}
+
+		if !isGPUDeviceClass(slice.Spec.Driver) {
```
enoodle: If we are here, can you also fix this function to only check for lower case GPU in the device class name? checking also for `nvidia.com/gpu` is redundant.

pkg/scheduler/cache/cluster_info/cluster_info.go
```diff
@@ -240,9 +241,51 @@ func (c *ClusterInfo) snapshotNodes(
...
+		log.InfraLogger.V(6).Infof("Failed to list ResourceSlices for DRA GPU counting: %v", err)
+		return
+	}
+
+	if len(resourceSlices) == 0 {
+		return
+	}
+
+	for nodeName, nodeInfo := range nodes {
+		draGPUsByClass := resources.CountNodeGPUsFromResourceSlicesByDeviceClass(nodeName, resourceSlices)
```
enoodle: Instead of going over all the nodes and for each node going over all the resources slices (nodes * slices) it is better to go once over the resource slices and map them to nodes and then for each node to read its list (nodes + slices )

pkg/scheduler/cache/cluster_info/cluster_info.go
```diff
@@ -240,9 +241,51 @@ func (c *ClusterInfo) snapshotNodes(
...
+	// Populate DRA GPU counts from ResourceSlices
+	c.populateDRAGPUs(resultNodes)
+
 	return resultNodes, minGPUMemory, nil
 }
 
+// For each resource type (e.g., "nvidia.com/gpu", "amd.com/gpu"), checks if extended resources exist.
+// If not, uses DRA counts from ResourceSlices.
+func (c *ClusterInfo) populateDRAGPUs(nodes map[string]*node_info.NodeInfo) {
+	resourceSlices, err := c.dataLister.ListResourceSlices()
```
enoodle: Maybe this interface can be changed to return slices per node and it will simplify this function.

pkg/scheduler/api/node_info/node_info.go
```diff
@@ -731,3 +748,56 @@ func (ni *NodeInfo) lessEqualTaskToNodeResources(
...
+		if draCount <= 0 {
+			continue
+		}
+
+		extendedCount := ni.Allocatable.Get(resourceName)
+
+		if extendedCount > 0 {
+			// Extended resources take precedence - don't set DRA for this resource
+			continue
+		}
```
enoodle: does this use case exist? for example with nvidia.com/gpu it is not possible to have both the device plugin and the dra plugin on the same node.

pkg/scheduler/api/node_info/node_info.go
SiorMeir: agreed and fixed

pkg/common/resources/dra_node_resources.go
SiorMeir: interesting, sounds good

pkg/scheduler/api/node_info/node_info.go
SiorMeir: This sounds weird but I think that it is possible.
We can remove & explicitly state that we track all gpu resources per node without distinction

pkg/common/resources/dra_node_resources.go
SiorMeir: no. AFAIK a physical GPU cannot be shared across nodes.
Removing

pkg/scheduler/cache/cluster_info/cluster_info.go
SiorMeir: acceppted

pkg/scheduler/api/node_info/node_info.go
SiorMeir: I don't see a place where we prevent that from our end, so I think this might serve as a soft guard.

pkg/scheduler/api/node_info/node_info.go
enoodle: @SiorMeir This PR is not using the different names, I don't see the use case for this super unlikely event. It can only happen on a custom machine that someone wants to mess with.
Lets simplify the code by just making it "DRAGPUs"

pkg/scheduler/cache/cluster_info/cluster_info.go
```diff
@@ -240,9 +241,69 @@ func (c *ClusterInfo) snapshotNodes(
...
+			draGPUsByClass[slice.Spec.Driver] += int64(len(slice.Spec.Devices))
+		}
+
+		// Add GPUs from all-nodes slices
+		for _, slice := range allNodesSlices {
+			if !resources.IsGPUDeviceClass(slice.Spec.Driver) {
+				continue
+			}
+			draGPUsByClass[slice.Spec.Driver] += int64(len(slice.Spec.Devices))
+		}
```
enoodle: Please remove this - currently it doesn't make sense to count that. We better not have non functioning code.

pkg/scheduler/api/node_info/node_info.go
enoodle: For example with the nvidia drivers it is written in the documentation. Also the DRA driver name is different than the extended resource name (`nvidia.com/gpu` vs `gpu.nvidia.com`) so I don't see the effectiveness of this.

pkg/common/resources/dra_node_resources.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		totalGPUs += countDevicesInSlice(slice)
+	}
+
+	return totalGPUs
+}
+
+func isSliceForNode(slice *resourceapi.ResourceSlice, nodeName string) bool {
+	if slice.Spec.NodeName != nil && *slice.Spec.NodeName != "" {
```
davidLif: Why do you need to check for `*slice.Spec.NodeName != "" `? It will fail on `*slice.Spec.NodeName == nodeName`

pkg/common/resources/dra_node_resources.go
```diff
@@ -0,0 +1,100 @@
...
+
+func countDevicesInSlice(slice *resourceapi.ResourceSlice) int64 {
+	if slice.Spec.Devices == nil {
+		return 0
+	}
+	return int64(len(slice.Spec.Devices))
+}
+
+// NodeGPUsByDeviceClass maps node name to device class to GPU count
+type NodeGPUsByDeviceClass map[string]map[string]int64
```
davidLif: I think that it is better to set the definitions at the start of the file unless there is a good reason to do otherwise. What do you think?

pkg/common/resources/dra_node_resources.go
```diff
@@ -0,0 +1,100 @@
...
+			result[nodeName][slice.Spec.Driver] += deviceCount
+		}
+	}
+
+	return result
+}
+
+// getNodeNamesForSlice returns the node names this slice applies to.
+// For AllNodes slices, returns empty slice (caller should handle specially).
+// For node-specific slices, returns a single-element slice.
```
davidLif: ResourceSlice also supports a NodeSelector for the case of multiple nodes. What do you think this function should return in this case?

pkg/scheduler/api/node_info/node_info.go
SiorMeir: removing

pkg/common/resources/dra_node_resources.go
```diff
@@ -0,0 +1,100 @@
...
+func getNodeNamesForSlice(slice *resourceapi.ResourceSlice) []string {
+	if slice.Spec.NodeName != nil && *slice.Spec.NodeName != "" {
+		return []string{*slice.Spec.NodeName}
+	}
+	// AllNodes slices are not returned here - they need special handling
+	return nil
+}
+
+// GetAllNodesSlices returns slices that apply to all nodes
```
davidLif: That's not exactly what this function is doing - it returns all sloces with AllNodes AND a gpu device class

pkg/scheduler/api/node_info/node_info_test.go
```diff
@@ -51,6 +51,12 @@ const (
 func nodeInfoEqual(l, r *NodeInfo) bool {
 	l.PodAffinityInfo = nil
 	r.PodAffinityInfo = nil
+	if l.DRAGPUsByResource == nil {
+		l.DRAGPUsByResource = make(map[v1.ResourceName]float64)
+	}
+	if r.DRAGPUsByResource == nil {
+		r.DRAGPUsByResource = make(map[v1.ResourceName]float64)
+	}
```
davidLif: Why twice?

pkg/common/resources/dra_node_resources.go
SiorMeir: this is pretty redundant
fixing

pkg/common/resources/dra_node_resources.go
SiorMeir: agreed, moving

pkg/common/resources/dra_node_resources.go
SiorMeir: We've removed allNodes use case so I'll remove this altogether

pkg/common/resources/dra_node_resources.go
SiorMeir: I think looping over and mapping from NodeSelector is the solution here.
We'll check that first

pkg/scheduler/api/node_info/node_info_test.go
SiorMeir: AI hiccup

PR #847, opened 2026-01-06
fix: Fraction bind rollback

itsomri:
## Description

In cases of bind failures *after* gpu reservation, the binder doesn't always remove the gpu group label from the pod, which leaves dangling reservation pods in the cluster. 
This PR:
1. Moves the cleanup logic to happen for all (fractional) failed binding attempts
2. Cleans up some logging and error handling in the binder - all info is still reported to logs, but some duplications were removed
3. Adds an e2e test that verifies there are no fractional over-allocations on GPU devices

## Checklist

- [x] Self-reviewed

### DIFF
--- a/pkg/binder/binding/binder.go
+++ b/pkg/binder/binding/binder.go
@@ -91,12 +91,10 @@ func (b *Binder) Rollback(ctx context.Context, pod *v1.Pod, node *v1.Node, bindR
 
 	var rollbackErrs []error
 
-	// Rollback plugins (K8sPlugins, GPUSharing, etc.)
 	if err := b.plugins.Rollback(ctx, pod, node, bindRequest, nil); err != nil {
 		rollbackErrs = append(rollbackErrs, fmt.Errorf("failed to rollback plugins for pod <%s/%s>: %w", pod.Namespace, pod.Name, err))
 	}
 
-	// GPU reservation cleanup only applies to shared GPU allocations
 	if common.IsSharedGPUAllocation(bindRequest) {
 		if err := b.resourceReservationService.RemovePodGpuGroupsConnection(ctx, pod); err != nil {
 			rollbackErrs = append(rollbackErrs, fmt.Errorf("failed to remove GPU group label from pod <%s/%s> during rollback: %w", pod.Namespace, pod.Name, err))


pkg/binder/plugins/gpusharing/gpu_sharing.go
```diff
@@ -101,3 +102,58 @@ func (p *GPUSharing) PostBind(
...
+	// Delete capabilities ConfigMap
+	capabilitiesConfigMapName, err := gpusharingconfigmap.ExtractCapabilitiesConfigMapName(pod, containerRef)
+	if err != nil {
+		// Annotation not set means PreBind never created the configmap
+		logger.V(1).Info("Rollback: could not extract capabilities configmap name, nothing to rollback",
+			"namespace", pod.Namespace, "name", pod.Name, "error", err)
+		return nil
+	}
+
+	if err := p.deleteConfigMap(ctx, pod.Namespace, capabilitiesConfigMapName); err != nil {
```
gshaibi:
```suggestion
	if err = p.deleteConfigMap(ctx, pod.Namespace, capabilitiesConfigMapName); err != nil {
```

pkg/binder/plugins/gpusharing/gpu_sharing.go
```diff
@@ -101,3 +102,58 @@ func (p *GPUSharing) PostBind(
...
+		return fmt.Errorf("failed to delete capabilities configmap %s/%s during rollback: %w",
```
gshaibi: What do you think about collecting this error and continue with the rollout attempt on this plugin?

pkg/binder/binding/binder.go
```diff
@@ -82,57 +76,59 @@ func (b *Binder) Bind(ctx context.Context, pod *v1.Pod, node *v1.Node, bindReque
...
+
+	var rollbackErrs []error
+
+	// Rollback plugins (K8sPlugins, GPUSharing, etc.)
+	if err := b.plugins.Rollback(ctx, pod, node, bindRequest, nil); err != nil {
+		rollbackErrs = append(rollbackErrs, fmt.Errorf("failed to rollback plugins for pod <%s/%s>: %w", pod.Namespace, pod.Name, err))
+	}
+
+	// GPU reservation cleanup only applies to shared GPU allocations
+	if common.IsSharedGPUAllocation(bindRequest) {
```
gshaibi, line 98: Out of curiosity, do you know why this is not part of a plugin? (gpu sharing plugin maybe?)

pkg/binder/binding/resourcereservation/resource_reservation.go
```diff
@@ -260,24 +262,11 @@ func (rsc *service) updatePodGPUGroup(
...
-	key := constants.GPUGroup
-	if isMultiFractionalPod {
-		multiGpuGroupLabelKey, _ := resources.GetMultiFractionGpuGroupLabel(gpuGroup)
-		key = strings.Replace(multiGpuGroupLabelKey, "/", "~1", -1)
-	}
-
-	// Create a JSON patch to remove the label
-	patch = []map[string]string{
+func (rsc *service) RemovePodGpuGroupsConnection(ctx context.Context, pod *v1.Pod) error {
+	patch := []map[string]string{
```
gshaibi: What about multi fractional pods?

pkg/binder/binding/binder.go
```diff
@@ -82,57 +76,59 @@ func (b *Binder) Bind(ctx context.Context, pod *v1.Pod, node *v1.Node, bindReque
...
+		if err := b.resourceReservationService.RemovePodGpuGroupsConnection(ctx, pod); err != nil {
+			rollbackErrs = append(rollbackErrs, fmt.Errorf("failed to remove GPU group label from pod <%s/%s> during rollback: %w", pod.Namespace, pod.Name, err))
+		}
+
+		if err := b.resourceReservationService.SyncForNode(ctx, bindRequest.Spec.SelectedNode); err != nil {
```
gshaibi, line 103: Isn't the sync on the `Bind` enough?

gshaibi (commented):
Looks very good

pkg/binder/binding/binder.go
itsomri: No, I think it should be, but it's out of scope for this

pkg/binder/binding/resourcereservation/resource_reservation.go
itsomri: The handling for multi fractional is essentially the same, since we need to atomically delete all references anyway. So I just remove the annotation

pkg/binder/binding/binder.go
itsomri: Which sync on bind?
We need to sync after removing the annotations, otherwise the reservation will remain dangling

pkg/binder/binding/resourcereservation/resource_reservation.go
gshaibi: But the label has a different key, no?

pkg/binder/binding/binder.go
gshaibi: I meant [this sync](https://github.com/NVIDIA/KAI-Scheduler/blob/e9decd1ed32edebd60a4c7d1d8bd7359c654ba1c/pkg/binder/binding/binder.go#L44).
It will remain dangling but will be removed upon the next bind request on that node, no?

pkg/binder/binding/resourcereservation/resource_reservation.go
itsomri: Hmm you're right, I missed that. Fixed + added tests

pkg/binder/binding/binder.go
itsomri: I prefer to clean up the reservation as soon as we can in this case

PR #820, opened 2025-12-30
feat: add clear error message for missing queue

gshaibi:
## Description

When a user misconfigures queues by specifying a non-existent queue or
a queue whose parent doesn't exist, jobs are now reported with a clear
error message instead of being silently skipped.

The error appears as:
- An event on the PodGroup
- A condition on the PodGroup with reason 'QueueDoesNotExist'
- An event on the pending pods

This helps users understand why their jobs aren't being scheduled.
## Related Issues

Fixes #

## Checklist

> **Note:** Ensure your PR title follows the [Conventional Commits format](https://github.com/NVIDIA/KAI-Scheduler/blob/main/CONTRIBUTING.md#pr-title-guidelines) (e.g., `feat(scheduler): add new feature`)

- [x] Self-reviewed
- [x] Added/updated tests (if needed)
- [x] Updated documentation (if needed)

## Additional Notes
image

```yaml
  status:
    conditions:
    - lastProbeTime: null
      lastTransitionTime: "2025-12-30T16:19:59Z"
      message: 'QueueDoesNotExist: Queue ''non-exist'' does not exist.'
      reason: Unschedulable
      status: "False"
      type: PodScheduled
    phase: Pending
    qosClass: BestEffort
```

image

```yaml
  status:
    conditions:
    - lastProbeTime: null
      lastTransitionTime: "2025-12-30T19:28:32Z"
      message: 'QueueDoesNotExist: Queue ''default'' has no parent queue.'
      reason: Unschedulable
      status: "False"
      type: PodScheduled
    phase: Pending
    qosClass: BestEffort
```

PR #858, opened 2026-01-08
feat: support n-level queue hierarchy

### LINKED ISSUE #621: Cannot use top-level queues directly
Hi, I am playing around with queues and have noticed my workloads that reference top level queues are not able to be scheduled. 

I see the following events in the pod:
```
Events:
  Type     Reason             Age                    From           Message
  ----     ------             ----                   ----           -------
  Warning  PodGrouperWarning  12m                    pod-grouper    Pod "qa-build-preemptible" not found
  Warning  Unschedulable      2m11s (x599 over 12m)  kai-scheduler  Unable to schedule pod
```

I have checked all the logs. The only concerning thing I see is an error in the pod-grouper pod saying "failed to find top owner" and "pod ... not found", like in the events above. The pod definitely exists (these are the events from the pod itself, after all), so I am not sure why it is not found. 

The quickstart guide tells us to reference the child queue of the default-parent-queue in our workloads, so is this a known limitation?

If so, why is that the case? I don't see any documentation about it, maybe I am missing something.
Also, a more informative pod event would be very helpful here.

Thanks!

gshaibi:
## Related Issues

Fixes #621

## Checklist

- [x] Self-reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Interpret negative or zero half-life value as disabled [#818](https://github.com/NVIDIA/KAI-Scheduler/pull/818) [itsomri](https://github.com/itsomri)
 - Handle invalid CSI StorageCapacities gracefully [#817](https://github.com/NVIDIA/KAI-Scheduler/pull/817) [rich7420](https://github.com/rich7420)
 - Embed CRD definitions in binary for env-test and time-aware-simulations to allow binary portability [#818](https://github.com/NVIDIA/KAI-Scheduler/pull/818) [itsomri](https://github.com/itsomri)
+- Fixed missing `podGrouper` configuration in Helm template that prevented podgrouper values from being applied [#860](https://github.com/NVIDIA/KAI-Scheduler/pull/860)
 
 ### Changed
 - Removed the constraint that prohibited direct nesting of subgroups alongside podsets within the same subgroupset.
--- a/deployments/kai-scheduler/templates/kai-config.yaml
+++ b/deployments/kai-scheduler/templates/kai-config.yaml
@@ -68,6 +68,23 @@ spec:
         {{- toYaml .Values.binder.resourceReservationPodResources | nindent 8 }}
       {{- end }}
 
+  podGrouper:
+    service:
+      enabled: true
+      image:
+        name: {{ .Values.podgrouper.image.name }}
+        repository: {{ .Values.global.registry }}
+        tag: {{ .Values.podgrouper.image.tag | default .Values.global.tag | default .Chart.AppVersion }}
+        pullPolicy: {{ .Values.podgrouper.image.pullPolicy | default .Values.global.imagePullPolicy }}
+      {{- if .Values.podgrouper.resources }}
+      resources:
+        {{- toYaml .Values.podgrouper.resources | nindent 8 }}
+      {{- end }}
+      {{- if .Values.podgrouper.affinity }}
+      affinity:
+        {{- toYaml .Values.podgrouper.affinity | nindent 8 }}
+      {{- end }}
+
   podGroupController:
     service:
       enabled: true


pkg/scheduler/actions/utils/job_order_by_queue.go
```diff
@@ -12,329 +12,410 @@ import (
...
+		node.needsReorder = false
+		return jo.getNextNode(pq)
 	}
-	if jobsOrder.departmentIdToDepartmentMetadata[department.UID].queuesPriorityQueue.Empty() {
-		log.InfraLogger.V(7).Warnf("Department: <%v> is active, yet no queues in department", department.Name)
+
+	if node.children.Empty() {
+		log.InfraLogger.V(7).Warnf("Queue node <%v> is active but has no children", node.queue.Name)
 		return nil
 	}
```
enoodle, line 253: should this always be prevented by `handlePopFromNode` ?

pkg/scheduler/actions/utils/job_order_by_queue.go
```diff
@@ -12,329 +12,410 @@ import (
...
+		}
 		return
 	}
 
-	jobsOrder.queueIdToQueueMetadata[queue.UID].shouldUpdateQueueShare = true
-}
+	parentQueueInfo, parentExists := jo.ssn.ClusterInfo.Queues[childQueue.ParentQueue]
+	if !parentExists {
+		log.InfraLogger.V(7).Warnf("Queue's parent doesn't exist. Queue: <%v>, Parent: <%v>",
+			childQueue.Name, childQueue.ParentQueue)
```
enoodle, line 195: We should make sure that the pods in that queue will have some kind of message / condition indicating this situation. Maybe open an issue for that?

pkg/scheduler/actions/utils/job_order_by_queue.go
```diff
@@ -12,329 +12,410 @@ import (
...
 
-func (jobsOrder *JobsOrderByQueues) getNextQueue(department *queue_info.QueueInfo) *queue_info.QueueInfo {
-	queue := jobsOrder.departmentIdToDepartmentMetadata[department.UID].queuesPriorityQueue.Peek().(*queue_info.QueueInfo)
-	if jobsOrder.queueIdToQueueMetadata[queue.UID].shouldUpdateQueueShare {
-		jobsOrder.updateTopQueueShare(queue, department)
-		return jobsOrder.getNextQueue(department)
+	// Get or create parent node
+	parentNode := jo.queueNodes[parentQueueInfo.UID]
+	parentNodeIsNew := parentNode == nil
+	if parentNode == nil {
```
enoodle:
```suggestion
	if parentNodeIsNew {
```

pkg/scheduler/actions/utils/input_jobs.go
```diff
@@ -15,21 +15,22 @@ type JobsOrderInitOptions struct {
 	FilterNonPreemptible     bool
 	FilterNonActiveAllocated bool
 	VictimQueue              bool
-	MaxJobsQueueDepth        int
+	// TODO: Rename to MaxJobsPerQueue as the name is misleading.
+	MaxJobsQueueDepth int
```
enoodle: Why is the name misleading? I think you should either suggest a better name or not set the comment. :)

pkg/scheduler/actions/utils/input_jobs.go
```diff
@@ -50,13 +51,21 @@ func (jobsOrder *JobsOrderByQueues) InitializeWithJobs(
...
+		if !jobsOrder.ssn.ClusterInfo.Queues[job.Queue].IsLeafQueue() {
 			continue
 		}
 
-		jobsOrder.addJobToQueue(job, jobsOrder.jobsOrderInitOptions.VictimQueue)
+		jobsOrder.addJobToQueue(job)
 	}
 
-	jobsOrder.buildActiveJobOrderPriorityQueues(jobsOrder.jobsOrderInitOptions.VictimQueue)
+	jobsOrder.buildActiveQueues()
```
enoodle, line 69: Can we just use `PushJob` and then we will not need `buildActiveQueues` at all?

pkg/scheduler/actions/utils/input_jobs.go
gshaibi: The name MaxJobsQueueDepth could be confusing in a hierarchical queues context, where "depth" typically refers to tree depth. Since this field actually limits the number of jobs considered per queue, MaxJobsPerQueue might be clearer.
I've added a TODO comment suggesting this rename rather than doing it in this PR to keep the diff focused.
If you think the old name fits better I'll remove the comment.

pkg/scheduler/actions/utils/input_jobs.go
gshaibi: Good suggestion. The difference is that `buildActiveQueues` immediately reorders the tree, while `PushJob` only marks it for reordering. 
Switching to PushJob would require changes to the existing flow, which I'd prefer to keep out of this PR to minimize scope.

pkg/scheduler/actions/utils/job_order_by_queue.go
gshaibi: I think this was handled already on #820

pkg/scheduler/actions/utils/job_order_by_queue.go
gshaibi: Fixed, thanks

pkg/scheduler/actions/utils/job_order_by_queue.go
gshaibi: You're right, it should, this is just a defensive check that I wonder if we need.
wdut?

pkg/scheduler/actions/utils/input_jobs.go
enoodle: I understand the possible confusion, I still think you should either rename it or leave it without the comment.

pkg/scheduler/actions/utils/input_jobs.go
enoodle: I don't see any changes that are needed, I tried it and all the tests seem to pass.
The reason is that when a job is popped from a queue that is marked for re-order then this is done and a job is popped again.
What do you think about making this change and deleting those two functions?

pkg/scheduler/actions/utils/job_order_by_queue.go
enoodle: :+1:

pkg/scheduler/actions/utils/job_order_by_queue.go
enoodle: maybe we should add a comment that this is an unhealthy situation that shouldn't happen? It might help us if it ever does happen.

pkg/scheduler/actions/utils/input_jobs.go
gshaibi: OK. I removed the comment.

pkg/scheduler/actions/utils/input_jobs.go
enoodle: Lets do it in another PR

enoodle (approved):

PR #881, opened 2026-01-15
fix: remove TODO in queue_info/queue_info.go

rich7420:
## Description

Reads child queues from `queue.Status.ChildQueues` instead of leaving it empty. The controller already maintains this via `childQueuesUpdater`, so we're just using that data now.

## Related Issues

Fixes #

## Checklist

- [x] Self-reviewed


rich7420:
cc @omer-dayan @itsomri @romanbaron

enoodle:
@rich7420 I am not sure that the todo comment is relevant :) there is some other piece of  code in the cluster info that updates this field. I think we can just remove the comment for now.

rich7420:
So should I close this PR?

rich7420:
@enoodle thanks for the review!

enoodle:
This is implemented here: https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/scheduler/cache/cluster_info/queue.go#L95-L103 
which is actually a better place because it doesn't rely on the queue controller so there will be no races.. I guess the comment should be removed.

rich7420:
@enoodle thanks for the review!

PR #885, opened 2026-01-15
feat(podgrouper): support owner-chain fallback for plugin selection

### LINKED ISSUE #856: Enhance PodGrouper plugin matching to fallback through owner chain

Enhance the PodGrouper plugin matching logic so that, after resolving the full owner reference chain for a pod (for example, `Pod → Job → JobSet → Kubeflow trainer`), the plugin is not selected solely based on the topmost owner type with an immediate fallback to the default plugin when no match is found. Instead, PodGrouper should iterate over the owner chain, starting from the topmost owner and moving down towards the pod, attempting to match a plugin for each owner type in order, using the existing plugin hub implementation (for example, the Job plugin registration in hub.go). Only if no plugin is found for any owner in the chain should the default plugin be used, so that existing workload-specific plugins continue to drive PodGroup creation even when new intermediate owner types (such as JobSet) are introduced without a dedicated plugin.

Useful references:

PodGrouper main flow and owner resolution: [pkg/podgrouper/podgrouper/podgrouper.go](https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/podgrouper/podgrouper/podgrouper.go#L54).

Plugin hub: [pkg/podgrouper/podgrouper/hub/hub.go](https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/podgrouper/podgrouper/hub/hub.go)

PodGrouper developer docs: [docs/developer/pod-grouper.md](https://github.com/NVIDIA/KAI-Scheduler/blob/main/docs/developer/pod-grouper.md)

### Why is this needed?

With the current behavior, extending RBAC and ownership chains to include new types (e.g., JobSet) can unintentionally degrade grouping semantics, because PodGrouper climbs to the new topmost owner, fails to match a plugin, and immediately falls back to the default per-pod grouping instead of leveraging an already available, more suitable plugin for an intermediate owner (such as batch/v1 Job). An owner-chain–aware fallback mechanism makes PodGrouper more robust to incremental integrations, prevents regressions in gang scheduling behavior when new workload types are added, and ensures that existing plugins continue to be applied wherever they can accurately model the desired scheduling semantics until dedicated plugins for new owners are implemented.

singh1203:
## Description

**feat(podgrouper)**: Implement owner-reference chain fallback for plugin selection

Enhance `PodGrouper.GetPGMetadata()` to iterate through the owner chain (`allOwners` slice)
from top-most owner down, selecting the highest GVK with a non-default plugin via
`PluginsHub.GetPodGrouperPlugin()`. Eliminates scheduling regressions when new owner types
gain RBAC visibility but lack dedicated plugins (e.g., JobSet → fallback to `batch/v1.Job`).

## Related Issues

Fixes #856

## Checklist

- [x] Self-reviewed
- [x] Added/updated tests  
- [ ] Updated documentation

## Breaking Changes

None

## Additional Notes

**Before**: `JobSet → no plugin → defaultgrouper` (per-pod grouping)  
**After**: `JobSet → no plugin → Job → job.NewK8sJobGrouper` (Job-level grouping)

Implementation via `selectPluginForOwners()` helper using `DefaultPluginsHub` type assertion.
No `PluginsHub` interface or plugin signature changes required.

## Release Notes

* **Improvements**
  * Enhanced plugin selection mechanism to intelligently handle complex ownership chains, with automatic fallback to intermediate owners when needed.

* **Tests**
  * Added test coverage for owner chain fallback scenarios to ensure reliable behavior in complex pod ownership configurations.

### DIFF
--- a/pkg/podgrouper/podgrouper/podgrouper.go
+++ b/pkg/podgrouper/podgrouper/podgrouper.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/NVIDIA/KAI-scheduler/pkg/podgrouper/podgroup"
 	pluginshub "github.com/NVIDIA/KAI-scheduler/pkg/podgrouper/podgrouper/hub"
+	"github.com/NVIDIA/KAI-scheduler/pkg/podgrouper/podgrouper/plugins/grouper"
 )
 
 type Interface interface {
@@ -77,10 +78,9 @@ func (pg *podGrouper) GetPodOwners(ctx context.Context, pod *v1.Pod) (
 
 func (pg *podGrouper) GetPGMetadata(ctx context.Context, pod *v1.Pod, topOwner *unstructured.Unstructured, allOwners []*metav1.PartialObjectMetadata) (*podgroup.Metadata, error) {
 	logger := log.FromContext(ctx)
-	ownerKind := metav1.GroupVersionKind(topOwner.GroupVersionKind())
-	plugin := pg.pluginsHub.GetPodGrouperPlugin(ownerKind)
+	plugin := pg.selectPluginForOwners(topOwner, allOwners)
 	logger.V(1).Info(fmt.Sprintf("Using %v plugin for pod.", plugin.Name()),
-		"pod", fmt.Sprintf("%s/%s", pod.Namespace, pod.Name), "topOwner", topOwner)
+		"pod", fmt.Sprintf("%s/%s", pod.Namespace, pod.Name), "topOwner", topOwner.GroupVersionKind())
 	return plugin.GetPodGroupMetadata(topOwner, pod, allOwners...)
 }
 
@@ -183,4 +183,32 @@ func (pg *podGrouper) handleGetOwnerError(
 	}
 	return nil, fmt.Errorf("failed to find %s: <%s/%s>, uid: <%v>: %w",
 		owner.Kind, pod.Namespace, owner.Name, owner.UID, err)
+}
+
+func (pg *podGrouper) selectPluginForOwners(
+	topOwner *unstructured.Unstructured,
+	allOwners []*metav1.PartialObjectMetadata,
+) grouper.Grouper {
+	hub, ok := pg.pluginsHub.(*pluginshub.DefaultPluginsHub)
+	if !ok {
+		return pg.pluginsHub.GetPodGrouperPlugin(metav1.GroupVersionKind{})
+	}
+
+	defaultPlugin := hub.GetDefaultPlugin()
+
+	// Build ownerGVKs in top-down order
+	ownerGVKs := make([]metav1.GroupVersionKind, 0, len(allOwners)+1)
+	ownerGVKs = append(ownerGVKs, metav1.GroupVersionKind(topOwner.GroupVersionKind()))
+
+	for i := len(allOwners) - 1; i >= 0; i-- {
+		ownerGVKs = append(ownerGVKs, metav1.GroupVersionKind(allOwners[i].GroupVersionKind()))
+	}
+
+	for _, gvk := range ownerGVKs {
+		plugin := pg.pluginsHub.GetPodGrouperPlugin(gvk)
+		if plugin != defaultPlugin {
+			return plugin
+		}
+	}
+	return defaultPlugin
 }
--- a/pkg/podgrouper/podgrouper/podgrouper_test.go
+++ b/pkg/podgrouper/podgrouper/podgrouper_test.go
@@ -374,3 +374,87 @@ func compareMetadata(expected, got *podgroup.Metadata) error {
 
 	return err
 }
+
+func TestPodGrouper_OwnerChainFallback_UsesIntermediatePlugin(t *testing.T) {
+	scheme := runtime.NewScheme()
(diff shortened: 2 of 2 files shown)


enoodle:
@coderabitai review

enoodle:
@coderabbitai review

pkg/podgrouper/podgrouper/podgrouper_test.go
```diff
@@ -374,3 +374,87 @@ func compareMetadata(expected, got *podgroup.Metadata) error {
...
+
+	// Create a real batch/v1 Job object
+	job := &unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": "batch/v1",
+			"kind":       "Job",
+			"metadata": map[string]interface{}{
+				"name":      "job-owner",
+				"namespace": "namespace",
+				"uid":       "uid-job",
```
enoodle, line 390: shouldn't the job object have the jobSet as owner?

pkg/podgrouper/podgrouper/podgrouper.go
```diff
@@ -184,3 +184,31 @@ func (pg *podGrouper) handleGetOwnerError(
 	return nil, fmt.Errorf("failed to find %s: <%s/%s>, uid: <%v>: %w",
 		owner.Kind, pod.Namespace, owner.Name, owner.UID, err)
 }
+
+func (pg *podGrouper) selectPluginForOwners(
+	topOwner *unstructured.Unstructured,
+	allOwners []*metav1.PartialObjectMetadata,
+) grouper.Grouper {
+	hub, ok := pg.pluginsHub.(*pluginshub.DefaultPluginsHub)
```
enoodle, line 192: I think that this is a bit missing the idea of having the plugins hub as an interface. I suggest that instead we will add the `GetDefaultPlugin` function to the interface.

pkg/podgrouper/podgrouper/podgrouper.go
```diff
@@ -184,3 +184,31 @@ func (pg *podGrouper) handleGetOwnerError(
...
+		return pg.pluginsHub.GetPodGrouperPlugin(metav1.GroupVersionKind{})
+	}
+
+	defaultPlugin := hub.GetDefaultPlugin()
+
+	// Build ownerGVKs in top-down order
+	ownerGVKs := make([]metav1.GroupVersionKind, 0, len(allOwners)+1)
+	ownerGVKs = append(ownerGVKs, metav1.GroupVersionKind(topOwner.GroupVersionKind()))
+
+	for i := len(allOwners) - 1; i >= 0; i-- {
```
enoodle, line 203: I think it would be more clear to use a function like `slices.Reveres()` and then there will be no need for this intermediate array and this for loop

pkg/podgrouper/podgrouper/podgrouper.go
```diff
@@ -77,10 +78,9 @@ func (pg *podGrouper) GetPodOwners(ctx context.Context, pod *v1.Pod) (
 
 func (pg *podGrouper) GetPGMetadata(ctx context.Context, pod *v1.Pod, topOwner *unstructured.Unstructured, allOwners []*metav1.PartialObjectMetadata) (*podgroup.Metadata, error) {
 	logger := log.FromContext(ctx)
-	ownerKind := metav1.GroupVersionKind(topOwner.GroupVersionKind())
-	plugin := pg.pluginsHub.GetPodGrouperPlugin(ownerKind)
+	plugin := pg.selectPluginForOwners(topOwner, allOwners)
 	logger.V(1).Info(fmt.Sprintf("Using %v plugin for pod.", plugin.Name()),
-		"pod", fmt.Sprintf("%s/%s", pod.Namespace, pod.Name), "topOwner", topOwner)
+		"pod", fmt.Sprintf("%s/%s", pod.Namespace, pod.Name), "topOwner", topOwner.GroupVersionKind())
 	return plugin.GetPodGroupMetadata(topOwner, pod, allOwners...)
```
enoodle, line 84: should we still send `topOwner` here even if it was not found to fit the plugin that we use eventually?

pkg/podgrouper/podgrouper/podgrouper.go
```diff
@@ -184,3 +184,31 @@ func (pg *podGrouper) handleGetOwnerError(
...
 		owner.Kind, pod.Namespace, owner.Name, owner.UID, err)
 }
+
+func (pg *podGrouper) selectPluginForOwners(
+	topOwner *unstructured.Unstructured,
+	allOwners []*metav1.PartialObjectMetadata,
+) grouper.Grouper {
+	hub, ok := pg.pluginsHub.(*pluginshub.DefaultPluginsHub)
+	if !ok {
+		return pg.pluginsHub.GetPodGrouperPlugin(metav1.GroupVersionKind{})
```
enoodle, line 194: shouldn't it use `metav1.GroupVersionKind(topOwner.GroupVersionKind())` here?

enoodle (commented):
Maybe this will be better implemented with the `skipTopOwner` plugin which knows to pass along the labels and annotations to the next next object in the ownership links.

PR #887, opened 2026-01-16
feat(admission): Add scheduler name match conditions webhooks

enoodle:
## Description

This change adds MatchConditions to the mutating and validating admission webhooks to ensure they only process pods with the correct scheduler name.

https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchconditions

Benefit for cluster stability:

- Webhooks are only invoked for KAI-scheduled pods, completely eliminating impact on other workloads

- When admission webhook server is down, non-KAI pods continue to be created normally

- Improved performance by reducing unnecessary webhook calls for unrelated pods

## Related Issues

Fixes https://github.com/NVIDIA/KAI-Scheduler/issues/883

## Checklist

- [ ] Self-reviewed

### DIFF
--- a/pkg/operator/operands/admission/resources.go
+++ b/pkg/operator/operands/admission/resources.go
@@ -8,12 +8,11 @@ import (
 	"fmt"
 	"strconv"
 
-	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/apimachinery/pkg/types"
-
 	admissionv1 "k8s.io/api/admissionregistration/v1"
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


pkg/operator/operands/admission/resources.go
```diff
@@ -180,6 +179,18 @@ func buildWebhookSelectors(kaiConfig *kaiv1.Config) (namespaceSelector *metav1.L
...
+// buildSchedulerNameMatchConditions creates match conditions to only run webhooks for pods with the correct scheduler name
+func buildSchedulerNameMatchConditions(kaiConfig *kaiv1.Config) []admissionv1.MatchCondition {
+	var matchConditions []admissionv1.MatchCondition
+	if kaiConfig.Spec.Global != nil && kaiConfig.Spec.Global.SchedulerName != nil {
+		matchConditions = append(matchConditions, admissionv1.MatchCondition{
+			Name:       "scheduler-name-match",
+			Expression: fmt.Sprintf("object.spec.schedulerName == '%s'", *kaiConfig.Spec.Global.SchedulerName),
+		})
+	}
+	return matchConditions
```
singh1203, line 191: I noticed when SchedulerName is nil, we just return an empty slice here.
If `matchConditions` is empty, wouldn't the webhook just run for all pods? That seems to defeat the whole point of this change. What happens in practice when someone doesn't set a scheduler name in the config? 
Should we maybe throw an error or at least warn about it?

enoodle: It can't be nil: https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/apis/kai/v1/global.go#L105

itsomri (approved):

PR #910, opened 2026-01-21
fix(operator): ClusterPolicy CDI parsing with gpu-operator > v25.10.0

enoodle:
## Related Issues

Fixes #

## Checklist

- [ ] Self-reviewed


pkg/operator/operands/binder/binder_test.go
```diff
@@ -137,7 +137,7 @@ var _ = Describe("Binder", func() {
 
 				It("detects CDI state with GPU Operator >= v25.10.0", func(ctx context.Context) {
 					clusterPolicy.Labels = map[string]string{
-						versionLabelName: gpuOperatorVersionDefaultCDIDeprecated,
+						versionLabelName: "v25.10.1",
```
gshaibi, line 140: It might also be beneficial to test the exact version if you want

gshaibi (approved):

pkg/operator/operands/binder/binder_test.go
enoodle: That was exactly the issue we had before ;-) It tested the exact version so we didn't see that it was the wrong function.

Issue #901, opened 2026-01-19
Add Horizontal Pod Autoscaling for the KAI Admission Service


davidLif:
### Proposal: Add Horizontal Pod Autoscaling for the KAI Admission Service

KAI includes an admission service responsible for handling the pod mutation and validation webhooks for pods that are scheduled by KAI.

Today, if a user observes high load on the admission service, they can manually increase the number of replicas via the KAI configuration. While this works, manually tuning replica counts is often suboptimal and reactive. In most cases, configuring Horizontal Pod Autoscaling (HPA) for the admission service deployment would provide a more robust and adaptive solution.

This enhancement proposes adding optional support for HPA on the admission service, allowing it to scale automatically based on load.

#### Open questions and considerations

- **Autoscaling metrics**  
  Which metrics best reflect the actual load on the admission service pods?  
  Examples to consider include:
  - CPU and/or memory utilization
  - Request rate to the webhook endpoints
  - Admission request latency
  - In-flight or queued admission requests (if available)

- **Manual replica configuration**  
  Users should still be able to explicitly configure a fixed number of replicas for the admission service via the KAI configuration.  
  We need to define how manual replica settings interact with HPA (e.g., mutually exclusive, or using the configured replica count as the HPA minimum).

### Why is this needed?

Horizontal pod autoscaling will allow the deployment to automaticly add/reduce the number of admission pods according to the pod webhoolks load.

rich7420:
@davidLif thanks for the issue
I think that, for autoscaling metrics, we should prioritize in-flight requests first, and use CPU and/or memory utilization as secondary metrics. 
Maybe the replica count set manually could be used as the minReplicas value in the HPA.

faizan-exe:
I agree with @rich7420 I am moving forward with this approach.

Also, should we enable setting admission replicaCount using values.yaml also?

davidLif:
> I agree with [@rich7420](https://github.com/rich7420) I am moving forward with this approach.
> 
> Also, should we enable setting admission replicaCount using values.yaml also?

Yes

faizan-exe:
Since we are going to be dealing with Custom Metrics API. I suggest we should be using KEDA instead of HPA

enoodle:
I think that KEDA is a bit of an overkill for this case and also we would like to avoid adding more dependencies when we can rely on existing ones. The PR with prometheus adapter seems to solve this well enough.
[...]

PR #911, opened 2026-01-21
feat(admission): enable hpa for KAI Admission Service

faizan-exe:
## Description

This PR enables Horizontal Pod Autoscaling for Admission Pods. Previously, user can only specifiy fixed number of replicas for admission pods using `kai-config.yaml` - now they have the option to enable autoscaling and avoid manual configuration.

The metrics for autoscaling are
-  `controller_runtime_webhook_requests_total`
- `process_cpu_seconds_total`

Prometheus Adapter is also introduced for interacting with Custom Metrics API.

## Related Issues

Issue #901 

## Checklist

- [Yes] Self-reviewed
- [Yes] Added/updated tests (if needed)
- [Yes] Updated documentation (if needed)

## Breaking Changes

No breaking changes. Backward Compatibility is being ensured.


pkg/apis/kai/v1/admission/admission.go
```diff
@@ -94,3 +101,30 @@ func (w *Webhook) SetDefaultsWhereNeeded() {
...
+	// +kubebuilder:validation:Optional
+	AverageRequestsPerPod *int32 `json:"averageRequestsPerPod,omitempty"`
+}
+
+// SetDefaultsWhereNeeded sets default fields for unset fields
+func (a *Autoscaling) SetDefaultsWhereNeeded() {
+	a.Enabled = common.SetDefault(a.Enabled, ptr.To(false))
+	a.MinReplicas = common.SetDefault(a.MinReplicas, ptr.To(int32(1)))
+	a.MaxReplicas = common.SetDefault(a.MaxReplicas, ptr.To(int32(5)))
+	a.AverageRequestsPerPod = common.SetDefault(a.AverageRequestsPerPod, ptr.To(int32(100)))
```
enoodle: Why 100?
Please use constants.

pkg/operator/operands/admission/resources.go
```diff
@@ -42,7 +44,10 @@ func (a *Admission) deploymentForKAIConfig(
 
 	deployment.Spec.Strategy.Type = appsv1.RecreateDeploymentStrategyType
 	deployment.Spec.Strategy.RollingUpdate = nil
-	deployment.Spec.Replicas = config.Replicas
+	// Only set replicas if HPA is disabled; HPA manages replicas automatically
```
enoodle: Yes, but the flow is clear enough without the comments. please remove.

pkg/operator/operands/common/common.go
```diff
@@ -36,6 +36,7 @@ var KaiServicesForServiceMonitor = []struct {
 	Namespaces    []string          // optional, if not provided, the service will be monitored in the namespace of the KAI config
 }{
 	{"queue-controller", "metrics", "queue-controller", nil, nil},
+	{"admission", "http-metrics", "admission", nil, nil},
```
enoodle: This prometheus instance is suppose to be used for the time aware scheduling feature only, so this is not needed here.

enoodle (changes requested):

docs/metrics/README.md
```diff
@@ -19,4 +19,21 @@ To enable the prometheus as a grafana datasource, if desired, apply grafana-data
...
+
+# Prometheus Adapter
+
+Prometheus Adapter exposes Prometheus metrics as Kubernetes custom metrics API, enabling HPA to scale based on custom metrics.
+
+## Installation
+
+Add the Prometheus Adapter Helm repository:
+```bash
+helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
```
itsomri, line 32: I think we can drop these two lines as they're duplicates from lines 4-5 in this file

docs/metrics/service-monitors.yaml
```diff
@@ -93,4 +93,22 @@ spec:
...
\ No newline at end of file
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: admission
+  namespace: kai-scheduler
+  labels:
+    accounting: kai-scheduler
```
itsomri, line 104: I think we should have an "accounting" prometheus that's separate from a general prometheus for metrics (and autoscaling).
The reasons are:
1. When using time aware fairness, the accounting prometheus is in the critical path of the scheduler. While the scheduler will work even if it's down, it could make very different scheduling decisions, causing unexpected preemptions. So we should keep it very minimal and keep load off of it.
2. The accounting prometheus, in some expected use cases, needs to be persistent, even for a year or more of data. Since you can't choose which metrics are persisted and which aren't, this means that every metric that it's collecting will take that space. But the more operational metrics (like the ones used for HPA, or other metrics like scheduler metrics) might not need this persistency, causing wasted storage space
3. The accounting prometheus can potentially have different sampling intervals than the general prometheus

Maybe we can find some design where the separation is configurable, or maybe it will just be easier to have them always separate.

What do you think?

pkg/operator/operands/admission/resources.go
```diff
@@ -139,6 +143,66 @@ func (a *Admission) serviceForKAIConfig(
...
+		},
+		{
+			Type: autoscalingv2.PodsMetricSourceType,
+			Pods: &autoscalingv2.PodsMetricSource{
+				Metric: autoscalingv2.MetricIdentifier{
+					Name: "cpu_utilization",
+				},
+				Target: autoscalingv2.MetricTarget{
+					Type:         autoscalingv2.AverageValueMetricType,
+					AverageValue: resource.NewMilliQuantity(int64(*config.Autoscaling.CPUUtilizationPercent)*10, resource.DecimalSI),
```
enoodle, line 197: Why is this multiplied by 10? What is the meaning of this Average Value here and why would it get down when scaling?

pkg/operator/operands/admission/resources.go
```diff
@@ -42,7 +44,9 @@ func (a *Admission) deploymentForKAIConfig(
 
 	deployment.Spec.Strategy.Type = appsv1.RecreateDeploymentStrategyType
 	deployment.Spec.Strategy.RollingUpdate = nil
-	deployment.Spec.Replicas = config.Replicas
+	if config.Autoscaling == nil || !*config.Autoscaling.Enabled {
+		deployment.Spec.Replicas = config.Replicas
```
enoodle, line 48: Maybe it should be set to nil otherwise? notice that it reads the current deployment in and only updates those fields.

docs/metrics/prometheus-adapter-values.yaml
```diff
@@ -0,0 +1,32 @@
+
+# Copyright 2025 NVIDIA CORPORATION
+# SPDX-License-Identifier: Apache-2.0
+
+prometheus:
+  url: http://prometheus-operated.kai-scheduler.svc.cluster.local
```
enoodle, line 6: We should use another instance

enoodle:
@faizan-exe Do you want to continue this ?

PR #932, opened 2026-01-26
docs: Added ray examples

itsomri:
## Related Issues

Fixes #

## Checklist

- [ ] Self-reviewed

### DIFF
--- a/examples/ray/ray-job.yaml
+++ b/examples/ray/ray-job.yaml
@@ -1,3 +1,6 @@
+# Copyright 2026 NVIDIA CORPORATION
+# SPDX-License-Identifier: Apache-2.0
+
 apiVersion: ray.io/v1
 kind: RayJob
 metadata:


examples/ray/ray-job.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: ray.io/v1
+kind: RayJob
+metadata:
+  name: rayjob-sample
+  annotations:
+    scheduling.run.ai/queue-name: "default"          # KAI queue name
```
rich7420:
```suggestion
```

examples/ray/ray-job.yaml
```diff
@@ -0,0 +1,44 @@
...
+spec:
+  entrypoint: python /home/ray/samples/sample_code.py
+  rayClusterSpec:
+    rayVersion: '2.46.0'
+    # Head group configuration
+    headGroupSpec:
+      rayStartParams: {}
+      template:
```
rich7420, line 17:
```suggestion
      template:
        metadata:
          labels:
            kai.scheduler/queue: default-queue
```

examples/ray/ray-job.yaml
```diff
@@ -0,0 +1,44 @@
...
+              requests:
+                cpu: "200m"
+    # Worker group configuration
+    workerGroupSpecs:
+    - replicas: 2
+      minReplicas: 1
+      maxReplicas: 5
+      groupName: gpu-workers
+      rayStartParams: {}
+      template:
```
rich7420, line 36:
```suggestion
      template:
        metadata:
          labels:
            kai.scheduler/queue: default-queue
```

rich7420 (commented):
Because KAI’s pod-grouper only looks for the queue in **labels**, not annotations (i think it literally reads `kai.scheduler/queue` from `topOwner.GetLabels()` or `pod.GetLabels()` in `CalcPodGroupQueue`), so any queue you put in an annotation won’t be used. With RayJob/KubeRay, the pods are created from the **head/worker pod template**, and the labels that end up on the actual pods come from `template.metadata.labels`; the RayJob’s `metadata.annotations` don’t get copied onto the pods, so KAI never sees them. If you only set the annotation (the older style), the job will still deploy and run, but since neither the top owner nor the pod has the queue **label**, KAI falls back to `DefaultQueueName` (`"default-queue"`), meaning the PodGroup is always scheduled into `default-queue`. In other words, the annotation isn’t “wrong” so much as it’s simply ignored by KAI’s queue resolution, and if you want a custom queue like `"my-queue"`, you’ll need to put `kai.scheduler/queue: my-queue` in the **pod template labels** so the created pods actually carry the label KAI reads. Plz correct me if i'm wrong. thanks!
ref: https://github.com/NVIDIA/KAI-Scheduler/blob/746e5fe93a177357bef67004cc48b791ae636296/pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go#L134-L147

examples/ray/ray-job.yaml
itsomri: Thanks! good catch. I fixed all the examples and validates them

enoodle:
You only need to set the labels on the top owner.

enoodle (approved):
I verified that this works

rich7420 (approved):
lg

PR #933, opened 2026-01-27
fix(e2e): dra tests

enoodle:
## Related Issues

Fixes #

## Checklist

- [ ] Self-reviewed

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,9 @@
+# Copyright 2026 NVIDIA CORPORATION
+# SPDX-License-Identifier: Apache-2.0
+
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+language: "en-US"
+
+reviews:
+  auto_review:
+    enabled: false
--- a/.github/workflows/post-coverage-comment.yaml
+++ b/.github/workflows/post-coverage-comment.yaml
@@ -20,14 +20,14 @@ jobs:
 
     steps:
       - name: Download PR number artifact
-        uses: actions/download-artifact@v6
+        uses: actions/download-artifact@v7
         with:
           name: pr-number-for-comment
           run-id: ${{ github.event.workflow_run.id }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Download coverage report artifact
-        uses: actions/download-artifact@v6
+        uses: actions/download-artifact@v7
         with:
           name: coverage-report-for-comment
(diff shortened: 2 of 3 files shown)


.github/workflows/on-pr.yaml
```diff
@@ -263,6 +263,8 @@ jobs:
         run: |
           helm upgrade -i kai-scheduler /mnt/images/kai-scheduler-$PACKAGE_VERSION.tgz -n kai-scheduler --create-namespace \
             --set "global.gpuSharing=true" --set "global.registry=localhost:30100" --debug --wait
+          kubectl create clusterrole pods-patcher --verb=patch --resource=pods
```
davidLif, line 266: Do we want to add this change to the hack installation as well?

davidLif (dismissed):

.github/workflows/on-pr.yaml
enoodle: :+1: fixed

test/e2e/suites/timeaware/timeaware_suite_test.go
```diff
@@ -43,9 +50,24 @@ var _ = BeforeSuite(func(ctx context.Context) {
 	By("Setting up test context")
 	testCtx = testcontext.GetConnectivity(ctx, Default)
 
+	By("Creating SchedulingShard for DRA nodes")
```
davidLif, line 53: Why do you create this shard for the timeaware tests?

test/e2e/suites/reclaim/hierarchy_level_fairness_test.go
```diff
@@ -39,6 +48,19 @@ var _ = Describe("Hierarchy level fairness", Ordered, func() {
 
 		BeforeAll(func(ctx context.Context) {
 			testCtx = testcontext.GetConnectivity(ctx, Default)
+
+			By("Creating SchedulingShard for DRA nodes")
```
davidLif, line 52: Why creating a shard here? Shouldn't this shard be created outside of spesific e2e tests?

enoodle: Two reason: 
1. It is not possible - there is no concept in ginkgo of "BeforeAllTestSuites"
2. We want to test that the logic works in a hybrid environment with both device plugin and DRA plugin enabled on different nodes - Those tests have logic that is dependant on them being able to fill the cluster which is not supported right now so they need this intevention. It is actually a bug in the design of the test cases

PR #955, opened 2026-02-03
fix: override DRA enable flag fix

SiorMeir:
## Related Issues

Fixes #

## Checklist

- [ ] Self-reviewed


enoodle:
Maybe we should just always set DRA to true in the snapshot tool? wouldn't it be simpler ?

SiorMeir:
@enoodle And when its off we record nothing? Wouldn't that be an issue?

@davidLif WDYT?

If we won't care this'll be a lot easier

enoodle:
@SiorMeir It will be like there are no available DRA resources...

PR #957, opened 2026-02-03
feat(pod-group): add validation on subgroup names

### LINKED ISSUE #948: Validate subgroup names are lowercase on podgroup creation

### Why is this needed?

faizan-exe:
## Description

This PR validates if the subGroup names are lowercase in admission webhook.

## Related Issues

Fixes #948 

## Checklist

- [Yes] Self-reviewed
- [Yes] Added/updated tests (if needed)
- [Nil] Updated documentation (if needed)

## Breaking Changes

No breaking changes.


faizan-exe:
PTAL. Thanks!
cc - @itsomri

itsomri (dismissed):
Looks great!

gshaibi (changes requested):
Hi @faizan-exe , thanks for the contribution!

The validation logic looks correct, but I think we should use a `+kubebuilder:validation:Pattern` marker on the `Name` field (and `Parent` field) in `podgroup_types.go` instead of adding this in the webhook.

The field already uses kubebuilder markers (`+kubebuilder:validation:MinLength=1`), so adding a `Pattern` constraint would enforce this at the CRD/OpenAPI schema level by the API server itself — before the request ever reaches the webhook. This is more idiomatic for Kubernetes APIs and provides defense-in-depth.

The same pattern should also be added to the `Parent *string` field, since it references subgroup names.

The webhook should still keep the cross-field validations (duplicate names, parent existence, cycle detection), as those can't be expressed declaratively.

After the change, run `make manifests` to regenerate the CRD.

pkg/apis/scheduling/v2alpha2/podgroup_types.go
```diff
@@ -118,7 +120,9 @@ type SubGroup struct {
 	MinMember int32 `json:"minMember,omitempty"`
 
 	// Parent is an optional attribute that specifies the name of the parent SubGroup
```
gshaibi: Please add a period.

pkg/apis/scheduling/v2alpha2/podgroup_types.go
```diff
@@ -109,7 +109,9 @@ func ParsePreemptibility(value string) (Preemptibility, error) {
 
 type SubGroup struct {
 	// Name uniquely identifies the SubGroup within the PodGroup.
+	// Must consist of lowercase alphanumeric characters or hyphens.
 	// +kubebuilder:validation:MinLength=1
```
gshaibi, line 113: That is not required anymore, right?

pkg/apis/scheduling/v2alpha2/podgroup_types.go
```diff
@@ -109,7 +109,9 @@ func ParsePreemptibility(value string) (Preemptibility, error) {
 
 type SubGroup struct {
 	// Name uniquely identifies the SubGroup within the PodGroup.
+	// Must consist of lowercase alphanumeric characters or hyphens.
```
gshaibi: Please consider rephrasing to something like:
```suggestion
	// Must be a valid DNS label (RFC 1123).
```

pkg/apis/scheduling/v2alpha2/podgroup_types.go
```diff
@@ -118,7 +120,9 @@ type SubGroup struct {
 	MinMember int32 `json:"minMember,omitempty"`
 
 	// Parent is an optional attribute that specifies the name of the parent SubGroup
+	// Must consist of lowercase alphanumeric characters or hyphens.
```
gshaibi: Please consider rephrasing to something like:
```suggestion
	// Must be a valid DNS label (RFC 1123).
```

gshaibi (commented):
Looks good, added a couple comments

gshaibi (changes requested):
Please run `make generate manifests`

gshaibi (dismissed):

itsomri (dismissed):

itsomri (dismissed):

gshaibi (dismissed):

enoodle (approved):

itsomri (approved):

enoodle:
continued in #1383

PR #969, opened 2026-02-05
feat: Track pod count as a scalar resource

itsomri:
## Description

This PR fixed max pod tracking on scheduling simulations.

Changes:
- "Pods" are now tracked as a scalar resource 
- Added an e2e test to verify preemption simulates pod limits correctly
- Changed node max pods predicate to handle edge cases for reservation pods

## Checklist

- [ ] Self-reviewed

### DIFF
--- a/test/e2e/suites/preempt/preempt_max_pods_test.go
+++ b/test/e2e/suites/preempt/preempt_max_pods_test.go
@@ -7,7 +7,6 @@ package preempt
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -213,9 +212,5 @@ var _ = Describe("Preemption with Max Pods Limit", Ordered, func() {
 
 		// Wait and verify pod remains unschedulable
 		wait.ForPodUnschedulable(ctx, testCtx.ControllerClient, fractionPod)
-
-		wait.WaitForEventInNamespace(ctx, testCtx.ControllerClient, fractionPod.Namespace, func(event *v1.Event) bool {
-			return event.Reason == "Unschedulable" && (strings.Contains(event.Message, "pod number exceeded") || strings.Contains(event.Message, "max pods"))
-		})
 	})
 })


itsomri:
@coderabbitai review

pkg/scheduler/api/resource_info/resource_requirment.go
```diff
@@ -16,6 +16,7 @@ import (
 const (
 	GPUResourceName    = "nvidia.com/gpu"
 	amdGpuResourceName = "amd.com/gpu"
+	PodsResourceName   = v1.ResourceName("pods")
```
gshaibi: Should we use `v1.ResourcePods`?

pkg/scheduler/api/resource_info/base_resources.go
```diff
@@ -55,6 +55,9 @@ func (r *BaseResource) Add(other *BaseResource) {
 	r.memory += other.memory
 	for rName, rrValue := range other.scalarResources {
 		r.scalarResources[rName] += rrValue
+		if r.scalarResources[rName] == 0 {
```
gshaibi, line 58: Just out of curiosity - what was the trigger for this change?

pkg/scheduler/api/resource_info/resource_requirment.go
```diff
@@ -153,10 +154,15 @@ func (r *ResourceRequirements) DetailedString() string {
...
 	for rName, rQuant := range r.scalarResources {
-		if rName == v1.ResourceEphemeralStorage || rName == v1.ResourceStorage {
+		switch rName {
+		case v1.ResourceEphemeralStorage, v1.ResourceStorage:
 			rQuant = rQuant / int64(MemoryToGB) // convert from milli-bytes to GB
+			messageBuilder.WriteString(fmt.Sprintf(", %s: %v (GB)", rName, rQuant))
+		case v1.ResourcePods:
+			messageBuilder.WriteString(fmt.Sprintf(", %s: %v", rName, rQuant))
+		default:
+			messageBuilder.WriteString(fmt.Sprintf(", %s: %v", rName, rQuant))
```
gshaibi: If pods and default cases are the same, can we use the default only?

pkg/scheduler/gpu_sharing/gpuSharing.go
```diff
@@ -12,15 +12,15 @@ import (
 	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/log"
 )
 
-type nodeGpuForSharing struct {
+type NodeGpuForSharing struct {
```
gshaibi: Why do we need it public?

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -262,4 +261,69 @@ func (pp *predicatesPlugin) evaluateTaskOnPredicates(
...
+	}
+
+	return nil
+}
+
+// willCreateNewGpuGroup determines if allocating this task will create a new GPU group
+// (and thus require a new reservation pod).
+func (pp *predicatesPlugin) willCreateNewGpuGroup(task *pod_info.PodInfo, node *node_info.NodeInfo) bool {
+	if pp.ssn == nil {
+		return false
```
gshaibi: Maybe we should return here `true` to be on the safe side?

pkg/scheduler/api/resource_info/base_resources.go
itsomri: In tests, there are DeepEqual comparisons that don't see resource=0 the same as not having the resource in the map. I don't really like adding this deletion but couldn't see a better option

pkg/scheduler/plugins/predicates/predicates.go
itsomri: Fixed

pkg/scheduler/gpu_sharing/gpuSharing.go
itsomri: Made private again

pkg/scheduler/api/resource_info/resource_requirment.go
itsomri: Fixed

pkg/scheduler/api/resource_info/resource_requirment.go
itsomri: done

PR #987, opened 2026-02-11
fix(scheduler): count MIG resources in preemption/reclaim GPU totals

boshu2:
# Draft PR — NVIDIA/KAI-Scheduler (base: v0.12)

## Title
fix(scheduler): count MIG resources in preemption/reclaim GPU totals

## Description

MIG-only pods request GPU via MIG scalar resources (for example, `nvidia.com/mig-3g.20gb`) while `Resource.GPUs()` reflects only non-MIG `nvidia.com/gpu`. In the preemption/reclaim pipeline, three call sites used `GPUs()` and treated MIG-only pods as 0 GPUs, which breaks GPU quota/preemption behavior for MIG workloads.

This PR switches those accounting paths to use the summed GPU total (`GetSumGPUs()`), so MIG-only requests are properly counted. It also adds targeted unit tests covering MIG-only pending/victim accounting to prevent regressions.

## Related Issues

Fixes #

## Checklist

- [x] Self-reviewed

## Breaking Changes

None.

## Additional Notes

- Intentional narrow scope: this PR addresses preempt/reclaim accounting only. Other `GPUs()` call sites (for example in topology-related code paths) are out of scope and can be tracked separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 - Fixed a bug in ray gang scheduling where not all worker groups' minMember would be respected [#962](https://github.com/NVIDIA/KAI-Scheduler/pull/962) [itsomri](https://github.com/itsomri)
+- Fixed a bug where MIG-only pods were treated as 0 GPUs in preemption/reclaim accounting.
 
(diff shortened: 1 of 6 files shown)


boshu2:
Closing for now: our in-cluster ocppoc validation on 2026-02-11 showed the tested MIG preemption scenario works on both v0.12.10 and this patch, so we do not currently have a demonstrated behavioral delta to justify merging this change.

PR #1001, opened 2026-02-12
Allow Grove plugin to ignore PodGangs with specific annotation

sanjaychatterjee:
## Description

This PR is to ensure that Grove PodGrouper plugin will ignore PodGangs which contain a `grove.io/ignore` annotation. With the upcoming Grove scheduler backend, Grove will start creating KAI PodGroups directly. Hence this specific Grove plugin needs to ignore all Grove PodGangs in future. This is a temporary measure to allow a smooth transition to the new scheduling backend framework in Grove.

## Related Issues

Fixes #

## Checklist

- [ ] Self-reviewed

### DIFF
--- a/pkg/podgrouper/podgrouper/plugins/grove/grove_grouper.go
+++ b/pkg/podgrouper/podgrouper/plugins/grove/grove_grouper.go
@@ -24,6 +24,10 @@ const (
 	labelKeyPodGangName = "grove.io/podgang"
 )
 
+const (
+	annotationKeyIgnorePodGang = "grove.io/ignore"
+)
+
 type GroveGrouper struct {
 	client client.Client
 	*defaultgrouper.DefaultGrouper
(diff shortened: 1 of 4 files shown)


pkg/podgrouper/podgrouper/plugins/grove/grove_grouper.go
```diff
@@ -66,6 +70,9 @@ func (gg *GroveGrouper) GetPodGroupMetadata(
 	if err != nil {
 		return nil, err
 	}
+	if _, ok := podGang.GetAnnotations()[annotationKeyIgnorePodGang]; ok {
+		return nil, fmt.Errorf("podgang %s/%s annotated to be ignored", pod.Namespace, podGangName)
```
romanbaron: Returning an error will cause the reconcile loop to re-queue the event and attempt to handle it later again. 

I think you need to change this area:
```https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/podgrouper/pod_controller.go#L113```

So it will return:
```
return ctrl.Result{}, nil
```

PR #1010, opened 2026-02-15
feat(operator): support plugin and action overrides in SchedulingShard spec

gshaibi:
## Related Issues

Fixes #

## Checklist

- [ ] Self-reviewed

## Summary by CodeRabbit

* **New Features**
  * Scheduler plugins and actions are now configurable per shard with support for enabling/disabling, priority ordering, and argument overrides.
  * Built-in plugins and actions are automatically configured with sensible defaults based on placement strategies (binpack/spread).
  * Custom plugins and actions can be added and integrated with priority-based ordering.

* **Tests**
  * Added comprehensive test coverage for plugin/action configuration, defaults, customization, and scheduling strategies.


pkg/apis/kai/v1/schedulingshard_types.go
```diff
@@ -67,11 +99,155 @@ type SchedulingShardSpec struct {
...
 	s.PlacementStrategy = common.SetDefault(s.PlacementStrategy, &PlacementStrategy{})
 	s.PlacementStrategy.SetDefaultWhereNeeded()
+
+	s.setDefaultPlugins()
+	s.setDefaultActions()
+}
+
+// Default priorities preserve the current hardcoded ordering.
+// Higher priority = runs first. Spaced by 100.
+var defaultPluginPriorities = map[string]int{
```
itsomri, line 134: Doesn't this add another place we need to update when adding plugins? Why not consolidate this with the default scheduler conf?

pkg/apis/kai/v1/schedulingshard_types.go
```diff
@@ -26,8 +28,38 @@ import (
...
+	// the default arguments for the plugin.
+	// +kubebuilder:validation:Optional
+	Arguments map[string]string `json:"arguments,omitempty"`
+}
+
+// ActionConfig allows overriding action settings in the scheduler configuration.
+type ActionConfig struct {
+	// Enabled controls whether this action is active. Defaults to true.
+	// +kubebuilder:validation:Optional
+	Enabled *bool `json:"enabled,omitempty"`
```
itsomri, line 55: What do you think about having negative or zero priority as "disabled" instead?

itsomri (approved):

pkg/apis/kai/v1/schedulingshard_types.go
gshaibi: I personally prefer the explicitness of the `enabled` field, though that’s mostly a matter of personal taste.

pkg/apis/kai/v1/schedulingshard_types.go
gshaibi: A similar map already existed before my change on https://github.com/NVIDIA/KAI-Scheduler/blob/ea08ecf01b0d4e2b3158b44df8acefcb3268cde3/pkg/operator/operands/scheduler/resources_for_shard.go#L150.
I’m not sure how frequently we add or remove plugins, and if we think consolidation is worthwhile, I’d suggest handling that in a separate PR.

PR #1035, opened 2026-02-18
fix: skip runtimeClassName injection when gpuPodRuntimeClassName is empty

yuanchen8911:
## Summary

When `gpuPodRuntimeClassName` is set to empty string (`""`), the admission webhook should not inject `runtimeClassName` into GPU pods. Currently, even with an empty value, the `Mutate` function still proceeds to evaluate the pod and may set `runtimeClassName` to an empty string.

This fix adds an early return in `RuntimeEnforcement.Mutate()` when `gpuPodRuntimeClassName` is empty, completely skipping the runtimeClassName injection.

## Problem

With GPU Operator v25.10.0+, `nvidia` is configured as the default containerd runtime. KAI scheduler's admission webhook injecting `runtimeClassName: nvidia` triggers the `management.nvidia.com` CDI management path, causing pod startup failures:

```
Error: failed to inject CDI devices: unresolvable CDI devices
management.nvidia.com/gpu=GPU-<UUID>: unknown
```

The `--gpu-pod-runtime-class-name` flag help text already documents "Set to empty string to disable", but the `Mutate` function did not check for this case.

## Changes

- `runtime_enforcement.go`: Add early return when `gpuPodRuntimeClassName` is empty
- `runtime_enforcement_test.go`: Add test case for empty string behavior

## Test plan

- [x] Existing unit tests pass
- [x] New test case verifies GPU pods are not mutated when `gpuPodRuntimeClassName` is empty

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Implemented block-level segmentation for pytorchjobs [#938](https://github.com/NVIDIA/KAI-Scheduler/pull/938) [itsomri](https://github.com/itsomri)
 
 ### Fixed
+- Fixed admission webhook to skip runtimeClassName injection when gpuPodRuntimeClassName is empty [#1035](https://github.com/NVIDIA/KAI-Scheduler/pull/1035)
 - Fixed helm uninstall does not remove webhooks [#959](https://github.com/NVIDIA/KAI-Scheduler/pull/959) [faizan-exe](https://github.com/faizan-exe)
 - Fixed security vulnerability where PodGang could reference pods in other namespaces, preventing cross-namespace manipulation
 - Fixed pod controller logging to use request namespace/name instead of empty pod object fields when pod is not found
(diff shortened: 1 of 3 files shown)


yuanchen8911:
Thanks for the review. The empty string is preserved through the operator pipeline because `SetDefault` only replaces `nil`, not empty string:

```go
// pkg/apis/kai/v1/common/set_default.go
func SetDefault[T any](target *T, value *T) *T {
    if target == nil {
        return value
    }
    return target  // *string("") is not nil, so it's kept
}
```

Full flow when user sets `admission.gpuPodRuntimeClassName: ""`:

1. Helm (after PR #972) renders `gpuPodRuntimeClassName: ""` in ConfigMap
2. Operator deserializes → `*string` pointing to `""` (not nil)
3. `SetDefaultsWhereNeeded` at `admission.go:70` → `SetDefault` keeps `""` because target is not nil
4. `buildArgsList` at `resources.go:362` → `config.GPUPodRuntimeClassName` is non-nil, so it passes `--gpu-pod-runtime-class-name ""`
5. **This fix** → `Mutate()` sees empty string, returns early, no `runtimeClassName` injection

Without this fix, step 5 would still proceed to evaluate the pod and call `setRuntimeClass(pod, "")`, which sets `runtimeClassName` to an empty string pointer rather than leaving it unset.

The `cmd/admission/app/options.go:70` default (`constants.DefaultRuntimeClassName`) only applies when the flag is not provided at all — since the operator explicitly passes the flag via `buildArgsList`, the pflag default is overridden.

PR #1058, opened 2026-02-24
feat(podgrouper): lws subgrouping

davidLif:
## Description

Implemented subgroup segmentation (with topology segment definitions) for leaderworkerset. Segmentation topology definitions are expected to be in the pod workers template annotations.

## Related Issues

Fixes #

## Checklist

- [x] Self-reviewed

### DIFF
--- a/pkg/podgrouper/podgrouper/plugins/leaderworkerset/lws_grouper.go
+++ b/pkg/podgrouper/podgrouper/plugins/leaderworkerset/lws_grouper.go
@@ -145,18 +145,18 @@ func calcLeaderReadyMinAvailable(pod *v1.Pod, fallbackSize int32) int32 {
 }
 
 func (lwsg *LwsGrouper) buildSubGroups(lwsJob *unstructured.Unstructured, pod *v1.Pod, replicasSize int) ([]*podgroup.SubGroupMetadata, error) {
-	subGroupPolicy, err := getSubGroupPolicy(lwsJob, replicasSize)
+	segmentationPolicy, err := getSegmentationPolicy(lwsJob, replicasSize)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get sub group policy for LWS %s/%s: %w", lwsJob.GetNamespace(), lwsJob.GetName(), err)
 	}
-	if subGroupPolicy == nil {
+	if segmentationPolicy == nil {
 		return buildSubGroupsWithoutSegmentation(replicasSize, pod), nil
 	}
 	topologyConstraints, err := getSegmentTopologyConstraints(pod, lwsJob)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get segment topology constraints for LWS %s/%s: %w", lwsJob.GetNamespace(), lwsJob.GetName(), err)
 	}
-	return buildSubGroupsWithSegmentation(subGroupPolicy, topologyConstraints, replicasSize, pod)
+	return buildSubGroupsWithSegmentation(segmentationPolicy, topologyConstraints, replicasSize, pod)
 }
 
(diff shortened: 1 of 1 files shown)


pkg/podgrouper/podgrouper/plugins/leaderworkerset/lws_grouper.go
```diff
@@ -121,73 +126,317 @@ func (lwsGrouper *LwsGrouper) getStartupPolicy(lwsJob *unstructured.Unstructured
...
 	}
+	return ""
+}
 
-	return &podgroup.SubGroupMetadata{
-		Name:           subGroupLeader,
+func getSubGroupPolicy(lwsJob *unstructured.Unstructured, replicasSize int) (*lws.SubGroupPolicy, error) {
+	segmentSizeInt64, foundSegmentDefinition, err := unstructured.NestedInt64(lwsJob.Object, "spec", "leaderWorkerTemplate", "subGroupPolicy",
+		"subGroupSize")
+	segmentSize := int(segmentSizeInt64)
```
SiorMeir: if an err occurers, will you be able to execute this?

pkg/podgrouper/podgrouper/plugins/leaderworkerset/lws_grouper.go
```diff
@@ -121,73 +126,317 @@ func (lwsGrouper *LwsGrouper) getStartupPolicy(lwsJob *unstructured.Unstructured
...
+		podGroupMetadata.MinAvailable = groupSize
+	}
+
+	return nil
+}
+
+func (lwsGrouper *LwsGrouper) buildSubGroups(lwsJob *unstructured.Unstructured, pod *v1.Pod, replicasSize int) ([]*podgroup.SubGroupMetadata, error) {
+	subGroupPolicy, err := getSubGroupPolicy(lwsJob, replicasSize)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get sub group policy for LWS %s/%s: %w", lwsJob.GetNamespace(), lwsJob.GetName(), err)
```
SiorMeir, line 150: wouldn't we want to fall back to no-segments if we have in issue with getting the policy?

pkg/podgrouper/podgrouper/plugins/leaderworkerset/lws_grouper.go
```diff
@@ -121,73 +126,317 @@ func (lwsGrouper *LwsGrouper) getStartupPolicy(lwsJob *unstructured.Unstructured
...
+	value, found, _ := unstructured.NestedString(workerTemplate, "metadata", "annotations", key)
+	if found {
+		return value
 	}
+	return ""
+}
 
-	return &podgroup.SubGroupMetadata{
-		Name:           subGroupLeader,
+func getSubGroupPolicy(lwsJob *unstructured.Unstructured, replicasSize int) (*lws.SubGroupPolicy, error) {
```
SiorMeir: maybe its worth adding a comment explaining the logic here?

pkg/podgrouper/podgrouper/plugins/leaderworkerset/lws_grouper.go
davidLif: No. If we encoutred an error while reading the policy, we have a problem understanding parts of the leaderWorkerSet object. In that case, it would be better to fail.

pkg/podgrouper/podgrouper/plugins/leaderworkerset/lws_grouper.go
davidLif: Fixed. This line will execute only if the error is nil

pkg/podgrouper/podgrouper/plugins/leaderworkerset/lws_grouper.go
davidLif: Renamed function and added a comment

PR #1060, opened 2026-02-24
fix(operator): add Ready condition and fix Reconciling condition for kstatus compatibility

### LINKED ISSUE #1042: Helm install with --wait times out on SchedulingShard CR readiness
### Description

When installing KAI scheduler via Helm with `--wait`, the install can time out waiting for the `SchedulingShard` CR to reach Ready status, even though all pods start successfully within seconds.

```
Error: resource not ready, name: default, kind: SchedulingShard, status: InProgress
context deadline exceeded
```

After the timeout, `helm list` shows the release as `failed`, but all KAI pods are running and the SchedulingShard becomes Ready shortly after. A subsequent `helm upgrade --reuse-values` succeeds immediately.

### Root Cause

The KAI chart creates a `SchedulingShard` CR as a regular resource (not a hook). With `--wait`, Helm monitors all resources including CRs with status conditions. The kai-operator needs to:
1. Start and become ready
2. Detect the SchedulingShard
3. Create child deployments (scheduler, binder, admission, etc.)
4. Wait for children to become ready
5. Update SchedulingShard status to Ready

On fresh installs (especially with image pulls on tainted nodes), this sequence can take longer than Helm expects, causing the `--wait` check to time out.

### Observed Behavior

- SchedulingShard created at `13:31:04Z`
- All conditions became True by `13:31:15Z` (~11 seconds)
- But Helm had already marked the release as `failed` due to timeout during initial reconciliation

### Environment

- KAI Scheduler: v0.12.14
- Kubernetes: v1.34
- Platform: EKS with tainted GPU nodes (H100)
- Install method: `helm upgrade --install kai-scheduler oci://ghcr.io/nvidia/kai-scheduler/kai-scheduler --version v0.12.14 --wait --timeout 10m`

### Suggested Fix

Consider one of:
1. **Don't include SchedulingShard in Helm's `--wait` tracking** — for example, create it as a post-install hook so Helm doesn't block on its readiness
2. **Mark the shard as Ready faster** — update status to Ready before waiting for all child deployments, since the operator will continue reconciling regardless
3. **Add `helm.sh/hook` annotations** to the SchedulingShard so Helm handles it separately from the main resource set

### Workaround

Skip `--wait` when installing KAI scheduler, or retry `helm upgrade` on failure.

gshaibi:
## Description

The `Reconciling` condition on SchedulingShard and Config CRs was permanently set to `True` and never cleared to `False` after reconciliation completed. This caused [kstatus](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/README.md) (used by Helm 4's `--wait` flag, ArgoCD, Flux, etc.) to permanently report the resources as `InProgress`, making `helm install --wait` time out even though all pods were running.

### Root cause

kstatus uses the ["abnormal-true" convention](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/README.md#conditions): if a `Reconciling` condition exists with `status: True`, the resource is considered `InProgress`. The operator set `Reconciling=True` at the start of reconciliation but never set it to `False` upon completion.

### Changes

**Fix `Reconciling` condition lifecycle:**
- `UpdateStartReconcileStatus` sets `Reconciling=True` (start of reconciliation) — unchanged behavior
- `ReconcileStatus` now sets `Reconciling=False` (end of reconciliation) — **new**

**Add `Ready` condition:**
- New `Ready` condition that mirrors the `Available` condition (`IsAvailable()` check)
- kstatus uses `Ready=True` as a fallback signal that a CR is fully reconciled
- Set as part of `ReconcileStatus`, before clearing `Reconciling`

**New condition constants:**
- `ConditionTypeReady` ("Ready")
- `Reconciling`, `Ready`, `NotReady` condition reasons

### kstatus behavior after fix

| Phase | `Reconciling` | `Ready` | kstatus result |
|-------|--------------|---------|----------------|
| Start of reconcile | `True` | (unchanged) | `InProgress` |
| End, system available | `False` | `True` | `Current` |
| End, system not available | `False` | `False` | `InProgress` |

## Related Issues

Fixes #1042

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. Existing conditions are preserved; only new conditions and status transitions are added.

## Additional Notes

- Both `SchedulingShard` and `Config` CRs benefit from this fix since they share the same `StatusReconciler` code path.
- The `Reconciling=False` condition is set **last** in `ReconcileStatus`, ensuring kstatus sees `InProgress` for the entire duration that other conditions are being updated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 bin/
+.claude/
 .idea/
 charts/
 cover.out
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,10 +32,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Added scale test environment setup script and updated service monitors for KAI scheduler [#1031](https://github.com/NVIDIA/KAI-Scheduler/pull/1031)
 - Implemented subgroups for leaderworkerset [#1046](https://github.com/NVIDIA/KAI-Scheduler/pull/1046) [davidLif](https://github.com/davidLif) 
 - Added discovery data to snapshot for more accurate debugging [#1047](https://github.com/NVIDIA/KAI-Scheduler/pull/1047) [itsomri](https://github.com/itsomri)
-- Implemented subgroup segmentation (with topology segment definitions) for leaderworkerset [#1058](https://github.com/NVIDIA/KAI-Scheduler/pull/10586) [davidLif](https://github.com/davidLif) 
+- Implemented subgroup segmentation (with topology segment definitions) for leaderworkerset [#1058](https://github.com/NVIDIA/KAI-Scheduler/pull/10586) [davidLif](https://github.com/davidLif)
 
 ### Fixed
 - Fixed operator status conditions to be kstatus-compatible for Helm 4 `--wait` support: added `Ready` condition and fixed `Reconciling` condition to properly transition to false after reconciliation completes [#1060](https://github.com/NVIDIA/KAI-Scheduler/pull/1060)
+- Fixed a bug where the node scale adjuster would not check if a pod was unschedulable before creating a scaling pod leading to unnecessary node scaling [#1094](https://github.com/NVIDIA/KAI-Scheduler/pull/1094) [slaupster](https://github.com/slaupster)
 - Fixed admission webhook to skip runtimeClassName injection when gpuPodRuntimeClassName is empty [#1035](https://github.com/NVIDIA/KAI-Scheduler/pull/1035)
 - Fixed topology-migration helm hook failing on OpenShift due to missing `kai-topology-migration` service account in the `kai-system` SCC [#1050](https://github.com/NVIDIA/KAI-Scheduler/pull/1050)
 - Fixed a bug where queue status did not reflect its podgroups resources correctly [#1049](https://github.com/NVIDIA/KAI-Scheduler/pull/1049)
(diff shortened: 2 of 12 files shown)


pkg/operator/controller/status_reconciler/status_reconciler.go
```diff
@@ -188,3 +204,35 @@ func (r *StatusReconciler) getDependenciesFulfilledCondition(ctx context.Context
 		LastTransitionTime: metav1.Now(),
 	}
 }
+
+func (r *StatusReconciler) getReadyCondition(ctx context.Context, gen int64) metav1.Condition {
```
enoodle, line 208: would it not be easier to call `getAvailableCondition` and change the type?

gshaibi: Its possible, we'll need to change the Type, Reason and Message, but I'm not sure I prefer it... But that is my personal taste and I don't have strong feelings about it, so if you want we can change it.

enoodle: The logic is the same - check isAvailable function and then return a condition with the results. The only difference is the constants that are used in the condition.

gshaibi: OK, I'll change it

gshaibi: done

enoodle (dismissed):

PR #1062, opened 2026-02-24
refactor(scheduler): adding resouce vector class

enoodle:
## Description

Introduces ResourceVector ([]float64) and ResourceVectorMap types as the foundation for vectorized resource accounting in the scheduler, as described in the [vectorization design doc](https://github.com/NVIDIA/KAI-Scheduler/blob/main/docs/developer/designs/vectorizing-resources/README.md)

## Checklist

- [x] Self-reviewed

### DIFF
--- a/pkg/scheduler/api/resource_info/resource_vector.go
+++ b/pkg/scheduler/api/resource_info/resource_vector.go
@@ -1,0 +1,323 @@
+// Copyright 2025 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+package resource_info
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/NVIDIA/KAI-scheduler/pkg/common/constants"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+type ResourceVector []float64
+
+type ResourceVectorMap struct {
+	resourceNames []string
+	namesToIndex  map[string]int
+}
+
+// NewResourceVectorMap creates a new ResourceVectorMap initialized with base resources.
+func NewResourceVectorMap() *ResourceVectorMap {
+	result := &ResourceVectorMap{
+		resourceNames: []string{},
+		namesToIndex:  make(map[string]int),
+	}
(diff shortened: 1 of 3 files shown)


pkg/scheduler/api/resource_info/resource_vector.go
```diff
@@ -0,0 +1,323 @@
...
+	vec := NewResourceVector(indexMap)
+	gpuIdx := indexMap.GetIndex(constants.GpuResource)
+	if gpuIdx >= 0 {
+		vec.Set(gpuIdx, 1)
+	}
+	return vec
+}
+
+// NewResourceVectorWithValues is a convenience function for tests only.
+func NewResourceVectorWithValues(milliCPU, memory, gpus float64, indexMap *ResourceVectorMap) ResourceVector {
```
SiorMeir, line 52: If this is for tests, Shouldn't we move this into a util in tests?
Claude code might get confused and use this :)

pkg/scheduler/api/resource_info/resource_vector.go
```diff
@@ -0,0 +1,323 @@
...
+		return float64(rQuant.MilliValue())
+	}
+	return float64(rQuant.Value())
+}
+
+func isGpuResource(resourceName string) bool {
+	return strings.HasSuffix(resourceName, constants.GpuResource)
+}
+
+func normalizeResourceName(resourceName string) string {
```
SiorMeir, line 196: Isn't there a scenario where we'll want to use this?

enoodle: this function is used: https://github.com/NVIDIA/KAI-Scheduler/pull/1062/changes#diff-1192ca31804645d442aa39831143571cb77e98586dfd6aa3d583c67886e0d845R70

pkg/scheduler/api/resource_info/resource_vector.go
enoodle: It will be used in the next PR, all of this is currently not called. If Claude is using functions please review the results.

SiorMeir (approved):

PR #1102, opened 2026-02-28
fix(chart): propagate global.nodeSelector to Config CR and operator deployments

yuanchen8911:
## Description

`global.nodeSelector` is defined in `values.yaml` but was not propagated to the Config CR template or applied to operator deployment pod specs. This prevents users from constraining KAI scheduler components to specific nodes using `nodeSelector`.

This fix adds `nodeSelector` support following the same pattern already used by `global.tolerations`:

1. **CRD schema** (`kai.scheduler_configs.yaml`): Added `nodeSelector` field to `spec.global` as `additionalProperties: string` map
2. **Go type** (`pkg/apis/kai/v1/global.go`): Added `NodeSelector map[string]string` field to `GlobalConfig` struct
3. **Helm template** (`kai-config.yaml`): Added conditional `nodeSelector` block using `toYaml`, matching the tolerations pattern
4. **Operator** (`pkg/operator/operands/common/common.go`): Applied `NodeSelector` from Config CR to deployment pod specs, alongside existing `Tolerations` propagation

## Related Issues

Fixes https://github.com/NVIDIA/KAI-Scheduler/issues/1101

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. This is a purely additive change. Existing deployments without `global.nodeSelector` configured will be unaffected.

## Additional Notes

The change mirrors the existing `tolerations` pattern exactly, ensuring consistency across the codebase.

### DIFF
--- a/deployments/kai-scheduler/crds/kai.scheduler_configs.yaml
+++ b/deployments/kai-scheduler/crds/kai.scheduler_configs.yaml
@@ -3244,6 +3244,12 @@ spec:
                     description: NodePoolLabelKey is the label name by with to filter
                       nodes, pods and other resources that the scheduler is watching
                     type: string
+                  nodeSelector:
+                    additionalProperties:
+                      type: string
+                    description: NodeSelector defines node selector for KAI operators
+                      & services
+                    type: object
                   openshift:
                     description: Openshift configures the operator to install on Openshift
                     type: boolean
--- a/pkg/apis/kai/v1/zz_generated.deepcopy.go
+++ b/pkg/apis/kai/v1/zz_generated.deepcopy.go
@@ -217,6 +217,13 @@ func (in *GlobalConfig) DeepCopyInto(out *GlobalConfig) {
 		*out = make([]string, len(*in))
 		copy(*out, *in)
 	}
+	if in.NodeSelector != nil {
+		in, out := &in.NodeSelector, &out.NodeSelector
+		*out = make(map[string]string, len(*in))
+		for key, val := range *in {
+			(*out)[key] = val
+		}
+	}
 	if in.Tolerations != nil {
 		in, out := &in.Tolerations, &out.Tolerations
 		*out = make([]corev1.Toleration, len(*in))


enoodle:
@yuanchen8911 Looks good, you need to run `make validate` for this to pass CI. Can you also update the changelog please?

yuanchen8911:
@enoodle Thanks for the review! I've addressed both items:

1. **`make validate`** — ran it locally, committed the regenerated code (16 files — mock updates, client regeneration). CI should now pass.
2. **Changelog** — already included in the initial commit under `## [Unreleased] ### Added`.

Could you approve the CI check when you get a chance? Thanks!

yuanchen8911:
> @enoodle Thanks for the review! I've addressed both items:
> 
> 1. **`make validate`** — ran it locally, committed the regenerated code (16 files — mock updates, client regeneration). CI should now pass.
> 2. **Changelog** — already included in the initial commit under `## [Unreleased] ### Added`.
> 
> Could you approve the CI check when you get a chance? Thanks!

deployments/kai-scheduler/crds/kai.scheduler_configs.yaml
```diff
@@ -9,7 +9,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.16.1
+    controller-gen.kubebuilder.io/version: v0.17.3
```
enoodle: I think this change explains the issues that the CI has. either downgrade to v0.16.1 or update the CI to use the newer version.

yuanchen8911:
@enoodle Good catch — the `make validate` diff was caused by my local Go 1.25 generating different mock output than the CI's Go 1.24 (mockgen parameter naming: `arg4` vs `state`). 

I've reverted the mock file to match upstream and kept only the CRD/RBAC/deepcopy regeneration. The `Validate & Test` failure should be resolved now.

The `Validate PR Title` failure — do you need a specific title format (e.g., conventional commits)?

enoodle:
:pray: the title fails probably because `helm` was not in the approve list for packages.

PR #1104, opened 2026-03-01
fix(binder): check active BindRequests before deleting reservation pods

### LINKED ISSUE #1103: Reservation pod premature deletion race condition during concurrent fraction binding
# Reservation Pod Premature Deletion Race Condition

## Problem Statement

The binder's resource reservation sync logic (`syncForPods`) prematurely deletes GPU reservation pods when concurrent bind operations race with informer cache propagation. This manifests as a flaky E2E test failure where reservation pods disappear during the binding of fractional GPU pods to the same node.

### The Deletion Logic

`syncForPods` in `pkg/binder/binding/resourcereservation/resource_reservation.go` decides to delete a reservation pod when it finds no "fraction pods" (user workload pods in Running/Pending phase) for that GPU group:

```go
for gpuGroup, reservationPod := range reservationPods {
    if _, found := fractionPods[gpuGroup]; !found {
        // DELETE reservation pod
    }
}
```

The pod list comes from `syncForGpuGroupWithLock`, which queries the informer cache using label selectors (`runai-gpu-group=<group>`).

### The Race

There are 3 independent triggers that call `SyncForGpuGroup`:

| Trigger | When | Location |
|---|---|---|
| `Binder.Bind()` → `SyncForNode()` | Start of pod binding | `binder.go:44` |
| `BindRequestReconciler.deleteHandler()` | BindRequest deleted | `bindrequest_controller.go:217` |
| `PodReconciler.syncReservationIfNeeded()` | Pod deleted/completed | `pod_controller.go:122` |

With `MaxConcurrentReconciles=10`, multiple BindRequests are reconciled in parallel. The race occurs as follows:

1. **Thread A** (binding Pod 1 to Node A, GPU group X):
   - `ReserveGpuDevice()` creates reservation pod for group X
   - `updatePodGPUGroup()` patches Pod 1 with label `runai-gpu-group=X`
   - The label patch is sent to the API server → API server persists it → Watch event is generated → informer cache receives the event

2. **Thread B** (binding Pod 2 to Node A, same GPU group X):
   - `Bind()` calls `SyncForNode("NodeA")`
   - `SyncForNode` lists pods on the node, finds the reservation pod (which has `runai-gpu-group=X`), extracts group X
   - Calls `SyncForGpuGroup("X")` → `syncForGpuGroupWithLock`
   - Lists pods with label `runai-gpu-group=X` from **informer cache**
   - **Cache lag**: Pod 1's label patch has not propagated to the cache yet
   - Finds: reservation pod (in binder namespace) → `reservationPods["X"]`
   - Does NOT find Pod 1 (label not in cache yet) → `fractionPods["X"]` is empty
   - **Deletes the reservation pod** ← BUG

The `gpuGroupMutex` provides per-group serialization but does NOT prevent this race because Thread B's `SyncForNode` observes the reservation pod (just created, already in cache) but not Pod 1's updated labels (patch not yet in cache).

### Why the Cache Shows the Reservation Pod but Not the Label

- The reservation pod is a **new object** (CREATE event) — informer receives it quickly
- Pod 1's label is an **update to an existing object** (UPDATE/PATCH event) — may be in a different event batch or processed after the CREATE
- The informer processes events sequentially per type, but CREATE and UPDATE events for different objects can have different propagation times

## Reproduction

The race can be reproduced with the following tests:

- **E2E stress test**: `366b3be2` (`test/e2e/suites/allocate/stress/stress_test.go`) — binds 32 fractional GPU pods concurrently to a single node and verifies no reservation pods are lost.
- **Integration test (envtest)**: `66b41b87` (`pkg/env-tests/reservation_race_scale_test.go`) — reproduces the race at scale with 32 fraction pods binding concurrently.

## Proposed Fix

Before deleting a reservation pod due to missing fraction pods, check if there are any **active (non-succeeded, non-failed) BindRequests** that reference this GPU group. If any exist, skip the deletion — a binding operation is in progress and the fraction pod label hasn't propagated yet.

See [`docs/developer/designs/fix-reservation-pod-race/design.md`](https://github.com/NVIDIA/KAI-Scheduler/blob/erez/fix-binder-reservation-pod-sync-race/docs/developer/designs/fix-reservation-pod-race/design.md) for the full design document.

enoodle:
## Description

Fixes a race condition where the binder's resource reservation sync logic (`syncForPods`) prematurely deletes GPU reservation pods during concurrent fractional GPU binding. The race occurs because the informer cache may show a newly-created reservation pod before the corresponding user pod's `runai-gpu-group` label patch has propagated, causing the sync to conclude no fraction pods exist for the GPU group and delete the reservation.

The fix adds a check for active (non-succeeded, non-failed) BindRequests referencing the GPU group before deleting a reservation pod. Since BindRequests are created before binding starts and persist until completion, they serve as a durable intent signal that prevents premature cleanup during the cache lag window.

## Related Issues

Fixes #1103

## Checklist

- [x] Self-reviewed

## Breaking Changes

None.

### Design Document
See [`docs/developer/designs/fix-reservation-pod-race/design.md`](https://github.com/NVIDIA/KAI-Scheduler/blob/erez/fix-binder-reservation-pod-sync-race/docs/developer/designs/fix-reservation-pod-race/design.md) for the full analysis.

### Verification
- **With fix**: 32/32 reservation pods survive concurrent binding
- **Without fix** (cherry-picked test to `origin/main`): 31/32 survive — 1 reservation pod is prematurely deleted, confirming the race

### Commits
| Commit | Description |
|---|---|
| `aafe0f2d` | Initial envtest reproducing the race at scale |
| `366b3be2` | E2E stress test for concurrent fraction binding |
| `66b41b87` | Stabilized integration test with mock device plugin |
| `ee29169f` | The fix: check active BindRequests before deletion |


pkg/binder/binding/resourcereservation/resource_reservation.go
```diff
@@ -208,6 +220,26 @@ func (rsc *service) syncForPods(ctx context.Context, pods []*v1.Pod, gpuGroupToS
...
+	if err := rsc.kubeClient.List(ctx, bindRequestList); err != nil {
+		return false, fmt.Errorf("failed to list BindRequests: %w", err)
+	}
+
+	for _, br := range bindRequestList.Items {
+		if br.Status.Phase == schedulingv1alpha2.BindRequestPhaseSucceeded ||
+			br.Status.Phase == schedulingv1alpha2.BindRequestPhaseFailed {
+			continue
+		}
+		if slices.Contains(br.Spec.SelectedGPUGroups, gpuGroup) {
```
davidLif, line 237: Maybe you should add an index to easily get all binding requests for the gpu group?

enoodle: Good call!

enoodle:
Using the indexer introduces the same bug :)

I will find some other way to avoid listing all the bind requests.

PR #1110, opened 2026-03-01
feat(podgrouper): Allow segment size annotation at the workerTemplate level 

davidLif:
## Description

Allow segment size annotation at the workerTemplate level (where the other segmentation labels are).

## Related Issues

Fixes #

## Checklist

- [x] Self-reviewed

### DIFF
--- a/pkg/scheduler/actions/common/solvers/accumulated_scenario_filters/node_affinities/node_affinities.go
+++ b/pkg/scheduler/actions/common/solvers/accumulated_scenario_filters/node_affinities/node_affinities.go
@@ -1,0 +1,194 @@
+// Copyright 2025 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+package node_affinities
+
+import (
+	"context"
+	"maps"
+	"slices"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	ksf "k8s.io/kube-scheduler/framework"
+	k8sframework "k8s.io/kubernetes/pkg/scheduler/framework"
+
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/actions/common/solvers/scenario"
+	"github.com/NVIDIA/KAI-scheduler/pkg/scheduler/api/common_info"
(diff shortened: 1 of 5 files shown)


pkg/podgrouper/podgrouper/plugins/leaderworkerset/lws_grouper.go
```diff
@@ -212,13 +212,33 @@ func getWorkerAnnotationValue(pod *v1.Pod, workerTemplate map[string]interface{}
...
+	if err != nil {
+		return nil, err
+	}
+	policy := lws.SubGroupPolicyTypeLeaderWorker
+	if found {
+		policy = lws.SubGroupPolicyType(policyStr)
+	}
+	return &lws.SubGroupPolicy{
+		SubGroupSize: segmentSize,
+		Type:         ptr.To(lws.SubGroupPolicyType(policy)),
```
gshaibi, line 237: nit
```suggestion
		Type:         ptr.To(policy),
```

pkg/podgrouper/podgrouper/plugins/leaderworkerset/lws_grouper.go
```diff
@@ -212,13 +212,33 @@ func getWorkerAnnotationValue(pod *v1.Pod, workerTemplate map[string]interface{}
...
+	}, nil
+}
+
+func getSegmentSize(lwsJob *unstructured.Unstructured, replicasSize int) (*int32, error) {
```
gshaibi, line 241: If you want, you can consider splitting it to `getSegmentSize` and `validateSegmentSize`, so the scope nesting can be reduced (return instead of nesting in `getSegmentSize`)

gshaibi (approved):

pkg/podgrouper/podgrouper/plugins/leaderworkerset/lws_grouper.go
davidLif: This is required - there is a type for the pinter

pkg/podgrouper/podgrouper/plugins/leaderworkerset/lws_grouper.go
davidLif: I don't think the validation has enough logic to separate it

PR #1123, opened 2026-03-02
fix(scheduler): Do not include resources with a count of 0.

### LINKED ISSUE #1120: Allow non-MIG jobs to run on node where MIG is disabed, but set to MigStrategy='mixed'

I was using MIG in our cluster before I decided to try KAI's ability to use cooperative GPU sharing.  I've since configured the nodes with "all-disabled" config, and disabled MIG. However:

1. Kubelet does not remove resources from a node, even if they are no longer reported by device plugins (https://github.com/kubernetes/kubernetes/issues/92396), so I still have mig device counts on the nodes, just the device count is 0. 
```
Capacity:
  cpu:                     192
  ephemeral-storage:       300602976Ki
  hugepages-1Gi:           0
  hugepages-2Mi:           0
  memory:                  1583754476Ki
  nvidia.com/gpu:          8
  nvidia.com/mig-1g.24gb:  0
  nvidia.com/mig-2g.48gb:  0
  pods:                    110
  rdma/ib0:                63
Allocatable:
  cpu:                     192
  ephemeral-storage:       292426574824
  hugepages-1Gi:           0
  hugepages-2Mi:           0
  memory:                  1583754476Ki
  nvidia.com/gpu:          8
  nvidia.com/mig-1g.24gb:  0
  nvidia.com/mig-2g.48gb:  0
  pods:                    110
  rdma/ib0:                63
```
2. The mig strategy was set to "mixed" (but shouldn't matter since it was disabled on the node).

On KAI version 0.12.15, the scheduler will not place non-MIG jobs on the node, reporting:
```
non MIG jobs cant run on a node with MigStrategy='mixed'
```

It would be useful to have the scheduler be able to place non-MIG jobs on nodes where MIG is disabled, regardless of MIG strategy.

### Why is this needed?

It is my understanding that gpu-operator applies the same MIG strategy to all nodes. If a cluster is to be setup with some MIG nodes (in mixed mode) and some MIG disabled nodes, the MIG strategy will be set to "mixed" across all nodes.

It would be helpful if KAI did not look only at MIG strategy to determine if a non-MIG pod can be placed, but also whether or not MIG is disabled.  

I believe this is partially implemented.  The `non MIG jobs cant run on a node with MigStrategy='mixed'` message is not only gated by a check to MIG strategy, but also on whether the `ni.IsMIGEnabled()` (whether the node is MIGed) returns true.  Unfortunately, `IsMIGEnabled` only checks to see if a MIG resource is reported on the node, but does not check device counts.

I think this could be fixed by either:
1. Adjust `IsMIGEnabled` so that an explicit check that count is > 0 is performed
2. Adjust `ResourceFromResourceList` from pkg/scheduler/api/resource_info/resource_info.go so that resources with rQuant == 0 are skipped.  This seems to be the best approach.  Elsewhere in resource_info.go resources in scalarResources are explicitly deleted if count is 0, just not in ResourceFromResourceList.

I'm happy to submit a short PR with either approaches if you approve.  I do not know how to easily test the changes, however, and do not have access to test clusters.

I'm able to work around this for now since I was able to disable MIG across the whole cluster (setting strategy to "none").

dougnd:
## Description

Kubelet does not remove resources, even when they are no longer advertised from a device plugin (see: https://github.com/kubernetes/kubernetes/issues/92396). 

This change prevents node resources with count 0 from impacting KAI's scheduling logic.

## Related Issues

Fixes #1120.

## Checklist

- [X] Self-reviewed


SiorMeir (approved):

dougnd:
Thanks for the review!  Let me know if there is anything else I should change.  

I noticed the Performance benchmarks seem to fail on my PR.  In the benckmark workflow (.github/workflows/benchmark.yaml), it looks like it:

1. Checks out PR branch using:
```
/usr/bin/git checkout --progress --force refs/remotes/pull/1123/merge
```

2. Runs benchmark
3. Checks out main
4. Runs benchmark
5. Checkout PR branch (restore), this time using:
```
git checkout fix-mig-detection
```
This fails with `error: pathspec 'fix-mig-detection' did not match any file(s) known to git`. 


Perhaps [this](https://github.com/NVIDIA/KAI-Scheduler/blob/main/.github/workflows/benchmark.yaml#L59) line could be changed to:
```
run: git checkout ${{ github.ref || github.sha }}
```

I believe this would then run the following instead:
```
git checkout  refs/remotes/pull/1123/merge
```

Or is this failure is expected for PRs from forks?

enoodle:
Thanks, We should fix this action on another PR.

Issue #1151, opened 2026-03-05
no way to unset admission.gpuPodRuntimeClassName during helm install in 0.13.0


julienmancuso:
### What happened?

https://github.com/NVIDIA/KAI-Scheduler/issues/1097 was opened because we kai-scheduler defaulted to nvidia.
to unset the pod runtime class name (to use the cluster default), a workaround was to set `admission.gpuPodRuntimeClassName=''` in 0.13.0-rc1

but in the latest 0.13.0, you can't do that anymore :
```
 Error creating pod : pods "trtllm-disagg-tp8-3-0-decode-0-decode-ldr-" is forbidden: pod rejected: RuntimeClass "''" not found
``` 

### Environment

- Kubernetes version 
- KAI Scheduler version 0.13.0
- Cloud provider or hardware configuration
- Tools that you are using KAI together with
- Anything else that is relevant

enoodle:
The way to set this may be changed to:
```
helm upgrade -i kai-scheduler oci://ghcr.io/nvidia/kai-scheduler/kai-scheduler -n kai-scheduler --create-namespace --set "global.gpuSharing=true" --version "v0.13.0" --set "admission.gpuPodRuntimeClassName="
```

I am looking into removing `nvidia` as the default.

keinanDaniel:
We will open another issue for the removal of the default, but this issue is done

PR #1161, opened 2026-03-08
fix(admission): remove nvidia as default gpuPodRuntimeClassName

cluster2600:
## Summary

- Changes the default `gpuPodRuntimeClassName` from `nvidia` to an empty string (disabled)
- Not every cluster has an `nvidia` RuntimeClass defined, and the current default causes pod creation failures on those setups
- Users who require the nvidia runtime class can still explicitly configure it via `--set admission.gpuPodRuntimeClassName=nvidia`

## Motivation

As raised in #1151, there is currently no straightforward way to unset the `gpuPodRuntimeClassName` during helm install in v0.13.0. The root cause is that `nvidia` is hardcoded as the default value in both the Helm chart (`values.yaml`) and the Go constant (`DefaultRuntimeClassName`). This is problematic for clusters that either don't use NVIDIA GPUs or have GPU support configured through the cluster's default runtime rather than a dedicated RuntimeClass.

## Changes

- `pkg/common/constants/constants.go`: Changed `DefaultRuntimeClassName` from `"nvidia"` to `""`
- `deployments/kai-scheduler/values.yaml`: Changed `gpuPodRuntimeClassName` default from `nvidia` to `""`
- `cmd/admission/app/options.go`: Updated the CLI flag help text to reflect the new default behaviour
- `pkg/admission/webhook/v1alpha2/runtimeenforcement/runtime_enforcement_test.go`: Updated test cases to use hardcoded `"nvidia"` rather than the constant, so they still excercise the configured runtime class injection path
- `CHANGELOG.md`: Documented the change under Unreleased > Fixed

## Test plan

- [x] Verified `go build ./...` completes successfully
- [x] Ran unit tests for `runtimeenforcement`, `admission`, `operator/operands/admission`, `binder/binding/resourcereservation`, and `binder/controllers` — all pass
- [x] Ran `helm lint` and `helm template` — chart renders `gpuPodRuntimeClassName: ""` as expected
- [ ] CI should confirm full test suite passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 - Updated resource enumeration logic to exclude resources with count of 0. [#1120](https://github.com/NVIDIA/KAI-Scheduler/issues/1120)
 - Fixed scheduling-constraints signature hashing for `Priority` and container `HostPort` by encoding full `int32` values, preventing byte-truncation collisions and flaky signature tests.
+- Changed default `gpuPodRuntimeClassName` from `nvidia` to empty string (disabled) so clusters without an `nvidia` RuntimeClass work out of the box [#1151](https://github.com/NVIDIA/KAI-Scheduler/issues/1151)
 
 ### Added
 - Added support for Ray subgroup topology-aware scheduling by specifying `kai.scheduler/topology`, `kai.scheduler/topology-required-placement`, and `kai.scheduler/topology-preferred-placement` annotations.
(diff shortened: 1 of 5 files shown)


enoodle:
> As raised in https://github.com/NVIDIA/KAI-Scheduler/issues/1151, there is currently no straightforward way to unset the gpuPodRuntimeClassName during helm install in v0.13.0.

That is not true, you can use `--set "admission.gpuPodRuntimeClassName="` .
As I wrote in the issue, I am looking into removing this as the default, but there are implications of this change with different versions of the GPU operator. I will create a PR in the future to make this update but currently there is a workaround.

cluster2600:
Ah fair enough, I didn't realise the empty string override worked for that. Thanks for clarifying!

Since you're planning to address this properly with the GPU operator implications in mind, I'll close this one — no point duplicating the effort.

Cheers

PR #1164, opened 2026-03-09
fix(admission): validate subgroup names are lowercase on PodGroup creation

### LINKED ISSUE #948: Validate subgroup names are lowercase on podgroup creation

cluster2600:
Closes #948.

## Problem

The scheduler internally lowercases subgroup parent names via `formatParentName()` in `pkg/scheduler/api/podgroup_info/subgroup_info/factory.go`, but this constraint was never validated at admission time. A user creating a PodGroup with mixed-case subgroup names (e.g. `Workers`) would have their resources accepted by the API but then silently hit parent-lookup failures during scheduling, which is quite confusing to diagnose.

## Solution

Added lowercase validation in the PodGroup validating webhook (`validateSubGroups`) that rejects:
- Subgroup names containing uppercase characters
- Parent references containing uppercase characters

Both `ValidateCreate` and `ValidateUpdate` paths are covered since they share the same validation function. This ensures the constraint is properly recognised and enforced before the resource reaches the scheduler.

## Changes

- `pkg/apis/scheduling/v2alpha2/podgroup_webhook.go` — added `strings.ToLower` checks for `SubGroup.Name` and `SubGroup.Parent` in `validateSubGroups()`
- `pkg/apis/scheduling/v2alpha2/podgroup_webhook_test.go` — updated existing test fixtures to use lowercase names (matching the new constraint), added three new test cases:
  - Uppercase subgroup name rejected
  - Mixed-case subgroup name rejceted
  - Uppercase parent reference rejected

## Test plan

- [x] All existing `TestValidateSubGroups` cases pass with lowercase fixture names
- [x] New validation cases cover uppercase name, mixed-case name, and uppercase parent

### DIFF
--- a/pkg/apis/scheduling/v2alpha2/podgroup_webhook.go
+++ b/pkg/apis/scheduling/v2alpha2/podgroup_webhook.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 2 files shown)


cluster2600:
Closing — upstream merged #1383 which adds kubebuilder validation patterns (`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`) on both `Name` and `Parent` fields in the CRD spec. This enforces lowercase at the API level, making the webhook-level validation here redundant. Cheers to @itsomri for landing it!

PR #1177, opened 2026-03-10
fix: allow device-plugin GPU requests on dual-mode nodes

yuanchen8911:
## Summary

Nodes running both the NVIDIA device-plugin (`nvidia.com/gpu` in allocatable) and the DRA GPU driver (ResourceSlices with `gpu.nvidia.com` driver) are incorrectly classified as DRA-only. This causes `PredicateByNodeResourcesType` to reject all device-plugin GPU pod requests with "device-plugin GPU requests cannot be scheduled on DRA-only nodes", even though the node can serve them via the device-plugin.

Additionally, `AddDRAGPUs` was called unconditionally on these dual-mode nodes, double-counting the same physical GPUs (device-plugin + DRA) in the allocatable pool, which could lead to GPU overcommitment.

This is common on clusters that deploy both `gpu-operator` (device-plugin) and `nvidia-dra-driver` simultaneously, such as EKS clusters with NVIDIA GPU Operator v25.10+ and DRA driver v25.12+.

## Changes

- Add `HasDevicePluginGPUs` field to `NodeInfo` to track whether a node advertises GPUs via the device-plugin
- Skip `AddDRAGPUs` on dual-mode nodes to prevent double-counting (same physical GPUs are advertised by both device-plugin and DRA)
- Update `PredicateByNodeResourcesType` to allow device-plugin GPU requests on dual-mode nodes (`HasDRAGPUs && HasDevicePluginGPUs`)
- Add predicate test for dual-mode node scheduling
- Add cluster snapshot test for dual-mode GPU accounting

## Assumptions

This change assumes that on dual-mode nodes, DRA ResourceSlices and device-plugin `nvidia.com/gpu` represent the **same physical GPUs**. If a future configuration exposes distinct GPU pools via each mechanism, the accounting policy would need revisiting.

## Test plan

- [x] `TestPredicateByNodeResourcesType_DRA`: DRA-only still blocked, device-plugin-only still allowed, dual-mode now allowed
- [x] `TestSnapshotNodesWithDRAGPUs`: dual-mode node reports 8 GPUs (not 16), `HasDevicePluginGPUs=true`
- [ ] Manual validation on EKS cluster with gpu-operator + nvidia-dra-driver (8x H100 node with both `nvidia.com/gpu: 8` and DRA ResourceSlices)

### DIFF
--- a/pkg/scheduler/api/node_info/node_info.go
+++ b/pkg/scheduler/api/node_info/node_info.go
@@ -98,6 +98,9 @@ type NodeInfo struct {
 
 	// HasDRAGPUs indicates GPUs were added via DRA ResourceSlices. Temporary fix - remove when device-plugin pods are supported on DRA nodes.
 	HasDRAGPUs bool
+	// HasDevicePluginGPUs indicates the node has GPUs advertised via the device-plugin (nvidia.com/gpu in allocatable).
+	// Used together with HasDRAGPUs to identify dual-mode nodes that support both device-plugin and DRA GPU requests.
+	HasDevicePluginGPUs bool
 
 	PodAffinityInfo pod_affinity.NodePodAffinityInfo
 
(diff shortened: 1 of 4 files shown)


enoodle:
I think that we would have preferred to implement something like https://github.com/kubernetes/enhancements/issues/5004 in KAI to handle this instead of this.
To my understanding if we have a node with both enabled they could both allocate the same device without notifying even if other devices are available. This means that we need to only allow one of them to handle the allocation and it should be the DRA driver because it has a richer API.

yuanchen8911:
Closing this PR based on feedback from the KAI scheduler and DRA driver teams.

**Context:** On our EKS cluster (K8s 1.35), we have both the device-plugin (`nvidia.com/gpu: 8` in allocatable) and the DRA driver (`gpuResourcesEnabledOverride=true`, ResourceSlices with `gpu.nvidia.com`) running on the same GPU nodes. Both DRA (ResourceClaim) and device-plugin (`nvidia.com/gpu: 1`) workloads work in practice — we ran CNCF conformance tests (gang scheduling, HPA) using `nvidia.com/gpu` and DRA validation using ResourceClaims on the same cluster.

However, the DRA team confirmed that device-plugin and DRA GPU allocation are incompatible on the same node — both systems advertise all GPUs independently, risking double-booking at the scheduler level. Our tests passed only because we didn't run both concurrently on the same node.

**The right path forward is KEP 5004** (`DRAExtendedResource` feature gate, K8s v1.35+), which will let the DRA driver honor `nvidia.com/gpu` extended resource requests alongside ResourceClaims, properly unifying accounting and obsoleting the need for the device-plugin. The KAI scheduler's existing check is intentionally conservative and correct for the current state of affairs.

For our immediate needs, we'll use KAI v0.12.17 (which predates the DRA-only node check) and plan to migrate fully to DRA once KEP 5004 reaches GA.

PR #1219, opened 2026-03-15
fix(scheduler): isolate DRA manager state from live informer updates

itsomri:
## Description

This PR Replaces the singleton DRA manager (backed by a live informer) with a session-scoped DRA manager that is isolated from concurrent informer updates during scheduling.

Previously, the DRA plugin's AssumeCache was backed by a live informer, meaning any API server update mid-session (e.g., binder writing allocation, user patching a claim) could silently corrupt assume state during scheduling simulations.
This caused ghost device entries in allocatedDevices and incorrect allocation decisions.

  Key changes:

  - Session-scoped DRA manager (dra_manager.go): Each scheduling session creates a fresh AssumeCache populated from a static snapshot of claims via a snapshotInformer (nil-event informer). No concurrent informer events can modify state during the session.
  - MutableDRAManager proxy: The upstream k8s DRA plugin stores SharedDRAManager() at init time. MutableDRAManager delegates to a swappable underlying manager, allowing the KAI DRA plugin to set the current
  session-scoped manager each session.
  - Removed restoreAllClaims(): No longer needed — fresh state each session eliminates divergent assume-cache state.
  - ResourceSliceTracker created once at plugin initialization, shared across sessions (read-only).
  - DRA objects in DataLister: ListResourceSlices() and ListDeviceClasses() added to the DataLister interface, used by snapshot plugin and cluster info snapshot.
  - Integration test fix: Claim status between test rounds is now read from the session-scoped DRA manager directly, instead of relying on OnSessionClose to sync back to ClusterInfo.

## Checklist

- [ ] Self-reviewed

### DIFF
--- a/pkg/scheduler/actions/integration_tests/integration_tests_utils/integration_tests_utils.go
+++ b/pkg/scheduler/actions/integration_tests/integration_tests_utils/integration_tests_utils.go
@@ -11,7 +11,6 @@ import (
 	. "go.uber.org/mock/gomock"
 	resourceapi "k8s.io/api/resource/v1"
 
-	"github.com/kai-scheduler/KAI-scheduler/pkg/common/k8s_utils"
 	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/actions/allocate"
 	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/actions/consolidation"
 	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/actions/preempt"
@@ -137,15 +136,10 @@ func runSchedulerOneRound(testMetadata *TestTopologyMetadata, controller *Contro
 	}
 	if len(testMetadata.TestDRAObjects.ResourceClaims) > 0 {
 		k8sPlugins := (*ssn).InternalK8sPlugins()
-		if k8sPlugins == nil {
+		if k8sPlugins == nil || k8sPlugins.SessionDRAManager == nil {
 			return
 		}
-		fh, ok := k8sPlugins.FrameworkHandle.(*k8s_utils.K8sFramework)
-		if !ok {
-			return
-		}
-		draManager := fh.SharedDRAManager()
-		claims, err := draManager.ResourceClaims().List()
+		claims, err := k8sPlugins.SessionDRAManager.ResourceClaims().List()
 		if err != nil {
 			log.InfraLogger.Errorf("Failed to list claims from DRA manager: %v", err)
 			return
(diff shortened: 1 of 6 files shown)


pkg/scheduler/plugins/dynamicresources/dynamicresources.go
```diff
@@ -57,46 +57,30 @@ func (drap *draPlugin) Name() string {
...
+	k8sPlugins := ssn.InternalK8sPlugins()
+	if k8sPlugins != nil && k8sPlugins.ResourceSliceTracker != nil {
+		drap.manager = k8s_utils.NewSessionDRAManager(
+			ssn.ClusterInfo.ResourceClaims,
+			k8sPlugins.ResourceSliceTracker,
+			k8sPlugins.InformerFactory,
+		)
+		if fh, ok := k8sPlugins.FrameworkHandle.(*k8s_utils.K8sFramework); ok {
+			fh.SetSharedDRAManager(drap.manager)
+		}
```
davidLif: Why not expose only a "getSessionDRAManager" function, and the internal logic will create a new manager if necessary?

pkg/scheduler/actions/integration_tests/integration_tests_utils/integration_tests_utils.go
```diff
@@ -134,20 +136,38 @@ func runSchedulerOneRound(testMetadata *TestTopologyMetadata, controller *Contro
...
+		for _, claimMetadata := range testMetadata.TestDRAObjects.ResourceClaims {
+			clusterClaim, ok := claimByName[claimMetadata.Namespace+"/"+claimMetadata.Name]
+			if !ok {
 				continue
 			}
-			clusterClaimStatus := clusterClaim.Status
-			if clusterClaimStatus.Allocation != nil || clusterClaimStatus.ReservedFor != nil || clusterClaimStatus.Devices != nil {
-				claim.ClaimStatus = clusterClaimStatus.DeepCopy()
+			status := clusterClaim.Status
+			if status.Allocation != nil || len(status.ReservedFor) > 0 {
```
davidLif: What about `clusterClaimStatus.Devices != nil `?

itsomri: Shouldn't really happen but added

pkg/scheduler/plugins/dynamicresources/dynamicresources.go
itsomri: I don't think it's a good enough reason to add to the already bloated session API. The proper way to solve this weird dependency is to remove dynamic resources completely from the k8s plugins, and only implement it in our own dynamicresources plugin (move the dynamic resources predicate from the generic k8s one to dra plugin). It's a bit of a refactor but I can give it a go if you want

davidLif: Can you give it a go with claude? If it's too long, I will approve

davidLif (approved):

PR #1223, opened 2026-03-17
fix(chart): auto-enable leader election when operator.replicaCount > 1

### LINKED ISSUE #1218: Auto-enable leader election when operator.replicaCount > 1

When operator.replicaCount is set to a value greater than 1, global.leaderElection should be automatically enabled (or the chart should fail with a validation error if it's
  explicitly set to false).

  Currently the chart defaults global.leaderElection: false and operator.replicaCount: 1 independently, with no coupling between them. This means a user can set operator.replicaCount:
  2 for HA without realizing they also need to set global.leaderElection: true, resulting in multiple operator instances concurrently reconciling the Config CR.

  Proposed behavior (either option):
  1. Auto-enable: If operator.replicaCount > 1, treat leaderElection as true regardless of the explicit value.
  2. Validation error: Add a helm template / install-time check that fails with a clear message like: "operator.replicaCount > 1 requires global.leaderElection to be true".

### Why is this needed?

 Running multiple operator replicas without leader election causes both instances to actively reconcile resources concurrently, which can lead to race conditions, conflicting updates,
   and resource flapping. This is a silent misconfiguration — nothing warns the user, and the chart templates successfully.

  Most Kubernetes operators built on controller-runtime default to leader election enabled and only allow disabling it for single-replica development setups. KAI's current behavior is
  the opposite and creates a footgun for anyone enabling HA.

david-gang:
## Description

When `operator.replicaCount` is set to a value greater than 1, leader election is now automatically enabled regardless of the `global.leaderElection` value. This prevents a silent misconfiguration where multiple operator replicas concurrently reconcile resources, causing race conditions and resource flapping.

## Related Issues

Fixes #1218

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. Single-replica deployments are unaffected. Multi-replica deployments that previously ran without leader election (a misconfiguration) will now correctly enable it.

## Additional Notes

- The fix uses Helm template functions to auto-enable leader election: `{{ or .Values.global.leaderElection (gt (int .Values.operator.replicaCount) 1) }}`
- Added 4 helm unit tests covering: default (single replica), explicit enable, auto-enable with replicaCount > 1, and combined scenario

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Users that have VPA installed on their cluster can now utilize it for proper vertical autoscaling
 - Added FOSSA scanning for the repository context. Scans will also be performed for submitted PRs. The results can be found [here](https://app.fossa.com/projects/custom%2B162%2Fgit%40github.com%3Akai-scheduler%2FKAI-Scheduler.git). [#1178](https://github.com/kai-scheduler/KAI-Scheduler/pull/1178) - [davidLif](https://github.com/davidLif) 
 - Added support for Ray subgroup topology-aware scheduling by specifying `kai.scheduler/topology`, `kai.scheduler/topology-required-placement`, and `kai.scheduler/topology-preferred-placement` annotations.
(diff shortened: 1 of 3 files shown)


enoodle (approved):

davidLif:
Hello @david-gang ,
Can you please resolve the merge conflict? After it, we will approve it again

david-gang:
> Hello @david-gang , Can you please resolve the merge conflict? After it, we will approve it again

done

PR #1229, opened 2026-03-17
fix(operator): prevent reconcile loop caused by status updates

cypres:
The kaiv1.Config object is in a constant reconcile loop in 0.13.0-0.13.1. I think it's related to the changes from #1060

Asked Claude to come up with a fix, this is what it found;

## Description

Commit 994f202a introduced a Reconciling condition that transitions True → False across each reconcile cycle:

  1. Start of reconcile: UpdateStartReconcileStatus patches Reconciling=True
  2. End of reconcile: deferred ReconcileStatus patches Reconciling=False

  Each status patch triggers a new Update event on the Config/SchedulingShard object, which re-queues a reconcile. The existing guard in reconcileCondition can't break this loop because it checks Status too — so False→True always looks like a genuine change and always patches.

  The result: an infinite patch → reconcile → patch → reconcile loop.

  ---
  The fix (hasReconcilingConditionForGeneration)

  UpdateStartReconcileStatus now checks whether a Reconciling condition already exists for the current generation before patching:

  func (r *StatusReconciler) UpdateStartReconcileStatus(...) error {
      if r.hasReconcilingConditionForGeneration(object) {
          return nil  // skip — we're in a status-triggered re-reconcile
      }
      // first reconcile for this generation: set Reconciling=True
      ...
  }

  Why this works:

  - Genuine spec change (e.g. generation goes 1→2): no Reconciling condition with ObservedGeneration=2 exists yet → proceeds, patches Reconciling=True ✓
  - Status-triggered re-reconcile (same generation): Reconciling condition was already set (to either True or False) in the previous cycle → guard fires, returns early, no patch → no new Update event → loop ends ✓
  - Cluster state change (e.g. a watched Deployment is deleted): still runs Deploy() and ReconcileStatus as normal — only UpdateStartReconcileStatus is skipped, so self-healing still works ✓

  The key distinction from the existing reconcileCondition guard: it only checks ObservedGeneration, not Status. So it fires whether the existing condition is True or False — exactly what's needed to break the alternating loop.

## Related Issues

Fixes #

## Checklist

- [X] Self-reviewed
- [X] Added/updated tests (if needed)
- [X] Updated documentation (if needed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Fixed rollback in scheduling simulations with DRA [#1168](https://github.com/NVIDIA/KAI-Scheduler/pull/1168) [itsomri](https://github.com/itsomri)
 - Fixed a potential state corruption in DRA scheduling simulations [#1219](https://github.com/kai-scheduler/KAI-Scheduler/pull/1219) [itsomri](https://github.com/itsomri)
 - Fixed operator reconcile loop caused by status-only updates triggering re-reconciliation. #1229 [cypres](https://github.com/cypres)
+- Fixed scheduler not starting on k8s clusters with DRA disabled, due to the ResourceSliceTracker not syncing. #1241 [cypres](https://github.com/cypres)
 
 ## [v0.13.0] - 2026-03-02
 ### Added
--- a/pkg/scheduler/k8s_internal/plugins/plugins.go
+++ b/pkg/scheduler/k8s_internal/plugins/plugins.go
@@ -6,9 +6,11 @@ package plugins
 import (
 	"context"
 
+	featureutil "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	resourceslicetracker "k8s.io/dynamic-resource-allocation/resourceslice/tracker"
+	"k8s.io/kubernetes/pkg/features"
 	"k8s.io/kubernetes/pkg/scheduler/apis/config"
 	k8sframework "k8s.io/kubernetes/pkg/scheduler/framework"
 	k8splfeature "k8s.io/kubernetes/pkg/scheduler/framework/plugins/feature"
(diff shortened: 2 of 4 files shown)


enoodle:
Hi @cypres !
Thanks for the fix
I see that it happens every ~30 seconds, I wonder why it is but I am not sure that it is due to status updates.
Can you also add some kind of tests to verify that this is indeed fixed and will not break again?

cypres:
> Hi @cypres ! Thanks for the fix I see that it happens every ~30 seconds

For us it happens 5 times per second, basically as fast as the loop can run.

> Can you also add some kind of tests to verify that this is indeed fixed and will not break again?

Yes, let me see what I can do.

cypres:
This test is broken; https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.13.2/pkg/operator/controller/schedulingshard_controller_test.go#L72-L83
It's not reporting a failure atm because there is no test suite for it, so it's not being run.

So when I try to add tests for the reconcile loop, I get stuck in a rabbit hole. I verified the changes in a dev cluster.

Can you help get this PR across the finish line @enoodle ?

My findings are troubling, basically the behavior we are seeing is that Kai Scheduler does not schedule workloads on Kubernetes clusters where DRA is not enabled. We are running this on a Kubernetes 1.31 cluster.

So `k api-versions | grep resource.k8s.io` returns nothing, and this causes the scheduler to not start, it just logs.
> kai-scheduler-default E0318 05:40:39.983023       1 reflector.go:205] "Failed to watch" err="failed to list *v1.ResourceSlice: the server could
not find the requested resource (get resourceslices.resource.k8s.io)" logger="UnhandledError" reflector="pkg/mod/k8s.io/client-go@v0.34.3/tools/cache/reflector.go:290"
type="*v1.ResourceSlice"

enoodle:
Hi @cypres 
Lets split it to two PRs that each solve a different problem that we have :sweat: 
I think the resource slice tracker one is very straight forward - we can merge it immediatly. We need to open a different PR for that and also update the changelog with this fix.
About the failing tests that are not running - I will take a look at this later today.

cypres:
Got it, split into #1241

cypres:
I refactored this PR to get rid of the predicate hack @enoodle

PR #1361, opened 2026-03-31
docs: replace invalid duration "1w" with "168h" in time-based-fairshare examples

JheSue:
## Problem

The time-based-fairshare documentation uses `windowSize: 1w` as an example value.
However, Go's `time.ParseDuration` does not support the `w` (week) unit —
only `ns`, `us`, `ms`, `s`, `m`, and `h` are valid.

When a user follows the documentation and sets `windowSize: 1w` in their
`SchedulingShard` CR, the kai-operator fails to list `SchedulingShard` resources
and enters a crash loop:
```
failed to list *v1.SchedulingShard: time: unknown unit "w" in duration "1w"
```

## Fix

Replace `1w` with the equivalent `168h` (7 × 24h) in all documentation examples.
The semantic meaning is preserved — one week is still the default.

## Changes

- `docs/time-based-fairshare/README.md` — replace `windowSize: 1w` with `windowSize: 168h`
- `examples/time-based-fairshare/README.md` — update default value in parameter table

## How to reproduce

1. Follow the time-based-fairshare setup guide
2. Set `windowSize: 1w` as shown in the docs
3. Observe kai-operator crash-looping with the above error

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,6 +15,9 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    # Enable only security updates for existing releases
+    # https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#open-pull-requests-limit-
+    open-pull-requests-limit: 0
     groups:
       kubernetes:
         patterns:


enoodle (approved):

enoodle:
Thanks @JheSue ! :pray:

itsomri:
Hi, this is probably a bug in code, not docs: we were supposed to use prometheus' duration type which allows for `w` https://pkg.go.dev/github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring@v0.88.0/v1#Duration

I'll take a look when I have the time

JheSue:
Hi, thanks for the pointer! I've investigated the issue and implemented a fix.
 I used model.Duration from github.com/prometheus/common/model, which is already an int64 type with built-in UnmarshalJSON/UnmarshalYAML that supports the full Prometheus duration format including w and d.
I've opened a new PR with the fix : #1371 
Feel free to close this one.

PR #1372, opened 2026-04-01
fix: Add resourceclaims/binding RBAC for DRA granular status authorization

praveen0raj:
## Summary

- Add `resourceclaims/binding` with `update` and `patch` verbs to the binder ClusterRole
- Updated both the kubebuilder RBAC marker in source code and the auto-generated manifest

## Why

Starting in Kubernetes v1.36, modifying `status.allocation` and `status.reservedFor` on a ResourceClaim requires additional permission on the synthetic `resourceclaims/binding` subresource. Without this change, KAI-Scheduler's binder will get 403 Forbidden when binding claims to nodes.

The existing `resourceclaims/status` permission is still required and kept as-is.

## Files changed

- `pkg/binder/controllers/bindrequest_controller.go` — added `+kubebuilder:rbac` marker
- `deployments/kai-scheduler/templates/rbac/binder.yaml` — added generated RBAC rule

## References

- kubernetes/kubernetes#138149
- kubernetes/kubernetes#134947
- https://github.com/kubernetes/website/pull/55070

## Test plan

- [x] Deploy ClusterRole without fix → `kubectl auth can-i update resourceclaims/binding` returns `no`
- [x] Deploy ClusterRole with fix → `kubectl auth can-i update resourceclaims/binding` returns `yes`
- [x] Existing `resourceclaims/status` permission unchanged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Fixed `windowSize` field in `SchedulingShard` CR to support Prometheus duration format (e.g. `1w`, `7d`). Previously, using `windowSize: 1w` as shown in the documentation caused the kai-operator to crash-loop with `time: unknown unit "w" in duration "1w"`.
 - Race condition where `SyncForGpuGroup` could prematurely delete reservation pods when the informer cache had not yet propagated GPU group labels on recently-bound fraction pods. The binder now checks for active BindRequests referencing the GPU group before deleting a reservation pod.
 - Fixed non-preemptible multi-device GPU memory jobs being allowed to exceed their queue's deserved GPU quota. The per-node quota check now correctly accounts for all requested GPU devices. [#1369](https://github.com/kai-scheduler/KAI-Scheduler/issues/1369)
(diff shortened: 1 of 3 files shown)


enoodle:
@praveen0raj what would happen if we use this rbac on k8s v1.31 for example? Will it fail? Should we guard this with some "if k8s version is  >=1.36"?

praveen0raj:
@enoodle 

No, it won't fail, and no version guard is needed.                                                                                                                                                                                                                                                                     
                                                                             
Kubernetes RBAC rules are declarative and permissive — a ClusterRole can reference resources that don't exist on the cluster. The API server simply ignores rules for unknown resources/subresources. It won't error when you apply the manifest, and it won't cause any runtime issues.                                                            
   
So on v1.31:                                                                                                                                                                                                                                                                                                                                        
  - The ClusterRole applies successfully with the extra rule                 
  - The resourceclaims/binding rule is effectively a no-op (nothing checks it)                                                                                                                                                                                                                                                                        
  - All existing permissions work exactly as before
                                                                                                                                                                                                                                                                                                                                                      
On v1.36+:                                                                 
  - The rule becomes active and grants the binder the new required permission                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                      
This is the same pattern used throughout the Kubernetes ecosystem — projects routinely ship RBAC for newer API features without version guards. Adding conditional logic would actually be worse: it would add Helm template complexity for zero practical benefit.

enoodle:
@praveen0raj Thank you for explaining that.
Could you please sign-off on the commit and add a changelog line?

praveen0raj:
@enoodle I have made changes. Thanks.

praveen0raj:
@enoodle There was some conflict in changelog. you have to approve it again. Thank you

praveen0raj:
Thanks  . it needs two approvals.

I am following the PR status until its get merged .  :)

Issue #1373, opened 2026-04-02
proposal(scheduler): cross-dimensional resource balance scoring to prevent GPU fragmentation


0x-auth:
### What you would like to be added?

A cross-dimensional resource balance scoring plugin that detects and prevents resource fragmentation at scheduling time, rather than fixing it after the fact.

Currently KAI supports bin-packing (fill nodes fully) and spread (distribute evenly), but neither considers the *shape* of resource usage across dimensions. A node can be 95% on GPU-Memory but 15% on GPU-Compute — bin-packing sees it as "mostly full" and spread sees it as "partially used." Both are wrong. That node has stranded compute.

The proposed plugin scores nodes by measuring how much *more balanced* they become after placing a pod, using cosine alignment between the pod's request vector and the node's free capacity vector, combined with Shannon entropy reduction.

### Why is this needed?

In GPU clusters running mixed inference workloads, resource fragmentation happens across dimensions, not just within them:

- Large model serving (LLaMA 70B, DeepSeek): VRAM full, GPU compute ~20% → compute is stranded
- Many small models (batch inference): GPU compute saturated, VRAM ~30% used → memory is stranded
- Mixed CPU+GPU nodes: CPU 90%, GPU 40% → GPU capacity wasted

This is the "jagged cluster" problem described in #1311, but caught at scheduling time instead of after the fact. Prevention > cure.

With the vectorized resource representation landing (#1353), adding a vector-based scoring plugin becomes natural — the infrastructure is already there.

### Proposed approach

Score each node using a 6D resource vector `[CPU, Memory, GPU-Compute, GPU-Memory, IOPS, Network]`:

```
alignment     = cosine_similarity(pod_request, node_free_capacity)
exhaustion    = φ × entropy_reduction(node_before, node_after) + utilization_delta
leak_penalty  = 0.15 × count(stranded_dimensions)

score = φ × alignment + exhaustion - leak_penalty + headroom × 0.3
```

Where φ = 1.618 (golden ratio) provides parameter-free weighting — no per-cluster tuning needed.

**Benchmark** (synthetic, 50 nodes × 500 pods): 13% improvement in resource balance vs LeastAllocated, zero stranded nodes in 4/5 scenarios. Scoring latency: ~115ns/op, zero allocations.

[...]

Issue #1353, opened 2026-03-30
refactor(scheduler): vectorize resource representation for performance at scale


enoodle:
## Overview

This issue tracks the multi-PR effort to migrate KAI Scheduler's resource representation from discrete struct-based types (`Resource`, `ResourceRequirements`) to vector-based representations (`ResourceVector`, `ResourceVectorMap`).

Design document: [Vectorizing Resource Representation](docs/developer/designs/vectorizing-resources/README.md)

## Motivation

Scheduler performance degrades significantly at scale:
- Full scheduling cycles take 3–4 minutes for 1000 nodes and can exceed 20 minutes for larger clusters
- The bottleneck is the scenario filtering phase of reclaim/preemption: resource comparisons iterate field-by-field over individual nodes and resources
- No bulk operations exist; every comparison allocates intermediate data

Vectorizing resources enables:
1. **Vectorized comparisons** — resources are flat `[]float64` arrays indexed by a shared `ResourceVectorMap`, enabling goroutine-based parallelism and future SIMD-style operations
2. **Efficient bin-packing** — normalized metrics (sum or dominant resource) for fast node sorting
3. **Scenario filtering acceleration** — pre-computed vectors allow quick feasibility checks without repeated struct traversal

Baseline benchmarks (before vectorization, 10-run average):

| Benchmark | Configuration | Time |
|-----------|---------------|------|
| ReclaimLargeJobs | 200 nodes | 816 ms |
| ReclaimLargeJobs | 500 nodes | 8.97 s |
| ReclaimLargeJobs | 1000 nodes | >40 min (timed out) |

### Phase 1: Type Introduction ✅
Introduce `ResourceVector` and `ResourceVectorMap` types with full arithmetic (`Add`, `Sub`, `LessEqual`, `Clone`, `SetMax`, `IsZero`) and conversion functions.

### Phase 2: Vector Map Generation ✅
Extend `ClusterInfoSnapshot` to build a `ResourceVectorMap` from cluster state. Pass shared map to all nodes and pods during session initialization.

### Phase 3: Pod & Node Info Vectorization ✅
Migrate read-path methods in `NodeInfo`, `PodInfo`, `PodGroupInfo`, and plugins to use vectors. Write-path methods update both Resource struct and vector fields during transition.

### Phase 4: Resource Struct Deprecation and Removal 🔄
[...]

Issue #1311, opened 2026-03-23
Scheduled consolidation/defragmentation of cluster


ghost:
### What you would like to be added?

A "cron job" that on a regular basis gracefully shuffles workloads to different nodes to create more fully empty nodes.

### Why is this needed?

Inference workloads or other workloads often can leave a cluster jagged with no room for a full node.

enoodle:
Hi @NVJCameron 
Great suggestion, it is  a good way to solve this problem of fragmentation.
How is your use case benefiting this more than the current consolidation action that moves jobs only when it can make space for pending ones? Will you be able to more easily scale nodes down if they were completely free of running jobs?

ghost:
Consolidation takes time and adds unwanted latency to scaling actions of additional inference pods. Also need the consolidation action to be soft, allowing replica count to be maintained instead of hard.

enoodle:
@NVJCameron Can you elaborate more on your 2nd point? What does "soft" mean exactly? Do you want it to be able to more "non-preemptible" jobs as long as you know that they will come up soon on a different node?

ghost:
Apologies likely not using the right term. I want to make sure the number of deployed pods doesn't go below the threshold - so something like a "maxSurge". I want to bring up a new pod and make sure its healthy before removing the older one.

enoodle:
@NVJCameron this sounds like a different feature request, I suggest that you create another issue for the "max surge" consolidation.

About the original feature request of different consolidation action - can you explain more about your use case? The types/sizes of inference workloads and maybe an example of how such defragmentation would happen.

ghost:
We often have individual containers (eg vLLM/SGLang) running with GPU x1,x2 x4, or x8 sizes. They scale up and down depending on usage of the model. The scaling down often leaves nodes partially used (think of two x4 config'd containers with only one being removed). The issue then arises when a bunch of x4 instances are half filling a bunch of nodes and a container workload that needs x8 GPU nodes then needs to scale, which then it can't because there are a bunch of half full nodes. Does that makes sense?

Opened the forked idea into another ticket here based on your previous comment: https://github.com/kai-scheduler/KAI-Scheduler/issues/1352

PR #1374, opened 2026-04-03
docs(scheduler): cross-dimensional balance scoring design

0x-auth:
## Description

Design proposal for a **cross-dimensional balance scoring plugin** (Lambda-G V3) that prevents resource fragmentation by combining post-placement variance minimization (60%) with cosine alignment (20%) as a directional tiebreaker.

Scope: CPU, Memory, and GPU-Memory (GPU-Compute not currently supported in KAI, as noted by @enoodle in #1373). The scoring function generalizes to any number of dimensions without modification.

**Key result:** V3 wins all 5 heterogeneous scenarios vs BalancedAllocation, DominantResource, LeastAllocated, and MostAllocated. 23% fewer stranded nodes, 53% less wasted capacity.

| Scenario | BalancedAlloc | DominantRes | **Lambda-G V3** |
|---|---|---|---|
| Mixed GPU — AI Workload | 78.7 | 79.4 | **81.8** |
| GPU — Inference Heavy | 81.3 | 76.6 | **81.9** |
| GPU — Training Heavy | 79.4 | 78.9 | **82.8** |
| CPU + Few GPUs | 72.3 | 68.8 | **74.7** |
| Scale (60n×300p) | 74.1 | 73.8 | **76.7** |

Design doc as requested in https://github.com/kai-scheduler/KAI-Scheduler/issues/1373#issuecomment-4181000472

cc @enoodle

## Related Issues

Refs #1373
Refs #1311 (cluster defragmentation)
Refs #1353 (vectorized resource representation)

## Checklist

- [x] Self-reviewed
- [ ] Added/updated tests (implementation PR)
- [x] Updated documentation (design doc added)

## Breaking Changes

None. This is a proposal-only PR — no code changes.

## Additional Notes

- Benchmark reproduction: https://github.com/0x-auth/lambda-g-auditor
- Koordinator also has an open proposal for the same algorithm: koordinator-sh/koordinator#2839
- Sub-microsecond scoring latency per node, zero heap allocations


docs/developer/designs/cross-dimensional-balance-scoring/README.md
```diff
@@ -0,0 +1,264 @@
+# Cross-Dimensional Resource Balance Scoring
+
+**Author:** Abhishek Srivastava (@0x-auth)  
+**Date:** 2026-04-03  
+**Status:** Proposal  
+**Issue:** [#1373](https://github.com/kai-scheduler/KAI-Scheduler/issues/1373)
+
+## Summary
+
+Add a scoring plugin that prevents resource fragmentation by combining post-placement variance minimization with cosine alignment between the pod's resource request vector and the node's free capacity vector. The plugin measures how much *more balanced* a node becomes after placing a pod, then uses directional alignment as a tiebreaker to steer pods toward nodes where their resource shape fills the gap.
```
enoodle, line 10: Why not add to the current nodeplacement plugin another option instead of creating a new plugin?

docs/developer/designs/cross-dimensional-balance-scoring/README.md
```diff
@@ -0,0 +1,264 @@
...
+|---|---|---|---|---|---|
+| LLaMA 70B serving | ~95% | ~20% | ~30% | ~60% | GPU-Compute |
+| Batch small-model inference | ~30% | ~85% | ~70% | ~20% | GPU-Memory, RAM |
+| CPU preprocessing pipeline | ~0% | ~0% | ~90% | ~25% | RAM, all GPU |
+
+This is the "jagged cluster" problem described in [#1311](https://github.com/kai-scheduler/KAI-Scheduler/issues/1311), but caught at scheduling time rather than repaired after the fact.
+
+### Why Now
+
+The vectorized resource representation landing in [#1353](https://github.com/kai-scheduler/KAI-Scheduler/issues/1353) makes a vector-based scoring plugin natural — the infrastructure for multi-dimensional resource vectors is already there.
```
enoodle, line 39: We could have implemented it before.
I don't think the design should be time based, it is just needed as is.

docs/developer/designs/cross-dimensional-balance-scoring/README.md
```diff
@@ -0,0 +1,264 @@
...
+
+In GPU clusters running mixed inference workloads, cross-dimensional fragmentation is common:
+
+| Workload Type | GPU-Mem | GPU-Compute | CPU | RAM | Stranded Resource |
+|---|---|---|---|---|---|
+| LLaMA 70B serving | ~95% | ~20% | ~30% | ~60% | GPU-Compute |
+| Batch small-model inference | ~30% | ~85% | ~70% | ~20% | GPU-Memory, RAM |
+| CPU preprocessing pipeline | ~0% | ~0% | ~90% | ~25% | RAM, all GPU |
+
+This is the "jagged cluster" problem described in [#1311](https://github.com/kai-scheduler/KAI-Scheduler/issues/1311), but caught at scheduling time rather than repaired after the fact.
```
enoodle, line 35: I am not sure that it is the same as what is described there, or that the problem there can be solved by node scoring at all.
The issue is relevant with only whole GPU allocation - the problem there is that after small jobs stop running "holes" are created.

docs/developer/designs/cross-dimensional-balance-scoring/README.md
```diff
@@ -0,0 +1,264 @@
...
+
+### Why Now
+
+The vectorized resource representation landing in [#1353](https://github.com/kai-scheduler/KAI-Scheduler/issues/1353) makes a vector-based scoring plugin natural — the infrastructure for multi-dimensional resource vectors is already there.
+
+## Proposal
+
+### Scope
+
+This proposal covers CPU, Memory, and GPU-Memory as scoring dimensions. GPU-Compute requests are not currently supported in KAI-Scheduler (as noted by @enoodle in [#1373](https://github.com/kai-scheduler/KAI-Scheduler/issues/1373)), but the scoring function generalizes to any number of dimensions without modification.
```
enoodle, line 45: I am not sure that we have to note it.. why not just support all the dimensions of the resource request vector?

docs/developer/designs/cross-dimensional-balance-scoring/README.md
```diff
@@ -0,0 +1,264 @@
...
+      − strand_penalty
+```
+
+#### Component 1: Post-Placement Variance Score (weight: 0.6)
+
+```
+after_frac[i] = (node.used[i] + pod.req[i]) / node.capacity[i]   for each active dimension
+mean          = average(after_frac)
+variance      = average((after_frac[i] - mean)² for each i)
+variance_score = max(0, (1.0 - variance × 4) × 100)
```
enoodle, line 65: why multiply the variance by 4? I see that in the k8s implementation they use the standard deviation instead (https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/noderesources/balanced_allocation.go#L248)

enoodle (commented):
I think that we will need to do a long overdue overhaul of the node scoring system in KAI-Scheduler for this to work, but we can design this separately in the mean time.

docs/developer/designs/cross-dimensional-balance-scoring/README.md
0x-auth: Good point — integrating as an additional scoring mode within the existing `nodeplacement` plugin makes more architectural sense than a standalone plugin. The scoring function is stateless and fits the same interface. I'll restructure the proposal as a new scoring option within `nodeplacement` rather than a separate plugin.

docs/developer/designs/cross-dimensional-balance-scoring/README.md
0x-auth: Agreed — removed the "Why Now" framing. The cross-dimensional balance problem exists independent of any particular PR timeline. The scoring function is useful as-is.

docs/developer/designs/cross-dimensional-balance-scoring/README.md
0x-auth: You're right — #1311 is about defragmenting after holes are created (descheduler-side), while this proposal prevents fragmentation at scheduling time. They're complementary but different problems. I'll decouple the reference to avoid confusion.

docs/developer/designs/cross-dimensional-balance-scoring/README.md
0x-auth: Makes sense. The scoring function already generalizes to any number of dimensions — it operates on whatever resource vector the node exposes. I limited the scope statement to be conservative, but there's no technical reason not to support the full vector. I'll update the doc to remove the dimension restriction.

docs/developer/designs/cross-dimensional-balance-scoring/README.md
0x-auth: The `× 4` scaling was chosen empirically to normalize variance into a [0, 100] score range. However, switching to standard deviation (as K8s does in BalancedAllocation) would be more consistent with the ecosystem and easier to reason about.

I'll align with the upstream stddev approach for the stability base, ensuring consistency with K8s `BalancedAllocation`, while retaining the Cosine Alignment component as the high-dimensional tiebreaker to address cross-dimensional fragmentation that stddev alone misses.

Reference: [kubernetes/balanced_allocation.go#L248](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/noderesources/balanced_allocation.go#L248)

0x-auth:
Understood — happy to align this proposal with whatever scoring architecture emerges from the overhaul. The core algorithm (variance base + alignment tiebreaker + penalties) is independent of the plugin framework, so it can adapt to whatever scoring interface KAI moves to. In the meantime I'll update the design doc based on your inline feedback.

0x-auth:
Reviving this now that v0.16.8 is out and the scheduler side is clearly moving. This design proposal has been open since April and I would like to unblock it with a concrete, minimal scope rather than let it sit.

@davidLif @enoodle picking up the two concerns you raised earlier, since I think they can both be resolved by narrowing scope:

**1. GPU sub-resource dependency (davidLif's point).** You were right that scoring across `gpu`, `gpu-compute`, and `gpu-memory` is premature while there is no user-facing way to request compute and memory separately. So I want to drop the GPU sub-dimensions from the initial version entirely. The scoping that survived the discussion (and that @Whatsonyourmind confirmed as the common case) is plain **CPU / Memory** cross-dimensional balance:

> pods with 2 CPU + 4Gi routed to nodes with 32 CPU + 128Gi available create the same fragmentation pattern, just on different units.

This needs no new resource fields, no GPU sub-resource support, and reduces to a well-defined 2D scoring function that generalizes to N dimensions later when the GPU sub-resources land.

**2. Where it fits.** The algorithm (variance base + alignment tiebreaker + penalties) is independent of the plugin framework, so it can either ship as a standalone Score plugin now or fold into the scoring overhaul enoodle mentioned. I do not want to build against the wrong interface.

So one specific question to unblock direction:

**Would a minimal CPU/Memory-only balance scoring plugin be reviewable against the current scheduler interface, or would you rather I hold and align it with the upcoming scoring overhaul?** Either answer works for me, I just want to build against the right target instead of guessing.

If it helps, I can cut the design doc down to the CPU/Memory-only scope and put up a small implementation PR so there is concrete code to react to rather than a design doc in the abstract. Happy to do that this week if there is appetite for it. Thanks.

davidLif:
Hello @0x-auth,

Do you have dat from real-life cases showing an improvment due to the proposed scoring plugin? How would this interact with the Dominant Resource Fairness (DRF) algorithem currently used by the scheduler?

0x-auth:
@davidLif answering both directly, including where the answer is no.

**Real life data: I do not have any.**

Everything in this proposal is simulation. `benchmark_simulation.py` is synthetic, and `benchmark_full.py` runs on a 3 node minikube, which is a real cluster but not a meaningful one. No production deployment, no production traces.

Context worth knowing, since it touches your question: @Whatsonyourmind had committed on #1373 to producing multi node mix benchmark numbers, and on 9 August formally withdrew that, saying no numbers were coming. So there is no pending external input that would answer this either. Nothing is sitting half finished.

If real data is a gate for this proposal, then it is gated, and I would rather say that than stretch what simulation shows.

**DRF interaction: not analysed.**

I have not worked out how this composes with DRF and there is nothing in the repo on it. Stating the obvious risk rather than pretending otherwise: DRF allocates on the dominant share, this scores on cross dimensional balance, and those two objectives can disagree. A pod whose dominant share points one way and whose shape points another would get pulled in opposite directions. I do not know today whether that is additive, whether one dominates, or whether it produces oscillation.

That is a real gap in the proposal and it needs doing before this is reviewable against the current scheduler. I am not going to guess at it in a comment.

**What did get done since your comment**

Two things on the code side, in `0x-auth/lambda-g-scheduler`:

The plugin here allocated on every score call. Two `make([]float64, ...)` calls building the same post placement vector, 96 B and 2 allocations per score.

```
before  117.70 ns/op  96 B/op  2 allocs/op
after    91.21 ns/op   0 B/op  0 allocs/op
```

All 8 tests still pass. This also means the "zero allocations" claim in #1373 was wrong when I wrote it, and is now correct.

On the scope narrowing from my August comment, the CPU and Memory only version stands. Dropping the GPU sub dimensions removes the dependency you raised and needs no new resource fields.

PR #1405, opened 2026-04-12
test: Concurrent job filler creation

itsomri:
## Description

This PR makes the filler job creation concurrent for e2e tests

## Checklist

- [x] Self-reviewed

### DIFF
--- a/test/e2e/modules/resources/fillers/filler_jobs.go
+++ b/test/e2e/modules/resources/fillers/filler_jobs.go
@@ -7,6 +7,7 @@ package fillers
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	. "github.com/onsi/gomega"
 	"golang.org/x/exp/maps"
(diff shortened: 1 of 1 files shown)


docs/troubleshooting/scale.md
```diff
@@ -0,0 +1,123 @@
...
+   spec:
+     queueDepthPerAction:
+       allocate: 50
+       preempt: 20
+       reclaim: 20
+       consolidation: 10
+       stalegangeviction: 30
+   ```
+
+5. **Scheduler pod resources (manual or VPA)** — Static requests/limits live on **`spec.scheduler.service.resources`** in the cluster **`Config`** CR (`pkg/apis/kai/v1/scheduler/scheduler.go`, referenced from `pkg/apis/kai/v1/config_types.go`). Defaults are set in code when fields are empty.
```
SiorMeir: QUESTION: Do we have any rec for resources for the scheduler?
Maybe we can add a "rule of thumb" for what is the order of magnitude which is relevant for some cluster scales

SiorMeir (approved):

PR #1435, opened 2026-04-16
fix: Exponential job solver

### LINKED ISSUE #1410: Improve performance for reclaim/preempt scenarios with many victims

The current solver/scenario builder for reclaim/preempt actions works by adding victims one by one until it finds a valid scenario (or exhaust all possible victims). This can lead, in some cases, to very long scheduling cycles. We should explore ways to make this more efficient.

### Why is this needed?

Improve usability of the scheduler

itsomri:
## Description

This PR changes the behavior of the JobSolver: instead of incremental, pod-by-pod search for a solution, it will now attempt solving in exponential steps (1 pod, 2 pods, 4 pods etc). If it fails, it will do a binary search from the last valid scenario until it finds the maximal solvable scenario.

This significantly improves reclaim/preempt/consolidate performance for large distributed jobs with many pending pods.

## Related Issues

Fixes #1410 

## Checklist

- [x] Self-reviewed

## Release Notes

* **Performance Improvements**
  * Improved scheduler job allocation efficiency, reducing processing time for scenario evaluations.

### DIFF
--- a/pkg/scheduler/actions/common/solvers/job_solver.go
+++ b/pkg/scheduler/actions/common/solvers/job_solver.go
@@ -44,39 +44,33 @@ func NewJobsSolver(
 	}
 }
 
+// Solve attempts to allocate all tasksToAllocate for pendingJob, finding victims as needed.
+// Uses exponential-then-binary search over k = number of leading tasks to allocate, probing
+// solvePartialJob at each k. Every search probe is discarded (session left clean) and hints
+// from successful probes are accumulated in state. After the search, a final solve is
+// attempted at n with the accumulated hints — this doubles as producing the live statement
+// returned on success and as a retry that recovers false negatives caused by hint gaps
+// during the search (a k-probe may fail with weak hints yet succeed once later probes
+// strengthen them). Matches prior all-or-nothing semantics: full n or nothing.
 func (s *JobSolver) Solve(
 	ssn *framework.Session, pendingJob *podgroup_info.PodGroupInfo) (bool, *framework.Statement, []string) {
 	state := solvingState{}
 	originalNumActiveTasks := pendingJob.GetNumActiveUsedTasks()
 
-	var statement *framework.Statement
-	var pendingTasks []*pod_info.PodInfo
 	tasksToAllocate := podgroup_info.GetTasksToAllocate(pendingJob, ssn.PodSetOrderFn, ssn.TaskOrderFn, false)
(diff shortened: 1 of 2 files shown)


pkg/scheduler/actions/common/solvers/job_solver.go
```diff
@@ -85,7 +83,93 @@ func (s *JobSolver) Solve(
...
+
+	for hi-lo > 1 {
+		mid := (lo + hi) / 2
+		if s.tryProbeAndDiscard(ssn, state, pendingJob, tasksToAllocate, mid) {
+			lo = mid
+		} else {
+			hi = mid
+		}
+	}
+	return lo
```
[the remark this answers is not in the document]
itsomri: Theoretically it might be possible in niche cases, but future scheduling attempts will retry the failed pods. Keeping this simple for now.

pkg/scheduler/actions/common/solvers/job_solver.go
```diff
@@ -85,7 +83,93 @@ func (s *JobSolver) Solve(
...
+) int {
+	n := len(tasksToAllocate)
+	if n == 0 {
+		return 0
+	}
+
+	lo := 0
+	var hi int
+	k := 1
+	for {
```
SiorMeir: for without condition?
is it like `while true`? if so maybe use that

pkg/scheduler/actions/common/solvers/job_solver.go
```diff
@@ -85,7 +83,93 @@ func (s *JobSolver) Solve(
...
+func (s *JobSolver) tryProbeAndDiscard(
+	ssn *framework.Session,
+	state *solvingState,
+	pendingJob *podgroup_info.PodGroupInfo,
+	tasksToAllocate []*pod_info.PodInfo,
+	k int,
+) bool {
+	result := s.probeAtK(ssn, state, pendingJob, tasksToAllocate, k)
+	if result == nil || !result.solved {
+		log.InfraLogger.V(5).Infof("No solution found for %d tasks out of %d tasks to allocate for %s",
```
SiorMeir, line 148: maybe DEBUG instead of trace?

SiorMeir (dismissed):

pkg/scheduler/actions/common/solvers/job_solver.go
itsomri: No `while` loops in go - `for` is `while`

SiorMeir (approved):

PR #1446, opened 2026-04-19
feat(chart): expose kai-operator log level via Helm values

### LINKED ISSUE #1445: feat(chart): expose kai-operator log level via Helm values (operator.zapLogLevel / --zap-log-level)

What you would like to be added?

Expose the kai-operator container log verbosity through the kai-scheduler Helm chart so installs can set it from `values.yaml` or `--set`, without hand-editing rendered manifests.

Concretely:

Add an `operator.zapLogLevel` value (default empty for backward compatibility).
When non-empty, append the operator CLI `flag --zap-log-level=<value>` to the operator Deployment args. The binary already supports this via controller-runtime `zap.Options / BindFlags` (not a custom -log-level flag).
Allowed values align with upstream: `debug`, `info`, `error`, `panic`, or a positive integer for custom verbosity; empty means omit the flag and keep current binary defaults.

### Why is this needed?

Operators need to raise or lower kai-operator logging (e.g. info in production, debug while troubleshooting) in a supported, chart-native way. Today there is no values.yaml knob, so teams patch Deployments or fork templates, which is fragile and easy to lose on upgrades. Wiring the existing --zap-log-level flag through Helm keeps behavior aligned with the operator binary, stays backward compatible when the value is left empty, and makes the setting repeatable across environments.

dttung2905:
## Description

Add an `operator.zapLogLevel` value (default empty for backward compatibility).
When non-empty, append the operator CLI flag `--zap-log-level=<value>` to the operator Deployment args. The binary already supports this via controller-runtime `zap.Options / BindFlags` (not a custom `-log-level` flag).
Allowed values align with upstream: `debug`, `info`, `error`, `panic`, or a positive integer for custom verbosity; empty means omit the flag and keep current binary defaults.

## Related Issues

Fixes https://github.com/kai-scheduler/KAI-Scheduler/issues/1445

## Checklist

- [x] Self-reviewed

## Breaking Changes
No it is backward compatible

### DIFF
--- a/deployments/kai-scheduler/values.yaml
+++ b/deployments/kai-scheduler/values.yaml
@@ -45,6 +45,9 @@ operator:
   probeBindAddress: ":8081"
   qps: 50
   burst: 300
+  # Zap log level for the operator (controller-runtime --zap-log-level). Leave empty to use binary defaults.
+  # When set: debug, info, error, panic. Example: "info"
+  zapLogLevel: ""
 
 podgrouper:
   enabled: true
--- a/deployments/kai-scheduler/values.yaml
+++ b/deployments/kai-scheduler/values.yaml
@@ -45,9 +45,9 @@ operator:
   probeBindAddress: ":8081"
   qps: 50
   burst: 300
-  # Zap log level for the operator (controller-runtime --zap-log-level). Leave empty to use binary defaults.
+  # Log level for the operator (currently mapped to controller-runtime --zap-log-level). Leave empty to use binary defaults.
   # When set: debug, info, error, panic. Example: "info"
-  zapLogLevel: ""
+  logLevel: ""
 
 podgrouper:
   enabled: true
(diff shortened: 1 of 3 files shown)


deployments/kai-scheduler/values.yaml
```diff
@@ -45,6 +45,9 @@ operator:
   probeBindAddress: ":8081"
   qps: 50
   burst: 300
+  # Zap log level for the operator (controller-runtime --zap-log-level). Leave empty to use binary defaults.
+  # When set: debug, info, error, panic. Example: "info"
+  zapLogLevel: ""
```
enoodle: Maybe here just call it "logLevel" so that we will not have a hard time if we decide to change logging library.

enoodle (changes requested):

deployments/kai-scheduler/values.yaml
dttung2905: Good point you brought up. Let me make change to it and also update the changelog in abit

enoodle (dismissed):

enoodle (approved):

gshaibi (approved):

PR #1472, opened 2026-04-23
chore(deps): bump k8s.io/* module group from v0.34.x to v0.35.4

gshaibi:
## Description

Bumps the kubernetes module group to v0.35.4 (k8s.io/kubernetes v1.35.4) and sigs.k8s.io/controller-runtime to v0.23.3. This unblocks importing KEP-4671 Workload API types, which is a prerequisite for #1317.

### Breaking API changes addressed

- **Scheduler framework types** (`Handle`, `SharedLister`, `NodeInfoLister`, `StorageInfoLister`, `WaitingPod`, `NominatingInfo`, `NodePluginScores`, `Extender`, `SharedDRAManager`, `APICacher`, `PreFilterResult`, `Plugin`) moved from ``k8s.io/kubernetes/pkg/scheduler/framework`` to ``k8s.io/kube-scheduler/framework``. Imports are updated and the ``K8sFramework`` stub in ``pkg/common/k8s_utils/framework_handle.go`` now implements the new ``Handle`` methods (``ProfileName``, ``SharedCSIManager``, ``WorkloadManager``, ``SignPod``) and returns the new ``ksf.Parallelizer`` interface.
- **controller-runtime v0.23** made ``admission.Validator`` and ``admission.Defaulter`` generic over the object type. ``PodValidator``, ``PodMutator``, ``QueueValidator`` and the ``PodGroup`` webhook methods now take typed parameters directly instead of ``runtime.Object`` and type assertions. ``ctrl.NewWebhookManagedBy`` calls now pass the object as the second argument.
- **``client.SubResourceClient``** now requires an ``Apply`` method; added to ``FakeSubResourceClient``.
- **DRA** ``cel.Features`` field renamed to ``EnableDRAConsumableCapacity`` in the scheduler feature struct.
- **DRA allocator** now strictly enforces ``ResourcePool.ResourceSliceCount``. The test fake in ``test_utils_builder.go`` now computes per-pool counts based on ``NodeName`` instead of using the total slice count for every pool.
- **``DynamicResourceAllocation``** feature gate is GA and ``LockToDefault: true`` in v1.35. The feature-gates test now asserts ``IsDynamicResourcesEnabled`` directly instead of probing the feature gate.

Also bumps envtest to ``release-0.23`` / ``1.35.0`` and the default kind node tag to ``v1.35.0``.

## Related Issues

Fixes #1466

## Checklist

- [x] Self-reviewed
- [x] Added/updated tests (if needed)
- [ ] Updated documentation (if needed)

[...]

PR #1475, opened 2026-04-24
feat(podgrouper): support upstream Kubernetes Workload API (KEP-4671)

gshaibi:
## Summary

Implements the Workload API translation layer described in \`docs/developer/designs/k8s-workload-api/README.md\`. When the upstream \`scheduling.k8s.io/v1alpha1\` Workload resource is exposed by the cluster (\`GenericWorkload\` feature gate, Alpha in k8s v1.35), the podgrouper maps \`Pod.spec.workloadRef\` to a KAI \`PodGroup\`:

- **Gang policy** → one KAI \`PodGroup\` per \`(Workload, podGroup, replicaKey)\`, \`MinMember = gang.minCount\`. Name: \`{workload}-{podGroup}-{replicaKey}\` (or \`{workload}-{podGroup}\` when \`replicaKey\` is empty).
- **Basic policy** → one KAI \`PodGroup\` per \`(Workload, podGroup)\`, \`MinMember = 1\`. All replica keys share the group.
- **Layered metadata**: the existing top-owner plugin runs first; Workload-derived overrides are applied on top (Name, MinAvailable, SubGroups cleared, Queue/Priority/Preemptibility/Topology/Labels/Annotations merged with Workload winning).
- **Opt-out**: \`kai.scheduler/ignore-workload-api=true\` on the Pod or its top owner short-circuits the override.
- **Instant recovery**: a secondary watch on \`Workload\` resources enqueues pending Pods referencing a previously-missing Workload.
- **Missing Workload**: treated as a soft failure — the Pod stays \`Pending\` without a retry loop and gets re-reconciled by the watcher once the Workload appears.

The upstream API isn't present on every cluster, so availability is detected via discovery at startup (mirrors the existing DRA pattern in \`pkg/common/feature_gates\`). When unavailable, the podgrouper runs unchanged.

## Related Issues

Builds on #1472 (k8s.io bump to v0.35.4, which is where the Workload API types became importable).

## Checklist

- [x] Self-reviewed
- [x] Added/updated tests (\`pkg/podgrouper/podgrouper/plugins/workload/workload_grouper_test.go\`)
- [x] Updated documentation — design doc already merged as part of the bump branch

## Breaking Changes

None. Behavior is gated on cluster capability and explicit opt-in via \`Pod.spec.workloadRef\`.

## Additional Notes

- Based on branch \`shaibi/issue-1466\` (the k8s v0.35 bump); the base comparison will collapse once #1472 merges and this PR is rebased onto main.
- \`make lint\` (golangci-lint v2.11.3) clean.
- \`make manifests\` regenerated RBAC to grant \`get;list;watch\` on \`scheduling.k8s.io/workloads\`.


docs/k8s-workload-api/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Opting out
+
+Set `kai.scheduler/ignore-workload-api: "true"` as an **annotation** on either the Pod or its top owner to bypass the Workload override and use only the top-owner-based grouping path.
+
+## Limitations
+
+- `Workload.spec` is **immutable** upstream once created — to change `gang.minCount` or restructure `podGroups`, delete and recreate the Workload. Mutating labels and annotations is fine and the changes propagate to the existing KAI PodGroup.
+- Workload `subGroups` are not yet part of the upstream API; if your top-owner controller would have produced KAI SubGroups, they're dropped when a Workload override is active.
+- Deleting a `Workload` does **not** delete the KAI PodGroup it created — running gangs aren't disrupted. New Pods referencing the deleted Workload stay `Pending` until a matching Workload reappears.
```
enoodle, line 63: How would the pod group get cleaned eventually?

pkg/podgrouper/podgrouper/podgrouper.go
```diff
@@ -81,7 +91,24 @@ func (pg *podGrouper) GetPGMetadata(ctx context.Context, pod *v1.Pod, topOwner *
...
+
+	base, err := plugin.GetPodGroupMetadata(topOwner, pod, allOwners...)
+	if err != nil {
+		return nil, err
+	}
+	if !pg.isWorkloadAPIEnabled {
+		return base, nil
+	}
+
+	merged, err := workload.ApplyOverride(ctx, base, pod, topOwner, pg.client)
```
enoodle, line 103: Why not create a workload based pod group if the pods are using a workload and otherwise create a pod group as usual? What are we taking from the original pod group? From the design it was not clear that we are using it at all.

pkg/podgrouper/integration_tests/workload_integration_test.go
```diff
@@ -0,0 +1,435 @@
...
+		pg := &schedulingv2alpha2.PodGroup{}
+		Eventually(func() error {
+			return k8sClient.Get(ctx, types.NamespacedName{Namespace: ns, Name: "serving-replicas"}, pg)
+		}, assertTimeout, assertInterval).Should(Succeed())
+		Expect(pg.Spec.MinMember).NotTo(BeNil())
+		Expect(*pg.Spec.MinMember).To(Equal(int32(1)))
+
+		err := k8sClient.Get(ctx, types.NamespacedName{Namespace: ns, Name: "serving-replicas-a"}, &schedulingv2alpha2.PodGroup{})
+		Expect(kerrors.IsNotFound(err)).To(BeTrue(), fmt.Sprintf("expected NotFound, got %v", err))
+	})
```
enoodle, line 67: Which pod is going to be the "owner" of the pod group?

docs/k8s-workload-api/README.md
gshaibi: Cleanup follows the standard KAI path: the PodGroup's OwnerReference is the pod's top owner, and Kubernetes garbage-collects the PodGroup via cascade delete when the top owner goes away.

The Workload resource is intentionally not the owner — it's treated as a policy template with an independent lifecycle (a Workload may outlive any single cohort of pods that references it, or be referenced by several cohorts over time). For pods owned by a controller (Deployment / Job / RayCluster / etc.), the controller is the lifecycle parent and cleanup is handled there.

For bare pods that reference a Workload, the recommended pattern would be to set the Workload as an OwnerReference on each pod if that makes sense to the user - the top-owner walk then lands on the Workload, the PodGroup is owned by the Workload, and deleting the Workload cleanly tears down both the pods and the PodGroup.

pkg/podgrouper/podgrouper/podgrouper.go
gshaibi: The layered design is deliberate and serves a few user-facing purposes:

1. **Backward compatibility for existing customers.** Today's customers configure scheduling attributes via top-owner metadata. With the layered design, adopting the Workload API for, say, gang sizing doesn't
force them to migrate every other attribute over — anything the Workload doesn't explicitly override still flows from the top owner, exactly as before.
2. **Workloads aren't always created by the user.** A top-owner controller (Kubeflow, Spark, Ray, etc.) might create the Workload on the user's behalf. Such a controller won't necessarily propagate every piece
of relevant metadata from the top owner into the Workload — but with the layered design, the user's top-owner-level configuration still reaches the PodGroup via the base path, and the Workload layers on
whatever the controller did set.
3. **OwnerReference.** The base path also produces the OwnerReference that we need for PodGroup's cleanup (as mentioned in the comment above)

pkg/podgrouper/integration_tests/workload_integration_test.go
gshaibi: The top owner of the pods

gshaibi:
Closing this - after a product discussion, we decided to drop the Workload API integration for now. The upstream API (KEP-4671) is still Alpha in 1.35 and the shape is expected to change before Beta.
Carrying a translation layer against a moving target risks user-visible churn (immutable spec semantics, evolving label/annotation contracts, and possibly the WorkloadRef field itself), so it's not the
right time to ship this.

We'll revisit once the API stabilizes upstream (Beta or clear stabilization signal). The design doc, prototype code, and integration tests on this branch stay as the starting point for that work.

Thanks @enoodle for the review - the ownership / lifecycle questions you raised are exactly the kind of contract decisions we'd want to revisit against a stable API rather than against a moving one.

Issue #1420, opened 2026-04-15
Support externally-created PodGroups for cross-workload atomic scheduling


enoodle:
### Real-world use cases

1. **Multi-workload gang scheduling**: Complex AI/ML pipelines increasingly span multiple Kubernetes resources — e.g., a post-training pipeline consisting of a RayJob for training + Grove for serving + batch/Job for evaluation. The current 1:1 mapping between top-owner and PodGroup means these cannot be scheduled atomically. Users need a way to express "all of these resources must get resources together or not at all."

2. **Grove integration**: The Grove project is already implementing external PodGroup creation for KAI in [grove#524](https://github.com/ai-dynamo/grove/pull/524), translating its PodGang concept into KAI PodGroups. A first-class mechanism would make this cleaner and benefit all integrators.

3. **Custom orchestrators**: Any platform that manages composite workloads (multi-component ML pipelines, coupled simulations, etc.) needs the ability to express "these pods from different controllers must be scheduled together."

## What you would like to be added?

Support for users to create PodGroup resources directly (externally), allowing multiple independent workloads (e.g., a RayJob + a batch/Job) to be grouped into a single PodGroup and scheduled atomically by KAI.

Today, PodGroups are created automatically by the **podgrouper** controller — one PodGroup per top-level workload owner. There is no supported way to group pods from *different* workload resources into a single PodGroup. Users who need atomic scheduling across multiple workloads (e.g., a post-training pipeline consisting of a RayJob for training, a Grove deployment for inference, and a batch/Job for evaluation) cannot express this with the current model.

### Proposed behavior

1. Users create a \`PodGroup\` CR directly, specifying \`queue\`, \`minMember\`, \`priorityClassName\`, \`subGroups\`, etc.
2. Pods from multiple workloads reference this PodGroup via the existing \`pod-group-name\` annotation.
3. The podgrouper controller **does not interfere** with pods that already reference an externally-managed PodGroup (does not create competing PodGroups for those pods).
4. The scheduler discovers and schedules the PodGroup normally — it already does this today since discovery is annotation-based.

#### 1. Podgrouper conflict — competing PodGroup creation

[...]

PR #1493, opened 2026-04-28
docs: design external pod group support

enoodle:
## Description

Design support for externally provided pod groups.

Today the pod grouper will overwrite pod group annotations on pods to point to a pod group that it creates but there are several use cases for externally created pod groups.

## Related Issues

Referes #1420 

## Checklist

- [x] Self-reviewed

### DIFF
--- a/docs/developer/designs/external-podgroups/README.md
+++ b/docs/developer/designs/external-podgroups/README.md
@@ -1,0 +1,185 @@
+# Externally-Created PodGroups
+
+## Overview
+
+KAI currently assumes that the podgrouper owns PodGroup creation for KAI-scheduled pods. The podgrouper derives a PodGroup from each pod's owner chain, creates or updates that PodGroup, and patches the pod with `pod-group-name` and, when relevant, `kai.scheduler/subgroup-name`.
+
+This blocks cross-workload gang scheduling. If multiple independent workloads should be scheduled as one atomic unit, an external controller or user needs to create one PodGroup and attach all participating pods to it without podgrouper rewriting the membership.
+
+## Goals
+
+- Support user-created or controller-created PodGroups.
+- Allow pods from multiple workloads to join the same PodGroup.
+- Prevent podgrouper from creating competing PodGroups for opted-out pods.
+- Prevent podgrouper from overwriting external `pod-group-name` and `kai.scheduler/subgroup-name`.
+- Detect pods that reference missing subgroups and prevent unsafe scheduling.
+- Preserve current auto-created PodGroup behavior unless users explicitly opt out.
+
+## Non-Goals
+
+- No automatic segmentation or `segmentSize` support.
+- No subgroup inference for external PodGroups.
+- No cross-namespace PodGroup membership.
+- No new lifecycle controller for external PodGroups.
+- No event fan-out to all participating workload owners.
+
+## API
+
+Pods still join a PodGroup with the existing annotation:
+
+```yaml
+metadata:
+  annotations:
+    pod-group-name: post-training-pipeline
+```
+
+Pods still join a subgroup with the existing label:
+
+```yaml
+metadata:
+  labels:
+    kai.scheduler/subgroup-name: train-workers-rack-0
+```
+
+A pod or any object in its owner chain may opt out of automatic podgrouper management with:
+
+```yaml
+metadata:
+  annotations:
+    kai.scheduler/skip-podgrouper: "true"
+```
+
+This annotation only tells podgrouper to leave the pod alone. It does not attach the pod to a PodGroup. Users must still set `pod-group-name`, and must set `kai.scheduler/subgroup-name` when using non-default subgroups.
+
+`PodGroup.spec.queue` is authoritative for scheduling. Pod queue labels are not used to select the queue for an external PodGroup.
+
+## Example
+
+```yaml
+apiVersion: scheduling.run.ai/v2alpha2
+kind: PodGroup
+metadata:
+  name: post-training-pipeline
+  namespace: default
+spec:
+  minMember: 5
+  queue: default-queue
+  priorityClassName: normal
+  topologyConstraint:
+    topology: cluster-topology
+    requiredTopologyLevel: topology.kubernetes.io/zone
+  subGroups:
+    - name: ray-head
+      minMember: 1
+    - name: ray-workers
+      minMember: 2
+      topologyConstraint:
+        topology: cluster-topology
+        requiredTopologyLevel: topology.kubernetes.io/rack
+    - name: evaluation
+      minMember: 2
+```
+
+```yaml
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: post-training-evaluation
+  annotations:
+    kai.scheduler/skip-podgrouper: "true"
+spec:
+  parallelism: 2
+  completions: 2
+  template:
+    metadata:
+      annotations:
+        pod-group-name: post-training-pipeline
+      labels:
+        kai.scheduler/subgroup-name: evaluation
+    spec:
+      schedulerName: kai-scheduler
+      containers:
+        - name: eval
+          image: ubuntu
+          command: ["sleep", "infinity"]
+      restartPolicy: Never
+```
+
+Other workload types, such as RayJob, use the same pattern: set `kai.scheduler/skip-podgrouper: "true"` on the workload or pod template, then set `pod-group-name` and subgroup labels on the generated pod templates.
+
+## Podgrouper Behavior
+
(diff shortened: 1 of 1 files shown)


docs/developer/designs/external-podgroups/README.md
```diff
@@ -0,0 +1,185 @@
...
+- Record an event on the PodGroup. Optionally mark the offending pod unschedulable when `markUnschedulable` is enabled.
+
+This requirement is part of the MVP because external PodGroups move subgroup assignment responsibility to users and external controllers. Silent ignore makes configuration mistakes hard to diagnose and can break gang semantics.
+
+## Lifecycle And Events
+
+External PodGroups are owned by whoever creates them:
+
+- External controllers should set an owner reference or delete the PodGroup themselves.
+- Manually-created PodGroups without owner references require manual cleanup.
```
SiorMeir, line 156: I think this will need to be emphasized here and  on the README after implementation - our podgroup controller will not handle manually / user created podgroups

docs/developer/designs/external-podgroups/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+A pod or any object in its owner chain may opt out of automatic podgrouper management with:
+
+```yaml
+metadata:
+  annotations:
+    kai.scheduler/skip-podgrouper: "true"
+```
+
+This annotation only tells podgrouper to leave the pod alone. It does not attach the pod to a PodGroup. Users must still set `pod-group-name`, and must set `kai.scheduler/subgroup-name` when using non-default subgroups.
```
SiorMeir, line 52: QUESTION: Does the skip-podgrouper: "true" setting propagate to linked subgroups?
I.E is it possible for a parent podgroup to be external & opted-out while its linked subgroups remain controlled?

enoodle: subgroups are not resources that we create (currently) but part of the PodGroup CRD - so they only exist within it.

docs/developer/designs/external-podgroups/README.md
enoodle: OK, it is written here already so I will keep this in mind

docs/developer/designs/external-podgroups/README.md
```diff
@@ -0,0 +1,185 @@
...
+## Scheduler Behavior
+
+The scheduler already discovers PodGroup membership from `pod-group-name`. However, the current behavior around invalid subgroup labels is unsafe for external PodGroups: if a pod has `kai.scheduler/subgroup-name` that does not exist in the referenced PodGroup, the pod is not assigned to a scheduler PodSet and can be ignored.
+
+This must become an explicit configuration error:
+
+- During snapshot construction, detect every pod whose `kai.scheduler/subgroup-name` does not match a subgroup in its referenced PodGroup.
+- Mark the PodGroup unschedulable and do not allocate any pod from that PodGroup in the cycle.
+- Add a PodGroup `SchedulingCondition` with reason such as `InvalidPodGroup` or `InvalidSubGroup`, and a message naming the pod and missing subgroup.
+- Record an event on the PodGroup. Optionally mark the offending pod unschedulable when `markUnschedulable` is enabled.
```
davidLif: Maybe the event and the condition should be handled by the podgroup controller, and the scheduler will just internally ignore these podgroups?

docs/developer/designs/external-podgroups/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+ApplyToCluster(metadata)
+assignPodToGroupAndSubGroup(pod, metadata)
+```
+
+Only the exact value `"true"` should skip. Missing or invalid values should preserve existing podgrouper behavior.
+
+## Scheduler Behavior
+
+The scheduler already discovers PodGroup membership from `pod-group-name`. However, the current behavior around invalid subgroup labels is unsafe for external PodGroups: if a pod has `kai.scheduler/subgroup-name` that does not exist in the referenced PodGroup, the pod is not assigned to a scheduler PodSet and can be ignored.
```
gshaibi: What about an invalid `pod-group-name`? Should it be treated the same and emit an error log and make the pod unschedulable?

enoodle: What is an invalid pod-group-name? It may just be a pod group that hasn't been created yet. Because the scheduler is scheduling pod groups and not pods, it will ignore the pod anyway.

enoodle: I will add that we will set an "unschedulable - missing pod group" condition **once** on the pod.

docs/developer/designs/external-podgroups/README.md
```diff
@@ -0,0 +1,185 @@
...
+Podgrouper should evaluate skip ownership before it applies PodGroup metadata or patches the pod:
+
+```text
+if pod schedulerName != configured scheduler:
+    return
+
+if pod has kai.scheduler/skip-podgrouper: "true":
+    return
+
+if pod is ownerless and already has pod-group-name:
```
itsomri, line 121: You would still want to evaluate and apply the podgroup metadata in case something changed

docs/developer/designs/external-podgroups/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+metadata = GetPGMetadata(pod, topOwner, allOwners)
+if metadata == nil:
+    return
+
+ApplyToCluster(metadata)
+assignPodToGroupAndSubGroup(pod, metadata)
+```
+
+Only the exact value `"true"` should skip. Missing or invalid values should preserve existing podgrouper behavior.
```
itsomri: What if the annotation exists but we can't parse the value? (in case of typo or something)
Skip podgroup creation and log an error/event?

enoodle: I will update this to skip on any value other than explicit "false" - If you don't want to skip you will not add the annotation at all

docs/developer/designs/external-podgroups/README.md
enoodle: @itsomri We will leave this case as it is today.

docs/developer/designs/external-podgroups/README.md
enoodle: I don't think that the pod group controller should list / update pods. I will change this here to only ignore those specific pods and not the whole pod groups - elastic pod groups can still schedule if some pods are missing.

docs/developer/designs/external-podgroups/README.md
enoodle: today we have this implemented in a way that will not re-create the pod group metadata in this case and we will not change it.

docs/developer/designs/external-podgroups/README.md
enoodle: @gshaibi I now think that it might not be a good idea to set this event - it can cause confusion when many pods are created an the pod grouper takes time to create the pod groups. It might be better just to ignore it as we do today. It relates to your latest fix of a pod grouper error https://github.com/run-ai/runai-engine/pull/1852

gshaibi: I agree

docs/developer/designs/external-podgroups/README.md
itsomri: Do you mean specifically for ownerless pods?
We do re-apply metadata for existing pod groups in ApplyToCluster: https://github.com/kai-scheduler/KAI-Scheduler/blob/main/pkg/podgrouper/podgroup/handler.go#L32

enoodle: Ok, but for ownerless pods with pod group name we already have an exception today

gshaibi: I agree with @enoodle, we can keep [this flow](https://github.com/kai-scheduler/KAI-Scheduler/blob/f4eb38ccc1afd232a69204fe60ec25e1276a2696/pkg/podgrouper/pod_controller.go#L220) out of this design's scope

docs/developer/designs/external-podgroups/README.md
gshaibi: My opinion: I don't think we should support values other than true. 
We don't try to fix typos on the key (`kai.scheduler/skip-podgroupeer`), so we shouldn't try to fix typos on the value either. 
Anything that isn't exactly true should be treated as if the annotation wasn't set.

Not a strong opinion though.

gshaibi (dismissed):

docs/developer/designs/external-podgroups/README.md
enoodle: So I will change back to activate on "true" and any typo will be considered as "false"

gshaibi (approved):

PR #1504, opened 2026-04-30
feat(binder): GPU_MEMORY_LIMIT binder plugin for shared GPU workloads

### LINKED ISSUE #1367: feat(binder): GPU_MEMORY_LIMIT binder plugin for shared GPU workloads
Sub-issue of #1364.

Depends on: #1365 (configurable binder plugins).

Implement a new binder plugin that injects a `GPU_MEMORY_LIMIT` environment variable into each container requesting a shared GPU. The value (in bytes) depends on the request type:
- **GPU memory request** — value is the explicitly requested memory amount
- **GPU fraction request** — value is computed after the scheduling decision, based on the fraction and the actual GPU's total memory on the assigned node

This enables external GPU memory isolation runtimes (e.g., HAMI Core) to enforce hard memory limits.

FouoF:
## Description

As the issue described.

## Related Issues

Fixes #1367 

## Checklist

- [x] Self-reviewed

## Breaking Changes

No


enoodle:
@FouoF Can you please rebase this on top of origin/main removing the commits already merged?

enoodle:
@FouoF Can you please implement it as a new plugin to the binder that will be "opt-in" - ie not enabled by default?
Currently it is added as part of the gpu sharing plugin, but I would like to keep all the GPU virtualization separate to avoid a mess in the future when more are implemented.

FouoF:
> @FouoF Can you please implement it as a new plugin to the binder that will be "opt-in" - ie not enabled by default? Currently it is added as part of the gpu sharing plugin, but I would like to keep all the GPU virtualization separate to avoid a mess in the future when more are implemented.

Thanks, I agree with the goal of keeping GPU virtualization separate from the GPU sharing plugin.

After looking deeper, I don't think making this a binder-only standalone plugin is actually independent in the current architecture. The virtualization logic still relies on several pieces owned by GPU sharing: the admission mutation that injects the env/configmap references, the `runai/shared-gpu-configmap` annotation, the GPU sharing ConfigMap lifecycle, and the binder plugin ordering where GPU sharing must create the ConfigMap before virtualization can update it.

So an opt-in binder plugin would currently be a hidden extension of `gpusharing`, not a truly separate virtualization plugin. That could be misleading and still leave us with tight coupling.

I think there are two reasonable options:
1. Keep this behavior explicitly under GPU sharing for now.
2. Do a larger follow-up refactor that separates the admission mutation, ConfigMap/env management, and binder logic for GPU virtualization as its own opt-in path.

Given that, I’d prefer not to present this as an independent binder plugin in this PR unless we also split the admission/configmap pieces.

enoodle:
@FouoF Lets present it as a dependent binder plugin then - it will only function well if the gpusharing plugin is also enabled.

FouoF:
@enoodle I have made it an independent plugin and it will check if the gpusharing plugin is enabled.

pkg/binder/plugins/gpusharing/gpu_sharing.go
```diff
@@ -76,7 +76,12 @@ func (p *GPUSharing) PreBind(
 		return err
 	}
 
-	return common.SetGPUPortion(ctx, p.kubeClient, pod, containerRef, bindRequest.Spec.ReceivedGPU.Portion)
+	err = common.SetGPUPortion(ctx, p.kubeClient, pod, containerRef, bindRequest.Spec.ReceivedGPU.Portion)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
enoodle: I think that we can revert this change, it will return err anyway.

FouoF: done

pkg/binder/common/gpu_access.go
```diff
@@ -58,6 +59,19 @@ func AddGPUSharingEnvVars(container *v1.Container, sharedGpuConfigMapName string
...
+		ValueFrom: &v1.EnvVarSource{
+			ConfigMapKeyRef: &v1.ConfigMapKeySelector{
+				Key: CudaDeviceMemoryLimit,
+				LocalObjectReference: v1.LocalObjectReference{
+					Name: sharedGpuConfigMapName,
+				},
+				Optional: ptr.To(true),
+			},
+		},
+	})
```
davidLif: Adding this env var definition here, forces every fractional pod to have a CudaDeviceMemoryLimit env var, even if the hami plugins are disabled.

pkg/binder/plugins/factory.go
```diff
@@ -101,6 +107,41 @@ func newGPUSharingPlugin(buildContext PluginBuildContext, arguments map[string]s
...
+	// hamicore requires gpusharing to have already created the configmap.
+	hamiCorePri := ptrDerefInt(config[HamiCorePluginName].Priority, 0)
+	gpuSharingPri := ptrDerefInt(config[GPUSharingPluginName].Priority, 0)
+	if gpuSharingPri <= hamiCorePri {
+		return fmt.Errorf("%q plugin requires %q to run before it (expected %q.priority > %q.priority, got %d <= %d)",
+			HamiCorePluginName, GPUSharingPluginName, GPUSharingPluginName, HamiCorePluginName,
+			gpuSharingPri, hamiCorePri)
+	}
+
+	return nil
```
davidLif, line 136: Do we want to define this relation in code, or add a more generalized way to express the plugin dependencies?

pkg/binder/plugins/factory.go
```diff
@@ -101,6 +107,41 @@ func newGPUSharingPlugin(buildContext PluginBuildContext, arguments map[string]s
...
+}
+
+func ptrDerefInt(v *int, defaultValue int) int {
```
davidLif: cat use the ptr library instead

pkg/binder/plugins/factory.go
```diff
@@ -124,3 +165,15 @@ func boolArgument(arguments map[string]string, name string) (bool, error) {
 	}
 	return parsed, nil
 }
+
+func boolArgumentOrDefault(arguments map[string]string, name string, defaultValue bool) (bool, error) {
```
davidLif: Where is this function used?

pkg/binder/plugins/factory.go
FouoF: This piece of code is essentially a fallback validation to prevent users from misconfiguring plugins. The root issue is that two peer-level scheduler plugins have an implicit dependency on each other, which is not a good practice regardless of the implementation details.

Rather than introducing an additional framework to explicitly model plugin dependencies, I think a better approach would be to move the environment variable injection logic out of the plugin layer entirely. However, that is already beyond the scope of this PR.

I’m also not sure whether the community’s direction would allow this kind of temporary workaround, but my view is that it would be more reasonable to revisit the problem once there are more plugins and the dependency issue becomes more concrete.

pkg/binder/common/gpu_access.go
FouoF: I have split this env into a standalone admission webhook. But it is still dependent on the gpu-sharing plugin.

pkg/binder/plugins/factory.go
FouoF: done

pkg/binder/plugins/factory.go
FouoF: removed

pkg/operator/operands/admission/resources.go
```diff
@@ -367,3 +372,14 @@ func buildArgsList(kaiConfig *kaiv1.Config, config *kaiv1admission.Admission) []
...
+func isHamiCoreEnabled(kaiConfig *kaiv1.Config) bool {
+	if kaiConfig.Spec.Binder == nil {
+		return false
+	}
+	pluginCfg, found := kaiConfig.Spec.Binder.Plugins[kaiv1binder.HamiCorePluginName]
+	if !found {
+		return false
+	}
+	return ptr.Deref(pluginCfg.Enabled, false)
+}
```
davidLif, line 406: Please add in the documentation for the hami plugins the requirment to set both the binder and admission hami flag as true

davidLif:
Also, the changelog conflict needs to be resolved. After that, I will approve

pkg/operator/operands/admission/resources.go
FouoF: Done

archlitchi:
@davidLif conflicts resolved, please take a look

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Added support for configuring scheduler log level and custom scheduler args via Helm values (`scheduler.args`) [#1452](https://github.com/kai-scheduler/KAI-Scheduler/pull/1452) [dttung2905](https://github.com/dttung2905)
 - Added `global.jsonLog` Helm value to enable JSON-formatted logging for use with log aggregation platforms
 - Added `crdupgrader.image.registry` Helm value to override `global.registry` for the `crd-upgrader` pre-install/pre-upgrade hook image, allowing the hook image to be served from a separate mirror without redirecting all chart images. [#1404](https://github.com/kai-scheduler/KAI-Scheduler/issues/1404)
+- Added an opt-in `hamicore` binder plugin (depends on `gpusharing`) to write the HAMI-core GPU memory limit (`CUDA_DEVICE_MEMORY_LIMIT`) for fractional GPU pods.
```
itsomri: This one should appear under the "unreleased" section above

itsomri (dismissed):

enoodle (changes requested):
Looks good, please fix the changelog comment

FouoF:
> Looks good, please fix the changelog comment

fixed

davidLif (dismissed):

enoodle (dismissed):

FouoF:
@davidLif @enoodle I have fixed an inconsistent test, please review again. 
By the way, Github always requests me to update the pr but some merge commit is no change. Is it something wrong?

davidLif (approved):

davidLif:
> @davidLif @enoodle I have fixed an inconsistent test, please review again. By the way, Github always requests me to update the pr but some merge commit is no change. Is it something wrong?

The merge commit itself doesn't have any code changes, but if you compare the code before/after the merge you will see changes.

FouoF:
@enoodle @itsomri PTAL

Issue #1502, opened 2026-04-29
Avoid solver-backed action work for victim-invariant pre-solver failures


enoodle:
### What you would like to be added?

Add an action-level eligibility guard before expensive solver-backed actions start victim simulation.

The guard should apply to actions that try to make pending jobs schedulable by changing the victim set:

- reclaim
- preempt
- consolidation

Before these actions create or run the generic solver for a candidate job, the scheduler should detect known pre-solver failures that cannot be fixed by evicting or pipelining victims.

Initial supported failure classes should be conservative:

- `VolumeBinding`: pod references a missing PVC.
- `ConfigMap`: pod references a missing required ConfigMap.
- `MaxNodePoolResources`: a task requests more of a resource than any node in the scheduler snapshot can provide.

The rule should be:

```text
known victim-invariant failure -> skip expensive action handling
victim-dependent failure       -> run existing action flow
unknown failure                -> run existing action flow
```

Skipped jobs must still expose the same useful fit/scheduling error they would have received later during allocation simulation.

This may require preserving structured pre-filter information, such as predicate name, status code, message, failed task, and failure classification, instead of relying only on the current `PrePredicateFn(... ) error` API.

### Why is this needed?

Today solver-backed actions can spend significant time building victim queues, probing scenarios, and simulating allocations for jobs that are impossible to schedule for reasons unrelated to victims.

For example, a pending job whose pod references a non-existing PVC reaches the `VolumeBinding` failure:

```text
persistentvolumeclaim "<name>" not found
```

Evicting other pods cannot create that PVC. The same applies to a missing required ConfigMap, and to a task that is larger than every node in the scheduler snapshot.

This wastes scheduler time and makes reclaim/preempt/consolidation slower on large clusters, while producing no different scheduling result.

The optimization should preserve behavior:

- jobs remain pending for the same reason;
- user-visible errors remain actionable;
- valid resource-pressure cases still enter the normal solver flow.

Non-goals for the first implementation:

- Do not skip generic pre-filter failures.
- Do not classify ordinary CPU/memory/GPU shortage as victim-invariant.
- Do not classify pod affinity/anti-affinity conflicts as victim-invariant.
- Do not classify topology capacity shortage as victim-invariant.
- Do not classify DRA capacity failures as victim-invariant.
- Do not add cross-session caching.

Acceptance criteria:

- Missing PVC jobs are skipped before solver work in reclaim, preempt, and consolidation.
[...]

PR #1513, opened 2026-04-30
docs: design pod invariant prefilters

enoodle:
## Description

Design pod invariant prefilters to quickly stop processing jobs that have no chance of getting scheduled through removing other pods.

## Related Issues

Refers #1502 

## Checklist

- [x] Self-reviewed

## Breaking Changes

NONE

## Additional Notes

Implementation will follow


docs/developer/action-victim-invariant-prefilter.md
```diff
@@ -0,0 +1,412 @@
...
+
+```text
+pkg/scheduler/api/types.go
+```
+
+Proposed types:
+
+```go
+type VictimInvariantPrePredicateFailure struct {
+	Task *pod_info.PodInfo
```
gshaibi: Why do we need it on the return value? Isn't it part of the caller arguments?

docs/developer/action-victim-invariant-prefilter.md
```diff
@@ -0,0 +1,412 @@
...
+
+- `VolumeBinding`
+- `ConfigMap`
+- `MaxNodePoolResources`
+
+It must not run all pre-filters and then classify their failures. Unknown or
+victim-dependent failures should remain invisible to this API.
+
+The hook follows the same status semantics as the existing
+`evaluateTaskOnPrePredicate` path:
```
gshaibi, line 158: Can we call the same function to ensure that?

docs/developer/action-victim-invariant-prefilter.md
```diff
@@ -0,0 +1,412 @@
...
+- initialize or reset the cache in `OnSessionOpen`;
+- cache only results whose inputs are stable for the intended cache lifetime;
+- delete entries after read if the cached result depends on state that can be
+  consumed or changed;
+- avoid caching results that depend on simulated victim changes unless the cache
+  key includes that state.
+
+For this feature, the predicates plugin cache should store results produced by
+real `predicate.PreFilter(task.Pod)` calls for the supported candidate
+predicates. It should not store synthetic shortcuts, because Kubernetes
```
gshaibi: What do you mean by `synthetic shortcuts`?

docs/developer/action-victim-invariant-prefilter.md
```diff
@@ -0,0 +1,412 @@
...
+pre-filters may populate per-pod cycle state that later filter logic expects.
+
+Suggested cache shape:
+
+```go
+type cachedPrePredicateResult struct {
+	required bool
```
gshaibi, line 217: What is it?

docs/developer/action-victim-invariant-prefilter.md
```diff
@@ -0,0 +1,412 @@
...
+	nodes    sets.Set[string]
+	status   *ksf.Status
+}
+
+type predicatesPlugin struct {
+	// existing fields...
+	prePredicateCache map[common_info.PodID]map[k8s_internal.PredicateName]cachedPrePredicateResult
```
gshaibi: Consider flattening the maps to reduce redirections and memory footprint.

Also, unrelated - do you think we can/should parallelize the predicate filtering? If so, we might want to use a thread-safe map here.

docs/developer/action-victim-invariant-prefilter.md
```diff
@@ -0,0 +1,412 @@
...
+
+## Initial Classifiers
+
+The first implementation supports only three conservative classifiers.
+
+`VolumeBinding` missing PVC:
+
+- predicate key is `VolumeBinding`;
+- status code is `UnschedulableAndUnresolvable`;
+- message identifies `persistentvolumeclaim "<name>" not found`.
```
gshaibi: Why do we need message identifiers?

docs/developer/action-victim-invariant-prefilter.md
enoodle: Its a really good question,
My thought was that the orignal task may not be the one we found to be blocking - but this is too confusing. I should be contained in the error message. I will remove it.

docs/developer/action-victim-invariant-prefilter.md
enoodle: In the design we just say that it will have the same semantics. I assume that each filter will know how to implement it best for itself.  The sensible implementation is to call the same function. This is why I also suggested to add the cache (see below)

docs/developer/action-victim-invariant-prefilter.md
enoodle: That the VictimInvariantPrePredicateFn returned something that is not compatible with the API that the plugin is expecting. I will improve the wording.

docs/developer/action-victim-invariant-prefilter.md
enoodle: This is just explaining what we expect to see in the error message. I will clarify this.

docs/developer/action-victim-invariant-prefilter.md
enoodle: It is caching the result of IsPreFilterRequired - I will add an explanation.

docs/developer/action-victim-invariant-prefilter.md
enoodle: Thanks, I will update this.

About parallelizing it - it is out of the scope for this design but we can always update the type in the future to support it. currently it is not helpful.

docs/developer/action-victim-invariant-prefilter.md
gshaibi: I meant, why do we need to match against a specific message? Isn't the `UnschedulableAndUnresolvable` status code enough?

enoodle: You are right, but this is the "design" for those pre filters - We want them to return a message that is good and will explain the problem that the pod has.

gshaibi (approved):

PR #1537, opened 2026-05-04
fix: Exponential solver regression

### LINKED ISSUE #1546: Failure to reclaim hero job on master
### What happened?

The scale test for hero job fails consistently. Investigate and fix

_No response_

### Environment

- Kubernetes version
- KAI Scheduler version
- Cloud provider or hardware configuration
- Tools that you are using KAI together with  - KOK
- Anything else that is relevant

### LINKED ISSUE #1591: scheduler: topology-required gang cannot preempt a fully-packed clique — pre-filter eliminates the domain before victim search runs
### What happened?

A gang-scheduled JobSet with `kai.scheduler/topology-required-placement`
cannot reclaim a topology domain that is 100% occupied by lower-priority
preemptible workloads, even when priority would normally allow it. The
clique is filtered out of the candidate list before the preempt action's
solver gets a chance to consider evicting the preemptible occupants.

#### Reproduction shape (v0.14.1)

A JobSet (gang of 11 pods, 1 GPU + 4 CPU + 140 GB RAM each, priority 65 in
queue `team-a`) carries:

```yaml
kai.scheduler/topology: cluster-topology
kai.scheduler/topology-required-placement: nvidia.com/gpu.clique
```

plus a `nodeAffinity` matching one of 15 allowed `nvidia.com/gpu.clique`
values.

Cluster state at scheduling time:

- `CLIQUE_PACKED` — 17 healthy nodes in `rack-A`, 100% occupied by 68
  running preemptible `low-priority-inference` pods (priority 30, same
  queue, preemptible by the gang on priority grounds — 65 > 30).
- `CLIQUE_PARTIAL` — 3 nodes with preemptible occupants, the rest
  unhealthy. Preemption frees at most 3 GPUs — gang of 11 cannot fit.
- All other allowed cliques have too few healthy nodes (large rack
  outage on `rack-C`, scattered `NotReady` nodes across `rack-D`,
  `rack-E`, `rack-F`, `rack-G`).

Expected: the scheduler evicts ≥11 `low-priority-inference` pods from
`CLIQUE_PACKED` and runs the 11-pod gang there.

Observed: the pod group ends `UnschedulableOnNodePool`, scheduler logs
`Didn't find a preemption strategy`. The snapshot shows the gang's pods
were evaluated against 19 specific nodes spanning 6 cliques — *every*
node in `CLIQUE_PACKED` is absent from the candidate list, with no
predicate failures logged for them, meaning they were eliminated before
predicates ran.

#### Root cause

In `pkg/scheduler/plugins/topology/job_filtering.go` at v0.14.1:

`calcSubTreeFreeResources` (lines 190–209) builds the domain's "free"
vector from idle + currently-releasing resources only:

```go
domain.IdleOrReleasingVector.Add(node.IdleVector)
domain.IdleOrReleasingVector.Add(node.ReleasingVector)
```

`checkJobDomainFit` (lines 352–370) then rejects any domain where the
job won't fit, via `getJobRatioToFreeResources` (lines 488–516):

```go
freeVal := domain.IdleOrReleasingVector.Get(i)
if freeVal <= 0 {
    resourceRatio = requiredResourceNotInDomainRatio   // = 1000.0
} else {
    resourceRatio = taskVal / freeVal
}
...
if getJobRatioToFreeResources(...) > maxAllocatableTasksRatio {  // = 1.0
    return TopologyFitError(...)  // domain dropped from candidates
}
```

For `CLIQUE_PACKED`:
- every node has `IdleVector[GPU] = 0` and `ReleasingVector[GPU] = 0`
- domain `freeVal[GPU] = 0` → ratio = 1000.0 → domain rejected.

`subSetNodesFn` is registered as the SubsetNodes function
(`topology_plugin.go:48`) and is reached by the preempt action via the
same `AllocateJob` path used by the regular allocator:

```
preempt.Execute
  → attemptToPreemptForPreemptor
  → solver.Solve (job_solver / by_pod_solver)
  → common.TryToVirtuallyAllocatePreemptorAndGetVictims
  → AllocateJob
  → ssn.SubsetNodesFn  ← topology pre-filter runs here, drops CLIQUE_PACKED
```

A clique filtered out at this point is invisible to the solver's victim
search. The preempt logic that *does* understand priority-based eviction
never gets a chance to look at the preemptible occupants on that clique.

A clique that is *partially* occupied (like `CLIQUE_PARTIAL`) passes the
pre-filter because `IdleOrReleasingVector` is non-zero — but preemption
then fails for that clique because too few victims exist there to cover
the gang.

#### Why the scenario-level topology filter doesn't save this

`accumulated_scenario_filters/idle_gpus/topology_aware_idle_gpus.go`
*does* track per-domain capacity and credits freed-victim GPUs back to
their domain as the scenario grows. But its result only gates whether a
scenario is "valid enough to simulate". The downstream simulation runs
`AllocateJob`, which calls `subSetNodesFn`, which rebuilds
`domain.IdleOrReleasingVector` fresh from current per-node
`Idle + Releasing` — i.e. only the evictions that have *actually* been
performed in the current statement. The solver evicts victims for one
potential-victim node at a time and rolls back per attempt
(`by_pod_solver.solveOnPotentialNodes`), so it never reaches a state
where 11 evictions on `rack-A` are pending in a single statement, except
via the binary-search `recordedVictimsTasks` accumulation path — which
has no mechanism to target a specific clique.

(There is also a `common.FeasibleNodesForJob` filter at preempt entry
that drops every GPU-requesting node with no idle+releasing capacity.
The solver does add nodes back as victims evict, so it isn't the same
hard block as `checkJobDomainFit`, but it's another place where "fully
packed by preemptibles" looks identical to "no capacity".)

#### Verifying

1. `kubectl get nodes -l nvidia.com/gpu.clique=<CLIQUE_PACKED>` — shows
   the healthy nodes in the clique.
2. `kubectl get pods -A -o wide | grep <CLIQUE_PACKED nodes> | grep Running`
   — shows them filled with preemptible workload.
3. Confirm the occupants' `priorityClassName` resolves to a priority
   lower than the pending gang's.
4. Scheduler log for the unschedulable pod group:
   `Didn't find a preemption strategy`. The pod's per-task fit-error
   trail contains zero references to nodes in `CLIQUE_PACKED`.

The preempt action should be able to reclaim a topology-required clique
that is fully occupied by lower-priority preemptible workloads. Filtering
out a clique whose every GPU is held by a preemptible occupant produces
the same outcome as if those occupants had a higher priority — which
defeats the priority/preemption model for the topology-required case.

At minimum, the topology pre-filter should consider preemptible-victim
capacity (not just current idle+releasing) when deciding whether a
domain is a viable candidate for a higher-priority job. PR #1574
(*"design: Solver refactor design"*) calls out this exact failure mode
in *Topology-constrained gang on topology domains* and proposes a
topology-first scenario generator that enumerates viable topology
domains first and derives the minimum victim set per domain. Filing
this as a standalone bug so the user-visible behavior has a tracking
item independent of the larger solver refactor — and to ask for
short-term mitigations that don't require waiting for that refactor.

#### Workarounds requested

Are there any flags, configuration, plugin orderings, annotations, or
operational workarounds in v0.14.1 — short of the refactor in #1574 —
that would let a topology-required gang reclaim a fully-packed clique?
What we've considered so far:

- Manually draining/cordoning a few nodes in `CLIQUE_PACKED` so the
  clique exposes non-zero idle GPU and survives the pre-filter, after
  which preempt can clean up the rest. Confirmed to work but requires
  operator intervention per incident.
- Scaling the preemptible workload down on the target clique by hand,
  same idea.

Any other options? In particular, is there any plugin/config that
either (a) makes `checkJobDomainFit` aware of preemptible-victim
resources, or (b) softens the `requiredResourceNotInDomainRatio = 1000.0`
behavior to give the preempt solver a chance to consider the domain?

### Environment

- Kubernetes version: 1.x (any recent version)
- KAI Scheduler version: v0.14.1
- Cloud provider or hardware configuration: NVIDIA GPU cluster with
  `nvidia.com/gpu.clique` topology labels.
- Workload kind: JobSet (gang-scheduled), 11 pods, 1 GPU each, with
  `kai.scheduler/topology-required-placement: nvidia.com/gpu.clique`.

### Related

- PR #1574 — *design: Solver refactor design.* Identifies this failure
  mode as the motivating example for a topology-first scenario generator
  ("victim-set-first generator emits mostly infeasible candidates and
  falls back to a full-set evict-everything candidate that
  over-evicts"). That PR is a design proposal, not a fix.

itsomri:
## Description

Fixes a scheduling bug where the by-pod solver fails to find multi-node reclaim victims in cases where the pending job needs to displace victims spread across multiple nodes. The regression test added in this branch (`Reclaim across many single-GPU nodes`) demonstrates the case: a pending job with N tasks where each task needs a victim from a different single-GPU node — the solver's per-potential-victim-node iteration only ever frees one node's victims at a time, so the simulation can satisfy at most one pending task and the scenario is reported infeasible even though a valid multi-node eviction exists.

The fix moves the search strategy out of `byPodSolver` and into `PodAccumulatedScenarioBuilder`:

- `PodAccumulatedScenarioBuilder` now hosts a sub-scenario emitter (new `sub_scenario_emitter.go`). Outer victim accumulation is unchanged — victims are still popped from the priority-ordered queue. When the outer filter accepts a state, the emitter groups the accumulated potential victims by node (preserving gang semantics through per-accumulator-call victim *units*), sorts node groups ascending by post-eviction capacity with insertion-order tiebreaks, and emits the smallest top-K node prefix whose combined freed capacity covers pending demand. If the simulation rejects that K, the next emission grows the prefix by one node.
- `byPodSolver.solve` collapses to a single path: evict the sub-scenario's recorded + potential victims, simulate, rollback on failure. The per-node iteration that caused the original bug is gone (~180 lines removed).

Why implemented this way:

1. **Performance.** The naive correctness fix — "if per-node fails, evict every accumulated potential victim at once and simulate" — works, but the simulation then runs over an O(victims) feasible-node set, and it can be too aggressive (evict some victims unnecessarily), and also scales badly. Builder-driven emission keeps the simulation's node set down to the prefix the emitter chose.
2. **Minimal behavior change.** The outer filters and validators are untouched. Sub-scenarios are constructed via the existing `NewByNodeScenario` / `AddPotentialVictimsTasks` APIs, so the solver, validator (`ssn.ReclaimScenarioValidatorFn`, etc.), and downstream cache ops see exactly the same shapes they always have. Recorded-victim flow, elastic-job task splitting, and gang eviction are all preserved.

## Related Issues

Fixes #1546

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. The change is internal to the solver/builder; the action APIs and scenario validator contracts are unchanged.

Two pre-existing unit tests had over-specified node assignments that depended on the OLD per-node iteration's tie-breaking order, and have been updated to reflect equally valid alternative solutions the new emitter produces:

- `TestConsolidation` test 16 ("jump over full nodes pods") — `running_job3` is now pipelined to node2 and `pending_job0` lands on node3 (the OLD code happened to swap these).
- `TestHandleScatteredNodesForGangPreempt` — the new emitter may evict up to 4 victims (and pipeline 2 of them back to their original now-free nodes) while walking K-prefixes to find the affinity-required pair, so the test's eviction/pipeline caps were tightened from 200/2 to 4/4 (the exact ceiling, with a comment explaining why).

## Additional Notes

**Performance** (`BenchmarkReclaimLargeJobs`, 3× iterations, Apple M3 Pro):

| Benchmark   | This PR |
|-------------|--------:|
| 100 Node    | ~0.4 s  |
| 500 Node    |  1.16 s |
| 1000 Node   |  4.63 s |

The cost scales roughly linearly with cluster size because the simulation's feasible-node set is bounded by the emitter's chosen prefix (∝ pending task count), not by the size of the accumulated victim pool.

**Correctness**

- New regression test `Reclaim across many single-GPU nodes` (commit `b04b3248`) demonstrates the fixed bug and passes.
- All unit and integration suites pass: `solvers`, `scenario`, `reclaim`, `preempt`, `consolidation`, `allocate`, plus the corresponding integration_tests variants.
- The benchmark expected eviction/pipeline caps in `reclaim_benchmark_test.go` (`numNodes * 10`) are loose; they can be tightened in a follow-up.

**Reviewer guidance**

- Start with `sub_scenario_emitter.go` — that file holds the new search logic and is the conceptual centerpiece.
- `by_pod_solver.go` is mostly deletions; the surviving `solve()` is short and matches a textbook "evict + simulate, rollback on failure" path.
- The *unit* concept (one accumulator-call boundary, identified by the scenario's representative `*PodGroupInfo`) is what preserves gang eviction for non-elastic jobs while still letting elastic jobs split into per-task units.
- Adversarial cases: the emitter's smart-skip is local-per-node and the growth strategy is linear (K → K+1). Pathological workloads could trigger many emissions before a viable one is found; the outer filter gates this in practice. We can revisit if profiles show regression on other workload shapes.

### DIFF
--- a/pkg/scheduler/actions/common/solvers/by_pod_solver.go
+++ b/pkg/scheduler/actions/common/solvers/by_pod_solver.go
@@ -139,7 +139,55 @@ func (s *byPodSolver) solveOnPotentialNodes(ssn *framework.Session, scenario *sc
 			return nil, err
 		}
 	}
+
+	return s.solveOnAllPotentialNodes(ssn, scenario, statement)
+}
+
+// solveOnAllPotentialNodes is a fallback used when no single host node's potential
+// victims are sufficient: it evicts every potential victim recorded in the scenario
+// at once and runs a single simulation. The scenario validator has already approved
+// the full potentials set, so this expands the simulator's view across nodes
+// without violating fair-share.
+func (s *byPodSolver) solveOnAllPotentialNodes(ssn *framework.Session, scenario *scenario.ByNodeScenario,
+	statement *framework.Statement) (*solutionResult, error) {
+	allPotentialVictims := scenario.PotentialVictimsTasks()
+	if len(allPotentialVictims) == 0 {
+		return nil, nil
+	}
+	log.InfraLogger.V(6).Infof(
+		"Per-node iteration exhausted; trying scenario with all potential victims evicted at once")
+
+	checkpoint := statement.Checkpoint()
+	pendingJob := scenario.GetPreemptor()
+	if err := common.EvictAllPreemptees(ssn, allPotentialVictims, pendingJob, statement, s.actionType); err != nil {
+		return nil, err
+	}
+	newFeasibleNodes := s.updateFeasibleNodes(ssn, allPotentialVictims)
+
+	victimTasks := getVictimTasks(scenario.RecordedVictimsTasks(), allPotentialVictims)
+	result := s.runSimulation(ssn, scenario, statement, victimTasks)
+	if result != nil {
+		return result, nil
+	}
+
+	s.feasibleNodesRollback(newFeasibleNodes)
+	if err := statement.Rollback(checkpoint); err != nil {
+		return nil, err
+	}
 	return nil, nil
+}
+
+func (s *byPodSolver) evictPotentialVictimsFromNode(
+	session *framework.Session, scenario *scenario.ByNodeScenario, statement *framework.Statement, nodeToTest string,
+) (*framework.Checkpoint, []*pod_info.PodInfo, error) {
+	recordedVictimsCheckpoint := statement.Checkpoint()
+	pendingJob := scenario.GetPreemptor()
+
+	potentialVictimsTasks := scenario.VictimsTasksFromNodes([]string{nodeToTest})
+	if err := common.EvictAllPreemptees(session, potentialVictimsTasks, pendingJob, statement, s.actionType); err != nil {
+		return nil, nil, err
+	}
+	return &recordedVictimsCheckpoint, potentialVictimsTasks, nil
 }
 
 func (s *byPodSolver) feasibleNodesRollback(newFeasibleNodes map[string]bool) {
@@ -158,19 +206,6 @@ func (s *byPodSolver) updateFeasibleNodes(ssn *framework.Session, victimTasks []
 		s.feasibleNodes[potentialVictimTasks.NodeName] = ssn.ClusterInfo.Nodes[potentialVictimTasks.NodeName]
 	}
 	return newFeasibleNodes
-}
-
-func (s *byPodSolver) evictPotentialVictimsFromNode(
-	session *framework.Session, scenario *scenario.ByNodeScenario, statement *framework.Statement, nodeToTest string,
-) (*framework.Checkpoint, []*pod_info.PodInfo, error) {
-	recordedVictimsCheckpoint := statement.Checkpoint()
-	pendingJob := scenario.GetPreemptor()
-
-	potentialVictimsTasks := scenario.VictimsTasksFromNodes([]string{nodeToTest})
-	if err := common.EvictAllPreemptees(session, potentialVictimsTasks, pendingJob, statement, s.actionType); err != nil {
-		return nil, nil, err
-	}
-	return &recordedVictimsCheckpoint, potentialVictimsTasks, nil
 }
 
 func (s *byPodSolver) handleScenarioSolution(
--- a/pkg/scheduler/actions/common/solvers/by_pod_solver.go
+++ b/pkg/scheduler/actions/common/solvers/by_pod_solver.go
@@ -60,27 +60,17 @@ func newByPodSolver(
 	}
 }
 
-// solve runs one simulation for the given scenario: evict all of its recorded and
-// potential victims, then try to virtually allocate the preemptor across the resulting
-// feasible nodes. The scenario builder owns the search strategy (which victims to put
-// in this scenario and in what order). solve is intentionally a single path with no
-// per-node fallbacks of its own.
-func (s *byPodSolver) solve(
-	session *framework.Session, scenario *scenario.ByNodeScenario,
-) *solutionResult {
+// solve evaluates a scenario's feasibility by simulating it: evicting its victims and simulating the allocation loop.
+// This simulates allocation order, node sorting, predicates, and all relevant scheduling logic, so the scheduler will
+// not perform evictions that will not result in successful allocation of the pending job. This also helps the scheduler
+// avoid evictions which are not relevant to the specific scenario.
+func (s *byPodSolver) solve(session *framework.Session, scenario *scenario.ByNodeScenario) *solutionResult {
 	statement := session.Statement()
 
 	pendingJob := scenario.GetPreemptor()
 	nextTaskToFindAllocation := scenario.PendingTasks()[len(scenario.PendingTasks())-1]
 
 	allVictims := getVictimTasks(scenario.RecordedVictimsTasks(), scenario.PotentialVictimsTasks())
-	if len(allVictims) == 0 {
-		// Nothing to evict. byPodSolver is only invoked when the preemptor needs
-		// to displace at least one task; pure idle-capacity fits are the allocate
-		// action's job, not ours. Signal failure so the builder advances.
-		statement.Discard()
-		return &solutionResult{false, nil, nil, nil}
-	}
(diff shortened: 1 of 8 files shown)


enoodle:
You probably need to add this commit: https://github.com/kai-scheduler/KAI-Scheduler/pull/1531/changes/c7966a2e03561f5687f061496edcd7e87b75ce3d
to make sure that the ReclaimLargeJob benchmarks run successfully on all sizes

davidLif (dismissed):

enoodle (approved):

Issue #549, opened 2025-09-30
Cannot install using helm chart with `Error: UPGRADE FAILED: resource mapping not found for name: "kai-config" namespace: "kai-scheduler" from "": no matches for kind "Config" in version "kai.scheduler/v1"`


averyngo34:
### What happened?

Installing v0.9.4 through helm failed with `Error: UPGRADE FAILED: resource mapping not found for name: "kai-config" namespace: "kai-scheduler" from "": no matches for kind "Config" in version "kai.scheduler/v1"` on Openshift 4.18.

Installing v0.6.14 completed successfully without problem.

```
$ helm upgrade kai-scheduler oci://ghcr.io/nvidia/kai-scheduler/kai-scheduler   --namespace kai-scheduler   --version v0.9.4
Pulled: ghcr.io/nvidia/kai-scheduler/kai-scheduler:v0.9.4
Digest: sha256:219db5b73b5d60aa09103bcb17a843eb860451d75efb46c80d0865b5ab0b48a1
Error: UPGRADE FAILED: resource mapping not found for name: "kai-config" namespace: "kai-scheduler" from "": no matches for kind "Config" in version "kai.scheduler/v1"
ensure CRDs are installed first
```

### What did you expect to happen?

Installing v0.9.14 completed successfully without problem.

### Environment

- Kubernetes version v1.31.10 - Openshift 4.18 
- KAI Scheduler version v0.9.14
- Cloud provider or hardware configuration RedHat Openshift on IBM Cloud
- Tools that you are using KAI together with: Helm
- Anything else that is relevant

itsomri:
Hi,
Thanks for opening this issue. I believe that we are missing SCC that's required for OCP installations.
Can you try to install version `v0.9.4-rc.6`? It adds SCC as seen in this PR: https://github.com/NVIDIA/KAI-Scheduler/pull/544

Having confirmation that it solves the issue will help push the fix.

Thans!

enoodle:
I think this should be solved in v0.9.5 thanks to @itsomri .

itsomri:
Closing for now as we got confirmation from other users that the issue is resolved

PR #1550, opened 2026-05-05
fix(chart): stop recreating kai-config CR on helm upgrade

### LINKED ISSUE #1536: fix(chart): kai-config Helm hook recreates SA on upgrade, causing leader election 401 when scheduler pods are not restarted
## Summary

On `helm upgrade`, if the upgrade does **not** change the scheduler Deployment's pod template (e.g. only admission resources, podgrouper config, or any other non-scheduler value is changed), the `scheduler` ServiceAccount is silently recreated with a new UID while the scheduler pods keep running with the old, now-invalid projected token. The result is a `401 Unauthorized` on the next API server call from those pods — most visibly in leader election:

```
E0504 09:36:52 leaderelection.go:448] error retrieving resource lock kai-scheduler/kai-scheduler: Unauthorized
```

The pod does **not** recover on its own in reasonable time. The kubelet only rotates projected tokens at ~80% of TTL (~48 min for a 1 h token).

**Workaround:** `kubectl rollout restart deployment/kai-scheduler-default -n kai-scheduler` immediately after the upgrade.

> **Note:** If the upgrade also changes the scheduler Deployment pod template (e.g. a new scheduler image tag), the rolling update creates fresh pods that receive a new token bound to the new SA UID — so the issue does not manifest in that case.

---

## Root Cause

The root cause is a chain of three interacting behaviors.

### 1. `kai-config.yaml` is a Helm hook with the default `before-hook-creation` delete policy

```yaml
# deployments/kai-scheduler/templates/kai-config.yaml
annotations:
  "helm.sh/hook": pre-install,pre-upgrade
  "helm.sh/hook-weight": "3"
  # no helm.sh/hook-delete-policy → defaults to before-hook-creation
```

Per Helm semantics, the default `hook-delete-policy` is `before-hook-creation`: Helm **deletes** the existing `kai-config` CR before creating a new one on every `helm upgrade`. The `kai-config` CR gets a brand-new UID on every upgrade, regardless of whether its spec changed.

### 2. The scheduler ServiceAccount is owned by `kai-config`

The operator's `ConfigReconciler` manages the scheduler SA as an operand owned by the `kai-config` CR (`pkg/operator/operands/scheduler/resources.go` + `deployable.go`). When `kai-config` is deleted, Kubernetes GC cascades and deletes the `scheduler` SA. When the new `kai-config` is created, the operator creates a new SA with a new UID.

### 3. The scheduler Deployment is not restarted unless its pod template changed

The scheduler Deployment is owned by the `SchedulingShard` CR (not `kai-config`), and `default-shard.yaml` carries `helm.sh/resource-policy: keep`, so the Deployment survives every upgrade. A rolling update is only triggered if the pod template spec changes. If the upgrade only touched unrelated values (admission resources, podgrouper config, etc.), the Deployment spec is unchanged — no rollout, pods keep running with a token bound to the old SA UID.

The API server rejects that token immediately (the old SA no longer exists). The `leaderelection` loop retries forever without restarting the pod.

### Sequence on an upgrade that does not change the scheduler pod template

```
helm upgrade (e.g. only admission resources changed)
  → Helm deletes kai-config CR          (before-hook-creation default)
    → GC deletes scheduler SA           (ownerRef cascade)
  → Helm creates new kai-config CR      (new UID)
    → operator creates new scheduler SA (new UID)
  → Scheduler Deployment unchanged → no rollout
    → pods carry old SA token → 401 on every API call
    → kubelet rotates token at ~80% TTL (~48 min) → eventually recovers
```

---

## Confirmation

- `kubectl get sa scheduler -n kai-scheduler` — `creationTimestamp` matches the upgrade time after every upgrade
- `kubectl auth can-i get leases --namespace kai-scheduler --as system:serviceaccount:kai-scheduler:scheduler` → **yes** (RBAC is fine; purely a token/auth issue)
- SA age resets on every upgrade
- Scheduler pods continue running (no rollout triggered when scheduler spec is unchanged)

---

## Proposed Fix

Remove the `helm.sh/hook` and `helm.sh/hook-weight` annotations from `kai-config.yaml`, making it a **regular Helm-managed resource** instead of a hook. Regular Helm resources are applied using patch/update semantics, preserving the `kai-config` CR UID across upgrades. GC is never triggered on the SA, and pods keep valid tokens.

CRD ordering is still guaranteed: `crd-upgrader` remains a `pre-install,pre-upgrade` hook (weight 2) and runs before the main Helm phase where `kai-config` would now be applied.

```diff
-  annotations:
-    "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-weight": "3"
```

This aligns with the pattern already used by `default-shard.yaml` and `default-queue.yaml` in the same chart.

---

## Impact

- Affects helm upgrades where the scheduler Deployment pod template is not changed (e.g. only admission, podgrouper, or other non-scheduler values are updated)
- Particularly severe in GitOps environments (ArgoCD) where syncs are frequent
- Silently breaks scheduling until manually detected and remediated
- All other components whose SAs are owned by `kai-config` (binder, podgrouper, queue-controller, etc.) are equally affected if their Deployments are also unchanged by the upgrade

gshaibi:
## Description

Fixes #1536: scheduler pods log `401 Unauthorized` for ~48 min after any `helm upgrade` that doesn't change the scheduler Deployment.

## Root cause

`kai-config` was a `pre-install,pre-upgrade` hook with the default `before-hook-creation` policy, so every upgrade deleted and recreated the CR. Operand `ServiceAccount`s have an `ownerReference` to `kai-config`, so GC cascaded SA deletion on each upgrade. The scheduler Deployment is owned by `SchedulingShard` (`resource-policy: keep`), so it isn't re-rolled with its SA — its pods kept tokens bound to the deleted SA UID and got `Unauthorized` on every API call (including leader election) until kubelet rotated the token at ~80% TTL (~48 min).

## Fix

Manage `kai-config` out of the Helm release so it's never delete-and-recreated:

- CR body moved to a `kai-scheduler.kai-config` named template (`templates/_helpers.tpl`).
- A `post-install,post-upgrade` Job (`kai-config-deployer`) renders the CR into a ConfigMap and applies it with `kubectl apply --server-side --force-conflicts`, and strips leftover `helm.sh/hook` annotations from any pre-existing CR. Reuses the existing `crd-upgrader` image (no new build).
- On uninstall, the CR is removed by the chart's existing `post-delete-cleanup` Job.
- `templates/kai-config.yaml` becomes a guarded test-only renderer (`_test.renderKaiConfig=true`) so the helm-unittest assertions keep targeting a real `Config` document.

The CR's UID stays stable, so operand SAs survive every upgrade.

## Checklist

- [x] Self-reviewed
- [x] Added/updated tests (helm-unittest covers all values flows; e2e upgrade asserts the CR generation advances before checking status)
- [x] Updated documentation (CHANGELOG entry under Fixed)

## Breaking Changes

None.


gshaibi:
GuyTodo: Needs to be validated against https://github.com/kai-scheduler/KAI-Scheduler/issues/549 scenario to make sure there is no regression

enoodle (approved):

Issue #1455, opened 2026-04-20
DRA plugin logs errors for completed/failed pods on every scheduling cycle


david-gang:
## Summary

On every scheduling cycle, the DRA plugin calls `assumePendingClaims()` which iterates over all pods in the scheduler snapshot — including pods in terminal states (`Succeeded`/`Failed`). For each completed pod, it attempts to look up the per-pod ResourceClaim (e.g. `<pod-name>-compute-domain-channel-<hash>`), which has already been correctly deleted by the DRA driver when the pod finished. This produces a flood of error log lines:

```
ERROR dynamicresources/dynamicresources.go:94  Failed to assume pending claim compute-domain-channel for pod default/<pod-name>: failed to get resource claim default/compute-domain-channel: could not find ResourceClaim "default/<pod-name>-compute-domain-channel-<hash>"
```

## Environment

- KAI Scheduler: v0.13.2
- NVIDIA DRA Driver: v25.12.0
- Kubernetes: vanilla k8s
- GPU: GB200 (Grace), arm64

## Root Cause

`assumePendingClaims()` (and its per-pod helper `assumePendingClaim()`) do not check pod phase before attempting to resolve ResourceClaims. Completed pods' per-pod ResourceClaims are intentionally cleaned up by the DRA driver, so the lookup always fails for any pod in a terminal state.

## Impact

- **Correctness**: none — running/pending jobs are unaffected
- **Observability**: high — every scheduling cycle generates one error log line per completed pod still present in the namespace. On a busy cluster with many completed pods not yet GC'd, this produces hundreds of spurious error lines per cycle, making real errors harder to spot

## Suggested Fix

Add a phase guard in `assumePendingClaims()` before processing each pod:

```go
if pod.Status.Phase == v1.PodSucceeded || pod.Status.Phase == v1.PodFailed {
    continue
}
```

This would skip terminal pods entirely and eliminate the spurious errors.

davidLif:
Hi @david-gang ,

THanks you very much for the detailed issue. Your suggested solution seems to be correct. I will validate and create a PR accordingly

PR #1554, opened 2026-05-06
fix(pod-group-controller): skip DRA ResourceClaim lookup for terminal pods

### LINKED ISSUE #1529: fix(podgroupcontroller): logs errors for completed/failed pods when ResourceClaims are already deleted
## Summary

The `podgroupcontroller` logs errors on every reconcile cycle for PodGroups whose pods have already completed, because it attempts to fetch DRA `ResourceClaim` objects that have already been cleaned up by the DRA driver.

```
ERROR    Failed to calculate metadata for pod <namespace>/<pod-name>
{"error": "failed to get resource claim <namespace>/<claim-name> for pod <namespace>/<pod-name>: ResourceClaim.resource.k8s.io \"<claim-name>\" not found"}
```

## Root Cause

`GetPodMetadata` in `pkg/podgroupcontroller/controllers/metadata/pod.go` calls `commonresources.FetchPodResourceClaims()` unconditionally for every pod, regardless of pod phase. For pods in terminal states (`Succeeded`/`Failed`), the DRA driver has already correctly deleted the per-pod `ResourceClaim`, so the lookup always fails.

The error propagates up through `addPodMetadata` -> `calculatePodGroupMetadata` -> `handlePodGroupStatus`, causing the reconciler to return an error and requeue.

## Relationship to #1455

This is the same root cause as #1455 (fixed in #1456), but in a different component. The scheduler's DRA plugin received a phase guard in `assumePendingClaims()`, but the `podgroupcontroller` metadata path was not updated.

## Suggested Fix

Add a phase guard in `GetPodMetadata` before fetching DRA claims:

```go
if pod.Status.Phase == v1.PodSucceeded || pod.Status.Phase == v1.PodFailed {
    return &PodMetadata{
        RequestedResources: v1.ResourceList{},
        AllocatedResources: v1.ResourceList{},
    }, nil
}
```

Or equivalently, skip `FetchPodResourceClaims` for terminal pods since `isActivePod` and `isAllocatedPod` would both return false anyway, making the claims unused.

## Impact

- **Correctness**: none — running/pending jobs are unaffected
- **Observability**: spurious ERROR log lines on every reconcile for any completed pod with DRA claims, making real errors harder to spot

david-gang:
## Description

`GetPodMetadata` in the `podgroupcontroller` called `FetchPodResourceClaims` unconditionally on every reconcile, including for pods in `Succeeded`/`Failed` phases. The DRA driver removes per-pod `ResourceClaim` objects as soon as a pod reaches a terminal phase (independent of pod deletion), so the lookup always failed with `NotFound` and produced spurious ERROR logs on every reconcile cycle until the pod object itself was eventually cleaned up — which can take an indefinite amount of time when `Job.spec.ttlSecondsAfterFinished` is unset.

```
ERROR  Failed to calculate metadata for pod <ns>/<pod>
{"error": "failed to get resource claim <ns>/<claim> for pod <ns>/<pod>: ResourceClaim.resource.k8s.io \"<claim>\" not found"}
```

The fix adds a phase guard at the top of `GetPodMetadata`. Terminal pods skip the ResourceClaim lookup and return empty `RequestedResources`/`AllocatedResources` — which is what the existing `isActivePod`/`isAllocatedPod` guards downstream would have produced anyway.

This is the same root cause as #1455 (fixed scheduler-side in #1456), but the `podgroupcontroller` metadata path was missed at the time.

## Related Issues

Fixes #1529

## Checklist

- [x] Self-reviewed

## Breaking Changes

None.

## Additional Notes

**Tests**

- `TestIsTerminalPod` — table-driven check across the four phases, matching the style of the existing `TestIsActivePod`/`TestIsPodAllocated`.
- `TestGetPodMetadata_TerminalPodSkipsResourceClaimLookup` — integration-style test with a `fake.ClientBuilder` and a Succeeded/Failed pod referencing a missing `ResourceClaim`. Verified to fail (nil-deref panic from `FetchPodResourceClaims`) without the fix and pass with it.

**Validation**

- `go test ./pkg/podgroupcontroller/...` ✅
- `go vet ./pkg/podgroupcontroller/...` ✅
- `gofmt -l pkg/podgroupcontroller/controllers/metadata/` clean


david-gang:
The failing **E2E Upgrade Tests** job is unrelated to this PR.

- Helm upgrade itself succeeds (REVISION 2, status: deployed). The timeout happens afterward, waiting for `KAIConfig.Status.Available=true` at `test/e2e/modules/wait/kai_config.go:41`.
- This PR only touches `pkg/podgroupcontroller/controllers/metadata/pod.go` (a phase guard in `GetPodMetadata`) — no path to KAIConfig reconciliation or operator startup.
- An in-flight chart fix ("fix(chart): stop recreating kai-config CR on helm upgrade") is currently iterating on this exact area on `main`, indicating the upgrade flow is broken independently of this change.
- All other jobs passed: unit tests, validate, build, regular E2E, FOSSA, coverage.

enoodle (dismissed):

davidLif (dismissed):

enoodle (approved):

PR #1556, opened 2026-05-06
fix(scheduler): account for native sidecar requests in pod resources

joeltg:
## Summary

`getPodResourceRequest` (in `pkg/scheduler/api/pod_info/pod_info.go`) treated every initContainer as sequential — `max(running_sum, init.Requests)` — which under-counts pods that use **native sidecars** (initContainers with `restartPolicy: Always`, [KEP-753](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/753-sidecar-containers/README.md)).

Per kubelet's admission accounting in [`k8s.io/component-helpers/resource.PodRequests`](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-helpers/resource/helpers.go) (`AggregateContainerRequests`), native sidecars run concurrently with regular containers, so:

- **Steady-state** running sum = main containers + all native sidecars
- **Init-phase peak** = for each non-restartable initContainer, `init.Requests + sum(native sidecars declared before it in spec order)` — those sidecars are already started and running concurrently with the init
- **Result** = `max(steady-state, init-phase peak)` plus pod overhead

Diverging from this lets the scheduler bind pods that kubelet then rejects with `OutOfCpu`/`OutOfGpu`/etc. at admission, leaving them in `phase=Failed`.

## Production symptom

We hit this on a GKE cluster running pods with a 250m-CPU `gke-gcsfuse-sidecar` (`restartPolicy: Always`) alongside a 4-CPU main container. KAI under-counted each pod by 250m. Across many pods on the same n4-standard-80 node, the cumulative phantom headroom exceeded the node's slack and KAI bound pods that kubelet then rejected with `OutOfCpu`. Once a pod transitioned to `Failed`, `IsActiveUsedStatus` excluded it from `nodeInfo.Requested` on the next cycle, freeing the same phantom slot, and KAI bound another pod into it — a livelock that left **200+ Failed pods on a single node** in our worst case.

After deploying this fix to our fork, Failed-pod accumulation on a comparable node dropped to single digits, and admission rejections went from constant to occasional.

## Implementation

Single pass over `Spec.InitContainers` in spec order:

- Native sidecar (`restartPolicy: Always`) → accumulate into a `sidecarPrefix` running-sum
- Regular initContainer → `initPhasePeak = max(initPhasePeak, init.Requests + sidecarPrefix)`

After the loop:
- `result += sidecarPrefix` (steady-state includes all sidecars)
- `result = max(result, initPhasePeak)` (peak across init phases)

This required a new `Add` method:
- `GpuResourceRequirement.Add` — sums `count`, `portion` (when both non-zero portions match, else error, mirroring `SetMaxResource`), `draGpuCounts`, `migResources`
- `ResourceRequirements.Add` — wrapper that sums both `BaseResource` and `GpuResourceRequirement`. Without this, `BaseResource.Add` was reachable via Go method promotion but silently dropped GPU/MIG resources from sidecars.

## Test plan

- [x] New unit cases in `TestGetPodResourceRequest` for:
  - Pod with one native sidecar (CPU+memory only) — verifies sidecar is *added* to running sum, not max'd
  - Regular init dominates and includes preceding native sidecar — verifies `init + sidecarPrefix` math
  - Native sidecar with GPU — verifies GPU/MIG half is summed (not silently dropped via method promotion)
- [x] All existing `pkg/scheduler/...` tests pass
- [x] Verified in production on Reflection AI fork (see "Production symptom" above)


pkg/scheduler/api/resource_info/resource_requirment.go
```diff
@@ -109,6 +109,18 @@ func (r *ResourceRequirements) SetMaxResource(rr *ResourceRequirements) error {
 	return r.GpuResourceRequirement.SetMaxResource(&rr.GpuResourceRequirement)
 }
 
+// Add sums `rr` into `r` across both halves. `BaseResource.Add` (the embedded
+// method picked up by Go's method promotion) only covers CPU/memory/scalar
+// resources; this wrapper additionally sums the `GpuResourceRequirement` so
+// callers don't silently drop GPU/MIG requests.
```
enoodle: It doesn't make sense for comments to explain that the function is needed because the inherited one is not good enough.

pkg/scheduler/api/pod_info/pod_info.go
```diff
@@ -378,20 +378,49 @@ func getPodGroupID(pod *v1.Pod) common_info.PodGroupID {
...
+		}
+		// Peak at this regular init's runtime = its own request plus every
+		// native sidecar declared before it (which is already running).
+		logErr(containerReq.Add(sidecarPrefix))
+		logErr(initPhasePeak.SetMaxResource(containerReq))
+	}
+	// Steady-state running sum: include all native sidecars.
+	logErr(result.Add(sidecarPrefix))
+	// Final: max(steady-state, init-phase peak).
+	logErr(result.SetMaxResource(initPhasePeak))
```
enoodle: Can you extract this to another function that will compute the native init containers effect. There you can go wild with the comments too, but it would be better to keep it to a more minimal amount and explain that this calculates the effects of long running init containers and link the KEP

enoodle:
Looks good,

I have a few minor comments
Can you also update the change log file  ? ( we are working on something to replace the need to update it , but it is not there yet)

enoodle (dismissed):

enoodle:
Thanks for the fix @joeltg !
We need you to "sign off" on the commits

joeltg:
Thanks for the quick review!

enoodle (dismissed):

pkg/scheduler/api/pod_info/pod_info.go
```diff
@@ -380,25 +380,50 @@ func getPodGroupID(pod *v1.Pod) common_info.PodGroupID {
...
-	for _, container := range pod.Spec.InitContainers {
-		err := result.SetMaxResource(resource_info.RequirementsFromResourceList(container.Resources.Requests))
-		if err != nil {
-			log.InfraLogger.Errorf("Failed to calculate pod required resources for pod %s/%s. Error: %s",
-				pod.Namespace, pod.Name, err.Error())
-		}
-	}
+	sidecarSum, initPhasePeak := initContainerEffects(pod)
+	logIfErr(pod, result.Add(sidecarSum))
+	logIfErr(pod, result.SetMaxResource(initPhasePeak))
```
davidLif, line 391: Logging the error here doesn't seem enough. How should we handle it?

joeltg: The existing `getPodResourceRequest` on main already uses this pattern for `SetMaxResource`. The error only fires on pods with mismatched fractional GPU portions across containers which is effectively malformed input.

Claude suggests:
```
  If we want stricter handling, I see two options, and I'd lean toward doing either in a follow-up since they're broader than this
  fix:

  1. Local guard: on error, treat the pod's resource request as a sentinel (e.g. mark it unschedulable inside PodInfo). Cheap, but
  there's no existing mechanism for "PodInfo built but invalid" — we'd be inventing one.
  2. Propagate: have getPodResourceRequest / NewTaskInfoWithBindRequest return an error so the caller can drop the pod. Cleaner, but
   touches every PodInfo construction site.
```

which would both be fairly large changes.

enoodle: Is shouldn't be handled here - but on admission. We also had this issue before this change.

joeltg:
any other feedback on this? would love to get it merged so we can rebase our fork!

davidLif (approved):

enoodle (approved):

PR #1565, opened 2026-05-07
fix(scheduler): prevent cross-queue consolidating reclaim from bypassing fair-share validation

piyushsprasad:
## Description

When `--allow-consolidating-reclaim=true` (the default), the proportion plugin's reclaim validator skips victim tasks in active allocated statuses from its resource accounting via `getResources()`. If all of a victim queue's tasks are re-pipelined during the reclaim simulation, the validator sees an empty victim-resource map and trivially approves — regardless of whether the reclaimer queue is over its deserved share.

We observed this cause an incident where 48 GPU of pre-training requests caused 376 GPU of running mid-training work to be evicted and restarted (evictedToRequestedRatio of 3.75x and 16x). Pre-training was already ~1100 GPU over its deserved share; mid-training was ~1700 GPU under. The evicted pods were successfully re-scheduled on new nodes in the next cycle — so from a resource accounting perspective, the consolidation worked as intended. But an over-fed queue should not be able to trigger disproportionate disruption to an under-fed queue.

### The fix

When all victims are re-pipelined and the resource map is empty, require the reclaimer queue to stay within its **deserved** quota for managed resources. This gates consolidation on whether the reclaimer is actually starved — preventing an over-fed queue from causing disproportionate disruption to get resources it isn't entitled to, while preserving consolidation for queues that genuinely need resources they're owed.

**Why deserved and not FairShare:** FairShare dynamically expands to absorb unused capacity. A queue can be over-deserved but under-FairShare — which is exactly what happened in the incident (pre-training deserved=5120, FairShare=6294, allocated=6215). Using deserved as the boundary prevents a queue from disrupting another queue precisely because that other queue is under-utilizing.

**Why only managed resources:** The deserved-quota check uses `isUnderDeservedForManagedResources`, which skips resources where `deserved=0` or `deserved=unlimited`. GPU-only queue configs (common in production) set `cpu.quota: 0` and `memory.quota: 0`. Without this scoping, running pods' CPU/Memory allocation would always exceed the zero deserved value, silently rejecting all consolidating reclaim.

### Why `IsActiveAllocatedStatus` is kept unchanged in `getResources()`

We initially narrowed `getResources()` to only skip `Pipelined` tasks (matching the flag's help text). This broke existing tests because the predicate serves two purposes:

1. **Skip re-placed victims** (`Pipelined`) — correct for resource accounting, the victim queue keeps those GPUs
2. **Skip not-yet-simulated victims** (`Running`, `Allocated`, etc.) — necessary because during incremental solving, the scenario can include victims whose tasks have not been speculatively evicted yet in the current solver pass (previous iteration's statement was discarded, reverting tasks to `Running`)

Properly separating these concerns (always pass full victim resources to the strategies, use a separate signal for "this victim was re-placed") would fix both the empty-map case and the partial re-pipelining case, but is a larger refactor. This PR addresses the most dangerous scenario (all victims invisible, no evaluation at all) with the deserved-quota guard.

### Changes

- Add `handleAllVictimsConsolidated` in `proportion.go`: when victims exist but `totalVictimsResources` is empty, check reclaimer against deserved quota for managed resources
- Add `isUnderDeservedForManagedResources` helper: compares allocated vs deserved only for resources with non-zero, non-unlimited quotas
- Add `reclaim_consolidation_empty_victim_map_total` metric with `reclaimer_queue` and `outcome` labels
- Add regression test mirroring the incident scenario (fails without the fix, passes with it)

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. The flag's default behavior is preserved for legitimate consolidation (reclaimer under deserved quota). Only the unfair case — reclaimer over deserved quota with all victims re-pipelined — is now rejected.

### DIFF
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -146,9 +146,11 @@ func (pp *proportionPlugin) reclaimableFn(
 	reclaimerInfo := pp.buildReclaimerInfo(scenario.GetPreemptor(), pp.minNodeGPUMemory)
 	totalVictimsResources := make(map[common_info.QueueID][]resource_info.ResourceVector)
 	victims := scenario.GetVictims()
+	skippedVictimCount := 0
 	for _, victim := range victims {
 		totalJobResources := pp.getVictimResources(victim)
 		if len(totalJobResources) == 0 {
+			skippedVictimCount++
 			continue
 		}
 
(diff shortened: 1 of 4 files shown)


pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -158,9 +160,72 @@ func (pp *proportionPlugin) reclaimableFn(
...
+}
+
+// isUnderDeservedForManagedResources checks allocated <= deserved only for
+// resources where the queue has a real quota (deserved > 0 and not unlimited).
+// Resources with deserved=0 or unlimited are not managed by the quota system
+// and should not block consolidation.
+func isUnderDeservedForManagedResources(allocated, deserved rs.ResourceQuantities) bool {
+	for _, resource := range rs.AllResources {
+		d := deserved[resource]
+		if d <= 0 || d == commonconstants.UnlimitedResourceQuantity {
```
enoodle, line 219: Does that mean that if I lower a queues resource quota to 0 it might suddenly be able to schedule more jobs?

PR #1578, opened 2026-05-11
feat(scheduler): Do not pipeline to long-terminating pods

### LINKED ISSUE #1492: Scheduler should detect repeatedly failed scheduling decisions and try alternatives (stale pipelining loop)
## Is your feature request related to a problem? Please describe.

When a pod is stuck in `Terminating` state on a node, the scheduler maps it to an internal `Releasing` status and counts its GPU memory as "soon to be available." If the pod never actually terminates, this creates an **infinite pipelining loop**:

1. The `gpusharingorder` scoring plugin takes precedence over resource availability — so the scheduler prefers nodes with existing GPU sharing activity, even when those resources are not actually free
2. Allocation detects no idle resources → forces `isPipelineOnly = true`
3. The Terminating pod never completes → loop repeats indefinitely — workloads are stuck in `Pipelined` forever

This can happen even when **other nodes in the cluster have available capacity** for the job.

The problem has two parts:

1. **Plugin precedence**: The [`gpusharingorder`](https://github.com/kai-scheduler/KAI-Scheduler/blob/main/pkg/scheduler/plugins/scores/scores.go) scoring plugin takes precedence over resource availability, so the scheduler will always prefer GPU sharing packing over immediate availability.

2. **Stateless scheduling**: The scheduler does not track previous decisions or their outcomes. When a decision leads to pipelining that never resolves, it keeps making the same failing decision indefinitely.

## Describe the solution you'd like

**Short-term**: Resource availability should be able to override GPU sharing packing preference. When a node's GPUs are blocked behind `Releasing` pods, nodes with immediately available resources should not lose to nodes with only releasing/pending resources.

**Long-term**: Introduce scheduling decision memory — if a workload has been pipelined to the same node repeatedly (N times or for X duration) without success, the scheduler should deprioritize that node and try alternatives. This would be a general-purpose fix for any scenario where the scheduler gets stuck in a loop.

**User-configurable plugin scores**: Consider making the scoring weights of plugins configurable by the user (e.g., via scheduler config or CRD). This would allow cluster administrators to tune the relative importance of GPU sharing packing vs. resource availability vs. other scoring factors based on their specific workload patterns and cluster characteristics.

## Describe alternatives you've considered

- **Manual intervention**: Deleting the stuck terminating pod unblocks the pipeline, but requires awareness and is not scalable.
- **Inverting plugin precedence entirely**: Making resource availability always win would solve this case but could hurt GPU utilization in healthy clusters where packing is desirable.

## Additional context

- **Related scoring code**: [`pkg/scheduler/plugins/scores/scores.go`](https://github.com/kai-scheduler/KAI-Scheduler/blob/main/pkg/scheduler/plugins/scores/scores.go)

davidLif:
## Description

Check for terminating pod deletion time. If it is more them configured (stuck-in-releasing-threshold), do not use it's resources to pipeline pods on.

## Related Issues

Fixes #1492

## Checklist

- [x] Self-reviewed

### DIFF
--- a/cmd/scheduler/app/options/options.go
+++ b/cmd/scheduler/app/options/options.go
@@ -70,6 +70,7 @@ type ServerOption struct {
 	AllowConsolidatingReclaim         bool
 	NumOfStatusRecordingWorkers       int
 	GlobalDefaultStalenessGracePeriod time.Duration
+	StuckInReleasingThreshold         time.Duration
 	PluginServerPort                  int
 	CPUWorkerNodeLabelKey             string
 	GPUWorkerNodeLabelKey             string
@@ -122,6 +123,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&s.AllowConsolidatingReclaim, "allow-consolidating-reclaim", true, "Do not count pipelined pods towards 'reclaimed' resources")
 	fs.IntVar(&s.NumOfStatusRecordingWorkers, "num-of-status-recording-workers", defaultNumOfStatusRecordingWorkers, "specifies the max number of go routines spawned to update pod and podgroups conditions and events. Defaults to 5")
 	fs.DurationVar(&s.GlobalDefaultStalenessGracePeriod, "default-staleness-grace-period", defaultStalenessGracePeriod, "Global default staleness grace period duration. Negative values means infinite. Defaults to 60s")
+	fs.DurationVar(&s.StuckInReleasingThreshold, "stuck-in-releasing-threshold", constants.DefaultStuckInReleasingThreshold, "Duration after a Running pod's deletionTimestamp before its resources are excluded from pipelining. Defaults to 2m.")
 	fs.IntVar(&s.PluginServerPort, "plugin-server-port", 8081, "The port to bind for plugin server requests")
 	fs.StringVar(&s.CPUWorkerNodeLabelKey, "cpu-worker-node-label-key", constants.DefaultCPUWorkerNodeLabelKey, "The label key for CPU worker nodes")
 	fs.StringVar(&s.GPUWorkerNodeLabelKey, "gpu-worker-node-label-key", constants.DefaultGPUWorkerNodeLabelKey, "The label key for GPU worker nodes")
--- a/pkg/apis/kai/v1/schedulingshard_types.go
+++ b/pkg/apis/kai/v1/schedulingshard_types.go
@@ -92,11 +92,6 @@ type SchedulingShardSpec struct {
 	// +kubebuilder:validation:Optional
 	MinRuntime *MinRuntime `json:"minRuntime,omitempty"`
 
-	// StuckInReleasingThreshold is the duration after a Running pod's deletionTimestamp
-	// before its resources are excluded from pipelining. Defaults to 2m.
-	// +kubebuilder:validation:Optional
-	StuckInReleasingThreshold *metav1.Duration `json:"stuckInReleasingThreshold,omitempty"`
-
 	// KValue specifies the kValue for the proportion plugin. Default is 1.0.
 	// +kubebuilder:validation:Optional
 	KValue *float64 `json:"kValue,omitempty"`
--- a/pkg/apis/kai/v1/schedulingshard_types.go
+++ b/pkg/apis/kai/v1/schedulingshard_types.go
@@ -92,6 +92,11 @@ type SchedulingShardSpec struct {
 	// +kubebuilder:validation:Optional
 	MinRuntime *MinRuntime `json:"minRuntime,omitempty"`
 
+	// StuckInReleasingThreshold is the duration after a Running pod's deletionTimestamp
+	// before its resources are excluded from pipelining. Defaults to 2m.
+	// +kubebuilder:validation:Optional
+	StuckInReleasingThreshold *metav1.Duration `json:"stuckInReleasingThreshold,omitempty"`
+
 	// KValue specifies the kValue for the proportion plugin. Default is 1.0.
 	// +kubebuilder:validation:Optional
 	KValue *float64 `json:"kValue,omitempty"`
--- a/pkg/scheduler/actions/utils/action.go
+++ b/pkg/scheduler/actions/utils/action.go
@@ -158,4 +158,3 @@ func getSumOfAvailableGPUs(ssn *framework.Session) (float64, int64) {
 
 	return sumOfAllAllocatableGPUs, sumOfAllAllocatableGPUsMemory
 }
-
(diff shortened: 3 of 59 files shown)


cmd/scheduler/app/options/options.go
```diff
@@ -70,6 +70,7 @@ type ServerOption struct {
 	AllowConsolidatingReclaim         bool
 	NumOfStatusRecordingWorkers       int
 	GlobalDefaultStalenessGracePeriod time.Duration
+	StuckInReleasingThreshold         time.Duration
```
itsomri, line 73: Let's try to be more elegant in our naming - how about `ReleasingTimeout` / `TerminationTimeout`?

pkg/apis/kai/v1/schedulingshard_types.go
```diff
@@ -92,6 +92,11 @@ type SchedulingShardSpec struct {
 	// +kubebuilder:validation:Optional
 	MinRuntime *MinRuntime `json:"minRuntime,omitempty"`
 
+	// StuckInReleasingThreshold is the duration after a Running pod's deletionTimestamp
+	// before its resources are excluded from pipelining. Defaults to 2m.
+	// +kubebuilder:validation:Optional
+	StuckInReleasingThreshold *metav1.Duration `json:"stuckInReleasingThreshold,omitempty"`
```
enoodle: It should be part of `Args` and not be mentioned individually so that this type will not explode with all the scheduler options

pkg/scheduler/api/pod_info/pod_info.go
```diff
@@ -169,10 +170,14 @@ func (pi *PodInfo) UpsertStorageClaim(claimInfo *storageclaim_info.StorageClaimI
 }
 
 func NewTaskInfo(pod *v1.Pod, draPodClaims []*resourceapi.ResourceClaim, vectorMap *resource_info.ResourceVectorMap) *PodInfo {
-	return NewTaskInfoWithBindRequest(pod, nil, draPodClaims, vectorMap)
+	return NewTaskInfoWithConfig(pod, nil, draPodClaims, vectorMap, commonconstants.DefaultStuckInReleasingThreshold)
```
enoodle: Is it ignoring the user configuration?

enoodle: We have so many NewTaskInfo.. functions, now another one. Do we need all of them? This PR also converts many to use the new one, maybe we should have just NewTaskInfo and NewTaskInfoDefaults for tests.

enoodle (changes requested):
I have two comments, one is to simplify things and one might require some more change.

pkg/scheduler/actions/utils/action.go
```diff
@@ -21,6 +21,10 @@ func GetVictimsQueue(
 	preemptees := map[common_info.PodGroupID]*podgroup_info.PodGroupInfo{}
 
 	for _, job := range ssn.ClusterInfo.PodGroupInfos {
+		// A job with any StuckInReleasing pod is not a valid victim
+		if hasStuckInReleasingPod(job) {
```
itsomri: I don't think that it make sense to exempt a job from being considered for eviction just because one of it's pods are stuck. You're basically providing incentives for users to get their pods stuck on eviction...

pkg/scheduler/api/pod_info/pod_info.go
```diff
@@ -169,10 +170,14 @@ func (pi *PodInfo) UpsertStorageClaim(claimInfo *storageclaim_info.StorageClaimI
...
 }
 
 func NewTaskInfoWithBindRequest(pod *v1.Pod, bindRequest *bindrequest_info.BindRequestInfo, draPodClaims []*resourceapi.ResourceClaim, vectorMap *resource_info.ResourceVectorMap) *PodInfo {
+	return NewTaskInfoWithConfig(pod, bindRequest, draPodClaims, vectorMap, commonconstants.DefaultStuckInReleasingThreshold)
```
itsomri: This surface is getting weird
What if we had only NewTaskInfo with `Options...` pattern?

pkg/scheduler/api/pod_info/pod_info.go
davidLif: No, it just sets the default value for the unitest using the function NewTaskInfo.

I want to refactor the whole different "NewTaskInfo" calls, but I think it should be part of a separate PR.

pkg/scheduler/api/pod_info/pod_info.go
davidLif: I want to refactor the whole different "NewTaskInfo" calls, but I think it should be part of a separate PR.

pkg/scheduler/actions/utils/action.go
davidLif: If we have a workload with pods terminating in a long time, I think that considering them as a victim and make other evictions based on the assumption that removing other pods of the workload will work is not reasonable. 

A job with one of it's pods terminating for a long time might be evicted by the stale gang eviction, which should disentivise the users to try something like that. The edge case here would be an elastic pod stuck in terminating state for a long time. We might want o add stale gang eviction for this case as well.

cmd/scheduler/app/options/options.go
davidLif: ReleasingTimeout / TerminationTimeout sound to me like a configuration for a  terminating pods eviction, rather then what we do here.

pkg/apis/kai/v1/schedulingshard_types.go
davidLif: Done

pkg/scheduler/actions/utils/action.go
itsomri: You can just not count their resources as "releasing" and the simulation will take care of the rest.
The job might not be violating gang requirements just because of one pod
You're letting jobs have a backdoor to be exempt from preemption

pkg/scheduler/api/pod_info/pod_info.go
itsomri: We can do a small refactor PR and then do this one

pkg/scheduler/actions/utils/action.go
davidLif: Even if they are not counted as releasing, the simulation will try to "evict" the pod (as it is part of the "active statuses, like running), and we already know that evicting this pod will fail.

pkg/scheduler/api/pod_info/pod_info.go
davidLif: Refactored

pkg/scheduler/api/pod_info/pod_info.go
davidLif: refactored

pkg/scheduler/actions/utils/action.go
davidLif: Removed

enoodle (approved):

itsomri (approved):

PR #1582, opened 2026-05-12
feat(metrics): add queue_metadata_name and queue_display_name labels

### LINKED ISSUE #1566: bug(metrics): inconsistent queue_name label value — scheduler emits DisplayName, queue-controller emits metadata.name
### What happened?

Prometheus metrics emitted by the **scheduler** and the **queue-controller** both use a label named `queue_name`, but they populate that label with **different values** for the same queue. This makes it impossible to reliably join across the two metric families in PromQL whenever a queue has `spec.displayName` set.

#### Scheduler metrics → use `spec.displayName` (falls back to `metadata.name`)

In [`pkg/scheduler/api/queue_info/queue_info.go`](https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/scheduler/api/queue_info/queue_info.go#L45-L49), `QueueInfo.Name` is built as:

```go
queueName := queue.Name
if queue.Spec.DisplayName != "" {
    queueName = queue.Spec.DisplayName
}
```

That `QueueInfo.Name` flows into `QueueAttributes.Name` and is passed verbatim as the `queue_name` label by [`pkg/scheduler/plugins/proportion/resource_division/resource_division.go:50-62`](https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/scheduler/plugins/proportion/resource_division/resource_division.go#L50-L62) for these metrics:

- `queue_fair_share_cpu_cores`
- `queue_fair_share_memory_gb`
- `queue_fair_share_gpu`
- `queue_cpu_usage`
- `queue_memory_usage`
- `queue_gpu_usage`

#### Queue-controller metrics → use `metadata.name`

[`pkg/queuecontroller/metrics/metrics.go` `SetQueueMetrics`](https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/queuecontroller/metrics/metrics.go#L142-L167) reads straight off the K8s resource:

```go
queueName := queue.Name  // *v2.Queue → metadata.name, never DisplayName
```

Affected metrics:

- `queue_info`
- `queue_deserved_gpus`
- `queue_quota_cpu_cores`
- `queue_quota_memory_bytes`
- `queue_allocated_gpus`
- `queue_allocated_cpu_cores`
- `queue_allocated_memory_bytes`

#### Documentation

[`docs/metrics/METRICS.md`](https://github.com/NVIDIA/KAI-Scheduler/blob/main/docs/metrics/METRICS.md) describes `queue_name` as *"Name of the Queue resource"* (i.e. `metadata.name`) — which matches the queue-controller behavior but **not** the scheduler behavior.

#### Concrete impact

Given a queue with `metadata.name: research-team-a` and `spec.displayName: Research Team A`:

```promql
# Returns nothing for queues whose displayName != metadata.name
queue_allocated_gpus / on(queue_name) queue_fair_share_gpu
```

Any dashboard or alert that joins quota/allocation (queue-controller) with fair-share/usage (scheduler) silently breaks for these queues.

A single, consistent value for the `queue_name` label across all KAI metrics.

### Backwards compatibility

Any change to the *value* of an existing label is a **breaking change** for users who already have dashboards, alerts, or recording rules keyed on the current behavior. So this needs to be done carefully — either as a clean break (with release notes / migration guidance) or additively.

#### Resolution paths

| Option | What changes | Pros | Cons / who breaks |
|---|---|---|---|
| **A. Align on `metadata.name` (scheduler changes)** | Scheduler stops substituting `DisplayName`; both sides emit `metadata.name` for `queue_name`. Optionally add a separate `queue_display_name` label on scheduler metrics so the display string is still queryable. | Matches docs. `metadata.name` is unique (`displayName` is not → label collisions are possible today). Joins work. Aligns with the rest of the API surface (PodGroup `spec.queue`, RBAC, parent-queue refs all use `metadata.name`). | Existing scheduler dashboards filtering on `queue_name="<display name>"` break. Needs CHANGELOG + migration snippet (`label_replace`, or switch the filter to the new `queue_display_name` label). |
| **B. Purely additive — keep current values, add new labels** | Leave `queue_name` exactly as it is on both sides. Add `queue_metadata_name` (always `metadata.name`) to scheduler metrics and `queue_display_name` to queue-controller metrics. Document the new labels as the join key going forward. | Zero breakage on day one. Users opt in to the new label as they migrate dashboards. | Cardinality goes up slightly. The pre-existing inconsistency on `queue_name` stays around indefinitely (until a future major version takes Option A). |

**Recommendation:** Land **Option B** first (purely additive, no breakage, gives users a working join key immediately), then plan **Option A** as the longer-term cleanup in a tagged release so the inconsistency on `queue_name` doesn't live forever.

Whichever option is chosen, the change must:

1. Land with a CHANGELOG entry describing the change and (for A) a migration snippet.
2. Update `docs/metrics/METRICS.md` so documented semantics match reality.
3. For A: be called out in release notes as a breaking change.

### Environment

- KAI Scheduler version: main (reproduces in current source)
- Reproducible with any queue that sets `spec.displayName` to something different from `metadata.name`

david-gang:
## Summary
- Implements [Option B](https://github.com/kai-scheduler/KAI-Scheduler/issues/1566#issue-) from #1566: every queue metric emitted by both the scheduler (`queue_fair_share_*`, `queue_*_usage`) and the queue-controller (`queue_info`, `queue_deserved_gpus`, `queue_quota_*`, `queue_allocated_*`) now carries two new labels — `queue_metadata_name` (always the Queue's `metadata.name`) and `queue_display_name` (the Queue's `spec.displayName`, empty when unset).
- The legacy `queue_name` label is left **unchanged** on both sides — queue-controller continues to emit `metadata.name`, the scheduler continues to emit `spec.displayName` with fallback to `metadata.name`. Existing dashboards and alerts keep working. New dashboards should join on `queue_metadata_name`.
- Tests for the queue-controller metrics and the new scheduler-side metrics test use `prometheus.Labels{}` maps so each label name → value binding is explicit; the scheduler-side test exercises the case where `queue_name != queue_metadata_name` to make slot order verifiable.
- `docs/metrics/METRICS.md` gains a new "Queue Identification Labels" section explaining the legacy inconsistency and the recommended migration. `CHANGELOG.md` updated under Unreleased / Added.

## Why preserve `queue_name`?
Changing the *value* of an existing label is a breaking change for users who already join on `queue_name` or filter dashboards by its current values. Option B is intentionally additive: it fixes the join problem on day one without breaking anything. Option A (aligning `queue_name` on `metadata.name` everywhere) is the cleaner long-term state and can be staged as a future tagged-release breaking change.

## Closes
Closes #1566

## Follow-up
While reviewing this change I noticed that `pkg/queuecontroller/metrics/metrics.go` builds the label-values slice positionally with `append([]string{queueName, queueName, queueDisplayName}, additionalMetricLabelValues...)` and feeds it to `WithLabelValues(...)`. That positional-slice pattern is exactly the class of bug Option B works around — a silent slot mismatch produces wrong label values with no compile-time signal. Switching `SetQueueMetrics` to build a `prometheus.Labels` map and use `With(...)` would eliminate that hazard at the source.

I'm keeping that refactor out of this PR to keep the diff focused on the user-visible fix. I plan to open a small follow-up titled `refactor(queue-controller): use prometheus.Labels map in queue metrics` that does only that cleanup.

## Test Plan
- [x] `go build ./...`, `go vet ./...`, `gofmt -l` all clean.
- [x] `go test ./pkg/scheduler/...` all pass (including new `pkg/scheduler/metrics/metrics_test.go`).
- [x] `go test ./pkg/queuecontroller/metrics/...` passes (existing + new `queue_display_name` case).
- [x] `KUBEBUILDER_ASSETS=... go test ./pkg/queuecontroller/controllers/...` passes (envtest-backed integration test, label assertions updated).
- [ ] Manual: deploy on a cluster with a queue whose `spec.displayName` differs from `metadata.name`, query Prometheus for `queue_allocated_gpus / on(queue_metadata_name) queue_fair_share_gpu` and confirm a non-empty result.

## Release Notes

* **New Features**
  * Enhanced queue metrics with new canonical identification labels (`queue_metadata_name` and `queue_display_name`) alongside the existing `queue_name` label for improved queue tracking and correlation.

* **Documentation**
  * Added guidance on queue identification labels with recommended PromQL join patterns and migration recommendations for dashboards and alerts.
  * Legacy labels maintained for backward compatibility with existing monitoring configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kai-scheduler/KAI-Scheduler/pull/1582)


enoodle:
@coderabbitai review

pkg/scheduler/plugins/proportion/resource_division/resource_division.go
```diff
@@ -49,13 +49,17 @@ func reportDivisionResult(queues map[common_info.QueueID]*rs.QueueAttributes) {
 
 		metrics.UpdateQueueFairShare(
 			queue.Name,
+			string(queue.UID),
+			queue.DisplayName,
 			cpuResourceShare.FairShare/resource_info.MilliCPUToCores,
 			memoryResourceShare.FairShare/resource_info.MemoryToGB,
 			gpuResourceShare.FairShare,
 		)
```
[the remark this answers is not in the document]
enoodle: thats the whole point of the PR..

david-gang:
@enoodle re the `queue.UID` confusion raised by CodeRabbit — it tripped me up too at first, but `QueueInfo.UID` is actually initialized from `metadata.name`, not the Kubernetes `metadata.uid`:

https://github.com/kai-scheduler/KAI-Scheduler/blob/313dddd14f140113f5853b8d6ac2d525fa6b9263/pkg/scheduler/api/queue_info/queue_info.go#L57-L58

```go
return &QueueInfo{
    UID: common_info.QueueID(queue.Name),
    ...
}
```

So `string(queue.UID)` in `resource_division.go` is the queue's `metadata.name`, and the PromQL join between scheduler and queue-controller metrics on `queue_metadata_name` does work as intended — CodeRabbit's concern is based on the assumption that `UID` means K8s uid, which isn't the case here.

That said, the naming is confusing enough that it caught both me and CodeRabbit. Would be good to either rename the field (e.g. `MetadataName` on `QueueInfo`/`QueueAttributes`) or at minimum add a doc comment clarifying that `UID` here is the K8s `metadata.name`, not the K8s `metadata.uid`. Happy to do that as a follow-up.

docs/metrics/METRICS.md
```diff
@@ -91,3 +95,35 @@ Business/Resource Labels:
...
+
+- **Queue-controller** sets `queue_name` to the Queue's `metadata.name`.
+- **Scheduler** sets `queue_name` to `spec.displayName` when set, falling back to `metadata.name`.
+
+For queues that set `spec.displayName` to something other than `metadata.name`, this means that a PromQL `on(queue_name)` join between scheduler and queue-controller metrics silently drops those queues.
+
+To make joins reliable without breaking existing dashboards that rely on the current `queue_name` values, two additional labels are emitted by all components:
+
+- `queue_metadata_name` is always the Queue's `metadata.name`. Use this as the join key.
+- `queue_display_name` is always the Queue's `spec.displayName` (empty when unset). Use this for human-readable presentation.
```
davidLif: I don't think this should be added to the docs.

docs/metrics/METRICS.md
```diff
@@ -91,3 +95,35 @@ Business/Resource Labels:
...
+Join scheduler and queue-controller metrics on `queue_metadata_name`:
+
+```promql
+queue_allocated_gpus / on(queue_metadata_name) group_left() queue_fair_share_gpu
+```
+
+### Migration
+
+The legacy `queue_name` label is preserved unchanged for backward compatibility. Existing dashboards continue to work. New dashboards and alerts should prefer `queue_metadata_name` as the canonical queue identifier.
+
```
davidLif: I don't think this should be added to the docs. No active migration steps are needed.

david-gang:
addressed the comments

davidLif (approved):

SiorMeir (approved):

PR #1599, opened 2026-05-20
perf(scheduler): skip redundant probe when partial allocation is insufficient

cparadis-nvidia:
## Description

`JobSolver.Solve` calls `searchMaxSolvableK` (introduced in #1435 to find the
largest `k` for which a partial probe succeeds) and then unconditionally
issues a final `probeAtK(ssn, &state, pendingJob, tasksToAllocate, n)`.
The previous early-return only triggered when `maxSolvedK == 0`, so any case
with `0 < maxSolvedK < n` paid for one extra full-scenario simulation that
was guaranteed to fail — `searchMaxSolvableK` already probed at `n` (during
its exponential phase, and again on each side of the binary search) and
recorded that it was not solvable.
This PR tightens the guard to `maxSolvedK < n`, returning early when no
full allocation is feasible and skipping the redundant final probe. The
returned `(solved, statement, victimNames)` tuple is unchanged in this
branch (`false, nil, calcVictimNames(state.recordedVictimsTasks)`), so the
all-or-nothing semantics of `Solve` are preserved.

## Checklist

- [x] Self-reviewed

## Summary by CodeRabbit

* **Bug Fixes**
  * Optimized the job scheduler solver to eliminate redundant re-simulation operations during resource allocation optimization.


enoodle:
@cparadis-nvidia Thanks, the performance improvement is visible, but I see it is failing some test cases for elastic workloads which might explains the original  intention there. Can you look into this and see if you can find a better condition there?

Issue #1573, opened 2026-05-10
metric: kai_pod_group_evicted_pods_total is hard to use — first-sample invisible, no owner label, counts pods not events


david-gang:
## Summary
The counter `kai_pod_group_evicted_pods_total` (`pkg/scheduler/metrics/metrics.go:204`, emitted from `pkg/scheduler/cache/status_updater/default_status_updater.go:120`) has four issues that together make standard PromQL unusable on it.

### 1. Single-eviction series are invisible to rate() / increase()
The series is created in Prometheus only on first increment, so it appears at value=N in a single scrape. `rate()` and `increase()` need at least 2 samples in the lookback window — a single jump from non-existent to N returns 0 forever.

Verified in production over a 24h window:
- `count(kai_pod_group_evicted_pods_total)` grew 91 → 94 (3 new series in the window).
- `count(rate(...[5m]) > 0)` = 0 for the same window.
- `count(increase(...[1h]) > 0)` = 0 for the same window.

So the new evictions are recorded but invisible to any standard rate-based dashboard or alert.

**Proposal**: pre-initialize the counter at 0 the moment the scheduler first observes a PodGroup, before any eviction. The series is then born at 0 in Prometheus; the first eviction takes it 0→N; `rate()`/`increase()` work from sample 2 onward.

### 2. The `uid` label adds churn for no real benefit
Label set today: `{podgroup, namespace, uid, nodepool, action}`. PodGroup is namespaced, so `{namespace, podgroup}` is already unique for the resource's lifetime. The PodGroup name already encodes the topOwner UID (see `defaultgrouper/default_grouper.go:95`: `pg-<topOwner-name>-<topOwner-UID>`), so when the workload is recreated the PG name already changes.

The only case `uid` disambiguates is: PodGroup is deleted/recreated while its topOwner stays alive with the same UID — niche internal-controller behavior. Meanwhile, every consumer pays the cost: each PodGroup recreation spawns a new series, multiplying issue (1).

**Proposal**: drop `uid` from the label set.

### 3. No owner workload label
For a JobSet the PodGroup name is `pg-<jobset-name>-<jobset-uid>-<replicatedjob>` (`pkg/podgrouper/podgrouper/plugins/jobset/jobset_grouper.go:81`). To roll up evictions per JobSet today, callers must `label_replace` with a UUID regex. Fragile, and doesn't extend to non-JobSet owners.

[...]

Issue #1189, opened 2026-03-11
Automatic subgroup creation per JobSet replica in the JobSet podgrouper


david-gang:
### What you would like to be added?

 When a JobSet has a replicatedJob with multiple replicas, the podgrouper should automatically create a **subgroup per replica** in the PodGroup and assign pods to their respective
  subgroup.

  ### Current behavior
  The JobSet podgrouper creates a flat PodGroup with a single `minAvailable` value calculated as `replicas × min(parallelism, completions)`. There is no subgroup structure — all pods
  from all replicas are treated as one flat group.

  ### Proposed behavior
  For a replicatedJob with `replicas: N` and `parallelism: P`, the podgrouper would generate:
  - A **parent subgroup** per replicatedJob
  - **N child subgroups**, one per replica, each with `minMember: P`
  - Pods automatically assigned to their replica's subgroup using the existing `jobset.sigs.k8s.io/job-index` label

  ### Example
  A replicatedJob `workers` with `replicas: 3, parallelism: 4` would produce subgroups like:

  ```yaml
  subgroups:
    - name: workers
    - name: workers-0
      parent: workers
      minMember: 4
    - name: workers-1
      parent: workers
      minMember: 4
    - name: workers-2
      parent: workers
      minMember: 4
```

  Pods would be labeled with kai.scheduler/subgroup-name: workers-<job-index> based on their jobset.sigs.k8s.io/job-index label.

### Why is this needed?

Each replica within a JobSet replicatedJob is a logically independent unit (e.g., a separate training job or inference instance). Today the podgrouper flattens all replicas into one PodGroup with no subgroup structure, losing this logical separation.

By automatically creating subgroups per replica, we enable users to take advantage of existing features that operate at the subgroup level — most notably topology-aware scheduling (TAS), where each replica can specify its own topology constraints (e.g., all pods of a replica on the same rack, while different replicas can land on different racks).

The JobSet podgrouper already has all the information needed to generate this structure (replica count, parallelism, and the `jobset.sigs.k8s.io/job-index` label on pods). This would require zero additional user configuration.

enoodle:
@david-gang this is a great feature to add!

davidLif:
[...]

PR #1604, opened 2026-05-20
docs(scheduler): design for workload-centric eviction metrics

david-gang:
## Description

This PR adds a design document under `docs/developer/designs/eviction-metrics/` for a redesign of the PodGroup eviction metric (`kai_pod_group_evicted_pods_total`).

The current metric records eviction events but cannot answer the operator question "how disrupted has *this workload* been recently?" with standard PromQL. Single-eviction series are invisible to `rate()`, there is no owner-workload label, the `uid` label causes per-recreation churn, and the counter conflates "pods evicted" with "eviction events".

The design proposes:

- Pre-initializing series at 0 on PodGroup observation, so `rate()` / `increase()` work from the first eviction onward.
- Dropping `uid`, adding `owner_kind` / `owner_name` sourced from the existing `kai.scheduler/top-owner-metadata` annotation so workload-level rollups (JobSet, Workspace, MPIJob, ...) become a single `sum by(owner_name)` in PromQL.
- A sibling `kai_pod_group_eviction_events_total` counter incremented once per (committed decision, affected PodGroup) — answering "how often was this workload disrupted" independently of "how many pods were churned".
- A `subgroup` label on the pods counter only (avoiding double-count at the PG level when a single decision spans multiple leaf subgroups in hierarchical PodGroups).

The design also flags what appears to be a real bug in the current emission: every per-pod eviction adds `EvictionGangSize` to the counter, so an N-pod gang writes N² to the counter instead of N. A separate `pods-counter-bug.md` in the same directory walks through this in detail. The design doc asks maintainers whether to fold the fix into the implementation PR or to open a separate bug issue and ship the fix on a faster timeline.

## Related Issues

Related to #1573 — this is the design proposal that addresses the four problems documented there.

Also relevant to #1189 — the proposed metric labels (`owner_kind`, `owner_name`, `subgroup`) are forward-compatible with the SubGroup hierarchy that #1189 will introduce for JobSet replicas. The multi-rack distributed-training case is called out explicitly as the driving operational pain.

## Checklist

- [x] Self-reviewed
- [ ] Added/updated tests (N/A — design doc only; tests will accompany the implementation PR)
- [x] Updated documentation (the doc itself)

## Breaking Changes

The follow-up implementation PR would introduce two changes that are technically breaking:

- **`uid` label removed** from `kai_pod_group_evicted_pods_total`. Real-world impact expected to be near-zero because (a) the metric is broken for rate-based queries today and (b) `uid` is an internal value, not a human-typed filter. A one-release feature-flag transition is offered as a mitigation if maintainers prefer.
- **Pods-counter behavioral fix** (no longer multiplied by gang size). For any current consumer this is a value change, though the previous values were not meaningful under any standard interpretation. See the linked `pods-counter-bug.md` for the analysis.

Both will be flagged in `CHANGELOG.md` when the implementation PR lands.

## Additional Notes

The doc has an explicit **Assumptions to validate** section and a **Sequencing question** for the gang-size scaling bug. Maintainer feedback on both would be valuable before opening the implementation PR.


docs/developer/designs/eviction-metrics/README.md
```diff
@@ -0,0 +1,325 @@
...
+
+Mitigation if `uid` removal is a concern: keep `uid` for one release behind a feature flag, then remove. Default to removed unless maintainers request otherwise.
+
+## Assumptions to validate
+
+This design rests on one remaining assumption about scheduler internals that comes from reading the code, not from end-to-end verification. Maintainer confirmation or correction would be appreciated before implementation; if it turns out to be wrong, the corresponding emission rule needs adjustment while the rest of the design (labels, aggregation properties, motivation) is unaffected.
+
+- **The current pods counter is effectively multiplied by gang size per eviction.** The per-pod loop in the preempt / reclaim / consolidation / stalegangeviction paths passes `EvictionGangSize = len(preempteeTasks)` on every iteration, and the cache's `Evicted()` adds that value to the counter on every call. A gang of N evicted pods therefore writes N² to the counter today. The full walk-through, affected actions, and the reasoning are in [pods-counter-bug.md](pods-counter-bug.md). The proposed "+1 per pod" rule in this design fixes it as a side effect.
```
enoodle: I think that the N² should be fixed independently of this design, and it doesn't have to be persisted here in the document for ever :)

docs/developer/designs/eviction-metrics/README.md
```diff
@@ -0,0 +1,325 @@
...
+For a cluster running 10,000 PodGroups (already an extreme), the events counter is ≤ 30k series and the pods counter is ≤ 150k series. Pre-initialization adds zero-valued series for PGs that never get evicted; the cost is bounded by the same `A × P` (and `A × P × S̄`) formula.
+
+For Deployments specifically, the per-pod PG model means `P` scales with pod count, not workload count. Operators querying by `owner_name` recover the workload view via aggregation — dashboards stay tractable.
+
+## Migration / breaking changes
+
+| Change | Type | Impact |
+|---|---|---|
+| Drop `uid` label | **Breaking** | Existing dashboard panels that filter on `uid="..."` return empty. In practice impact is expected to be near-zero because (a) the metric is broken for rate-based queries today, so consumers using it are limited, and (b) `uid` is an internal value, not a human-typed filter. CHANGELOG entry required. |
+| Add `owner_kind`, `owner_name`, `subgroup` | Additive | Changes series identity, but queries that `sum by (...)` without these labels keep working. |
```
enoodle: `owner_kind` is the full GVK ?

docs/developer/designs/eviction-metrics/README.md
```diff
@@ -0,0 +1,325 @@
...
+```
+
+## Cardinality
+
+Worst-case series count per cluster, with `A` = number of `action` values (≤ 3), `P` = active PodGroups, `S̄` = average leaf SubGroups per PG (typically 1, ≤ 5):
+
+- `kai_pod_group_eviction_events_total`: `A × P` series.
+- `kai_pod_group_evicted_pods_total`: `A × P × S̄` series.
+
+For a cluster running 10,000 PodGroups (already an extreme), the events counter is ≤ 30k series and the pods counter is ≤ 150k series. Pre-initialization adds zero-valued series for PGs that never get evicted; the cost is bounded by the same `A × P` (and `A × P × S̄`) formula.
```
enoodle: 10K different workloads is not an extreme, there are clusters with 100K pod groups where most are pending.

Should we also initialize the per subgroup metric?

docs/developer/designs/eviction-metrics/README.md
```diff
@@ -0,0 +1,325 @@
...
+
+Single PG, `owner_kind=InferenceJob`, `owner_name=infer-y`. Pods counter has three series, one per leaf SubGroup; events counter has one series:
+
+| Metric | Series | Sum |
+|---|---|---|
+| `kai_pod_group_evicted_pods_total{subgroup="prefill"}` | 3 | |
+| `kai_pod_group_evicted_pods_total{subgroup="decode"}` | 4 | |
+| `kai_pod_group_evicted_pods_total{subgroup="preprocessor"}` | 1 | |
+| `sum by (owner_name)` | | **8 pods** |
+| `sum by (owner_name, subgroup)` | | exact role breakdown |
```
enoodle: add which metric for clarity

docs/developer/designs/eviction-metrics/README.md
```diff
@@ -0,0 +1,325 @@
...
+| Label | Source | Notes |
+|---|---|---|
+| `podgroup` | `PodGroup.Name` | Unchanged. |
+| `namespace` | `PodGroup.Namespace` | Unchanged. |
+| `nodepool` | PodGroup label (existing helper) | Unchanged. |
+| `action` | scheduler action type | Unchanged. Values: `preempt`, `reclaim`, `consolidation`, `stalegangeviction`. |
+| `owner_kind` | `kai.scheduler/top-owner-metadata` annotation on the PodGroup | NEW. The Kind of the top-level workload object (`JobSet`, `Deployment`, `MPIJob`, …). |
+| `owner_name` | `kai.scheduler/top-owner-metadata` annotation on the PodGroup | NEW. The Name of the top-level workload object. Re-submissions of the same name aggregate together. |
+| `subgroup` | `kai.scheduler/subgroup-name` label on the evicted pod (pods counter only) | NEW. The leaf SubGroup the evicted pod belongs to. `""` for flat PodGroups. |
+| `uid` | — | REMOVED. |
```
enoodle: I don't know if anyone if using it but we need to verify that. We will have to announce that we change this.

enoodle (commented):
Thank you for designing this.
I think that we should separate the bug fix from the metric improvement design, and merge them separately.

docs/developer/designs/eviction-metrics/README.md
david-gang: Just the Kind (e.g. `JobSet`, `Deployment`) — not the full GVK. The annotation carries `group`, `version`, and `kind` separately, so we have a choice; I'd prefer Kind only because it's the simplest:

- Matches the kube-state-metrics convention (`kube_pod_owner` etc. use bare Kind), so PromQL joins with kube-state-metrics work without `label_replace` to strip group/version.
- Stable across CRD version bumps (`v1alpha2 → v1beta1` won't churn the label).
- Human-typeable in queries — `owner_kind="JobSet"` is what operators expect.

Collision risk (two CRDs sharing a Kind across groups) is theoretical for top-owner kinds; if it ever shows up we can add an `owner_group` label later without breaking existing queries.

If you're OK with this, I'll update the doc to make it explicit.

docs/developer/designs/eviction-metrics/README.md
david-gang: You're right on both points — fixed in the doc:

- Reframed cardinality around 100K PGs (not 10K) and corrected `A` from 3 to 4 (`preempt`, `reclaim`, `consolidation`, `stalegangeviction`). Dropped the bogus S̄ ≤ 5 cap — there's no `MaxItems` validation on `SubGroups`. Honest framing: `L = total leaf subgroups` scales with workload structure — flat PGs give `L = P`, multi-replica JobSets contribute one leaf per replica (single digits to low tens), inference pipelines a handful. At 100K PGs the events counter is ~400k series; the pods counter is ~400k for mostly-flat populations and a multiple of that for hierarchical workloads.

  On cost, I was too cavalier in the first draft — corrected it. Sample storage *is* cheap (Gorilla XOR encoding compresses constant-zero sequences to ~1 bit/sample), but each active series still costs ~3KB of head-block RAM regardless of value, and **Grafana Cloud / managed Prometheus offerings bill by active series count** — so pre-init does add real recurring cardinality cost even before any eviction. Operators on managed plans should size accordingly. The doc now flags this trade-off explicitly. I don't see a way around it: any alternative that skips pre-init leaves the metric broken in exactly the way #1573 describes.

- Yes, pre-init the per-subgroup pods counter too. `PodGroup.Spec.SubGroups` declares leaf subgroups upfront, so on `OnAdd` we enumerate `(action × leaf-subgroup)` and create a zero series for each. Without this, the first eviction of any subgroup hits the first-sample-invisible problem this design is meant to fix.

docs/developer/designs/eviction-metrics/README.md
enoodle: The workload is not expected to change CRD version after creation.. and you can group your metrics view on that as well.
There are some cases where the same kind name is used by different operators (mpi / kubeflow for example) so maybe we can add the version / group in another label?

david-gang: Good point — the data is already on the annotation (the podgrouper writes `kind`, `name`, `uid`, `group`, `version` via `pkg/podgrouper/topowner/metadata.go`), so this is free to expose.

Pushed [5e102ac](https://github.com/kai-scheduler/KAI-Scheduler/pull/1604/commits/5e102ac7) adding `owner_group` as a separate label (kept `owner_kind` as just the Kind, not a full GVK string). Skipped `owner_version` for the reason you gave — workload objects don't change CRD version after creation, so it'd add cardinality without disambiguating anything. The version field is still in the annotation, so a follow-up can add the label later if a real need shows up, without breaking existing series.

docs/developer/designs/eviction-metrics/README.md
```diff
@@ -0,0 +1,332 @@
...
+> 4. After commit succeeds → events counter increments once per distinct PodGroup that had at least one pod committed in this decision.
+
+So a decision evicting M pods across K distinct PodGroups produces M increments to the pods counter and K increments to the events counter. A rolled-back decision produces zero of either.
+
+#### Pre-initialization at 0
+
+When the scheduler first observes a PodGroup (cache event-handler `OnAdd`), both counters are initialized to 0 across all label combinations that can ever fire for that PG. The series is then born at 0 in Prometheus and a first eviction takes it `0 → N`, making `rate()` / `increase()` correct from the first eviction onward.
+
+- **Events counter**: one series per `action` value (subgroup is not a label on this counter).
+- **Pods counter**: one series per `(action, leaf-subgroup)` combination. Leaf subgroups are declared upfront in `PodGroup.Spec.SubGroups`, so they can be enumerated at `OnAdd` time without waiting for pods. PodGroups with no `SubGroups` field pre-init a single `subgroup=""` series per action.
```
davidLif: Some subgroups can be added dynamicly to the podgroup. This case should also be handled

david-gang: Good catch — updated in a1972ec9.

Pods-counter pre-init now also runs on `OnUpdate` for newly added leaf SubGroups (events counter unchanged — no `subgroup` label). Also drop series on PodGroup `OnDelete` via `DeletePartialMatch`.

david-gang:
Hi,  I addressed all comments on this PR. Do you see anything outstanding before this PR can be pushed to main ?

david-gang:
> Hi, I addressed all comments on this PR. Do you see anything outstanding before this PR can be pushed to main ?

ping @enoodle and @davidLif anything missing from my side? when can we get this merged?

davidLif:
> > Hi, I addressed all comments on this PR. Do you see anything outstanding before this PR can be pushed to main ?
> 
> ping @enoodle and @davidLif anything missing from my side? when can we get this merged?

Sorry, will review today

davidLif (approved):

davidLif (dismissed):

docs/developer/designs/eviction-metrics/README.md
```diff
@@ -0,0 +1,333 @@
...
+
+### Labels
+
+| Label | Source | Notes |
+|---|---|---|
+| `podgroup` | `PodGroup.Name` | Unchanged. |
+| `namespace` | `PodGroup.Namespace` | Unchanged. |
+| `nodepool` | PodGroup label (existing helper) | Unchanged. |
+| `action` | scheduler action type | Unchanged. Values: `preempt`, `reclaim`, `consolidation`, `stalegangeviction`. |
+| `owner_group` | `kai.scheduler/top-owner-metadata` annotation on the PodGroup | NEW. The API Group of the top-level workload object (`kubeflow.org`, `jobset.x-k8s.io`, `apps`, …). Disambiguates Kinds shared across operators (e.g., `MPIJob` from kubeflow vs mpi-operator). `""` for core/v1 objects. |
```
gshaibi, line 78: What about `owner_id`? In the current state, how can we distinguish between separate owners with the same name in the same namespace?

david-gang: Kubernetes uniqueness is (API group, resource, namespace, name), not name+namespace alone. Two live objects of the same group/kind cannot share a name in the same namespace ([Object Names and IDs](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/), [API Concepts](https://kubernetes.io/docs/reference/using-api/api-concepts/)).

So with the proposed labels, separate owners are already distinguishable:

Same ns, same name, different kind: Deployment/train vs JobSet/train → owner_kind
Same ns, same kind name, different group: MPIJob under kubeflow.org vs mpi-operator.kubeflow.org → owner_group
Same name in different namespaces → namespace
The only case owner_id / owner UID would add is not simultaneous: JobSet train is deleted and a new JobSet train is created later (new UID). Kubernetes allows reusing the name; UID exists specifically to tell those generations apart.

That reuse is intentional in this design (“re-submissions of the same name aggregate as one logical workload”). If we also want instance-level queries (“this run vs the previous run”), we can add owner_uid as a follow-up — but it isn’t needed to disambiguate two owners that exist at the same time.

gshaibi: I think that may cause a degradation, as currently we support distinguishing between two workloads with the same name/namespace.
IMO we should include `owner_uid` from day 1, unless there is a special reason not to.

david-gang: Added `owner_uid` from day one and updated the design accordingly.

davidLif (approved):

PR #1617, opened 2026-05-23
feat(scheduler): Jobset as a single podgroup 

### LINKED ISSUE #1189: Automatic subgroup creation per JobSet replica in the JobSet podgrouper

davidLif:
## Description

Rewrite the jobset podgrouper plugin to make the resulting podgroup use subgroups, instead of creating multiple podgroups. This also allows the different subgroups to have different propertites (for example, now some of the minMember of a spesific replicatedJob can be set using an annotation in the replicatedJob template.

## Related Issues

Fixes #1189

## Checklist

- [x] Self-reviewed


pkg/podgrouper/podgroup/handler.go
```diff
@@ -98,13 +98,17 @@ func (h *Handler) createPodGroupForMetadata(podGroupMetadata Metadata) *scheduli
...
 		},
 		Spec: schedulingv2alpha2.PodGroupSpec{
-			MinMember:         ptr.To(podGroupMetadata.MinAvailable),
 			Queue:             podGroupMetadata.Queue,
 			PriorityClassName: podGroupMetadata.PriorityClassName,
 			SubGroups:         []schedulingv2alpha2.SubGroup{},
 			Preemptibility:    podGroupMetadata.Preemptibility,
 		},
 	}
+	if podGroupMetadata.MinSubGroup != nil {
```
itsomri, line 107: Why do we need these guards here?
And if we do: feels like if we get this unwanted result (both minSubGroup and MinAvailalbe defined), we should be more loud about this issue. As it is, it will silently ignore some results from the plugin

itsomri (dismissed):

itsomri (approved):

PR #1622, opened 2026-05-24
fix(scheduler): kai_pod_group_evicted_pods_total inflated by gang size

### LINKED ISSUE #1620: metric: kai_pod_group_evicted_pods_total inflated by gang size (N pods → N²)
## Summary

`kai_pod_group_evicted_pods_total` increments by `gang_size` on every per-pod eviction emit instead of by 1. For an N-pod gang from a single PodGroup, the counter grows by **N²** instead of N. For decisions spanning multiple PodGroups, each PG's counter grows by `(its tasks in the batch) × N`.

## Walk-through

A preemption of 4 pods from the same PodGroup goes through `EvictAllPreemptees` in `pkg/scheduler/actions/common/action.go`:

```go
preempteeTasks = [pod1, pod2, pod3, pod4]   // N = 4

for _, task := range preempteeTasks {       // runs 4 times
    stmt.Evict(task, message, eviction_info.EvictionMetadata{
        EvictionGangSize: len(preempteeTasks),  // = 4, every iteration
        ...
    })
}
```

Each `stmt.Evict` reaches the cache, which calls `defaultStatusUpdater.Evicted()` in `pkg/scheduler/cache/status_updater/default_status_updater.go`, which calls:

```go
metrics.RecordPodGroupEvictedPods(..., evictionMetadata.EvictionGangSize)
```

And `RecordPodGroupEvictedPods` in `pkg/scheduler/metrics/metrics.go` does:

```go
counter.Add(float64(count))   // count = EvictionGangSize = 4
```

So the loop produces:

| Iteration | counter delta | running total |
|-----------|--------------|---------------|
| 1 | +4 | 4 |
| 2 | +4 | 8 |
| 3 | +4 | 12 |
| 4 | +4 | 16 |

**Evicting 4 pods writes 16 to the counter.** N=10 pods writes 100. N=100 writes 10,000.

For decisions spanning multiple PodGroups (which happens through `scenario.RecordedVictimsTasks()` flattening pods from multiple recorded victim jobs into one slice passed to `EvictAllPreemptees`), the same multiplier applies per PG: if 6 of the 10 tasks belong to PG-A and 4 to PG-B, then PG-A's counter grows by `6 × 10 = 60` and PG-B's by `4 × 10 = 40`.

## Affected actions

All four eviction-emitting actions go through the same code path or an equivalent one:

| Action | Path | Status |
|--------|------|--------|
| `preempt` | by-pod solver → `EvictAllPreemptees` | inflated |
| `reclaim` | by-pod solver → `EvictAllPreemptees` | inflated |
| `consolidation` | by-pod solver → `EvictAllPreemptees` | inflated |
| `stalegangeviction` | own per-task loop with the same shape (`pkg/scheduler/actions/stalegangeviction/stalegangeviction.go`) | inflated |

Consolidation is arguably the worst-affected by intent — its purpose is moving multiple pods at once to free nodes, so gang sizes for consolidation are routinely >1. Any panel showing "pods evicted via consolidation" is off whenever consolidation touches more than one pod.

## Why this likely went unnoticed

The inflation collapses when `EvictionGangSize == 1`. Every existing test fixture in the repo uses `EvictionGangSize: 1`, and any single-pod eviction path produces 1²=1, so the counter looks correct. The bug only manifests for gang-scheduled multi-pod evictions and for consolidation.

## Reproduction

A focused unit test reproduces the bug end-to-end against `defaultStatusUpdater.Evicted()` — the same function the action layer reaches through `stmt.Evict`. The test simulates the `EvictAllPreemptees` loop exactly: one `Evicted()` call per pod in the gang, each carrying `EvictionGangSize = N`.

`pkg/scheduler/cache/status_updater/default_status_updater_test.go`:

```go
func TestEvicted_PodsCounter_GangSizeBug(t *testing.T) {
    const gangSize = 4

    pg := &enginev2alpha2.PodGroup{
        ObjectMeta: metav1.ObjectMeta{
            Name:      "gang-bug-pg",
            Namespace: "gang-bug-ns",
            UID:       "gang-bug-uid",
        },
    }

    kubeClient := fake.NewSimpleClientset()
    kubeAiSchedClient := kubeaischedfake.NewSimpleClientset()
    recorder := record.NewFakeRecorder(100)
    statusUpdater := New(kubeClient, kubeAiSchedClient, recorder, 1, false, nodePoolLabelKey)

    // Simulate EvictAllPreemptees: one stmt.Evict per pod in the gang,
    // every call carrying EvictionGangSize = len(preempteeTasks).
    for i := 0; i < gangSize; i++ {
        statusUpdater.Evicted(pg, eviction_info.EvictionMetadata{
            Action:           "preempt",
            EvictionGangSize: gangSize,
            Preemptor:        &types.NamespacedName{Namespace: "preemptor-ns", Name: "preemptor"},
        }, "gang-bug evicted")
    }

    value, found := getEvictedPodsCounterValue(t, pg.Name, pg.Namespace, string(pg.UID), "default", "preempt")
    assert.True(t, found, "counter sample for the test PodGroup was not emitted")
    assert.Equal(t, float64(gangSize), value,
        "expected counter to equal the number of pods evicted (%d), got %v — gang of N is writing N² instead of N",
        gangSize, value)
}
```

The test asserts the expected behavior (counter == N == 4) and fails on `main` with the actual value:

```
default_status_updater_test.go:868:
    Error: Not equal:
           expected: 4
           actual  : 16
    Messages: expected counter to equal the number of pods evicted (4), got 16 — gang of N is writing N² instead of N
```

The 4 → 16 jump is exactly N² as predicted by the walk-through.

## Environment

Reproduced on `main` (commit `08625cfa`).

david-gang:
## Description

`kai_pod_group_evicted_pods_total` was incremented by `EvictionGangSize` (= N) on every per-pod eviction emit. For an N-pod gang the counter grew by N² instead of N, and for batches spanning multiple PodGroups each PG inflated by `(its_tasks × batch_size)`. All eviction-emitting actions were affected: `preempt`, `reclaim`, `consolidation`, and `stalegangeviction`.

**Root cause** — `EvictionMetadata.EvictionGangSize` is informational (populates the per-pod event annotation `num-evicted-pods`), but was also passed as the `count` to `metrics.RecordPodGroupEvictedPods` at `pkg/scheduler/cache/status_updater/default_status_updater.go:126`. Each call to `defaultStatusUpdater.Evicted` represents exactly one pod evicted (the cache calls it per pod via `sc.evict`), so the counter should increment by 1, not by N. The const binding `evictionGangSize = "num-evicted-pods"` makes the original informational intent of the field clear.

**Fix** — drop the `count` parameter from `RecordPodGroupEvictedPods` and use `Inc()` so the API is structurally per-pod and the misuse vector is removed. Update the single caller. `EvictionGangSize` stays on `EvictionMetadata` (still used for the event annotation, line 107, unchanged).

Only one caller of the metric function exists, and only one implementation of the `StatusUpdater` interface — so fixing the boundary fixes every affected action with no further changes needed in `EvictAllPreemptees`, `stalegangeviction.handleStaleJob`, `Statement.Evict`, or `Cache.Evict`.

## Related Issues

Fixes #1620

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. `RecordPodGroupEvictedPods` is an internal scheduler-package function with a single caller; its signature change is internal.

### Tests added in `pkg/scheduler/cache/status_updater/default_status_updater_test.go`

1. **`TestEvicted_PodsCounter_GangSizeBug`** — verbatim from the issue. Simulates the `EvictAllPreemptees` loop: 4 `Evicted` calls each with `EvictionGangSize=4`. Pre-fix wrote 16, asserts == 4. Fails on `main`.
2. **`TestEvicted_PodsCounter_SinglePod`** — `EvictionGangSize=1`, regression guard for the case that was always correct (1²=1 masked the bug).
3. **`TestEvicted_PodsCounter_MultiplePodGroups`** — simulates `RecordedVictimsTasks()` flattening across PGs: 6 emits for PG-A and 4 for PG-B with `EvictionGangSize=10`. Pre-fix gave 60/40, asserts 6/4.
4. **`TestEvicted_PodsCounter_ActionLabelIsolation`** — two `Evicted` calls with different `Action` values; confirms the `action` label still partitions counters.
5. **`TestEvicted_EventAnnotationPreservesGangSize`** — verifies `num-evicted-pods=N` still appears on the per-pod event annotation, so the informational use of `EvictionGangSize` is preserved.

### Verification

- New tests pass; pre-fix `TestEvicted_PodsCounter_GangSizeBug` reproduces the exact 4→16 the issue describes.
- Full `pkg/scheduler/...`, `pkg/binder/...`, `pkg/operator/...`, `pkg/queuecontroller/...`, `pkg/podgrouper/...`, `pkg/podgroupcontroller/...`, `pkg/admission/...`, `pkg/env-tests/...` green under envtest.
- `make lint` clean (golangci-lint v2.11.3 via Docker).
- Helm chart tests pass.
- `CHANGELOG.md` updated under `Unreleased → Fixed`.

### DIFF
--- a/pkg/scheduler/metrics/metrics.go
+++ b/pkg/scheduler/metrics/metrics.go
@@ -365,9 +365,9 @@ func RegisterPreemptionAttempts() {
 	preemptionAttempts.Inc()
 }
 
-// RecordPodGroupEvictedPods records the number of pods evicted for a pod group
-func RecordPodGroupEvictedPods(name, namespace, uid, nodepool, action string, count int) {
-	podGroupEvictedPodsTotal.WithLabelValues(name, namespace, uid, nodepool, action).Add(float64(count))
+// RecordPodGroupEvictedPods records a single pod eviction for a pod group.
+func RecordPodGroupEvictedPods(name, namespace, uid, nodepool, action string) {
+	podGroupEvictedPodsTotal.WithLabelValues(name, namespace, uid, nodepool, action).Inc()
 }
 
 func IncScenarioSearchJobs[A ~string](action A, result string, reducedBudget bool) {
--- a/pkg/scheduler/metrics/metrics.go
+++ b/pkg/scheduler/metrics/metrics.go
@@ -365,8 +365,8 @@ func RegisterPreemptionAttempts() {
 	preemptionAttempts.Inc()
 }
 
-// RecordPodGroupEvictedPods records a single pod eviction for a pod group.
-func RecordPodGroupEvictedPods(name, namespace, uid, nodepool, action string) {
+// IncPodGroupEvictedPods records a single pod eviction for a pod group.
+func IncPodGroupEvictedPods(name, namespace, uid, nodepool, action string) {
 	podGroupEvictedPodsTotal.WithLabelValues(name, namespace, uid, nodepool, action).Inc()
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Fixed scheduler nil-pointer panic in the preempt scenario builder when a (partial) job has no tasks to allocate (`NewIdleGpusFilter` dereferenced a nil scenario); added the missing nil-guard matching the sibling filters [#1664](https://github.com/kai-scheduler/KAI-Scheduler/issues/1664) [sam-huang1223](https://github.com/sam-huang1223)
 - Fixed default node-scale-adjuster image name (`node-scale-adjuster` → `nodescaleadjuster`) so it matches the image published to GHCR
 - Fixed duplicate GPU reservation pods being created for a single `gpu-group` on a node (each reserving a different physical GPU), which corrupted the scheduler's fractional-GPU accounting and left devices unschedulable. Reservation pods are now named deterministically per (node, gpu-group) and treat AlreadyExists as success, so concurrent or retried binds collide on one object instead of duplicating [#1673](https://github.com/kai-scheduler/KAI-Scheduler/issues/1673)
+- Fixed `kai_pod_group_evicted_pods_total` counter being inflated by gang size. The metric was incremented by `EvictionGangSize` (= N) on every per-pod eviction emit, so an N-pod gang eviction wrote N² to the counter instead of N (and a cross-PodGroup batch of size N inflated each PG's counter by `tasks_in_pg × N`). All eviction-emitting actions (preempt, reclaim, consolidation, stalegangeviction) were affected. [#1620](https://github.com/kai-scheduler/KAI-Scheduler/issues/1620)
 
 ## [v0.15.0] - 2026-05-20
 
--- a/pkg/scheduler/cache/status_updater/default_status_updater.go
+++ b/pkg/scheduler/cache/status_updater/default_status_updater.go
@@ -123,7 +123,6 @@ func (su *defaultStatusUpdater) Evicted(
 		string(evictedPodGroup.UID),
 		nodepool,
 		evictionMetadata.Action,
-		evictionMetadata.EvictionGangSize,
 	)
 }
 
--- a/pkg/scheduler/cache/status_updater/default_status_updater.go
+++ b/pkg/scheduler/cache/status_updater/default_status_updater.go
@@ -117,7 +117,7 @@ func (su *defaultStatusUpdater) Evicted(
 		message)
 
 	nodepool := utils.GetNodePoolNameFromLabels(evictedPodGroup.Labels, su.nodePoolLabelKey)
-	metrics.RecordPodGroupEvictedPods(
+	metrics.IncPodGroupEvictedPods(
 		evictedPodGroup.Name,
 		evictedPodGroup.Namespace,
 		string(evictedPodGroup.UID),
--- a/pkg/scheduler/cache/status_updater/default_status_updater_test.go
+++ b/pkg/scheduler/cache/status_updater/default_status_updater_test.go
@@ -5,14 +5,18 @@ package status_updater
 
 import (
 	"errors"
+	"fmt"
 	"sync"
 	"testing"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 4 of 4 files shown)


enoodle (changes requested):
This is great, Thanks!
Can you refactor the tests so that they will address the code and not the "bug fix"  ? I am not sure that we need a test to replicate this bug as it is not a specific scenario. We can have tests to verify the metric change / event

david-gang:
> This is great, Thanks! Can you refactor the tests so that they will address the code and not the "bug fix" ? I am not sure that we need a test to replicate this bug as it is not a specific scenario. We can have tests to verify the metric change / event

addressed the comment.

enoodle (dismissed):

davidLif (dismissed):

enoodle (dismissed):

davidLif (dismissed):

david-gang:
@davidLif this pr was nearly merged and it did not merge due to some flakieness in the e2e tests

```
deployment.apps/registry created
deployment.apps/registry condition met
Run kubectl port-forward -n kube-registry deploy/registry 5000:5000 &
/home/runner/work/_temp/1c6d5fc7-5ed8-42b9-b12b-9c6d5bc4d185.sh: line 2: /mnt/images/docker_images.tgz: No such file or directory
```

can you please take a look here?

pkg/scheduler/metrics/metrics.go
```diff
@@ -365,9 +365,9 @@ func RegisterPreemptionAttempts() {
 	preemptionAttempts.Inc()
 }
 
-// RecordPodGroupEvictedPods records the number of pods evicted for a pod group
-func RecordPodGroupEvictedPods(name, namespace, uid, nodepool, action string, count int) {
-	podGroupEvictedPodsTotal.WithLabelValues(name, namespace, uid, nodepool, action).Add(float64(count))
+// RecordPodGroupEvictedPods records a single pod eviction for a pod group.
+func RecordPodGroupEvictedPods(name, namespace, uid, nodepool, action string) {
```
gshaibi: Can you please rename the function accordingly? (to something like `IncPodGroupEvictedPods`)

gshaibi (commented):
Thanks for the fix!

pkg/scheduler/metrics/metrics.go
david-gang: fixed

enoodle (dismissed):

enoodle (approved):

gshaibi (approved):

PR #1628, opened 2026-05-26
feat(scheduler): add snapshot redaction CLI tool

### LINKED ISSUE #1606: Add a reduction tool for snapshots

The snapshot tool is widely used to demonstrate issues and replicate real-world production clusters in a reproducible way. However, it the snapshot contains some potentially sensitive info that not all KAI users are comfortable sharing: pod, job, namespace and queue names, image names, volume names, secrets metadata, etc.
All of this info is metadata that doesn't effect the actual scheduling logic, and would work the same even if the names are obfuscated.
If we identify all these bits of potentially sensitive info, it should be trivial to build a tool to find-and-replace it.
This tool should also keep a translation table back from the obfuscated keys to the real ones: that way, if we identify issues in some jobs in the obfuscated snapshot, the original user can identify the original ones.

### Why is this needed?

This will improve our ability to assist and replicate issues of real users with potentially sensitive use cases

maishivamhoo123:
Resolves #1606 
This PR introduces a standalone post-processing CLI tool to redact sensitive Kubernetes metadata (like Pod names, Namespaces, ConfigMaps, etc.) from generated scheduler snapshots, allowing users to safely share their snapshots for debugging.

* **Redactor Package:** Added `pkg/scheduler/plugins/snapshot/redactor` to handle the core obfuscation logic for `RawKubernetesObjects`.
* **CLI Wrapper:** Added `cmd/snapshot-redact/main.go` to safely unzip, process, and re-zip the snapshots while generating a reusable translation table (`translation-table.json`).
* **Tests:** Added comprehensive unit tests in `redactor_test.go` to ensure deterministic obfuscation and correct cross-resource mapping.

* Added new unit tests for the `redactor` package.
* Verified `make build validate test` passes locally.

## Release Notes

* **New Features**
  * Added snapshot redaction tool to obfuscate sensitive identifiers and metadata in Kubernetes snapshots
  * Generates translation tables for consistent redaction mapping
  * Supports `--dry-run` and `--force` modes with optional redaction reports
  * Provides verbosity options and redaction statistics summary

[Review Change Stack](https://app.coderabbit.ai/change-stack/kai-scheduler/KAI-Scheduler/pull/1628?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


maishivamhoo123:
@coderabbitai review

maishivamhoo123:
@itsomri , @enoodle and @team can you please review this PR ?
Thank you!

cmd/snapshot-redact/main.go
enoodle, line 1: Why is this needed?  It would be easier to add an option to the snapshot plugin to output the redacted version and then the normal tool will handle it.

pkg/scheduler/plugins/snapshot/redactor/redactor.go
```diff
@@ -0,0 +1,647 @@
...
+					pod.Status.ContainerStatuses[i].ImageID,
+					"imageid",
+				)
+			}
+		}
+	}
+}
+
+// redactPodSpec redacts sensitive information in PodSpec
+func (r *Redactor) redactPodSpec(spec *corev1.PodSpec) {
```
enoodle: NodeName is missing

pkg/scheduler/plugins/snapshot/redactor/redactor_test.go
enoodle, line 1: should have a test that makes sure that the snapshot tool can still work with the redacted version - that we get the same scheduling results.

cmd/snapshot-redact/main.go
maishivamhoo123: @enoodle, thank you for your review ,
You're absolutely right that adding a plugin flag would be simpler initially.
reason behind choosing the CLI approach because of the operational benefits 
Plugin flag demerits:
1. Users must re-run expensive snapshot capture  if redaction logic changes
2. Redaction decision is forced at capture time, not at sharing time
3. Redaction bugs could affect the scheduler plugin's reliability
4. Changes to redaction logic require scheduler redeployment

Standalone tool advantages:
1. Post-processing: redact existing snapshots without re-capture
2. Operational flexibility: users choose when and if to redact
3. Isolation: redaction bugs don't impact scheduler plugin
4. Independent deployment: redaction updates don't require scheduler changes
6. Generates translation table for easier debugging
7.
I believe the standalone approach better serves long-term 
production operations. However, I respect your expertise in the project's needs.

Happy to discuss if you want to add the plugin flag approach or standalone tool.

pkg/scheduler/plugins/snapshot/redactor/redactor.go
```diff
@@ -0,0 +1,650 @@
...
+func NewRedactor() *Redactor {
+	return &Redactor{
+		translationTable: make(map[string]string),
+		counters:         make(map[string]int),
+		stats:            RedactionStats{},
+	}
+}
+
+// Obfuscate checks if a string is already translated. If not, it creates a new one.
+func (r *Redactor) Obfuscate(original, prefix string) string {
```
enoodle: To my understanding the table doesn't capture the prefix so it will merge different types:
Obfuscate("worker-1", "node") -> "node-1"
Obfuscate("worker-1", "pod") -> "node-1"
Will this cause issues ?

maishivamhoo123: @enoodle yaa it will cause a issue. I am sorry i missed it .  Thank you for pointing this out . let me fix it and raise me a new pr .
Thank you.

maishivamhoo123:
@enoodle , @itsomri and @team   can you please review this PR. Thank you!

pkg/scheduler/plugins/snapshot/redactor/redactor_test.go
enoodle: @maishivamhoo123 Did you add such test?

maishivamhoo123: Yes @enoodle, initially I had missed Some part of this scenario, but I’ve now added the complete test coverage for it as well.
Could you please take a look at this test function?

```go
func TestSchedulingDecisionsConsistentAfterRedaction(t *testing.T) {
	snap := &snapshot.Snapshot{
		RawObjects: &snapshot.RawKubernetesObjects{
			Pods: []*corev1.Pod{
				// Pod 1: Must match nodes with workload-type=general
				{
```

Thank you!

itsomri:
Hi @maishivamhoo123 , thank you so much for this contribution!
This looks good but there are some relationships that we need to preserve. For example, it seems that there's some bug in the queue names translation: the queue spec.Parent remains unobfuscated, so the queues' hierarchy gets broken. When the scheduler sees a queue but can't find the parent, it ignores it, so this breaks behavior.
There are a few more fields like that:

``` 
BindRequest.Spec.PodName
BindRequest.Spec.SelectedNode
```

And there might be some more - I'm trying to map them.

I have another idea: what if the obfuscation method would use some deterministic hash function on the string input, so it will always output the same value? Would that make implementation easier? We can let the user add a secret salt value to improve obfuscation. Let me know what you think.

I'm attaching a snapshot from one of our scale test so you can experiment with it.

Thank you!
[snapshot-e2e-50.zip](https://github.com/user-attachments/files/28352265/snapshot-e2e-50.zip)

maishivamhoo123:
> Hi @maishivamhoo123 , thank you so much for this contribution! This looks good but there are some relationships that we need to preserve. For example, it seems that there's some bug in the queue names translation: the queue spec.Parent remains unobfuscated, so the queues' hierarchy gets broken. When the scheduler sees a queue but can't find the parent, it ignores it, so this breaks behavior. There are a few more fields like that:
> 
> ```
> BindRequest.Spec.PodName
> BindRequest.Spec.SelectedNode
> ```
> 
> And there might be some more - I'm trying to map them.
> 
> I have another idea: what if the obfuscation method would use some deterministic hash function on the string input, so it will always output the same value? Would that make implementation easier? We can let the user add a secret salt value to improve obfuscation. Let me know what you think.
> 
> I'm attaching a snapshot from one of our scale test so you can experiment with it.
> 
> Thank you! [snapshot-e2e-50.zip](https://github.com/user-attachments/files/28352265/snapshot-e2e-50.zip)

@itsomri Thank you for your review .
you idea sound really interesting. let me try if i will get any problem or i will let you know. Thank you.

maishivamhoo123:
@itsomri and @team can you please review this pr thank you ?

Issue #1400, opened 2026-04-09
Design for limits per resource type


keinanDaniel:
What you would like to be added?
queues should be able to set quotas by nvidia.com/gpu type, like H100 or B200 or etc.

Why is this needed?
allow usage of KAI in clusters containing multiple types of GPUs (a common occurence out in the wild)

Points of considuration
* How do differenciate between the different resources? Labels? Dra attributes?
* Only Limits, or the full resource queue parameters (quota, over-quota weight, etc) ? please look for current parctices accros the landscape.
* How easy it would be to use the same implementation to allow other extended resources/other dra device classes?
* What happnes if there are quotas per gpu type, but the workload requested "generic type"?
* What happnes to the gpu-memory and gpu-compute calculations? Currently, they are translated to the "gpu-resources".
* Please detail the different user stories as part of the desgin, and try to validate it with thous that requested this feature.

PR #1635, opened 2026-05-27
docs: propose AccountedResource API

enoodle:
## Description

Adds a design proposal for the AccountedResource API for KAI.

The document covers:
- background and motivation for arbitrary accounted resources;
- concrete use cases and how the proposed API solves them;
- proposed Queue and AccountedResource API shape;
- examples for generic GPUs, GPU types, GPU memory, GPU compute, DRA, annotations, and EFA;
- validation and user feedback behavior;
- phase boundaries between limit-only accounting and later fairness/reclaim work.

This is a docs-only design PR. It does not implement the API.

## Related Issues

Related to #1400

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. Design document only.

## Additional Notes

The proposal intentionally scopes phase 1 to limit enforcement. Fairness, reclaim, status, metrics, and migration from existing spec.resources are called out as later design work.

### DIFF
--- a/docs/developer/designs/accounted-resource-api.md
+++ b/docs/developer/designs/accounted-resource-api.md
@@ -607,6 +607,34 @@ Scheduler behavior:
   through the existing `OverLimit` path. The message should identify the queue,
   `resourceRef`, current usage, candidate request, and limit.
 
+## Queue Admission Signaling
+
+This design must integrate with the queue-admission signaling work tracked in
+https://github.com/kai-scheduler/KAI-Scheduler/issues/1615.
+
+Issue #1615 covers a scheduler behavior problem: pods that cannot run because of
+queue quota or capacity limits should not look like ordinary node-resource
+failures. If KAI marks those pods as unschedulable in the same way it marks pods
+that lack cluster capacity, autoscalers such as Karpenter may provision nodes
+even though the workload is blocked by queue policy and cannot be admitted by
+adding capacity.
+
+`AccountedResource` limits introduce new queue-policy rejection reasons. Any
+solution for #1615, whether it uses scheduling gates or another admission
+signal, must apply globally to all queue-capacity failures, including:
+
+- current CPU, memory, and GPU limits from `spec.resources`;
+- every finite limit in `spec.accountedResources`;
+- default limits from `AccountedResource.spec.defaults.limit`;
+- validation failures where a queue references an unresolved
+  `AccountedResource` and jobs in that queue are blocked from scheduling.
+
+The important boundary is that AccountedResource over-limit failures are not
+node-fit failures. They should be surfaced through the same global
+queue-admission mechanism that #1615 defines, so AccountedResource limits do not
+create a new path that accidentally triggers scale-up for jobs blocked by queue
+policy.
+
 ## Phase Boundaries
 
 Phase 1 intentionally stops at limit enforcement. The proposal defines how to
--- a/docs/developer/designs/accounted-resource-api.md
+++ b/docs/developer/designs/accounted-resource-api.md
@@ -334,6 +334,21 @@ Queue semantics:
   workload, both limits are enforced.
 - `deserved` and `overQuotaWeight` are intentionally not exposed in phase 1.
 
+### Hierarchical Queue Semantics
+
+`spec.accountedResources` may be configured on any queue in the hierarchy,
+including parent queues. Parent queue limits act as aggregate ceilings for the
+subtree rooted at that queue. Leaf or child queue limits act as local ceilings.
+
+When KAI evaluates a candidate allocation for a workload in a queue, the
+allocation must fit the effective limit of that queue and every ancestor queue.
+For each queue in the path, the effective value is the explicit queue entry when
+present, otherwise `AccountedResource.spec.defaults.limit`.
+
+For example, if parent queue `research-team` has `gpu-h200` limit `10`, and
+children `ml-team` and `cv-team` each have `gpu-h200` limit `8`, each child is
+capped at `8`, while the combined subtree usage is capped at `10`.
+
 ### AccountedResource CRD
 
 An `AccountedResource` defines how one logical resource is charged.
--- a/docs/developer/designs/accounted-resource-api.md
+++ b/docs/developer/designs/accounted-resource-api.md
@@ -596,6 +596,28 @@ Phase 1 should reuse KAI's existing GPU accounting logic:
 No `AccountedResource.status` field is required in phase 1 to expose this
 classification.
 
+## Shared DRA ResourceClaims
+
+DRA `ResourceClaim`s may be shared by more than one pod. For
+`AccountedResource` accounting, KAI should only allow a shared `ResourceClaim`
+to be used by pods that belong to the same KAI queue.
+
+If a pod references an already allocated or shared `ResourceClaim` that is also
+used by a pod from another queue, KAI should reject scheduling that pod. This
+keeps `AccountedResource` usage attribution unambiguous and prevents one queue
+from consuming another queue's accounted allocation.
+
+When multiple pods in the same queue share the same allocated `ResourceClaim`,
+KAI should account the selected devices and capacities once for that queue,
+based on the claim allocation, rather than multiplying the same claim allocation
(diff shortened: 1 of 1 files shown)


docs/developer/designs/accounted-resource-api.md
```diff
@@ -0,0 +1,628 @@
...
+- Preserve the existing `spec.resources` behavior.
+- Avoid ResourceGroup-style registration requirements in phase 1.
+- Keep fairness and reclaim out of phase 1 so the limit API can be reviewed
+  independently.
+
+## Use Cases
+
+### Generic GPU Limit Plus Per-Type Limits
+
+An administrator wants queues to keep using generic GPU accounting for existing
```
kshitizlohia1994, line 61: is this intended to support `accountedResources` only on leaf queues or also on parent queues as org-wide ceilings? if parent `research-team` has `gpu-h200` limit: 10 and children `ml-team` / `cv-team` each have `gpu-h200` limit: 8, is the effective org cap 10, 16, or only per child 8?

docs/developer/designs/accounted-resource-api.md
```diff
@@ -0,0 +1,628 @@
...
+- CRD schema and CEL validate structure, quantity syntax, key syntax, and basic
+  required fields.
+- Admission webhook validates cross-field and cross-object rules, including
+  source one-of rules, source/quantity compatibility, duplicate queue refs, and
+  existing `AccountedResource` references when visible.
+- Scheduler validates runtime and cache-dependent state.
+
+Scheduler behavior:
+
+- If a queue references an `AccountedResource` that the scheduler cannot
```
kshitizlohia1994, line 634: curious to know: does broken resourceRef blocks scheduling for all jobs in the queue (even on limit: -1)? is this following the same precedence when queue has missing parent queue?

docs/developer/designs/accounted-resource-api.md
enoodle: Yes it will support it on any queue. I will make it clear in the design doc.

docs/developer/designs/accounted-resource-api.md
enoodle: Yes, it should block and we should specify that this is the reason in the "Unschedulable" messages / events.

davidLif (dismissed):

docs/developer/designs/accounted-resource-api.md
```diff
@@ -0,0 +1,677 @@
...
+
+### Hierarchical Queue Semantics
+
+`spec.accountedResources` may be configured on any queue in the hierarchy,
+including parent queues. Parent queue limits act as aggregate ceilings for the
+subtree rooted at that queue. Leaf or child queue limits act as local ceilings.
+
+When KAI evaluates a candidate allocation for a workload in a queue, the
+allocation must fit the effective limit of that queue and every ancestor queue.
+For each queue in the path, the effective value is the explicit queue entry when
```
gshaibi, line 345: So opted-out queues get the default limit, right?
I think we should consider to make opted-out queues unlimited in that sense instead of falling back to the default set on the AccountedResource CR.

What do you think?

docs/developer/designs/accounted-resource-api.md
```diff
@@ -0,0 +1,677 @@
...
+- If a queue references an `AccountedResource` that the scheduler cannot
+  resolve, this is a queue configuration error. KAI blocks scheduling for all
+  jobs in that queue until the reference is resolved, even if the queue entry
+  uses `limit: "-1"`.
+- Blocked jobs should surface an Unschedulable message and event with a
+  distinct reason such as `AccountedResourceNotFound`. The message should
+  include the queue name and missing `resourceRef`.
+- The queue should get a condition such as
+  `AccountedResourcesResolved=False`, reason `AccountedResourceNotFound`.
+- If an existing `AccountedResource` cannot be evaluated for a candidate because
```
gshaibi: IMHO we should be restrictive here and not allow quota bypass.
Alternatively, we might want to set something on the Queue's status that mentions its limit is not fully enforced or something.

But I don't have strong opinion here, so whatever you think.

docs/developer/designs/accounted-resource-api.md
enoodle: The default limit can be unlimited (I will make sure it is mentioned) - and so in effect opted-out queues will be unlimited, unless the AccountedResource explicitly changed that. This way opted-out queues will not get too much power - taking resources that we are actually trying to limit.

docs/developer/designs/accounted-resource-api.md
enoodle: I agree.. I am not sure what I meant to say here exactly, what does "cannot be evaluated for a candidate" means - I will remove it so that w we stay with the other points.

docs/developer/designs/accounted-resource-api.md
```diff
@@ -0,0 +1,674 @@
...
+```
+
+Core semantics:
+
+- `spec.sources` is required.
+- Each source sets exactly one of `resourceName`, `dra`, or `podAnnotation`.
+- `match` decides when the source applies to a candidate placement.
+- `quantity` decides how much to charge when the default quantity is not enough.
+- Sources are additive.
+- If multiple sources in the same `AccountedResource` match, KAI sums them.
```
gshaibi, line 378: What is the use case here?

enoodle: @gshaibi This is less for answering a specific usecase and more for clarifying the behavior. Between counting only the one first listed and summing them the latter seemed more reasonable.

gshaibi: I'm not sure I understand what you mean by "between counting only the first listed".
I can imagine a resource request that can be matched by several sources (such as both by DRA's opaque config and by DRA's selected capacity) and accounting it as a sum of both sources seems incorrect.
How did you imagine it?

enoodle: @gshaibi In this design we only define one way to describe DRA resource claims, the way that the resource claim itself is counted is out of scope.

This line is more about what would happen if one pod requests multiple resources that are accounted by the same AccountedResource. For example requesting GPUs both by using extended resource (`nvidia.com/gpu: 1`) and with a DRA resource claim. The only use case that I can think of is for the DRA CPU driver - but even there it is only a temporary solution.

So just to answer the "what if" question I remarked that we would sum all the resources. My previous comment mentioned that I saw 2 options:
* Only counting the first one listed on the AccountedResource resource
* Summing the values of all the resources requested through all channels

I chose the latter - summing the values.

gshaibi: OK, I got your intention. You aimed to sum *separate* resources. 
I think that is good for now. 
In the future, if we encounter *same* resources expressed by multiple `AccountedResource.spec.sources`, we might want to expand the API to support it.

docs/developer/designs/accounted-resource-api.md
```diff
@@ -0,0 +1,674 @@
...
+The API should let administrators define these accounting rules once, then let
+each queue opt into the resources it wants to limit.
+
+## Design Goals
+
+- Support arbitrary resource names, including CPU, memory, pods, Kubernetes
+  extended resources, and GPU resources.
+- Support placement-derived accounting, such as counting `nvidia.com/gpu` as
+  `gpu-h200` only when the selected node has `nvidia.com/gpu.product: H200`.
+- Support DRA resources through device class and selected device attributes.
```
gshaibi, line 48: Would we support ResourceClaims shared among pods?
We can also handle it on a separate effort.

docs/developer/designs/accounted-resource-api.md
enoodle: Maybe it is just an issue of defining the AccountedResource correctly. I don't think we should handle it actually.

docs/developer/designs/accounted-resource-api.md
enoodle: Currently for GPUs we only allow pods to share a resource claim if they belong to the same queue. I think this will have to extend to all AccountedResource's . I will update the design accordingly.

gshaibi (approved):

PR #1636, opened 2026-05-27
fix: Inject the nvidia runtime by default only to fractional pods.

### LINKED ISSUE #1543: feat(admission): scope nvidia runtime class injection to fraction pods only
## What you would like to be added?

KAI's admission webhook currently injects `runtimeClassName: nvidia` into every
pod that requests GPU access — both whole-GPU pods (`nvidia.com/gpu` resource)
and fraction pods (`gpu-fraction` / `gpu-memory` annotations). The binder
separately injects `runtimeClassName: nvidia` into GPU reservation pods.

Live cluster testing shows the runtime class is only needed for fraction pods.
For whole-GPU and reservation pods it is at best redundant, at worst breaks pod
startup.

Direction:

1. Admission webhook: stop injecting `runtimeClassName` for whole-GPU pods; keep
   injecting it for fraction pods only.
2. Binder: stop injecting `runtimeClassName` on reservation pods. Default the
   `binder.resourceReservation.runtimeClassName` Helm value to empty.
3. Reframe the admission flag: rename `admission.gpuPodRuntimeClassName` →
   `admission.gpuFractionRuntimeClassName` (or keep the old name with a
   deprecation alias) so the configuration semantics make it clear the flag only
   affects fraction pods. Default stays `nvidia` for fraction pods.
4. Remove the OpenShift-specific `--set admission.gpuPodRuntimeClassName=null`
   workaround note from the README — it was needed because the injection
   affected whole-GPU pods on CDI-enabled clusters; that case goes away.

## Why is this needed?

KAI injects `runtimeClassName: nvidia` as a defensive default added in v0.9
([PR #602](https://github.com/NVIDIA/KAI-Scheduler/pull/602)) based on an
anticipated change in GPU Operator behavior. Live testing on multiple GPU
Operator versions shows the assumption was incorrect for whole-GPU and
reservation pods, and the injection is currently the root cause of #1097 and
#1151.

### What the toolkit actually configures (verified, GPU Operator v25.10.1, containerd)

The NVIDIA Container Toolkit writes a containerd drop-in config on every GPU
node. The content depends on whether CDI is enabled.

**CDI disabled** — toolkit makes nvidia the default containerd runtime:
```toml
[plugins."io.containerd.cri.v1.runtime".containerd]
  default_runtime_name = "nvidia"
```
Every pod on a GPU node runs through the nvidia container runtime regardless of
`runtimeClassName`. KAI's injection is redundant.

**CDI enabled** (default since GPU Operator v25.10.0):
```toml
[plugins."io.containerd.cri.v1.runtime"]
  enable_cdi = true
[plugins."io.containerd.cri.v1.runtime".containerd]
  default_runtime_name = "runc"
```
For pods that go through the device plugin (`nvidia.com/gpu` request), the
device plugin adds CDI annotations to the container spec and containerd injects
the GPU natively — no `runtimeClassName` needed. Explicitly setting
`runtimeClassName: nvidia` routes the pod through the nvidia-container-runtime's
`management.nvidia.com` CDI path, causing:
```
Error: failed to inject CDI devices: unresolvable CDI devices
management.nvidia.com/gpu=GPU-<UUID>: unknown
```

**GPU Operator v26.3.0+ with NRI Plugin** — the nvidia runtime class object is
not created at all; setting it causes pod rejection. Verified on GPU Operator
v26.3.1: reservation pods (whole-GPU path) work without a runtime class, but
fraction pods get no GPU injection because the runtime class workaround that
exists in CDI mode isn't available either.

### Why fraction pods are different

Fraction pods request fractional GPU access via KAI annotations rather than
through a `nvidia.com/gpu` resource. Their pod spec has empty resources, so the
device plugin's `Allocate()` is never called for them and no CDI annotations are
added. KAI's binder sets `NVIDIA_VISIBLE_DEVICES` via a ConfigMap instead, and
that env var is only meaningful to the legacy nvidia-container-runtime — which
is invoked via `runtimeClassName: nvidia`.

Verified on GPU Operator v25.10.1 with CDI enabled:
- Fraction pod with no `runtimeClassName` → no GPU injection at all (no
  `/dev/nvidia*`, no `nvidia-smi`).
- Fraction pod with `runtimeClassName: nvidia` → nvidia-container-runtime reads
  `NVIDIA_VISIBLE_DEVICES`, injects the GPU via CDI, pod sees devices correctly.

So the runtime class is genuinely load-bearing for fraction pods on CDI-enabled
clusters and must be preserved by default. On GPU Operator v26.3.0+ with NRI,
fraction pods are broken regardless because the runtime class object doesn't
exist — that's a separate gap tracked in **#1544**.

### Summary

| Scenario | Effect of KAI injecting `runtimeClassName: nvidia` |
|---|---|
| Containerd, CDI disabled | Redundant — nvidia is already the default runtime |
| Containerd, CDI enabled (v25.10.0+ default) | **Harmful** for whole-GPU pods — routes to `management.nvidia.com` CDI path, causing injection failures |
| GPU Operator v26.3.0+ with NRI Plugin (verified on v26.3.1) | **Harmful** — runtime class object does not exist, pod rejected |
| Fraction pods, CDI enabled | **Necessary** — device plugin doesn't add CDI annotations for these pods |
| Fraction pods, NRI Plugin (verified on v26.3.1) | Cannot be set — runtime class doesn't exist; fraction pods broken regardless (see **#1544**) |

## Related

- #1097 — disabling runtime class override was ignored in v0.13.0-rc1
- #1151 — empty-string workaround in v0.13.0; comment thread explicitly planned this issue
- https://github.com/NVIDIA/KAI-Scheduler/pull/1035 — fix that made empty string work
- https://github.com/NVIDIA/KAI-Scheduler/pull/602 — original PR that added the default
- **#1544** — follow-up to remove the runtime class dependency for fraction pods entirely (CDI/NRI native injection)

davidLif:
## Description

Inject the nvidia runtime by default only to fractional pods. Deprecate the old runtime config field

## Related Issues

Fixes #1543 

## Checklist

- [x] Self-reviewed


cmd/binder/app/options.go
```diff
@@ -106,9 +105,6 @@ func InitOptions(fs *pflag.FlagSet) *Options {
 	fs.Var(&options.Plugins,
 		"plugins",
 		"JSON-serialized binder plugin configuration keyed by plugin name")
-	fs.StringVar(&options.RuntimeClassName,
```
enoodle, line 109: Don't remove this option - It allows overriding the runtime class of the reservation pods and can be useful in some scenarios.

davidLif: But according to the ticket https://github.com/kai-scheduler/KAI-Scheduler/issues/1543, we wanted to stop injecting a runtimeclass to reservation pods. Do we still want to keep the option to inject them with a runtime class? If we need to inject reservation pods, does that mean that we want to inject the same runtimeclass to whole gpu pods?

enoodle: It's an option to change that, the default should be empty - not to add runtime class

davidLif: Changed according to the comment

README.md
```diff
@@ -94,12 +94,8 @@ Follow the instructions [here](docs/developer/building-from-source.md)
 
 ### OpenShift
 
-When `gpu-operator` <v25.10.0 is installed, the following flag should be added to the installation command:
+If CDI is enabled, add `--set binder.cdiEnabled=true` to the installation command.
 
-```
---set admission.gpuPodRuntimeClassName=null
```
enoodle, line 100: I think that we will still need it for fractions on Openshift

davidLif: Fixed

enoodle (approved):

PR #1638, opened 2026-05-27
docs: cross topology example

enoodle:
## Description

Example and unit test for jobs that use more than one topology CRD, one for each sub group.

## Related Issues

Fixes #

## Checklist

- [x] Self-reviewed

## Breaking Changes

None

### DIFF
--- a/hack/setup-e2e-cluster.sh
+++ b/hack/setup-e2e-cluster.sh
@@ -11,6 +11,7 @@ CLUSTER_NAME=${CLUSTER_NAME:-e2e-kai-scheduler}
 
 REPO_ROOT=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/..
 : ${FEATURE_CONFIG:="default"}
+KIND_CONFIG=${KIND_CONFIG:-""}
 
 : ${KIND_K8S_TAG:="v1.35.0"}
 : ${KIND_IMAGE:="kindest/node:${KIND_K8S_TAG}"}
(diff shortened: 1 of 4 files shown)


hack/setup-e2e-cluster.sh
```diff
@@ -54,28 +59,36 @@ while [[ $# -gt 0 ]]; do
...
 done
 
-GENERATED_KIND_CONFIG=$(mktemp "${TMPDIR:-/tmp}/kind-config-XXXXXX.yaml")
-trap "rm -f \"$GENERATED_KIND_CONFIG\"" EXIT
-${REPO_ROOT}/hack/generate-kind-config.sh \
-    --feature-config "$FEATURE_CONFIG" \
-    --k8s-version "$KIND_K8S_TAG" \
-    --output "$GENERATED_KIND_CONFIG"
+if [[ -n "$KIND_CONFIG" && "$FEATURE_CONFIG" != "default" ]]; then
+  echo "--feature-config cannot be used together with --kind-config"
```
davidLif, line 77: You renamed the feature-config flag to kind-config

hack/setup-e2e-cluster.sh
```diff
@@ -54,28 +59,36 @@ while [[ $# -gt 0 ]]; do
...
-
 # Deploy local image registry
 echo "Deploying local image registry..."
 kubectl apply -f ${REPO_ROOT}/hack/local_registry.yaml
 kubectl wait --for=condition=available --timeout=60s deployment/registry -n kube-registry
 
 # Install the fake-gpu-operator to provide fake GPU resources for the e2e tests
 DRA_PLUGIN_ENABLED="false"
-if [ "$FEATURE_CONFIG" = "dra-enabled" ]; then
+if [ -z "$KIND_CONFIG" ] && [ "$FEATURE_CONFIG" = "dra-enabled" ]; then
```
davidLif: Why does the KIND_CONFIG here nees to be empty?

hack/setup-e2e-cluster.sh
enoodle: Thank you, you are right, I meant only to add one.

hack/setup-e2e-cluster.sh
enoodle: because we check that both are not set together it is not required here.

hack/setup-e2e-cluster.sh
enoodle: I see that it was only removed from the doc, not from the actual options, I will fix it! FEATURE_CONFIG and KIND_CONFIG are two different variables but only one of them can be defined

hack/setup-e2e-cluster.sh
```diff
@@ -54,20 +60,28 @@ while [[ $# -gt 0 ]]; do
...
+  CLUSTER_KIND_CONFIG="$GENERATED_KIND_CONFIG"
+fi
 
 kind create cluster \
-    --config "$GENERATED_KIND_CONFIG" \
+    --config "$CLUSTER_KIND_CONFIG" \
     --image "${KIND_IMAGE}" \
     --name "$CLUSTER_NAME"
 
-rm -f "$GENERATED_KIND_CONFIG"
```
davidLif, line 69: And what code would clear the generated kind config after the change?

davidLif (approved):

PR #1648, opened 2026-05-29
perf(scheduler): shrink cached pod objects

### LINKED ISSUE #1646: Scheduler: strip unused Pod fields in cache transforms to reduce memory
Parent: #1634

## Goal
Reduce retained scheduler Pod cache size by stripping fields that scheduler code does not use before Pods enter the informer store.

## Scope
- Add a scheduler Pod cache transform using client-go informer transform support.
- Preserve metadata required for cache keys and scheduler behavior: name, namespace, UID, resource version, generation, labels, annotations, owner references, finalizers, deletion timestamp, and creation timestamp.
- Preserve scheduling-relevant spec/status fields: scheduler name, node name, priority, priority class name, containers and init containers with names/resources, init restart policy, volume mounts, host ports, configmap refs from `env` and `envFrom`, resource claims, volumes, affinity, node selector, tolerations, topology spread constraints, scheduling gates, overhead, runtime class name, phase, conditions, nominated node name, and start time.
- Preserve ephemeral-container volume mounts and configmap refs from `env` and `envFrom` because the scheduler ConfigMap predicate inspects them.
- Strip large unused fields such as literal env values, commands, args, images, probes, lifecycle hooks, security contexts, managedFields, and other non-scheduling fields after coverage proves they are unused.

## Acceptance Criteria
- Scheduler behavior is unchanged for resource requests, PodGroup membership, node assignment, affinity, configmap predicate behavior, host-port signature, DRA claim handling, GPU annotations, and status handling.
- Transform tests prove stripped fields are removed and preserved fields remain available.
- Transform does not mutate the input object.
- Scale/heap evidence compares scheduler Pod cache memory before and after the transform on many active Pods with large env values.

enoodle:
## Description

To reduce the scheduler memory footprint we will transform the pods that the scheduler saves in its internal cache to a minimal size required for scheduling

## Related Issues

Fixes #1646 

## Checklist

- [x] Self-reviewed
- [x] Measure memory consumption change - It is only directly impacted by the pods environment variables that are filtered. For the scale tests this is not reducing the critical paths of memory consumption (in the specific scale tests that we currently have in this repository)


pkg/scheduler/cache/cache.go
```diff
@@ -156,6 +156,10 @@ func newSchedulerCache(schedulerCacheParams *SchedulerCacheParams) *SchedulerCac
 	)
 
 	sc.informerFactory = informers.NewSharedInformerFactory(sc.kubeClient, 0)
+	if err := setSchedulerPodTransform(sc.informerFactory.Core().V1().Pods().Informer()); err != nil {
```
gshaibi, line 193: Should we put it behind a configuration? It might effect performance on large clusters and I think we should have a way to disable it by demand.

gshaibi (dismissed):

davidLif (approved):

PR #1661, opened 2026-06-02
feat(chart): Add ability to change security and resources for the Helm chart

treydock:
## Description

Add several new configuration options for Helm.

* Ensure all Job and Deployment resources can support securityContexts
* Ensure Jobs and Deployments can have resources defined

## Checklist

- [x] Self-reviewed


enoodle:
@treydock Why add resources configuration to those helm jobs and not just set some constant? I am not sure that it is really helpful to play with this value and maybe we  should keep it a little bit more simple by just setting a value.

treydock:
> Why add resources configuration to those helm jobs and not just set some constant? I am not sure that it is really helpful to play with this value and maybe we should keep it a little bit more simple by just setting a value.

Setting some value is good but I'd imagine it's possible the value set would need to be overridden.  I figured leaving empty leaves it undefined like before but allows sites that want resource limits to be able to set them.  We use Kyverno to ensure things we deploy are setting resource requests and limits so having some value set is necessary but also can't assume a hardcoded default is always sufficient I'd imagine.

enoodle (approved):

itsomri (approved):

PR #1662, opened 2026-06-03
docs: NUMA awareness plugin design

itsomri:
## Description

Add a proposal for NUMA aware plugin.

## Related Issues

Fixes #

## Checklist

- [ ] Self-reviewed


docs/developer/designs/numa-topology/README.md
```diff
@@ -0,0 +1,824 @@
...
+- **Stay correct under concurrency and preemption.** Concurrent placement decisions must not
+  over-commit a NUMA zone, and preempting or reclaiming for a topology-sensitive pod must avoid
+  evicting victims that would not actually free a usable aligned slot.
+- **Keep adoption cost low.** Build on the standard `NodeResourceTopology` tooling already common
+  in the ecosystem; require no mandatory new cluster components, keeping richer accuracy and
+  broader policy coverage as opt-in enhancements.
+
+## Non-Goals
+
+- **Fractional / MIG GPU sharing.** Only whole-GPU (`RequestTypeRegular`, integer
```
gshaibi: What if the node's NUMA zones doesn't contain GPU? Should we try to numa-align non-whole-gpu workloads there?

enoodle: If the node is in "restricted" or "single-numa" we have to do that otherwise kubelet might fail the pod after binding and we don't want this.

docs/developer/designs/fake-nrt/README.md
enoodle, line 1: @itsomri Do you want this to be implemented in this repository  ? Maybe it should be somewhere else as a stand alone / part of FGO  ?

docs/developer/designs/numa-placement-exporter/README.md
```diff
@@ -0,0 +1,195 @@
...
+- Influencing the kubelet's placement. The agent is read-only with respect to allocation; it
+  observes and reports, it does not hint or pin.
+- Supporting pods the kubelet does not NUMA-align (non-Guaranteed). Those have no exclusive
+  placement to report.
+
+## Design Details
+
+### Architecture
+
+A DaemonSet on each NUMA/GPU node. Each instance:
```
enoodle, line 56: Is it NUMA or GPU ? 
How are the NUMA nodes going to be labeled?

docs/developer/designs/numa-topology/README.md
itsomri: Actually I think that NUMA policy is only enforced for resources that appear in the NRT - notice that pods were able to schedule in our test cluster before we fixed the GPU topology publishing. It's weird but it is what it is. We will be able to enforce better policies when we get to design scheduler-side numa policies - v1 only aims to match the kubelet decisions

docs/developer/designs/fake-nrt/README.md
itsomri: Yeah it's a part of FGO. I will remove this doc from the repo

docs/developer/designs/numa-placement-exporter/README.md
itsomri: NUMA. I don't really know yet, will mark it as an open question.

docs/developer/designs/numa-topology/README.md
```diff
@@ -0,0 +1,781 @@
...
+### Caveats
+
+- **Accuracy depends on a healthy agent.** With the flag on but the agent absent or badly lagging,
+  reconstruction degrades to *predicted-only* — the very mode this avoids. The operator only enables
+  the flag alongside the agent; the plugin should additionally treat a pod running well beyond the
+  agent's report interval with no observed annotation as an agent-health signal (log/metric), so the
+  degradation is visible rather than silent.
+- **A pod with neither observed nor predicted placement** is omitted from the subtraction (never
+  guess a zone) → a transient per-zone over-report on its zone. With the agent covering all pods this
+  is limited to the bind→observe window of KAI's own pods, where the predicted record covers it.
```
enoodle, line 757: This is the only gap we still have then? If the agent didn't put the annotation (and the NRT was not updated) on a pod before the scheduler takes a snapshot then we will be at risk of submitting a pod that will be rejected?
Can we somehow give the agent some more time? skip the current scheduling session a few times maybe?

itsomri: There's also the possibility of races when multiple pods are scheduled to the same node at the same time - the order of pods' allocation can, in some cases, matter for NUMA placement (theoretically there could be cases where allocating pod A and then pod B is possible, but B and then A is not, and bind order is not guaranteed today).
Regarding the mitigation for this gap: I think that skipping the whole session is problematic in large clusters: when you have thousands of nodes, you could have 2-3 stale nodes at any given time just by chance, or you could have a few malfunctioning NUMA placement exporters - you wouldn't want the entire cluster to be affected by that. You also don't want to completely ignore the node, because you risk unnecessary evictions. Maybe we can convert all NUMA-sensitive allocations on stale nodes to pipelined, and let the scheduler self-correct on following cycles? That way you "defer" the actual placement without being destructive.

enoodle: Sounds good. We can also only create one bind request per node per session, and leave the other pods in "pipelined" status for the next session to mitigate the first problem that you raised.

itsomri: I updated the design doc, marked it as "follow up" until we get a chance to experiment with it

enoodle (approved):

PR #1670, opened 2026-06-07
refactor(api): remove never-populated PodGroup status fields

### LINKED ISSUE #1650: PodGroup status.phase and pod-state counters are defined in the API but never populated
### Summary
The `PodGroup` CRD exposes `status.phase` (type `PodGroupPhase`) and the per-state pod counters `status.running` / `succeeded` / `failed` / `pending` in its served schema, but no controller ever writes them. Consumers reading `podgroup.status.phase` to determine whether a job is running/completed/failed get an empty string in all states.

### Version
Observed on a live cluster running **v0.14.1**. Verified against source at **v0.14.1** and current `main` (`v0.14.0-92-g0204b4cc`) — behavior is identical on both.

### API definition
`pkg/apis/scheduling/v2alpha2/podgroup_types.go`:
```go
type PodGroupStatus struct {
    Phase      PodGroupPhase ...         // line 144
    Conditions []PodGroupCondition ...   // line 148
    Running    int32 ...                 // line 152
    Succeeded  int32 ...                 // line 156
    Failed     int32 ...                 // line 160
    Pending    int32 ...                 // line 164
    ...
}
type PodGroupPhase string                // line 176 — bare type, no constants defined
```
`PodGroupPhase` has **no constants** (no `PodGroupRunning`, `PodGroupPending`, etc.). These fields appear to be leftovers from the kube-batch/Volcano `PodGroup` lineage (file header is `Copyright 2017 The Kubernetes Authors`) and are present unchanged since the repo's initial commit.

These fields are part of the **served** CRD schema (`deployments/kai-scheduler/crds/scheduling.run.ai_podgroups.yaml`, `phase`/`running`/`succeeded`/`failed`/`pending`, each with a description), so they're discoverable and look authoritative to API consumers.

### No writer exists
The only writers of `PodGroup.status` in the tree are:
- `getStatusWithMetadata` (`pkg/podgroupcontroller/controllers/patcher/pod_group.go:38-50`) — sets only `ResourcesStatus.{Requested, Allocated, AllocatedNonPreemptible}`.
- the scheduler status updater (`pkg/scheduler/cache/status_updater/default_status_updater.go`) — sets only `Status.SchedulingConditions` (no reference to `Phase`).

There is no assignment to `Status.Phase`, `Status.Running/Succeeded/Failed/Pending`, or `Status.Conditions` anywhere in non-test code.

### Live-cluster evidence (v0.14.1)
- **Running job**: no `status.phase` field; `status.resourcesStatus.requested`/`allocated` populated (cpu/gpu/rdma). (PodGroup was gang-pending/unschedulable per `schedulingConditions`, yet still had requested/allocated.)
- **Completed job**: `status: { resourcesStatus: {} }` — empty, no phase.
- **Failed job**: `status: { resourcesStatus: {} }` — empty, no phase.

`status.phase` was empty in all three states.

### Downstream impact
Tooling that reads `podgroup.status.phase` to distinguish live vs. terminated PodGroups silently receives `""` for every PodGroup in every state. The fields are a trap: they're in the schema and documented, but inert. Consumers must instead fall back to `status.resourcesStatus.requested` (summed over active pods) and `status.schedulingConditions`, which is undocumented as the intended liveness signal.

### The gap
`phase` isn't just unwired — there's no defined semantics behind it. There are no `PodGroupPhase` constants and no transitions, so it's unclear what the field is even meant to express in KAI's model: what "Running" means for a gang (all `minMember` running, or any?), how elastic / subgroup PodGroups map onto a single phase, when it would move to `Completed` vs. `Failed`. Volcano's original semantics don't obviously carry over. The counters (`running`/`succeeded`/`failed`/`pending`) are a simpler case — the controller already iterates every pod — but they're equally inert today.

### Question
What's the intended direction for these fields? Broadly the options seem to be: (a) populate `phase` and the counters with semantics defined for KAI's gang model, or (b) deprecate/remove the unused fields and document `resourcesStatus` + `schedulingConditions` as the supported way to determine PodGroup liveness. Happy to help once there's a direction.

david-gang:
## Description

The `PodGroup` (`scheduling.run.ai/v2alpha2`) status schema exposes `phase`, `running`, `succeeded`, `failed`, and `pending`, but no controller ever populates them — they are inert leftovers from the kube-batch/Volcano lineage. They appear authoritative in the served schema yet always read empty/zero, trapping consumers.

This PR removes those fields (and the `PodGroupPhase` type). Per the issue discussion, liveness should be determined via `status.resourcesStatus` and `status.schedulingConditions`, which remain. The `conditions` field is kept untouched.

- Removed `status.phase` / `running` / `succeeded` / `failed` / `pending` and the `PodGroupPhase` type from `pkg/apis/scheduling/v2alpha2/podgroup_types.go`
- Regenerated CRD manifest and deepcopy (`make generate manifests`)
- Updated two tests that used the removed fields as arbitrary filler to use the surviving `Conditions` field, preserving each test's original assertion

## Related Issues

Fixes #1650

## Checklist

- [x] Self-reviewed

## Breaking Changes

These fields are removed from the served `v2alpha2` schema. Because `v2alpha2` is an **alpha** version (no compatibility guarantee under the Kubernetes API deprecation policy) and the fields belong to `status` (reconstructed by controllers, never written here), no new API version or conversion is required. On upgrade the CRD applies cleanly; existing PodGroups carry no data in these fields, so nothing is lost. Any consumer reading `status.phase`/counters (always empty today) must switch to `status.resourcesStatus` + `status.schedulingConditions`.

## Additional Notes

CHANGELOG updated under a new `Removed` section.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Updated Go toolchain and base build images to v1.26.3.
 - **Breaking:** The podgroup produced for JobSet is now produces as a single PodGroup per JobSet with a two-level SubGroup hierarchy (one parent SubGroup per `replicatedJob`, one leaf SubGroup per replica) regardless of `startupPolicyOrder`. The `kai.scheduler/batch-min-member` annotation on the JobSet now overrides the root `minSubGroup`; the same annotation on `replicatedJobs[].template.metadata.annotations` overrides the leaf `minMember` (defaulting to `template.spec.parallelism`). [#1617](https://github.com/kai-scheduler/KAI-Scheduler/pull/1617) [davidLif](https://github.com/davidLif)
 
+### Removed
+- Removed the never-populated `status.phase`, `status.running`, `status.succeeded`, `status.failed`, and `status.pending` fields (and the `PodGroupPhase` type) from the `PodGroup` (`scheduling.run.ai/v2alpha2`) schema. No controller ever wrote them; use `status.resourcesStatus` and `status.schedulingConditions` to determine PodGroup liveness. [#1650](https://github.com/kai-scheduler/KAI-Scheduler/issues/1650) [david-gang](https://github.com/david-gang)
+
 ### Fixed
 - Fixed post-delete cleanup hook hardcoding `kai-scheduler` namespace instead of Helm release namespace on `helm uninstall` [#1619](https://github.com/kai-scheduler/KAI-Scheduler/pull/1619) [dttung2905](https://github.com/dttung2905)
 - Improved solver performance in some large reclaim scenarios [#1627](https://github.com/kai-scheduler/KAI-Scheduler/pull/1627) [itsomri](https://github.com/itsomri)
--- a/deployments/kai-scheduler/crds/scheduling.run.ai_podgroups.yaml
+++ b/deployments/kai-scheduler/crds/scheduling.run.ai_podgroups.yaml
@@ -212,17 +212,6 @@ spec:
                       type: string
                   type: object
                 type: array
-              failed:
-                description: The number of pods which reached phase Failed.
-                format: int32
-                type: integer
-              pending:
-                description: The number of pods which reached phase Pending.
-                format: int32
-                type: integer
-              phase:
-                description: Current phase of PodGroup.
-                type: string
               resourcesStatus:
                 description: Status of resources related to pods connected to this
                   pod group.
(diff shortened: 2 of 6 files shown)


enoodle:
```
Error: vet: pkg/queuecontroller/controllers/suite_test.go:248:5: unknown field Running in struct literal of type v2alpha2.PodGroupStatus
```
@david-gang seems that this test also needs updating

david-gang:
> ```
> Error: vet: pkg/queuecontroller/controllers/suite_test.go:248:5: unknown field Running in struct literal of type v2alpha2.PodGroupStatus
> ```
> 
> @david-gang seems that this test also needs updating

fixed. thanks

enoodle (approved):

gshaibi (approved):

PR #1672, opened 2026-06-08
feat(admission): add opt-in deviceaccess plugin for NVIDIA_VISIBLE_DEVICES

gshaibi:
## Description

Adds an **opt-in** `deviceaccess` admission plugin that controls how pods may use the
`NVIDIA_VISIBLE_DEVICES` environment variable. When enabled it:

1. **Validates** — rejects pods that set `NVIDIA_VISIBLE_DEVICES` to a value other than
   `void`/`none` (or via a `valueFrom` reference) on any non-fraction container, since such
   values bypass/conflict with NVIDIA's device plugin.
2. **Mutates** — injects `NVIDIA_VISIBLE_DEVICES=void` into every container that does **not**
   request a GPU (whole-GPU or MIG), blocking those containers from accessing GPUs on the node.

The GPU-fraction container legitimately sets/uses this variable (via the GPU-sharing
ConfigMap) and is exempt in both paths, identified via `GetFractionContainerRef`.

Disabled by default; OSS clusters are unaffected unless explicitly enabled:
- CLI: `--block-nvidia-visible-devices`
- Config: `spec.admission.blockNvidiaVisibleDevices: true`
- Helm: `--set global.blockNvidiaVisibleDevices=true`

A single flag gates both behaviors. The admission operand emits the flag when the config
field is set, following the existing `hamicore`/`gpuSharing` pattern.

## Related Issues

N/A

## Checklist

- [x] Self-reviewed
- [x] Added/updated tests (plugin Validate + Mutate unit tests, `IsMigResource`, operand + helm unit tests, e2e suite)
- [x] Updated documentation (CHANGELOG)

## Breaking Changes

None — the feature is opt-in and disabled by default.

## Additional Notes

- New e2e suite `test/e2e/suites/admission/deviceaccess` toggles the flag at runtime and
  asserts: rejection of `all`/`1`/`1,2`, admission of `void`/`none`, a flag-off case, and that
  a non-GPU sidecar gets `NVIDIA_VISIBLE_DEVICES=void` while the GPU container does not.
- Adds `resources.IsMigResource` + `constants.NvidiaMigResourcePrefix` (needed by the mutation
  to recognise MIG device requests).

### DIFF
--- a/pkg/apis/kai/v1/admission/admission.go
+++ b/pkg/apis/kai/v1/admission/admission.go
@@ -32,9 +32,8 @@ type Admission struct {
 	// +kubebuilder:validation:Optional
 	GPUSharing *bool `json:"gpuSharing,omitempty"`
 
-	// BlockNvidiaVisibleDevices enables the device access plugin, which validates that
-	// pods do not override the NVIDIA_VISIBLE_DEVICES environment variable with
-	// values that conflict with NVIDIA's device plugin.
+	// BlockNvidiaVisibleDevices prevents pods from overriding the NVIDIA_VISIBLE_DEVICES
+	// environment variable, which would conflict with NVIDIA's device plugin.
 	// +kubebuilder:validation:Optional
 	BlockNvidiaVisibleDevices *bool `json:"blockNvidiaVisibleDevices,omitempty"`
 
--- a/pkg/apis/kai/v1/admission/admission.go
+++ b/pkg/apis/kai/v1/admission/admission.go
@@ -32,6 +32,12 @@ type Admission struct {
 	// +kubebuilder:validation:Optional
 	GPUSharing *bool `json:"gpuSharing,omitempty"`
 
+	// BlockNvidiaVisibleDevices enables the device access plugin, which validates that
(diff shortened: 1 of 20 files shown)


pkg/apis/kai/v1/admission/admission.go
```diff
@@ -32,6 +32,12 @@ type Admission struct {
 	// +kubebuilder:validation:Optional
 	GPUSharing *bool `json:"gpuSharing,omitempty"`
 
+	// BlockNvidiaVisibleDevices enables the device access plugin, which validates that
```
itsomri: It's not really a plugin though is it?
Maybe just document it with a simple "prevents overriding the NVIDIA_VISIBLE_DEVICES env var" or something like this?

itsomri (dismissed):

pkg/apis/kai/v1/admission/admission.go
gshaibi: sure, will do

gshaibi: Done — reworded to describe the behavior (no longer calls it "a plugin") in dd0f59b0.

itsomri (approved):

Issue #1598, opened 2026-05-20
Support gpu NUMA topology aware scheduling


Hagay-RunAI:
### What you would like to be added?

Currently KAI Scheduler is not NUMA topology aware which means that it may place a workload (Pod) into a node where the CPU cores, CPU memory and GPUs are not on the same NUMA node which degrades the perfomance of that workload (Pod).

KAI scheduler places multi-GPU workloads on nodes where the aggregate free GPU count satisfies the request, but the per-NUMA-node free GPU distribution does not. Kubelet's TopologyManager (in case of for example: restricted policy) rejects every pod on admission, and KAI scheduler retries to the same node until backoff is exhausted, causing deterministic job failure. 
Adding NUMA topology awarness so KAI scheduler knows which GPUs are attched to which NUMA node, and how much CPU cores and memory are avaialble on each NUMA node can asisst the scheduler to choose nodes where there's enough GPU, CPU core and CPU memory on the same NUMA node or minimal span of NUMA nodes. 

KAI Scheduler must support the different NUMA topology aware scheduling policies: None, Best-effort, Restricted and SingleNumaNode.

### Why is this needed?

Currently KAI Scheduler is not NUMA topology aware which means that it may place a workload (Pod) into a node where the CPU cores, CPU memory and GPUs are not on the same NUMA node which degrades the perfomance of that workload (Pod).

PR #1674, opened 2026-06-09
feat: NUMA awareness plugin

itsomri:
## Description

This PR adds a numa plugin implementation as described in #1662. A few caveats:
- This PR does not implement the predicted placement fallback strategy - that's in a follow up PR, so reclaim/preempt cases will not work without a NUMA placement agent daemonset (which is out of scope as well for this PR)
- This PR is not intended to work efficiently in high scale environments. Potential performance improvements were already considered and documented, and will be implemented once we have a better idea of where the bottlenecks are. These include:
  - Representing NUMA zone resource availability using resource vectors
  - Pre-calculating things at plugin open instead of at allocation evaluation time - for example, the preferred placement of a pod on a node is static per (pod resource request, node NRT allocatable resources) - can be cached pretty easily
- Part of the feature (cross-cycle predicted placement persistence) is meant to be implemented in another PR to keep the review scope manageable. Since the plugin is disabled by default, the changelog entry will be added when the implementation of v1 is completed.

## Related Issues

Implements #1662 #1598

## Checklist

- [x] Self-reviewed

## Release Notes

* **New Features**
  * Added NUMA-aware scheduling plugin that enforces Topology Manager policies (single-numa-node, restricted) to optimize pod placement across NUMA zones.
  * Scheduler now tracks per-zone resource availability and enforces per-node NUMA topology constraints during task placement decisions.
  * Support for observing actual pod NUMA placement via optional placement agent, enabling improved scheduling accuracy.


pkg/scheduler/plugins/numa/evaluator.go
```diff
@@ -0,0 +1,334 @@
...
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/api/node_info"
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/api/pod_info"
+)
+
+type resourceAmounts = map[v1.ResourceName]resource.Quantity
```
enoodle: isn't it similar to ResourceList from "k8s.io/api/core/v1"  ? why not use the existing one that is well known?

pkg/scheduler/plugins/numa/evaluator_test.go
```diff
@@ -0,0 +1,221 @@
...
+		allocation, admit := restrictedEvaluator{}.evaluate(node, noIgnoreList, []resourceAmounts{req(gpu, "6", "cpu", "24")})
+		assert.True(t, admit)
+		assert.Equal(t, []int{0, 1}, allocation.ZoneIndices(), "spans both NUMA zones")
+
+		gpu0, gpu1 := amountAt(allocation, 0)[gpu], amountAt(allocation, 1)[gpu]
+		totalGPU := gpu0.Value() + gpu1.Value()
+		assert.Equal(t, int64(6), totalGPU, "the full GPU request is allocated across the mask")
+	})
+
+	t.Run("reject: 4-GPU + 1-CPU footgun", func(t *testing.T) {
```
enoodle, line 149: Please improve the naming of scenario and function names here.

pkg/scheduler/plugins/numa/evaluator.go
itsomri: No reason actually. Changed

pkg/scheduler/plugins/numa/evaluator_test.go
itsomri: Changed to `TestRestrictedAllocatableVsAvailable`

enoodle (approved):

PR #1694, opened 2026-06-15
fix(binder): prevent duplicate reservation pods for a single gpu-group

KaiPilotBot:
# Description
Backport of #1693 to `v0.15`.


enoodle:
We don't support v0.15 anymore

PR #1696, opened 2026-06-15
docs(scheduler): add reclaim generator portfolio design

enoodle:
## Description

Adds a standalone developer design for a bounded reclaim scenario generator portfolio.

- Documents the generator abstraction, plugin registration/order, and synchronous driver loop.
- Defines the time-only budget model, including alpha/experimental budget configuration knobs.
- Specifies production metrics, test coverage, rollout criteria, and alternatives.
- Keeps generator selection tied to normal scheduler plugin enablement and order.

## Related Issues

N/A

## Checklist

- [x] Self-reviewed

## Breaking Changes

None.

## Additional Notes

Docs-only PR. Verified with `git diff --check origin/main...HEAD`.

### DIFF
--- a/docs/developer/designs/reclaim-generator-portfolio-design.md
+++ b/docs/developer/designs/reclaim-generator-portfolio-design.md
@@ -175,31 +175,31 @@ kind: SchedulingShard
 spec:
   scenarioSearchBudgets:
     maxActionSearchDuration:
-      default: "1s"
-      reclaim: "2s"
-      preempt: "1s"
-      consolidation: "1s"
-    maxJobSearchDuration: "250ms"
+      default: "5m"
+      reclaim: "5m"
+      preempt: "5m"
+      consolidation: "5m"
+    maxJobSearchDuration: "4m"
     minJobSearchDuration: "0s"
     maxGeneratorSearchDuration:
-      default: "250ms"
-      NodeLocalGreedy: "50ms"
-      MultiNodeGang: "250ms"
+      default: "2m"
+      NodeLocalGreedy: "30s"
+      MultiNodeGang: "2m"
 ```
 
 Initial defaults when the block is omitted:
 
 | Field | Default | Meaning |
 | --- | --- | --- |
-| `maxActionSearchDuration.default` | `"1s"` | fallback action budget |
-| `maxActionSearchDuration.reclaim` | `"2s"` | reclaim-specific action budget |
-| `maxActionSearchDuration.preempt` | `"1s"` | preempt-specific action budget |
-| `maxActionSearchDuration.consolidation` | `"1s"` | consolidation-specific action budget |
-| `maxJobSearchDuration` | `"250ms"` | per-job budget cap |
+| `maxActionSearchDuration.default` | `"5m"` | fallback action budget |
+| `maxActionSearchDuration.reclaim` | `"5m"` | reclaim-specific action budget |
+| `maxActionSearchDuration.preempt` | `"5m"` | preempt-specific action budget |
+| `maxActionSearchDuration.consolidation` | `"5m"` | consolidation-specific action budget |
+| `maxJobSearchDuration` | `"4m"` | per-job budget cap |
 | `minJobSearchDuration` | `"0s"` | disabled best-effort job floor |
-| `maxGeneratorSearchDuration.default` | `"250ms"` | fallback generator budget |
-| `maxGeneratorSearchDuration.NodeLocalGreedy` | `"50ms"` | narrow generator budget |
-| `maxGeneratorSearchDuration.MultiNodeGang` | `"250ms"` | wide generator budget |
+| `maxGeneratorSearchDuration.default` | `"2m"` | fallback generator budget |
+| `maxGeneratorSearchDuration.NodeLocalGreedy` | `"30s"` | narrow generator budget |
+| `maxGeneratorSearchDuration.MultiNodeGang` | `"2m"` | wide generator budget |
 
 All values are strings parsed with Go's standard `time.ParseDuration`, which supports units such as `ms`, `s`, `m`, and `h`. Values must parse as non-negative durations. For max budgets, `0s` means unlimited and is used for legacy-equivalent support configurations. For `minJobSearchDuration`, `0s` disables the best-effort floor. If configured, `minJobSearchDuration` must be lower than `maxJobSearchDuration` unless `maxJobSearchDuration` is unlimited. Unknown action or generator names should be rejected during configuration validation so misspelled knobs do not silently fail.
 
@@ -345,7 +345,7 @@ Initial implementation criteria:
 
 Default tuning criteria:
 
-- Defaults are tuned against the 500-node snapshot and representative production-like snapshots.
+- Defaults are tuned against the 1000-node large-job benchmark and representative production-like snapshots.
 - Reduced-budget user messages are accurate and only emitted for reduced-budget jobs.
 - Metrics show `deadline_exhausted`, `generators_exhausted`, `no_generator`, and `not_attempted` outcomes clearly by action and, where applicable, generator.
 - Legacy-equivalent configuration is documented as current emitter plus unlimited budgets; bounded-search misses are terminal for the current scheduling attempt.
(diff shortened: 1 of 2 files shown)


docs/developer/designs/reclaim-generator-portfolio-design.md
```diff
@@ -0,0 +1,343 @@
...
+
+1. Every accepted solution is fully simulated and validator-approved.
+2. Bounded-portfolio negative results can be approximate; they are not complete unschedulability proofs.
+3. Victim batches remain gang-preserving units, including multi-node jobs from the #1537 fix.
+4. `Solve` remains all-or-nothing; no partial-probe state is committed unless the full job is solved.
+5. The post-simulation validator remains the final authority for consolidation, proportion, and other post-eviction side effects.
+
+### Mechanism
+
+`solvePartialJob` keeps its skeleton. The scenario source changes from a single exhaustive emitter to an ordered portfolio of generators. The portfolio owns generator iteration, action filtering, budget checks, and stop reasons.
```
itsomri: I think that we should include in this design, even if it's not implemented in v1, a generator-selector mechanism, which will be able to select (and order) scenario builders based on the job (and maybe cluster state?)

docs/developer/designs/reclaim-generator-portfolio-design.md
```diff
@@ -0,0 +1,343 @@
...
+
+When the driver reaches the effective deadline without a validated solution, the action reports "no solution" as an incomplete result. The result reason must distinguish at least deadline exhaustion, generator exhaustion, no applicable generator, and not-attempted jobs so metrics and reduced-budget messages are accurate.
+
+### Search Result Contract
+
+The driver returns a structured result, not just `nil`. The result reason is the single source of truth for metrics, reduced-budget messages, and rollout criteria:
```
itsomri, line 117: We should also export (not sure if here) which generator solved the job - this could be useful in the future when we want to validate different generator implementations

docs/developer/designs/reclaim-generator-portfolio-design.md
```diff
@@ -0,0 +1,343 @@
...
+// It performs no simulation; it only decides which victim sets are worth trying.
+type ScenarioGenerator interface {
+    Name() string
+    Next() *scenario.ByNodeScenario // nil when exhausted
+}
+```
+
+A generator is built per probe from a shared solve context: partial pending job, recorded victims, feasible nodes, victim queue, gang constraints, topology constraints, and action type.
+
+For Phase 1, `Next()` intentionally stays simple and does not accept a deadline or context. The generator contract is that `Next()` is cheap and incremental: it may compute or return the next candidate, but it must not run simulation, scan an unbounded search space, or perform expensive blocking work before returning. The initial `NodeLocalGreedy` and `MultiNodeGang` generators must be structured around this contract. If a future generator needs expensive candidate construction, the interface must be revisited so candidate generation can receive a deadline/context or otherwise poll the search budget internally.
```
itsomri, line 143: Do we keep the assumption that scenarios are only additive? If we do, we should state it on the interface. If not, we need to re-thing the scenario filters

docs/developer/designs/reclaim-generator-portfolio-design.md
```diff
@@ -0,0 +1,343 @@
...
+## Motivation
+
+The current reclaim path can spend unbounded synchronous scheduler time trying to prove that no valid victim set exists. The scale-test failure is a negative case: the pending job is unschedulable by construction, but reclaim drains a wide scenario search before giving up. The bounded generator portfolio makes that failure mode bounded and observable while preserving the safety property that any accepted reclaim solution is fully simulated and validator-approved.
+
+### Goals
+
+- Bound pathological reclaim, preempt, and consolidation scenario search by time budget.
+- Keep every accepted solution validator-approved; never accept a scenario from a shortcut alone.
+- Preserve the #1537 gang/topology correctness fix by keeping whole victim gangs intact.
+- Restore fast common-case behavior with a narrow `NodeLocalGreedy` generator before wider search.
```
itsomri, line 49: I suggest "InductiveScenarioBuilder" as a name for the old one

docs/developer/designs/reclaim-generator-portfolio-design.md
enoodle: I will add this to the future possible enhancements

docs/developer/designs/reclaim-generator-portfolio-design.md
enoodle: this (generator, job) tuple information doesn't really fit metrics, so we will have to either collect this from logs or export it in some way. I don't think this should be part of this design.

docs/developer/designs/reclaim-generator-portfolio-design.md
enoodle: Good point, I will add that currently it is part of the contract for this function, but we will update it in the future for something like https://github.com/kai-scheduler/KAI-Scheduler/pull/1614 so that the generator can also say "I want to start over from a different place"

docs/developer/designs/reclaim-generator-portfolio-design.md
enoodle: @itsomri `ScenarioBuilder` postfix is a bit redundant. I am not sure that `Inductive` is more descriptive than `NodeLocalGreedy` but it is also not the best of names.

docs/developer/designs/reclaim-generator-portfolio-design.md
enoodle: I updated the design to move the responsibility to filter the scenarios to the scenario generator- it should, internally, keep the scenarios monotonic before using the scenario filters. Then it can create sub-scenarios from them that don't have to be so.

itsomri: Sounds good

itsomri (dismissed):

itsomri (dismissed):

docs/developer/designs/reclaim-generator-portfolio-design.md
```diff
@@ -0,0 +1,387 @@
...
+| 1 | `NodeLocalGreedy` | Restore the cheap pre-#1537 node-local scenario shape for common cases and the scale-test failure. |
+| 2 | `MultiNodeGang` | Wrap today's wide accumulated scenario builder while preserving the #1537 whole-gang behavior. |
+
+The negative result is intentionally approximate when produced by the bounded portfolio. If the budget expires or the registered generators do not cover the shape, the scheduler may report no solution even if a solution exists. This is acceptable only because accepted positives remain fully simulated and validator-approved, and because the behavior is explicitly bounded and observable.
+
+### Limitations/Risks and Mitigations
+
+| Risk | Mitigation |
+| --- | --- |
+| False negative when a valid scenario exists after the budget expires. | Report deadline/generator exhaustion through metrics; keep budget knobs alpha/experimental while tuning defaults; add future generators based on observed misses. |
```
SiorMeir, line 79: I think its fair to ask the user for a few "rounds" and timeout adjustmets to mitigate this risk

docs/developer/designs/reclaim-generator-portfolio-design.md
```diff
@@ -0,0 +1,387 @@
...
+
+For Phase 1, `Next()` intentionally stays simple and does not accept a deadline or context. The generator contract is that `Next()` is cheap and incremental: it may compute or return the next candidate, but it must not run simulation, scan an unbounded search space, or perform expensive blocking work before returning. The initial `NodeLocalGreedy` and `MultiNodeGang` generators must be structured around this contract. If a future generator needs expensive candidate construction, the interface must be revisited so candidate generation can receive a deadline/context or otherwise poll the search budget internally.
+
+Generators are responsible for scenario quality, not just scenario enumeration. They should use cheap necessary-condition checks before emitting a candidate, because every emitted scenario can trigger a full simulation and validator pass. Low-quality candidates waste the bounded search budget and can hide useful later generators behind avoidable simulation work.
```
SiorMeir, line 145: Will we be able (at a later phase, maybe) to serialize scenarios? 
I'm thinking about metrics & caching that might benefit from that.

docs/developer/designs/reclaim-generator-portfolio-design.md
enoodle: I don't understand, this is not interactive.

docs/developer/designs/reclaim-generator-portfolio-design.md
enoodle: I don't see a reason that will prevent that

SiorMeir (approved):

PR #1708, opened 2026-06-15
chore(deps): drop github.com/NVIDIA/gpu-operator dependency

gshaibi:
## What

The operator depends on `github.com/NVIDIA/gpu-operator` only for the `ClusterPolicy` CRD struct — scheme registration plus watching/reading the CR to auto-detect CDI (`pkg/operator/operands/binder/resources.go`, `pkg/operator/controller/config_controller.go`). It never compiles any upstream gpu-operator runtime code.

This PR replaces that import with a minimal, hand-maintained mirror of the `ClusterPolicy` types under `third_party/nvidia/gpu-operator/api/nvidia/v1/`, and removes `github.com/NVIDIA/gpu-operator` (and its NVIDIA indirect deps) from the module graph.

## Why

1. **The reported CVEs don't apply.** The gpu-operator / NVIDIA Container Toolkit CVEs live in upstream runtime code we never link — we only embed CRD struct definitions.
2. **The dependency can't be bumped to a fixed version.** gpu-operator publishes CalVer tags (`v24.x`/`v25.x`/`v26.x`) without the matching `/vN` module-path suffix that Go's semantic import versioning requires, so it only ever resolves at `v1.x` pseudo-versions — which CVE scanners always compare as lower than the advisory's fixed version (e.g. 25.3.2). There is no importable version that satisfies the scanner.

## How

- New `third_party/nvidia/gpu-operator/api/nvidia/v1` package: trimmed `ClusterPolicy`/`ClusterPolicyList`/`ClusterPolicySpec`/`CDIConfigSpec`/`OperatorSpec` mirror, registering the **identical** `nvidia.com/v1` GVK. The controller-runtime client watches/lists/gets the same cluster objects; undeclared fields are ignored on decode, so behavior is unchanged.
- Repointed the 7 importing files from `github.com/NVIDIA/gpu-operator/api/nvidia/v1` to the mirror.
- `go mod tidy` drops `gpu-operator`, `k8s-kata-manager`, `k8s-operator-libs`.

## Verification

- `go build ./cmd/operator/... ./pkg/operator/... ./third_party/...`
- `go test ./pkg/operator/operands/binder/... ./pkg/operator/operands/common/...` pass (these exercise the `ClusterPolicy` fixture with `Operator.DefaultRuntime`/`CDI`).


third_party/nvidia/gpu-operator/api/nvidia/v1/types.go
```diff
@@ -0,0 +1,97 @@
...
+
+KAI-scheduler only ever compiled these CRD struct definitions (no upstream
+runtime code), so the reported gpu-operator / container-toolkit CVEs never
+applied to its binaries. Mirroring the types locally removes the dependency from
+the module graph entirely.
+
+The GroupVersionKind is identical to upstream (nvidia.com/v1, Kind=ClusterPolicy)
+so the controller-runtime client watches, lists and gets the very same cluster
+objects; fields we do not declare are ignored when decoding live objects.
+*/
```
SiorMeir, line 16: Had to read this 3 times to understand 😅 
I think we can remove/shorten the first paragraph

SiorMeir (dismissed):

SiorMeir (approved):

Issue #1719, opened 2026-06-17
Deduplicate equivalent reclaim scenario candidates before simulation


enoodle:
### What you would like to be added?

Implement scenario deduplication for the current reclaim scenario generator implementation.

The current implementation can emit equivalent scenario candidates more than once, which causes the scheduler to spend simulation and validation budget on repeated victim sets. Add a per-job/probe deduplication cache at the scenario portfolio/generator boundary, before full scheduler simulation.

Related design context: https://github.com/kai-scheduler/KAI-Scheduler/pull/1696

Acceptance criteria:

- Add a canonical scenario fingerprint for ByNodeScenario candidates.
- The fingerprint is deterministic and order-independent for equivalent victim sets.
- The fingerprint includes all scenario data that affects simulation, including victim tasks/jobs and node assignment data.
- Deduplicate both within a single generator and across portfolio generators such as NodeLocalGreedy, MultiNodeGang, and future generators.
- Scope the cache to the current job/probe search, not global scheduler state.
- Keep generator-side quality checks intact: generators should still build scenarios incrementally using accumulated filters and avoid obviously invalid or low-quality candidates before emission.
- Add tests or benchmark instrumentation showing duplicate candidates are skipped before simulation.

### Why is this needed?

This issue exists in the current implementation: equivalent reclaim scenarios can be emitted and simulated repeatedly instead of being skipped before simulation.

In the 100-node unschedulable reclaim benchmark, BenchmarkReclaimUnschedulableDistributedJob_100Node produced 1,207 scenario candidates, of which 1,020 were duplicate scenarios, leaving only 187 unique simulations. Track 1 generation-level dedupe reduced that case to 187 candidates with 0 duplicates, confirming that the duplicate work is real and avoidable.

This matters because every emitted scenario can trigger expensive scheduler simulation and validator work. It becomes more important with multiple scenario generators, since NodeLocalGreedy, MultiNodeGang, and future case-specific generators may discover the same effective victim set through different heuristics.

Issue #1660, opened 2026-06-02
scale: reclaim failure test first regresses in v0.14.3


enoodle:
The scale test

`Kwok scale test / Big cluster / Whole GPU Tests / Reclaim / Single GPU Reclaimees / measure reclaim failure time / measure time for reclaim to fail on distributed job last pod`

times out on `main`.

Replay of a snapshot captured from the failing scale-test cluster shows that the first bad `v0.14` hotfix is `v0.14.3`; `v0.14.3` and `v0.14.4` get stuck in `reclaim`.

This regression starts in the `v0.14.2 -> v0.14.3` release window, which includes:
- `fix: Exponential solver regression v0.14` (`#1595`)
- `perf: Re-introduce empty victim guard v0.14` (`#1602`)

Related branch scope for the exponential solver fix:
- `main` via `#1537`
- `v0.15` / `v0.15.0`
- `v0.14` / `v0.14.3` / `v0.14.4`

This issue records the regression boundary and the affected scale test. It does not yet isolate which commit in that release window causes the reclaim stall.

enoodle:
Tracking the reclaim generator portfolio work for this issue.

Design:
- #1696

Full implementation reference:
- #1716

Implementation PR stack:
- [x] #1740 - Scenario search budget config
- [x] #1741 - Scenario generator registration
- [x] #1742 - Scenario search budget model
- [x] #1743 - Wire budgets into solver
- [x] #1744 - Built-in scenario generators
- [x] #1745 - Portfolio driver
- [x] #1746 - Scenario search metrics
- [ ] #1747 - Reduced-budget and unresolved explainability
- [x] #1748 - Reclaim benchmark and topology coverage

PR #1727, opened 2026-06-19
perf(scheduler): deduplicate equivalent reclaim scenario candidates before simulation

kshitizlohia1994:
## Description

Addresses #1719.

In the 100-node unschedulable reclaim benchmark, 1,207 scenario candidates were generated of which 1,020 (84%) were duplicates — the same victim set reached via different accumulation orderings, each triggering full scheduler simulation unnecessarily.

### What this PR does

**Adds `Fingerprint() uint64` to `ByNodeScenario`** (`pkg/scheduler/actions/common/solvers/scenario/by_node_scenario.go`):
- Hashes preemptor UID + sorted victim job entries (each with sorted task UIDs) + sorted recorded-victim UIDs using FNV-64a
- Deterministic and order-independent: same victim set in any insertion order → same fingerprint
- Includes task UIDs (not just counts), so scenarios selecting different tasks from the same job — i.e. different node assignments — get distinct fingerprints

**Deduplicates in `solvePartialJob`** (`pkg/scheduler/actions/common/solvers/job_solver.go`):
- Maintains a `map[uint64]struct{}` scoped to the current job/probe search
- Skips simulation for any scenario whose fingerprint was already processed
- Sits above all generators — catches duplicates within a single generator and across the portfolio (NodeLocalGreedy, MultiNodeGang, and future generators)
- Generator-side quality checks remain intact

**Adds `scenarios_deduped_by_action` Prometheus counter** (`pkg/scheduler/metrics/metrics.go`):
- Mirrors `scenarios_simulation_by_action` and `scenarios_filtered_by_action`
- Lets operators observe deduplication rate in production

### Tests

- **6 fingerprint unit tests** in `by_node_scenario_test.go`: same victim set different order → same fingerprint; different victim sets → different fingerprints; same job different tasks (different node assignment) → different fingerprints; empty scenario is stable; different preemptors are distinct
- **`TestReclaimDeduplicatesEquivalentScenarios`** in `reclaim_benchmark_test.go`: 20-node topology where the distributed job needs 5 of 20 nodes — multiple accumulation orderings produce duplicate K-prefix sub-scenarios — asserts `scenarios_deduped_by_action` counter increments > 0, directly proving dedup fires

## Related Issues

Part of #1719

## Checklist

- [x] Self-reviewed

## Breaking Changes

None

### DIFF
--- a/pkg/scheduler/actions/common/solvers/job_solver.go
+++ b/pkg/scheduler/actions/common/solvers/job_solver.go
@@ -186,8 +186,16 @@ func (s *JobSolver) solvePartialJob(ssn *framework.Session, state *solvingState,
 	scenarioBuilder := NewPodAccumulatedScenarioBuilder(
 		ssn, partialPendingJob, state.recordedVictimsJobs, s.generateVictimsQueue(), feasibleNodeMap)
 
+	seenScenarios := make(map[uint64]struct{})
 	for scenarioToSolve := scenarioBuilder.GetValidScenario(); scenarioToSolve != nil; scenarioToSolve =
 		scenarioBuilder.GetNextScenario() {
+		fp := scenarioToSolve.Fingerprint()
+		if _, seen := seenScenarios[fp]; seen {
+			metrics.IncScenarioDedupedByAction()
+			continue
+		}
+		seenScenarios[fp] = struct{}{}
+
 		scenarioSolver := newByPodSolver(feasibleNodeMap, s.solutionValidator, ssn.AllowConsolidatingReclaim(),
 			s.actionType)
 
--- a/pkg/scheduler/actions/common/solvers/scenario/by_node_scenario.go
+++ b/pkg/scheduler/actions/common/solvers/scenario/by_node_scenario.go
@@ -4,6 +4,10 @@
 package scenario
 
 import (
+	"fmt"
+	"hash/fnv"
+	"sort"
+
 	"golang.org/x/exp/maps"
 	"golang.org/x/exp/slices"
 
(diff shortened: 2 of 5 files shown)


enoodle:
@kshitizlohia1994 Thank you for opening this one.
This area is going to have a refactor done (#1716) and I prefer to build a fix on top of that instead of the current implementation that will still have limits that makes the scheduler unreliable at scale. (see https://github.com/kai-scheduler/KAI-Scheduler/issues/1660)
With the current implementation where we only have one scenario builder / generator it makes more sense in my opinion to fix it so that it will not generate duplicated scenarios than to let it do that and then filter them on the next layer. After #1716 we will have 2 or more (hopefully if someone uses KAI and think of a way to model their cluster and jobs better they can create a plugin for themselves that will be even better than the KAI ones ) and then it will make sense to also filter between them.
So I am sorry but we can't merge it currently, but maybe you can save this work and re-apply it soon?

kshitizlohia1994:
> @kshitizlohia1994 Thank you for opening this one. This area is going to have a refactor done (#1716) and I prefer to build a fix on top of that instead of the current implementation that will still have limits that makes the scheduler unreliable at scale. (see #1660) With the current implementation where we only have one scenario builder / generator it makes more sense in my opinion to fix it so that it will not generate duplicated scenarios than to let it do that and then filter them on the next layer. After #1716 we will have 2 or more (hopefully if someone uses KAI and think of a way to model their cluster and jobs better they can create a plugin for themselves that will be even better than the KAI ones ) and then it will make sense to also filter between them. So I am sorry but we can't merge it currently, but maybe you can save this work and re-apply it soon?

Makes sense to build on top of #1716 rather than patching the current implementation. I'll close this pr for now. Thanks

PR #1731, opened 2026-06-21
feat: add kai-pending agent skill

### LINKED ISSUE #1516: Skill to investigate a job's pending status

We want to have a skill that can thoroughly investigate the job's status and reasons for it even if currently KAI's system is not exporting the reasons well.
A bonus option is to help it suggest to open an issue in case it was not trivial to find the reason to increase KAI's explain-ability.

### Why is this needed?

A very common question on the issues / slack channels is "why is my job pending"

somanythingstodo:
## Description

Here are instructions for AI agents investigating why a job is Pending on a KAI-Scheduler cluster. KAI-specific causes are kept separate from generic Kubernetes issues, which are triaged out rather than guessed at. It is meant to be extensible — new cases are added as additional reference docs.

The `scripts/free_capacity.py` script is optional, but for the generic "no nodes with enough resources" branch it computes the real free capacity per node (from kubectl dumps of nodes and pods), so the agent reads numbers instead of guessing.

Also the "Fix" section in each playbook is advisory and is open to discussion.

## Related Issues

Fixes #1516

## Checklist

- [ ] Self-reviewed


enoodle:
@somanythingstodo We should have something that read logs and the jog status - makes sure that the pod group got created (if not - looks at podgrouper logs etc..)
If there is no event / condition explaining the Unschedulable reason we need to look at the scheduler logs, and also if there was a binding issue read the binder logs if the message is not clear enough.

.agents/skills/kai-pending/references/nonpreemptible-over-quota.md
enoodle, line 1: this for example will be explained directly in the "Unschedulable" condition that the scheduler emits, so we don't need to overload the agent with extra information about it. It should know to first read the reason there.

enoodle (commented):
I think that some of the "references" files are unnecessary - They explain things that the scheduler already explains in the Unschedulable messages

somanythingstodo:
@enoodle could you take a look when you get a chance?

I tried to keep the skill extensible — new cases can be added as extra reference docs. 
There's also an optional helper script `free_capacity.py`, it is useful for the case of generic the scheduling verdict "no nodes with enough resources were found..."

A few design decisions I'd appreciate your feedback on:

1. Is a python helper acceptable here at all? I think for an agent skill it might be fine ..
2. free_capacity.py reads json dumps (pods and nodes) which could be a problem on large clusters, maybe it should use scheduler snapshot (/get-snapshot) instead?
3. For fractional GPU it parses the gpu-fraction annotation, but it does not know how much fractional
GPU is actually free inside existing reservation pods — is there a good source for the per GPU
fractional accounting?
4. For fair-share / priority starvation I currently just parse the queue specs (overQuotaWeight,
allocated). Maybe there is a better way, I noticed there is fairshare-simulator, but haven't explored it yet ..

Thanks a lot!

enoodle:
> @enoodle could you take a look when you get a chance?
> 
> I tried to keep the skill extensible — new cases can be added as extra reference docs. There's also an optional helper script `free_capacity.py`, it is useful for the case of generic the scheduling verdict "no nodes with enough resources were found..."
> 
> A few design decisions I'd appreciate your feedback on:
> 
> 1. Is a python helper acceptable here at all? I think for an agent skill it might be fine ..
> 2. free_capacity.py reads json dumps (pods and nodes) which could be a problem on large clusters, maybe it should use scheduler snapshot (/get-snapshot) instead?
> 3. For fractional GPU it parses the gpu-fraction annotation, but it does not know how much fractional
>    GPU is actually free inside existing reservation pods — is there a good source for the per GPU
>    fractional accounting?
> 4. For fair-share / priority starvation I currently just parse the queue specs (overQuotaWeight,
>    allocated). Maybe there is a better way, I noticed there is fairshare-simulator, but haven't explored it yet ..
> 
> Thanks a lot!

1. Python is good, the agent should know how to use it
2. I haven't looked at this script yet, but what does it do that the scheduler doesn't know already?
3. I am not sure what the fractional gpu reference goal is - maybe if we make it more clear we will know what it needs better.
4. There are metrics / queue status that expose this

.agents/skills/kai-pending/references/nonpreemptible-over-quota.md
somanythingstodo: thanks, I removed that case and some others (QueueDoesNotExist / OverLimit / NonPreemptibleOverQuota are clear from the reason)

somanythingstodo:
@enoodle Thanks for your comments. I dropped the per-reason reference docs for the clear typed reasons.

Honestly, I work mostly with Slurm, so I carried over how I triage pending there: misconfig (time/partition/qos), low priority + fair-share, account/cluster limits. 
KAI exposes far more on the pod/podgroup status and a lot of that Slurm matrix has no equivalent here. 

If you can say which pending cases actually show up most in practice, I'll be happy to shape the skill around those.

On your points:
- logs - I added a step regarding that to skill, but my doubt is that kubectl logs are short-lived and in prod logs should be checked in Loki/Elastic/etc.
- script and scheduler info - you are right, scheduler knows all of this, it just doesn't expose it by default (per-node fit only at high verbosity, which I doubt prod runs, or in status with detailed-fit-errors=true). Should I drop script and point at detailed-fit-errors, or keep it?
- fractional: neither the script nor the scheduler exposes fractional GPU usage well in logs today, so the skill can't really decide fractional fit. I can try to add more detailed info in scheduler logs or in status regarding fractional GPU usage, if that is useful. (in another PR of course)

enoodle:
> * logs - I added a step regarding that to skill, but my doubt is that kubectl logs are short-lived and in prod logs should be checked in Loki/Elastic/etc.

Usually in KAI when there is a failure the logs will repeat themselves because the scheduler will retry on every scheduling session.

> * script and scheduler info - you are right, scheduler knows all of this, it just doesn't expose it by default (per-node fit only at high verbosity, which I doubt prod runs, or in status with detailed-fit-errors=true). Should I drop script and point at detailed-fit-errors, or keep it?

If the scheduler already emitted this data it is better to use that so that we will only have one place with this logic implemented.

> * fractional: neither the script nor the scheduler exposes fractional GPU usage well in logs today, so the skill can't really decide fractional fit. I can try to add more detailed info in scheduler logs or in status regarding fractional GPU usage, if that is useful. (in another PR of course)

I am still not sure what is the goal here. Do you mean that the workload fails to be allocated because of cluster fragmentation? (eg there are 2 GPUs and the workload is asking for 0.4 but there are already two fractions of 0.7 and 0.9 running) we should expose this information from the scheduler and then read it in the skill, but it should be done in another PR.

.agents/skills/kai-pending/SKILL.md
```diff
@@ -0,0 +1,92 @@
+---
+name: kai-pending
+description: Use when a KAI-Scheduler (Run:ai / NVIDIA) pod or PodGroup is stuck Pending and you need to know why - GPU jobs that won't start, queue quota/limit, fair-share, gang scheduling, fractional GPU, node-pool affinity, or scheduling gates. Reads the PodGroup's scheduling verdict; an optional python helper adds per-node free-capacity math.
```
enoodle: remove the Run:ai/ NVIDIA

.agents/skills/kai-pending/references/fair-share.md
```diff
@@ -0,0 +1,26 @@
...
+tell you who is holding capacity right now):
+
+```bash
+kubectl get queues -o json | jq -r '.items[] | "\(.metadata.name) parent=\(.spec.parentQueue) alloc=\(.status.allocated["nvidia.com/gpu"] // 0) req=\(.status.requested["nvidia.com/gpu"] // 0)"'
+```
+
+1. Your queue shows `allocated` ~0 while `requested` > 0; a **sibling** (same `parentQueue`) whose
+   `allocated` holds the GPUs = fair-share starvation, working as designed.
+2. `overQuotaWeight` / priority (from the spec) only explain **why** the split landed that way. The
+   same live numbers are also exported as Prometheus metrics if a metrics stack is present.
```
enoodle: If not - it can read it directly from the scheduler pod.

somanythingstodo:
>  
> > * logs - I added a step regarding that to skill, but my doubt is that kubectl logs are short-lived and in prod logs should be checked in Loki/Elastic/etc.
> 
> Usually in KAI when there is a failure the logs will repeat themselves because the scheduler will retry on every scheduling session.

Make sense, I also thought about it, but concern was that if there are thousands of pending jobs, the message still could be missed.

> 
> > * script and scheduler info - you are right, scheduler knows all of this, it just doesn't expose it by default (per-node fit only at high verbosity, which I doubt prod runs, or in status with detailed-fit-errors=true). Should I drop script and point at detailed-fit-errors, or keep it?
> 
> If the scheduler already emitted this data it is better to use that so that we will only have one place with this logic implemented.

Yes, agreed , will remove it.

> 
> > * fractional: neither the script nor the scheduler exposes fractional GPU usage well in logs today, so the skill can't really decide fractional fit. I can try to add more detailed info in scheduler logs or in status regarding fractional GPU usage, if that is useful. (in another PR of course)
> 
> I am still not sure what is the goal here. Do you mean that the workload fails to be allocated because of cluster fragmentation? (eg there are 2 GPUs and the workload is asking for 0.4 but there are already two fractions of 0.7 and 0.9 running) we should expose this information from the scheduler and then read it in the skill, but it should be done in another PR.
> 

Yes, I meant fragmentation here ..
I will check how I can expose it

.agents/skills/kai-pending/SKILL.md
somanythingstodo: done

.agents/skills/kai-pending/references/fair-share.md
somanythingstodo: thanks, added

somanythingstodo:
> > > * logs - I added a step regarding that to skill, but my doubt is that kubectl logs are short-lived and in prod logs should be checked in Loki/Elastic/etc.
> > 
> > 
> > Usually in KAI when there is a failure the logs will repeat themselves because the scheduler will retry on every scheduling session.
> 
> Make sense, I also thought about it, but concern was that if there are thousands of pending jobs, the message still could be missed.
> 
> > > * script and scheduler info - you are right, scheduler knows all of this, it just doesn't expose it by default (per-node fit only at high verbosity, which I doubt prod runs, or in status with detailed-fit-errors=true). Should I drop script and point at detailed-fit-errors, or keep it?
> > 
> > 
> > If the scheduler already emitted this data it is better to use that so that we will only have one place with this logic implemented.
> 
> Yes, agreed , will remove it.
> 
> > > * fractional: neither the script nor the scheduler exposes fractional GPU usage well in logs today, so the skill can't really decide fractional fit. I can try to add more detailed info in scheduler logs or in status regarding fractional GPU usage, if that is useful. (in another PR of course)
> > 
> > 
> > I am still not sure what is the goal here. Do you mean that the workload fails to be allocated because of cluster fragmentation? (eg there are 2 GPUs and the workload is asking for 0.4 but there are already two fractions of 0.7 and 0.9 running) we should expose this information from the scheduler and then read it in the skill, but it should be done in another PR.
> 
> Yes, I meant fragmentation here .. I will check how I can expose it

@enoodle 
So I removed the script and made a related cleanup for "references" files. 
The smart models can probably figure out all these pending reasons without the skill, but even for such a model the explicit steps would help.

Regarding the fragmentation - I can make another PR with some additional logging, if you think it could be a useful contribution.
Thanks for your review!

somanythingstodo:
rebased

.agents/skills/kai-pending/references/fair-share.md
```diff
@@ -0,0 +1,41 @@
...
+2. `overQuotaWeight` / priority (from the spec) only explain **why** the split landed that way. The
+   same live numbers are also exported as Prometheus metrics; with no metrics stack, read them
+   directly from the scheduler pod - it logs the per-queue allocation each session.
+
+## Why the holder isn't evicted for you
+
+KAI evicts a holder only if it is **preemptible** (priorityClass `value < 100`), you **outrank** it,
+and it has run past its protection window. So the pod stays Pending when the holder is
+non-preemptible (`value >= 100`, never evicted), or preemptible but still young - check the holder's
+`spec.priorityClassName` -> `value` and the queue's `preemptMinRuntime` / `reclaimMinRuntime`.
```
enoodle: I think that fair share and priority are mixed here in a wrong context.

The Preempt action removes lower priority jobs from a queue to make place for a higher priority job from the same queue.
The Reclaim action removes jobs from queues that are over their quota (usually over their fair-share quota) to make place for jobs from other queues that are under their quota. The reclaim action doesn't care about priority, it takes a very minor part in ordering the tasks of queues that can be removed.

.agents/skills/kai-pending/references/fair-share.md
```diff
@@ -0,0 +1,41 @@
...
+   `allocated` holds the GPUs = fair-share starvation, working as designed.
+2. `overQuotaWeight` / priority (from the spec) only explain **why** the split landed that way. The
+   same live numbers are also exported as Prometheus metrics; with no metrics stack, read them
+   directly from the scheduler pod - it logs the per-queue allocation each session.
+
+## Why the holder isn't evicted for you
+
+KAI evicts a holder only if it is **preemptible** (priorityClass `value < 100`), you **outrank** it,
+and it has run past its protection window. So the pod stays Pending when the holder is
+non-preemptible (`value >= 100`, never evicted), or preemptible but still young - check the holder's
```
enoodle: whats "young" ?

.agents/skills/kai-pending/references/fair-share.md
```diff
@@ -0,0 +1,41 @@
...
+tell you who is holding capacity right now):
+
+```bash
+kubectl get queues -o json | jq -r '.items[] | "\(.metadata.name) parent=\(.spec.parentQueue) quota=\(.spec.resources.gpu.quota // 0) alloc=\(.status.allocated["nvidia.com/gpu"] // 0) req=\(.status.requested["nvidia.com/gpu"] // 0)"'
+```
+
+(`quota` is from the **spec** - status has no deserved/quota field, only `allocated` / `requested`.)
+
+1. Your queue shows `allocated` ~0 while `requested` > 0; a **sibling** (same `parentQueue`) whose
+   `allocated` holds the GPUs = fair-share starvation, working as designed.
```
enoodle: The calculated fair share is part of the metrics exposed by the scheduler - maybe should look at this as well.

.agents/skills/kai-pending/references/fair-share.md
somanythingstodo: yes, thanks, I re-read docs and a code, I messed up here

.agents/skills/kai-pending/references/fair-share.md
somanythingstodo: I was thinking of reclaimMinRuntime, but here it is not clear

.agents/skills/kai-pending/references/fair-share.md
somanythingstodo: Right. So I rewrote the whole fair-share.md reference file to remove any ambiguity, now LLM should look into scheduler logs, check for reclaim attempts (and should stop guessing about fair share calculation)

enoodle (dismissed):

enoodle:
Thanks @somanythingstodo !

.agents/skills/kai-pending/SKILL.md
```diff
@@ -0,0 +1,89 @@
...
+kubectl get podgroup "$PG" -n <ns> -o json \
+  | jq '.status.schedulingConditions[] | {nodePool, reasons: (.reasons | unique_by(.message))}'
+```
+
+- no `pod-group-name` annotation -> never grouped -> step 5 (pod-grouper).
+- `schedulingConditions` empty -> no verdict yet; re-check. Persists -> scheduler down, or gated (step 1).
+- populated -> `reasons[]` repeats per cycle; `unique_by(.message)` dedupes, the richest message
+  (e.g. `MaxNodePoolResources`) is the verdict. Take `.reason` -> step 3.
+
+## 3. Act on the reason - read its `.message`, then:
```
gshaibi: I suggest the agent won't act automatically, maybe guiding the user regarding possible solutions would fit better

.agents/skills/kai-pending/references/gang.md
```diff
@@ -0,0 +1,16 @@
+# gang: can't assemble
+
+Verdict message: `Resources were found for N pods while M are required for gang scheduling`.
+Gang is all-or-nothing - KAI places the PodGroup only if `>= minMember` pods fit **at once**; here
+each pod fits alone but the cluster can't host `minMember` together.
+
+## Steps
+
+1. Read `M - N` from the message = how many more simultaneous slots are needed.
+2. Count nodes with free GPU (`capacity - used`, step 4 detail) >= per-pod request; compare to `minMember`.
```
davidLif: Why count only the gpus? Also, you can't compare the amout of resources to minMember

.agents/skills/kai-pending/references/node-fit.md
```diff
@@ -0,0 +1,14 @@
+# node-fit: too big for any single node
+
+Verdict message: e.g. `... in a single node ... topped at N` / `MaxNodePoolResources`. The pod is
+bigger than any single node, and a pod cannot span nodes.
+
+## Steps
+
+1. Compare `requests: gpu=N` against the largest node's `capacity` in the step 4 fit detail.
+2. Check the CPU / memory reasons too - the blocker may be CPU or memory, not GPU.
```
davidLif: And what about extended resources? I would try and write it differently ("compare the resources spesified" or something similar that we would seee as working).

davidLif (commented):
Do we have anything we can try this skill on? Maybe try for snapshots based on the scheduler integration tests or the e2e in some way?

.agents/skills/kai-pending/SKILL.md
somanythingstodo: right, thanks, here in "Act on the reason" - I just asked LLM to continue investigation, like read the logs and so on. 
But for QueueDoesNotExist, OverLimit, NonPreemptibleOverQuota  that is actually guidance for user. I rewrote that.

.agents/skills/kai-pending/references/gang.md
somanythingstodo: thanks, I was mostly focused on GPUs in my tests, so missed that.

.agents/skills/kai-pending/references/node-fit.md
somanythingstodo: rewrote, thank you

somanythingstodo:
> Do we have anything we can try this skill on? Maybe try for snapshots based on the scheduler integration tests or the e2e in some way?

@davidLif 
I tested mostly manually on test cluster, but later asked LLM to make some automatization around most of the cases.

From one side, making auto tests for agents skills looks like overkill and not easy to do, mainly because models are non-deterministic, and we even can't fix exact model against which we are running (if it is not a local deployment).
And actual testing it is not easy - sometimes LLM just try to guess the reason from test environment, from namespace name for example.

But from other side, keeping something in repo without tests means, that this will always be somehow outdated, there will be always new cases, that are not covered (like new bounded reclaim generator portfolio).

So may be it should be very generic, because KAI has very good status reasons and logging (comparing to SLURM for example), and basically I want LLM just to do the following:
	- find PodGroup
	- read all statuses
	- if status is not enough - read scheduler logs
	- (and may be if binding fails - look for BindRequest and binder)


Test cases, that I ended up are like this (I don't know, where to put that code actually, may be e2e code, but not sure if it fits. I can publish right now in separate repo for reference, but may be you have better idea):
```
	- 01-no-node-fits — pod needs 9 GPU, node has 8
	- 02-over-limit — queue gpu.limit=0
	- 03-queue-full-wait — queue capped at 8; A runs, B waits 
	- 04-preempt — low-prio job holds 8 GPU, high-prio preempts it
	- 05-queue-not-exist — label points at a missing queue
	- 06-nonpreemptible-over-quota — non-preemptible job, queue gpu.quota=0
	- 07-gang-wont-assemble — gang minMember=5, only 4 pods fit
	- 08-fair-share-starved — sibling queue (oqw=100) takes all GPU; light queue (oqw=1) waiter starves.
	- 09-fractional-no-reservation — gpu-fraction pod can't get a reservation pod (all GPUs full).
	- 10-pinned-to-full-pool — pod pinned by nodeSelector to a full pool while another pool is free
	- 11-scheduling-gated — pod with spec.schedulingGates
	- 12-no-queue-label — pod with no kai.scheduler/queue label
	- 13-fit-detail-hidden — fragmented free GPU (2 per node) vs a 4-GPU ask
	- 14-bind-failure — "scheduled but not Running": bind-stage failure (Actually need to test this better)
	- 15-reclaim-deserved — 08 + deserved quota: claimer-1 reclaims a pod (GuaranteeDeservedQuota), claimer-2 fails the CanReclaimResources fairShare gate
	- 16-reclaim-cousins — claimer/holder are cousins (different parents, one root)
	- 17-reclaim-cross-tree — two independent trees, all equal weights: each claimer moves fair share (8/24 → 16/16), two reclaims, then equal-split stop
	- 18-reclaim-no-parent — parent-less leaf queues
	- 19-reclaim-skipped-signature — two identical claimers whose reclaim fails
	- 20-no-node-fits-cpu — pod needs 500 CPU cores
	- 21-gang-extended-resource — gang minMember=3, each pod wants 1 example.com/widget, only 2 widgets in cluster 
```

gshaibi (dismissed):
We introduced a new way to add changelog entries, can you please check our [contribution docs](https://github.com/kai-scheduler/KAI-Scheduler/blob/main/CONTRIBUTING.md#contributing-changes)? Its basically running `make changelog` to add an entry.
Can you please adjust the PR accordingly?

Other than that, LGTM.
Thank you @somanythingstodo 🙇

somanythingstodo:
@gshaibi 
thanks, updated with `make changelog` (instead of updating changelog file directly)

enoodle (approved):

gshaibi (approved):

PR #1734, opened 2026-06-22
perf(scheduler): skip redundant full-job probe when allocation is infeasible

gshaibi:
## Description

`JobSolver.Solve` runs `searchMaxSolvableK`, an exponential-doubling + binary search that finds the largest `k` for which a probe (a full eviction/allocation scenario simulation) succeeds. That search is only correct under the assumption that solvability is **monotonic** in `k` (if `k` tasks can't be placed, neither can `k' > k`).

Previously, after the search, `Solve` always ran one more probe at the full count `n`:

```go
maxSolvedK := s.searchMaxSolvableK(...)
if maxSolvedK == 0 {
    return false, nil, ...
}
result := s.probeAtK(ssn, &state, pendingJob, tasksToAllocate, n) // always at n
```

When `0 < maxSolvedK < n`, the monotonicity assumption the search already relies on guarantees the probe at `n` fails — so this was an extra, guaranteed-failing scenario simulation on the unsolvable path.

This PR gates the final probe on `maxSolvedK == n` (returning early otherwise). The probe at `n` is still performed when the full set is solvable, because the search discards every `Statement` to keep the session clean, so the committable live statement must be rebuilt.

`tasksToAllocate` represents the next all-or-nothing allocation unit (gang minimum gap, or one elastic step), so there is no partial-commit case being lost.

## Related Issues

Fixes #

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. The early return produces the same `(false, nil)` result the failing probe would have returned.

## Additional Notes

- Behavior-preserving optimization under the monotonicity assumption the binary search already depends on; the win is avoiding one full scenario simulation per job that can be partially but not fully placed.
- No new dedicated test: the change is output-identical, so a regression would surface in the existing end-to-end solver suites rather than a new unit test. Verified locally:
  - `go test ./pkg/scheduler/actions/common/solvers/...`
  - `go test ./pkg/scheduler/actions/{reclaim,preempt,consolidation,allocate}/...`
  - all green; `gofmt`/`go vet` clean.


gshaibi:
Duplicates https://github.com/kai-scheduler/KAI-Scheduler/pull/1599

PR #1749, opened 2026-06-22
perf(scheduler): avoid duplicate full job solver probe

enoodle:
## Description

Avoids redundant work in the `JobSolver` probe sequence.

`searchMaxSolvableK` uses disposable partial probes to discover a solvable prefix and carry victim hints forward. The solver then runs `probeAtK(n)` to create the live statement for the complete job. Previously, the partial search could also probe `n`, causing the complete allocation to be simulated and discarded before immediately being simulated again.

This PR:

- Limits the disposable search to partial probes where `k < n`.
- Skips partial search for single-task jobs.
- Stops exponential partial search when its next probe reaches or exceeds `n`, instead of treating `n` as a synthetic failed upper bound and binary-searching toward it.
- Leaves the complete `probeAtK(n)` exclusively to the solve path, where it runs once and its live statement can be returned.

For example, an all-successful four-task search changes from `1, 2, 4, 4` to `1, 2, 4`. It does not add an early return after a failed partial upper bound; once any partial probe succeeds, the complete job still receives its final probe.

References:

- Reclaim generator portfolio design: #1696 (`docs/developer/designs/reclaim-generator-portfolio-design.md`)

## Related Issues

N/A

## Checklist

- [x] Self-reviewed

## Breaking Changes

None.

## Additional Notes

The branch is rebased directly onto `main`; the PR now contains only the focused solver optimization, regression tests, and changelog fragment.

Verification:

- `go test ./pkg/scheduler/actions/common/solvers -count=1`
- `git diff --check origin/main...HEAD`


pkg/scheduler/actions/common/solvers/job_solver.go
```diff
@@ -162,21 +162,23 @@ func (s *JobSolver) solvePendingJobWithGenerator(
...
-		ssn, state, pendingJob, tasksToAllocate, jobBudget, availableGenerator, generatorBudget,
-	)
-	if maxSolvedK == 0 {
-		if searchResult == nil {
-			searchResult = terminalSearchResult(SearchResultGeneratorsExhausted, jobBudget.ReducedBudget())
+	if n > 1 {
+		maxSolvedK, searchResult := s.searchMaxSolvableK(
+			ssn, state, pendingJob, tasksToAllocate, jobBudget, availableGenerator, generatorBudget,
+		)
+		if maxSolvedK == 0 {
```
davidLif, line 169: Why does it have to be equal to 0 instead of smaller then n-1? Why should we attempt to allocate all 3 pods if we cann't find a solution for 2?

.changes/unreleased/fixed-20260921-170136.yaml
SiorMeir, line 1: Add PR no. & Author (it'll reflect in the final changelog)

PR #1764, opened 2026-06-24
fix(scheduler): filter unreclaimable reclaim victims

### LINKED ISSUE #1750: Reclaim abandoned when a lower-priority protected queue exists, even though an over-quota sibling is reclaimable
## What happened?

A pending job in a queue that is **under its deserved quota** (and therefore entitled to reclaim) fails to reclaim resources from an **over-quota sibling queue**, when an unrelated **lower-priority queue that is under its deserved quota** (and therefore not reclaimable) also exists in the cluster. The reclaim is abandoned even though reclaiming the over-quota sibling is a valid solution.

### Reproduction

A deterministic unit test (plus a passing control) is on this branch:

[`pkg/scheduler/actions/reclaim/reclaim_anchor_victim_test.go`](https://github.com/david-gang/KAI-Scheduler/blob/dgang/reclaim-protected-anchor-repro/pkg/scheduler/actions/reclaim/reclaim_anchor_victim_test.go)

```bash
go test ./pkg/scheduler/actions/reclaim/ -run TestReclaimAnchorVictim -v
```

Queue hierarchy under test (all queues under their fair share; cluster has a small amount of spare capacity):

```
root
├── org-parent (priority 110)
│   ├── reclaimer-queue  (under deserved quota -> starved, entitled to reclaim)
│   └── batch-queue      (over deserved quota of 0 -> reclaimable; sibling of reclaimer-queue)
└── protected-queue      (priority 100, under deserved quota -> NOT reclaimable)
```

- `TestReclaimAnchorVictim` asserts the correct behavior and currently **FAILS** (the reclaimer stays `Pending`).
- `TestReclaimAnchorVictim_ControlNoProtectedQueue` is identical except `protected-queue` is removed; it **PASSES** (the reclaimer reclaims `batch-queue` and is pipelined).

The only difference between the two is the presence of the lower-priority protected queue, which isolates the defect.

### Observed mechanism

Instrumenting the reclaim path (on a production snapshot and in the test) shows:

- The victim queue ordering pops the **protected-queue** subtree first. Both `protected-queue` and `org-parent` are under their fair share and starved, so the ordering falls to the priority tiebreak, and `protected-queue` (priority 100) sorts before `org-parent` (priority 110).
- Because the protected queue's victims are accumulated first, its node sorts first in the sub-scenario emitter's candidate list and is included in **every** prefix `[0..K)` the emitter produces.
- The reclaimer task is therefore always placed in a scenario whose evicted victim is a protected-queue pod. The reclaim scenario validator (`proportion.Reclaimable`) only ever receives the protected-queue victim set and rejects it (the protected queue is under its deserved quota). A `batch-queue`-only victim set is never offered to the validator.
- Net effect: the reclaim is abandoned even though reclaiming `batch-queue` is valid. On a large cluster this also makes each reclaim attempt expensive (many rejected scenarios per pending job).

This was originally surfaced from a production cluster snapshot where a starved training queue could not reclaim from an over-quota batch queue while a lower-priority production queue sat under its deserved quota; it was reduced to the synthetic test above.

`reclaimer-job` reclaims a pod from the over-quota `batch-queue` and is pipelined. The presence of an unrelated, lower-priority, non-reclaimable queue should not prevent a valid reclaim from a different (reclaimable) queue.

## Environment

- Kubernetes version: N/A (reproduced as a scheduler unit test)
- KAI Scheduler version: reproduces on `main` (and v0.14.x)
- Cloud provider or hardware configuration: N/A
- Tools that you are using KAI together with: N/A
- Anything else that is relevant: reproduction is a pure unit test under `pkg/scheduler/actions/reclaim`

enoodle:
## Description

Filters reclaim victim candidates that are definitely not eligible for the deserved-quota reclaim path. This prevents an unrelated under-deserved queue from being selected as an anchor victim and blocking a valid reclaim from an over-quota queue.

The proportion plugin now registers a reclaim victim filter, while the decision logic lives under `pkg/scheduler/plugins/proportion/reclaimable/filter_victims.go`.

## Related Issues

Fixes #1750

## Checklist

- [x] Self-reviewed

## Breaking Changes

None.

## Additional Notes

Validation:

- `go test ./pkg/scheduler/plugins/proportion/reclaimable`
- `go test ./pkg/scheduler/plugins/proportion`
- `go test ./pkg/scheduler/actions/reclaim`

### DIFF
--- a/pkg/scheduler/actions/reclaim/reclaim_anchor_victim_test.go
+++ b/pkg/scheduler/actions/reclaim/reclaim_anchor_victim_test.go
@@ -8,6 +8,7 @@ import (
 
 	. "go.uber.org/mock/gomock"
 	"gopkg.in/h2non/gock.v1"
+	"k8s.io/utils/ptr"
 
 	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/actions/reclaim"
 	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/api/pod_status"
@@ -17,8 +18,6 @@ import (
 	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/test_utils/nodes_fake"
 	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/test_utils/tasks_fake"
 )
-
-func intPtr(i int) *int { return &i }
 
 // Queue hierarchy under test:
 //
@@ -91,10 +90,10 @@ func TestReclaimAnchorVictim(t *testing.T) {
 			"node-spare": {GPUs: 1},
 		},
 		Queues: []test_utils.TestQueueBasic{
-			{Name: "org-parent", DeservedGPUs: 6, GPUOverQuotaWeight: 1, ParentQueue: "root", Priority: intPtr(110)},
-			{Name: "reclaimer-queue", DeservedGPUs: 4, GPUOverQuotaWeight: 1, ParentQueue: "org-parent", Priority: intPtr(75)},
-			{Name: "batch-queue", DeservedGPUs: 0, GPUOverQuotaWeight: 1, ParentQueue: "org-parent", Priority: intPtr(25)},
-			{Name: "protected-queue", DeservedGPUs: 4, GPUOverQuotaWeight: 1, ParentQueue: "root", Priority: intPtr(100)},
+			{Name: "org-parent", DeservedGPUs: 6, GPUOverQuotaWeight: 1, ParentQueue: "root", Priority: ptr.To(110)},
+			{Name: "reclaimer-queue", DeservedGPUs: 4, GPUOverQuotaWeight: 1, ParentQueue: "org-parent", Priority: ptr.To(75)},
+			{Name: "batch-queue", DeservedGPUs: 0, GPUOverQuotaWeight: 1, ParentQueue: "org-parent", Priority: ptr.To(25)},
+			{Name: "protected-queue", DeservedGPUs: 4, GPUOverQuotaWeight: 1, ParentQueue: "root", Priority: ptr.To(100)},
 		},
 		Departments: []test_utils.TestDepartmentBasic{
 			{Name: "root", DeservedGPUs: 10},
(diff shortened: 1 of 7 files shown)


david-gang:
It solves our issue on the original snapshot as well. Great Work!

pkg/scheduler/actions/reclaim/reclaim_anchor_victim_test.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/actions/reclaim"
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/api/pod_status"
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/constants"
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/test_utils"
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/test_utils/jobs_fake"
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/test_utils/nodes_fake"
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/test_utils/tasks_fake"
+)
+
+func intPtr(i int) *int { return &i }
```
davidLif: Use the ptr module instead

pkg/scheduler/plugins/proportion/reclaimable/filter_victims.go
```diff
@@ -0,0 +1,65 @@
...
+	reclaimerQueue, reclaimeeQueue := r.getLeveledQueues(queues, reclaimer.Queue, reclaimeeQueueID)
+	if reclaimerQueue == nil || reclaimeeQueue == nil {
+		return true
+	}
+
+	if !canUseGuaranteeDeservedQuotaStrategy(reclaimer, reclaimerQueue) {
+		return true
+	}
+
+	return canBeDeservedQuotaReclaimCandidate(reclaimer, reclaimeeQueue)
```
davidLif, line 33: I would think that we would need to test both options for all the unequal queues in the Hierarchy, not only for the first divergence

pkg/scheduler/plugins/proportion/reclaimable/filter_victims.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func canUseGuaranteeDeservedQuotaStrategy(
```
davidLif: What if the reclaimer satisfies the MaintainFairShareStrategy, but not the GuaranteeDeservedQuotaStrategy?

enoodle: Our goal in the filter is to know if the reclaimee cannot under any circumstances be reclaimed. 

If the potential reclaimee queue is above its deserved quota then it might be reclaimed, if the reclaimer has the GuaranteeDeservedQuotaStrategy available.
=> One way we know for sure that it won't be reclaimed is if the reclaimee queue itself is not over deserved quota. this is true if `canUseGuaranteeDeservedQuotaStrategy` returns true on the reclaimer.

MaintainFairShareStrategy is not something we have to check for here for the reclaimer because on the reclaimer side we should assume it is tested before the action starts working on the reclaimer -if it doesn't have MaintainFairShareStrategy it will not be able to reclaim anything.

We are a too conservative with the use of canUseGuaranteeDeservedQuotaStrategy - even if it fails we might be able to filter the job - if the queue is below its fair share.

pkg/scheduler/plugins/proportion/reclaimable/filter_victims.go
enoodle: currently in the `Reclaimable` function we check with similar logic: https://github.com/kai-scheduler/KAI-Scheduler/blob/main/pkg/scheduler/plugins/proportion/reclaimable/reclaimable.go#L80

pkg/scheduler/actions/reclaim/reclaim_anchor_victim_test.go
enoodle: :+1:

pkg/scheduler/plugins/proportion/reclaimable/filter_victims.go
enoodle: I updated the usage of this function to:
```
	if !canUseGuaranteeDeservedQuotaStrategy(reclaimer, reclaimerQueue) {
		return canBeMaintainFairShareReclaimCandidate(reclaimeeQueue)
	}
```

davidLif (dismissed):

davidLif (approved):

PR #1779, opened 2026-06-28
feat(chart): support externally managed KAI resources

gshaibi:
## Description

Support embedding the KAI Helm chart as a subchart when KAI resources are managed externally, while preserving standalone defaults.

- Add `priorityClasses.enabled`, defaulting to `true`, and gate the existing `build`, `train`, `inference`, and `build-preemptible` PriorityClasses without changing their names or values.
- Preserve `Config/kai-config` during uninstall when `kaiConfigDeployer.enabled=false`.
- Replace namespace-wide Deployment cleanup with `app.kubernetes.io/managed-by=kai-operator` selection.
- Apply the cleanup ownership label through the common operator Deployment factory, including dynamically named scheduler Deployments for SchedulingShards.

The previous post-delete hook used `kubectl delete deployment --all`, which could remove unrelated workloads in a shared namespace, and always deleted `kai-config` even when its lifecycle was external to the chart.

## Related Issues

N/A

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. Existing standalone behavior remains unchanged because `priorityClasses.enabled` and `kaiConfigDeployer.enabled` both default to `true`.

## Additional Notes

Validation:

- `helm unittest -f 'tests/**/*_test.yaml' .` — 112 tests passed
- `helm lint .` — passed
- Default and external-management `helm template` renders — passed
- `go test ./pkg/operator/operands/common ./pkg/operator/operands/scheduler` — passed

A v0.16 backport commit is prepared locally and will be published after this change lands on `main`.

### DIFF
--- a/pkg/operator/operands/common/common.go
+++ b/pkg/operator/operands/common/common.go
@@ -28,6 +28,11 @@ import (
 
 var controllerTypes = []string{"Deployment", "DaemonSet"}
 
+const (
+	OperatorManagedByLabelKey   = "app.kubernetes.io/managed-by"
+	OperatorManagedByLabelValue = "kai-operator"
+)
+
 // PodDisruptionBudgetImplementedServices lists operand resource names with operator-side PDB creation.
 var PodDisruptionBudgetImplementedServices = map[string]struct{}{
 	"admission": {},
@@ -139,6 +144,7 @@ func DeploymentForKAIConfig(
 		return nil, err
 	}
 	deployment := deploymentObj.(*appsv1.Deployment)
+	deployment.Labels[OperatorManagedByLabelKey] = OperatorManagedByLabelValue
 	deployment.TypeMeta = metav1.TypeMeta{
 		Kind:       "Deployment",
 		APIVersion: "apps/v1",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ## [Unreleased]
 
 ### Added
+- Added `priorityClasses.enabled` Helm value (default `true`) for installations that manage KAI PriorityClasses externally.
 
 ### Changed
 
 ### Fixed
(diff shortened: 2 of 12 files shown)


pkg/operator/operands/common/common.go
```diff
@@ -28,6 +28,11 @@ import (
 
 var controllerTypes = []string{"Deployment", "DaemonSet"}
 
+const (
+	OperatorManagedByLabelKey   = "app.kubernetes.io/managed-by"
```
SiorMeir, line 32: NIT: this is a k8s global. Is there a way to import this from the k8s package instead of hard-coding it?

README.md
```diff
@@ -86,6 +86,22 @@ Run the following command after replacing `<VERSION>` with the desired release v
 helm upgrade -i kai-scheduler oci://ghcr.io/kai-scheduler/kai-scheduler/kai-scheduler -n kai-scheduler --create-namespace --version <VERSION>
 ```
 
+### Embedding the chart
+
+The chart can be embedded as a subchart while Config and cluster-scoped resources are managed externally:
```
SiorMeir: NIT: maybe make it a bit clearer here that doing this will remove kai's operator from these responsibilities

SiorMeir (dismissed):

README.md
gshaibi: KAI Operator was never responsible for these actually.
I agree with you that this might be a bit confusing, and I think we should just drop this doc modification until we get a proper request.
What do you think?

pkg/operator/operands/common/common.go
gshaibi: Unfortunately I couldn't find an exported constant at our current imported packages ):

README.md
SiorMeir: sure - If we target these specific values and not looking at the "bigger" scope of responsibility that's ok.

In the future I suggest we'll add a "manage it yourself" doc with a link to the main README

SiorMeir (approved):

PR #1788, opened 2026-06-28
feat(scheduler): add topology level aliases

### LINKED ISSUE #1498: feat: Support user-friendly names for topology levels in Topology CRD

Add an optional `alias` field to each level entry in the Topology CRD:

```
spec:
  levels:
    - labelKey: "accelerator.nvidia.com/rack"
      alias: "rack"
    - labelKey: "kubernetes.io/hostname"
      alias: "node"
```
When a user specifies a topology constraint on their workload, the scheduler should accept either the `labelKey` or the `alias` and resolve it to the correct topology level for scheduling decisions. `alias` values must be unique within a Topology resource.
Existing Topology CRs without the alias field remain valid and fully functional — no migration is needed.

### Why is this needed?

Topology levels in the KAI Topology CRD are defined using raw node label keys (e.g., `accelerator.nvidia.com/rack`). These are verbose and opaque to end users who need to reference them when specifying topology constraints on their workloads. Admins should be able to expose intuitive aliases like `rack` or `node` without changing the underlying label structure.

This is particularly relevant for platforms built on top of KAI, such as Run:ai, where:

- End users submit workloads without knowledge of the underlying cluster's node label taxonomy
- Different node pools (scheduling shards) may use different label keys to represent the same logical topology level (e.g., `rack`). Supporting aliases in the Topology CRD lays the groundwork for users to express a topology constraint once using a friendly name, and have it resolved to the correct label key per node pool — though cross-shard resolution is out of scope for this issue

SiorMeir:
## Description

Adds **topology level aliases**: an optional `alias` on each `Topology` level (`kai.scheduler/v1alpha1`), so a workload can express a topology constraint with a friendly name (e.g. `rack`) instead of the verbose node label key. Aliases are defined with the topology and are **cluster-global**.

- **API:** `Topology.spec.levels[].alias` (optional). When a level has no alias, behavior is unchanged.
- **One-to-one, enforced by a validating webhook:** a new `Topology` validating webhook (hosted by the admission service) rejects duplicate alias names and any alias that collides with a `nodeLabel`.
- **Mutable:** the `Topology.spec.levels` immutability rule is relaxed to freeze only the `nodeLabel` structure, so aliases can be edited; the nodeLabel-uniqueness CEL rule is updated to compare on `nodeLabel`.
- **Resolution happens once, at the source.** When the cluster snapshot is built, every job's constraint level strings (`RequiredLevel`/`PreferredLevel`) are resolved to canonical node labels (`TopologyConstraintInfo.ResolveAliases` → `PodGroupInfo.ResolveTopologyAliases` → applied in `cluster_info.Snapshot()` before constraint signatures are computed). Every downstream consumer — the topology plugin **and** the idle-GPU topology-aware scenario filter — then reads canonical labels and needs no alias handling. Raw label keys always pass through unchanged.

The `SchedulingShard` CRD is unchanged.

> **Why resolve at the source?** An earlier revision resolved aliases inside the topology plugin. Review surfaced that the idle-GPU topology-aware filter independently indexes node labels by the constraint level string and would silently break for aliased workloads. Rather than resolve per-consumer (fragile — the next consumer reintroduces the bug), resolution now happens once when the snapshot is built, so every current and future consumer is correct by construction.

## Related Issues

Resolves #1498

## Checklist

- [x] Self-reviewed

### Tests
- **Unit:** `TopologyConstraintInfo.ResolveAliases` (alias→label, pass-through, signature invalidation, nil-safe); `PodGroupInfo.ResolveTopologyAliases` (root + subgroups + podsets, per-topology maps, pass-through); webhook validator (duplicate alias, alias/nodeLabel collision, re-point on update); topology plugin matching unaffected.
- **E2E (thin):** the test Topology carries an alias on the rack level; a workload references it by alias and co-locates into one rack.

## Breaking Changes

None. The new field is optional and absent ⇒ unchanged behavior; existing objects remain valid with no migration. The only change to an existing rule relaxes `Topology.spec.levels` immutability (strictly more permissive).

## Additional Notes

- New `Topology` validating webhook hosted by the admission service (`kai.scheduler/v1alpha1` scheme added; one cluster-scoped `ValidatingWebhookConfiguration` for `topologies`, generated by the operator).
- Performance: resolution runs once per snapshot (not on the per-node hot path); per-topology alias maps are built once, lookups are O(1), and it's a no-op when no topology declares an alias.


test/e2e/suites/allocate/topology/topology_test.go
```diff
@@ -251,6 +251,57 @@ var _ = Describe("Topology", Ordered, func() {
 		})
 	})
 
+	// Thin end-to-end check that a Topology level alias resolves through the real scheduler.
+	// The test Topology carries an alias on the rack level (see rd.CreateRackZoneTopology);
+	// one-to-one validation is covered by unit tests.
```
SiorMeir: remove

test/e2e/modules/resources/rd/topology.go
```diff
@@ -29,6 +29,9 @@ const (
 	TestZoneLabelKey = "e2e-topology-label/zone"
 	TestRackLabelKey = "e2e-topology-label/rack"
 	NodeNameLabelKey = "kubernetes.io/hostname"
+	// TestRackAlias is a friendly alias for TestRackLabelKey, set on the test Topology so workloads
+	// can reference the rack level by alias. Benign for tests using the raw key (pass-through).
```
SiorMeir: remove

pkg/scheduler/plugins/topology/topology_plugin.go
```diff
@@ -119,3 +121,42 @@ func (*topologyPlugin) addNodeDataToTopology(topologyTree *Info, topology *kaiv1
...
+		if aliases == nil {
+			aliases = map[string]string{}
+		}
+		aliases[level.Alias] = level.NodeLabel
+	}
+	return aliases
+}
+
+// resolveLevel maps a topology-constraint level string to its canonical node label key.
+// An alias resolves to its configured label; any other string passes through unchanged.
```
SiorMeir: shorten this:

`Return label[alias] or label directly based on input`

pkg/scheduler/plugins/topology/topology_plugin.go
```diff
@@ -119,3 +121,42 @@ func (*topologyPlugin) addNodeDataToTopology(topologyTree *Info, topology *kaiv1
...
+func (info *Info) resolveLevel(level string) string {
```
SiorMeir: maybe `extractLabel`?

SiorMeir:
@coderabbitai review

pkg/apis/kai/v1alpha1/topology_types.go
```diff
@@ -57,8 +57,8 @@ type TopologySpec struct {
 	// +listType=atomic
 	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:MaxItems=16
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="field is immutable"
-	// +kubebuilder:validation:XValidation:rule="size(self.filter(i, size(self.filter(j, j == i)) > 1)) == 0",message="must be unique"
+	// +kubebuilder:validation:XValidation:rule="self.map(l, l.nodeLabel) == oldSelf.map(l, l.nodeLabel)",message="nodeLabel structure is immutable; only aliases may be edited"
+	// +kubebuilder:validation:XValidation:rule="size(self.filter(i, size(self.filter(j, j.nodeLabel == i.nodeLabel)) > 1)) == 0",message="nodeLabel must be unique"
```
gshaibi, line 61: Neat!

pkg/scheduler/plugins/topology/job_filtering.go
```diff
@@ -44,7 +45,10 @@ func (t *topologyPlugin) subSetNodesFn(
 		return []node_info.NodeSet{nodeSet}, nil
 	}
 
-	id, level, validNodes := lowestCommonDomainID(nodeSet, topologyTree.TopologyResource.Spec.Levels, subGroup.GetTopologyConstraint())
+	// Resolve constraint aliases to canonical node label keys once per subgroup, before any matching.
+	resolvedConstraint := topologyTree.resolveConstraint(subGroup.GetTopologyConstraint())
```
gshaibi: Do we also need to use it on https://github.com/kai-scheduler/KAI-Scheduler/blob/04e8472c359b7f62cd412b92011076565b820efa/pkg/scheduler/actions/common/solvers/accumulated_scenario_filters/idle_gpus/topology_aware_idle_gpus.go#L220?

SiorMeir: this was a really good catch - yes, we needed that there.

I've re worked the solution to do the alias translation on a more global level so we would not have to worry about doing it at the point we use the label.

itsomri (approved):

pkg/scheduler/api/podgroup_info/job_info.go
```diff
@@ -165,6 +165,36 @@ func (pgi *PodGroupInfo) GetAllPodSets() map[string]*subgroup_info.PodSet {
 	return pgi.PodSets
 }
 
+// ResolveTopologyAliases rewrites every subgroup/podset topology constraint's level strings to the
+// canonical node labels, using per-topology alias maps (topology name -> alias -> nodeLabel). It is
+// applied once when the snapshot is built, so every downstream consumer (topology plugin, solvers)
+// reads canonical labels and never has to resolve aliases itself.
+func (pgi *PodGroupInfo) ResolveTopologyAliases(aliasesByTopology map[string]map[string]string) {
```
itsomri, line 172: What if a podgroup refers to a non existing topology level because of the alias? Will the user get an indication?

SiorMeir: It'll error out the same way "regular" topology annotations that point to topologies that don't exist

We translate these aliases early on so they should behave the same

PR #1697, opened 2026-06-15
docs(scheduler): update semi-preemptible HLD with minSubGroups and multi-level tree semantics

SiorMeir:
## Description

Updates the semi-preemptible design document to reflect the current API (which now includes `minSubGroups`) and resolves open questions from issues #1585 and #1596.

Key changes:
- Clarifies that the core/elastic pod split applies exclusively at `minMember` leaf PodSets — `minSubGroups` intermediate nodes define scheduling gates only, not preemption thresholds
- Defines multi-level tree behavior: semi-elasticity is a pod-level concept; subgroups are atomic scheduling units (not semi-elastic)
- Adds immutability constraint: `minMember` and `minSubGroups` must not increase on semi-preemptible PodGroups post-creation (validation webhook requirement)
- Documents future `minNonPreemptible` field: scope, work required, and the three-tier pod ordering complexity it introduces

## Related Issues

Fixes #1585
Fixes #1596

## Checklist

- [ ] Self-reviewed
- [ ] Added/updated tests (if needed)
- [x] Updated documentation (if needed)

## Breaking Changes

None — documentation only.

## Additional Notes

This is an HLD update only. Technical design and implementation are tracked separately.

PR #1804, opened 2026-07-01
feat(scheduler): semi-preemptible podgroup mode

SiorMeir:
## Description

Adds a third preemptibility mode, **`semi-preemptible`**, to KAI Scheduler.

A semi-preemptible PodGroup is non-preemptible up to its **minimal satisfying shape** — computed
recursively as the `minSubGroup` highest-priority child subgroups at each SubGroupSet and `minMember`
pods at each leaf PodSet — and everything beyond that minimum is **elastic**: allocated over-quota and
reclaimed/preempted first. This lets inference and elastic-training workloads guarantee a minimum while
bursting opportunistically. Elasticity applies at every level of the subgroup tree (whole surplus
subgroups burst/reclaim like surplus pods do at a leaf). No new public API fields — reuses the existing
`preemptibility` field/label, `minMember`, and `minSubGroup`.

Implements the approved HLD (PR #1697). Highlights:
- `SemiPreemptible` preemptibility value + `IsSemiPreemptibleJob()`.
- New `podgroup_info.GetCoreTasks` — single source of truth for the non-preemptible core set, mirroring
  the eviction protection walk so quota accounting and eviction agree by construction.
- Tree-aware quota accounting (only the core counts toward `AllocatedNotPreemptible`) and eviction that
  offers only the elastic surplus (the core is never a victim; phase-3 full-eviction is skipped).
- Semi-preemptible jobs admitted as reclaim/preempt victim candidates (core protected downstream).
- Immutability webhook: rejects increases to `minMember`/`minSubGroup` on a running semi-preemptible
  PodGroup.
- **Automated segmentation is out of scope and mutually exclusive** with semi-preemptible, soft-enforced
  in the PodGrouper via a non-blocking `PodGrouperWarning` event (a fully-gang auto-segmented tree has no
  surplus, so semi-preemptible would be inert). Hand-authored `minSubGroup` trees are the supported combo.

## Related Issues

Design: #1697

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. The mode is strictly opt-in; existing workloads are unaffected.

## Additional Notes

- User-facing docs: `docs/elastic/README.md` (Semi-Preemptible Workloads) + `examples/semi-preemptible/`;
  limitation note in `docs/topology/segments.md`.
- Tests: unit tests across `podgroup_info`, `proportion`/`capacity_policy`, `v2alpha2` webhook, and the
  podgrouper groupers/controller; plus an E2E two-step expand→scale-down lifecycle spec in
  `test/e2e/suites/preempt/`.
- Verified locally: `make validate` and `make test` both green.

## Summary by CodeRabbit

* **New Features**
  * Added support for a new `semi-preemptible` workload mode, including API validation, scheduling behavior, and examples.
  * Documented how to enable this mode and how its core vs. elastic portions behave.
  * Added end-to-end coverage for semi-preemptible preemption scenarios.

* **Bug Fixes**
  * Semi-preemptible workloads now emit warnings when combined with automatic segmentation.
  * Improved quota and eviction handling so protected core tasks are kept while elastic surplus is reclaimed first.


SiorMeir:
@coderabbitai review

pkg/apis/scheduling/v2alpha2/podgroup_webhook.go
```diff
@@ -59,6 +68,65 @@ func (_ *PodGroup) ValidateUpdate(ctx context.Context, _ *PodGroup, podGroup *Po
...
+				"cannot increase minSubGroup (%s -> %s) on subgroup %q of a semi-preemptible PodGroup",
+				formatInt32Ptr(oldSG.MinSubGroup), formatInt32Ptr(newSG.MinSubGroup), newSG.Name)}
+		}
+	}
+
+	return nil
+}
+
+// increasedInt32 reports whether the value grew from old to updated. A nil old with a non-nil updated is
+// treated as an increase (unset -> set raises the minimal satisfying set); other transitions to nil are not.
```
SiorMeir: Simplify naming - this will remain as a util function no need to over-verbose it

pkg/apis/scheduling/v2alpha2/podgroup_webhook.go
```diff
@@ -59,6 +68,65 @@ func (_ *PodGroup) ValidateUpdate(ctx context.Context, _ *PodGroup, podGroup *Po
...
+			return &semiPreemptibleImmutabilityError{msg: fmt.Sprintf(
+				"cannot increase minMember (%s -> %s) on subgroup %q of a semi-preemptible PodGroup",
+				formatInt32Ptr(oldSG.MinMember), formatInt32Ptr(newSG.MinMember), newSG.Name)}
+		}
+		if increasedInt32(oldSG.MinSubGroup, newSG.MinSubGroup) {
+			return &semiPreemptibleImmutabilityError{msg: fmt.Sprintf(
+				"cannot increase minSubGroup (%s -> %s) on subgroup %q of a semi-preemptible PodGroup",
+				formatInt32Ptr(oldSG.MinSubGroup), formatInt32Ptr(newSG.MinSubGroup), newSG.Name)}
+		}
+	}
```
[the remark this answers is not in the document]
SiorMeir: These scenarios are not dealt in this logic by design.

pkg/apis/scheduling/v2alpha2/podgroup_webhook_validation_errors.go
```diff
@@ -41,6 +41,13 @@ type missingMinMemberError struct{ msg string }
 
 func (e *missingMinMemberError) Error() string { return e.msg }
 
+// semiPreemptibleImmutabilityError is returned when an update increases minMember or minSubGroup on a
+// semi-preemptible PodGroup. Raising either would reclassify already-running over-quota elastic
+// pods/subgroups as core, silently growing the minimal satisfying set and breaking quota invariants.
```
SiorMeir: remove

SiorMeir: or at least slim down

pkg/podgrouper/podgroup/metadata.go
```diff
@@ -39,6 +39,24 @@ type Metadata struct {
...
 	Topology               string
+
+	// Warnings collects non-blocking messages surfaced as Warning events on the pod by the controller.
+	// It is an internal field, not part of any public API surface.
+	Warnings []string
+}
+
+// semiPreemptibleSegmentedWarning is emitted when a workload requests both automated segmentation and
+// semi-preemptible. The two are mutually exclusive: a fully-gang auto-segmented tree has no surplus, so
+// semi-preemptible is inert and the workload behaves as non-preemptible.
```
SiorMeir: remove

pkg/podgrouper/podgroup/metadata.go
```diff
@@ -39,6 +39,24 @@ type Metadata struct {
...
+const semiPreemptibleSegmentedWarning = "semi-preemptible is not compatible with automatic segmentation " +
+	"(kai.scheduler/segment-size): the segmented tree is fully gang and has no elastic surplus, so the " +
+	"workload will behave as non-preemptible"
+
+// WarnIfSemiPreemptibleSegmented appends a warning when a segmented workload is also semi-preemptible.
```
SiorMeir: remove

pkg/scheduler/api/podgroup_info/job_info.go
```diff
@@ -221,6 +225,17 @@ func (pgi *PodGroupInfo) IsWithinPreemptionDelay(now time.Time) bool {
 	return end != nil && now.Before(*end)
 }
 
+func (pgi *PodGroupInfo) IsSemiPreemptibleJob() bool {
+	return pgi.Preemptibility == enginev2alpha2.SemiPreemptible
+}
+
+// HasEvictableTasks reports whether any of this job's allocated tasks may be taken as a victim.
+// Preemptible jobs offer all of them; semi-preemptible jobs offer their elastic surplus only
+// (GetTasksToEvict decides which); non-preemptible jobs offer none.
```
SiorMeir: remove

pkg/podgroupcontroller/utilities/pod-group/preemptible.go
```diff
@@ -19,14 +19,17 @@ import (
 	pg "github.com/kai-scheduler/KAI-scheduler/pkg/common/podgroup"
 )
 
-func IsPreemptible(ctx context.Context, podGroup *v2alpha2.PodGroup, kubeClient client.Client) (bool, error) {
+// GetPreemptibility resolves the podgroup's effective preemptibility, falling back to its priority
+// when the field is unset.
```
SiorMeir: remove

pkg/scheduler/api/podgroup_info/job_info.go
SiorMeir: remove

pkg/scheduler/api/podgroup_info/job_info.go
```diff
@@ -501,21 +516,13 @@ func (pgi *PodGroupInfo) IsStale() bool {
...
-		if !podSet.IsGangSatisfied() {
-			return true
-		}
-	}
-	return false
+	return !pgi.IsGangSatisfied()
 }
 
+// IsGangSatisfied walks the SubGroupSet tree rather than the flat PodSets map, so a job with
+// minSubGroup is satisfied once that many children are - not once every leaf is.
```
SiorMeir: remove

docs/developer/designs/semi-preemptible/README.md
```diff
@@ -15,7 +15,7 @@ We want to add a new 3rd mode, named **semi-preemptible**, where the podgroup is
 - Change nothing for existing workloads — the mode is strictly opt-in.
 
 **Non-Goals**
-- **Automated segmented subgroups** (the `kai.scheduler/segment-size` annotation path). Automated segmentation is **out of scope** and mutually exclusive with semi-preemptible; the combination is soft-enforced (see [Automated Segmentation](#automated-segmentation-out-of-scope)). Subgroup-level elasticity is supported only for **hand-authored** subgroup trees.
+- **Automated segmented subgroups** (the `kai.scheduler/segment-size` annotation path) are not a design target: the core/elastic split is specified against **hand-authored** subgroup trees. The two are not rejected or warned, and the outcome follows from the tree each grouper emits. A LeaderWorkerSet segmented tree is fully gang (every segment's `minAvailable` equals its size), so there is no surplus and semi-preemptible is inert. A segmented `PyTorchJob` with `minReplicas` leaves its trailing segments at `minAvailable: 0`, so those pods are elastic and semi-preemptible behaves as designed.
```
davidLif, line 18: Are we sure about the behavior for PyTorchJob? This sound strange

pkg/apis/scheduling/v2alpha2/podgroup_webhook.go
```diff
@@ -59,6 +68,63 @@ func (_ *PodGroup) ValidateUpdate(ctx context.Context, _ *PodGroup, podGroup *Po
...
+	if updated == nil {
+		return false
+	}
+	if old == nil {
+		return true
+	}
+	return *updated > *old
+}
+
+func formatInt32Ptr(v *int32) string {
```
davidLif, line 121: Use the ptr package instead

docs/developer/designs/semi-preemptible/README.md
```diff
@@ -135,11 +149,25 @@ reconciles because core membership was ranked by live allocation ratio.
...
-resource vector per job (`coreResourceQuantities` in the proportion plugin). It publishes that
-amount at the **PodGroup level**, and the podgroupcontroller **consumes** it instead of re-deriving
-which pods are "core". This removes the divergence and the flip by construction, and stays fully
-**Pod-agnostic** — no pod-level labels or markers (an anti-pattern for this feature).
+The **scheduler is the single source of truth** for *which pods are core*; the podgroupcontroller
+remains the single source of truth for *how much those pods cost*. The scheduler publishes the core
+pod names at the PodGroup level, and the controller sums exactly that set through its existing
+per-pod accounting path. This removes the divergence and the flip by construction.
+
+The scheduler deliberately does **not** publish a resource amount, even though it has one
```
itsomri, line 157: That's not a good reasoning. We can translate the resource vector to resource quantities, and anyway these are the ones that matter for quotas. This entire paragraph is on the wrong premise. I'd just leave it (calculating the allocated resources) as something that's out of scope for this PR

docs/developer/designs/semi-preemptible/README.md
```diff
@@ -135,11 +149,25 @@ reconciles because core membership was ranked by live allocation ratio.
...
+(`coreResourceQuantities` in the proportion plugin). That vector is `rs.ResourceQuantities` — CPU in
+millicores, memory in MB, GPU as fractions, and only those three dimensions — whereas
+`ResourcesStatus.Allocated` is a `v1.ResourceList` in native pod-request units, built by
+`metadata.calculatedAllocatedResources`, carrying GPU-sharing and DRA extras the scheduler's vector
+does not model. Publishing the scheduler's vector into `AllocatedNonPreemptible` would place two
+incompatible unit systems in sibling fields of the same struct, and `queuecontroller` **sums** both
+into the queue rollup — so the mismatch would not merely lose fidelity, it would produce a wrong
+rollup. Naming pods keeps the resource math in the one component that owns the units.
+
+Pod *names* are not pod *markers*: nothing is written to the pods themselves, so the feature stays
```
itsomri, line 167: This one is also weak, I don't get what you mean here. Why are pod-level labels/annotations anti-pattern? What does "directly debuggable" mean here? What does a resource vector have to do with any of this

pkg/scheduler/framework/session.go
```diff
@@ -520,6 +520,9 @@ func closeSession(ssn *Session) {
 	// Push all jobs for status update into the channel
 	resolveDetailedFitErrors := ssn.RecomputeDetailedFitErrors
 	for _, job := range ssn.ClusterInfo.PodGroupInfos {
+		if job.IsSemiPreemptibleJob() {
+			job.CorePodNames = podgroup_info.GetCorePodNames(job, ssn.TaskOrderFn)
```
davidLif, line 524: Why fill the CorePodNames field only for semi-preemtable podgroups? I would expect the concept to be valid in all cases.
All 'MinAvaliable' (directly or by subgroups) pods are core pods. we just treat them differently for each preemptability mode:

Non-preemptable - Do not touch any of the pods, including core pods
Semi-preemptable - Do not touch only these pods, ohers are preemptable
Preemptable - You can preempt all pods, but the core pods are the "ride or die together" pods.

I am afraid that if we have Different ways to calculate and store the "MinAvaliable" pods for each preemptability mode, code divergence and bug might arise.

docs/developer/designs/semi-preemptible/README.md
SiorMeir: This is over sloppified. 

My point here is that we don't want to interact directly with pods more than we have to, and keep pod name as something we should rely on.

I'll remove this entire paragraph

pkg/apis/scheduling/v2alpha2/podgroup_webhook.go
SiorMeir: Done

docs/developer/designs/semi-preemptible/README.md
SiorMeir: I agree. I'll separate this into a different PR & combine @davidLif suggestion of merging Core pods accounting into a unified way.

pkg/scheduler/framework/session.go
SiorMeir: The scope was not to refactor any existing flows, but I have to agree - we have to unify these logics into one place. 
As i've answered in omri's comment, I'll do that in a following PR to reduce review complexity

docs/elastic/README.md
```diff
@@ -19,3 +19,60 @@ kubectl apply -f pytorch-elastic.yaml
...
 And, if additional resources are available, the workload will be able to add 2 additional workers.
 If resources are requested by more prioritized workload, KAI Scheduler will be able to evict only part of its pods and the workload will continue running.
+
+## Semi-Preemptible Workloads
+
+A **semi-preemptible** workload keeps its **minimal required shape** non-preemptible and in-quota, while everything above that minimum runs **elastically** — allocated over-quota and reclaimed/preempted first. This lets inference and elastic-training workloads guarantee a minimum while bursting opportunistically when spare capacity exists.
+
+The core (non-preemptible) set is the tree's minimal satisfying set, computed recursively:
+- at a leaf PodSet, the `minMember` highest-priority pods are core; pods beyond `minMember` are elastic;
+- at a mid-level SubGroupSet, the `minSubGroup` highest-priority child subgroups are core; additional scheduled subgroups are elastic and reclaimed **as a whole subgroup** (never split).
```
davidLif: Never split? an elastic subgroup for an elastic parent will not be able to reclaim a single pod if it's "good enoght"?

SiorMeir: No. 
Elasticity within the subgroup is as implemented today. The idea that we tried to convey here is that subgroups will also adhere to the opportunistic semi-preemptible logic (expand on availability, shrink on reclaim from other jobs)

I'll make that clearer

pkg/scheduler/api/podgroup_info/allocation_info.go
```diff
@@ -175,6 +184,19 @@ func collectTasksFromPodSet(ps *subgroup_info.PodSet, taskOrderFn common_info.Le
 	return getTasksFromQueue(taskPriorityQueue, numTasksToAllocate)
 }
 
+// collectElasticTaskFromPodSet grabs one ready task from ps as extra elastic capacity, regardless
+// of whether ps has met its own minAvailable: the parent's gang requirement is already satisfied by
+// other members by the time this runs, so ps is never required to complete its own gang here. A
+// subgroup that can never reach its minAvailable (a broken/orphaned gang) must still be able to burst.
+func collectElasticTaskFromPodSet(ps *subgroup_info.PodSet, taskOrderFn common_info.LessFn, mode TaskAllocationMode) []*pod_info.PodInfo {
```
itsomri: Can you explain this change? are we allowing sub-groups to be preempted below their minAvailable? Why exactly?

pkg/scheduler/api/podgroup_info/eviction_info.go
```diff
@@ -21,20 +22,68 @@ func GetTasksToEvict(job *PodGroupInfo, subGroupOrderFn, taskOrderFn common_info
...
 
 	jobHasMoreActiveTasksAfterEviction := len(tasks) < job.GetActiveAllocatedTasksCount()
 	return tasks, jobHasMoreActiveTasksAfterEviction
 }
 
+// collectOrphanEviction returns the allocated tasks of a member that holds no core slot and has not
+// formed its gang. Those pods deliver nothing - no gang of their own, no core slot - so a
+// semi-preemptible job gives them up before any real surplus. Recurses into core SubGroupSets, where
+// the same rule applies one level down.
+func collectOrphanEviction(
```
itsomri: 1. "Orphan" implies pods without owner subgroup, which is not the case. Maybe "stale" pods, as in "pods that belong to a stale subgroup" (we already use "stale" for the same context for podgroups, see stalegangeviction)
2. These need to be evicted by the scheduler not just as part of reclaims, otherwise we could get obscure deadlocks in the cluster between jobs

itsomri: (Maybe you did that, I'm sorry if some of my comments miss some context in the PR, it's hard to keep it all in my head and I have to start somewhere)

pkg/scheduler/plugins/proportion/capacity_policy/capacity_policy.go
```diff
@@ -51,21 +51,28 @@ func (cp *CapacityPolicy) IsTaskAllocationOnNodeOverCapacity(task *pod_info.PodI
...
-		result := checkFn(requestedShare, job)
-		if !result.IsSchedulable {
-			log.InfraLogger.V(5).Infof("Job: <%v/%v> is over capacity. Reason: %v", job.Namespace, job.Name, result.Message)
-			return result
-		}
+// coreRequiredQuota is the share to charge against the not-preemptible quota. A semi-preemptible job
+// that already meets its minimum is bursting, so the incoming share is elastic and charges nothing;
+// otherwise (gang phase) the whole share is core.
+func coreRequiredQuota(requestedShare rs.ResourceQuantities, job *podgroup_info.PodGroupInfo) rs.ResourceQuantities {
+	if job.IsSemiPreemptibleJob() && podgroup_info.IsMinRequirementSatisfied(job) {
```
itsomri, line 73: I think this is wrong:
Let's assume a podgroup with an allocated core set. This returns 0 event if there are extra pending pods for this podgroup that will get allocated, so they're allowed to allocate. Then, in the next cycle, we could get a different set of pods that will be considered the "core", right? which could mean that we're suddenly overquota?

pkg/scheduler/api/podgroup_info/allocation_info.go
SiorMeir: Not preemption — allocation. And it's wrong; removed.

What it did: during the parent's elastic phase, `collectElasticTaskFromPodSet` handed semi-preemptible jobs one pod at a time for a PodSet still below its own minAvailable. Every other preemptibility mode goes through collectTasksFromPodSet, which asks for minAvailable - allocated in one batch and takes nothing when the gang can't be completed. A partial gang does no work, so those pods held resources for nothing — and collectOrphanEviction targets exactly that set, so allocation and eviction were pointing at the same pods from opposite ends.

It was written to separate core from elastic subgroups, but that separation already exists: `collectTasksFromSubGroupSet` gates the gang phase on `minSubGroup`, and `SubGroupOrderFn` orders unsatisfied subgroups first, so they're still filled first.

This doesn't affect `collectOrphanEviction`: subgroups still drop below their floor from pod loss, so that path stands on its own merits. Answering your eviction_info.go comment separately.

pkg/scheduler/plugins/proportion/capacity_policy/capacity_policy.go
SiorMeir: As discussed, we'll leave that open for now.
We'll mark this in a separate issue

pkg/scheduler/api/podgroup_info/eviction_info.go
SiorMeir: Both taken.

1. Renamed to `collectStaleEviction` in `67588c5c8` — agreed, "orphan" says "no owning subgroup", which isn't this.

2. We'll deal with this in the following accounting PR

docs/developer/designs/semi-preemptible/README.md
SiorMeir: Checked it against `pytorch_grouper.go` — the sentence is accurate but buries the interesting part:

- `:192` sets `MinSubGroup: numSegments`, so **every** segment is required and the subgroup-level core/elastic split is inert.
- `:196-208` — segments at or past `ceil(workerMinAvailable / segmentSize)` get `MinAvailable: 0`.

So the elasticity is pod-level *inside* trailing segments, not subgroup-level. And the rounding means `workerMinAvailable: 5` with `segmentSize: 4` gives a core of 8, not 5 — segmentation inflates the minimum to a segment boundary.

That predates this PR and is exactly why segmented trees are a non-goal here. Want it spelled out in the doc?

pkg/scheduler/framework/session.go
SiorMeir: Updating this one: core membership is now pinned to `status.schedulingState.corePods` instead of being recomputed each session (`7b5116c13`), so the "different ways to calculate" divergence you flagged is narrower than when I first replied — the stability mechanism now exists in one place.

Still deferred to the follow-up: publishing `CorePods` for *all* preemptibility modes. This line still gates on `IsSemiPreemptibleJob()`, which is your actual point.

docs/developer/designs/semi-preemptible/README.md
```diff
@@ -209,31 +229,25 @@ if state := originalStatus.SchedulingState; state != nil && state.NonPreemptible
...
 - **Ownership**: the scheduler and controller already patch `PodGroup.Status` on disjoint paths
   (`schedulingConditions` / annotations vs `resourcesStatus`); adding scheduler-written
   `schedulingState` introduces no new write-conflict class.
-- **Staleness**: the value refreshes each scheduler session and only changes when allocation changes
-  (which is scheduler-driven). `LastEvaluatedSession` is the future hook if staleness detection is
-  ever needed.
-- **Defense-in-depth**: Option A removes the need for both sides to recompute core independently. If
-  they ever must, core ordering should be deterministic *and reproducible* — rank by static priority
-  then name rather than by live allocation ratio (the root cause of the flip).
+- **Staleness**: the set refreshes each scheduler session and only changes when core membership
```
itsomri, line 242: This isn't really in the api right?

SiorMeir: nope. this `LastEvaluatedSession` isn't in the API. The only field there is PodGroupSchedulingState.CorePods (podgroup_types.go:205)

itsomri (approved):

PR #1809, opened 2026-07-01
perf(scheduler): reduce min-runtime reclaim memory usage

### LINKED ISSUE #1808: Reduce reclaim memory retained by the min-runtime protection cache

Refactor the min-runtime reclaim protection check to remove the pending-job/victim-job boolean cache.

The effective reclaim min-runtime duration is already cached by pending queue and victim queue in `resolver.reclaimMinRuntimeCache`. After resolving that duration, reclaim protection can be calculated directly from the victim's last-start timestamp on every check:

```go
protectedUntil := victim.LastStartTimestamp.Add(minRuntime.Duration)
return time.Now().Before(protectedUntil)
```

The refactor should:

- Retain the existing queue-pair duration cache.
- Remove `reclaimProtectionCache`, including its session lifecycle and insertion/lookup paths.
- Re-evaluate protection from the current victim timestamp instead of retaining the first boolean result for each job pair.
- Add a scalable benchmark matching the 500-node `reclaims many single GPU Jobs` scale test and a regression test for repeated protection checks.

### Why is this needed?

The current cache is keyed by pending job and victim job, so its retained state grows as `O(pending jobs × victim jobs)`. The 500-node scale test has eight GPUs per node, 4,000 running single-GPU jobs, and 4,000 pending single-GPU jobs. Reclaim can therefore populate approximately 16 million cache entries in one scheduler session.

This workload does not configure a nonzero reclaim min-runtime, so those entries primarily retain duplicate `false` results even though all pending jobs share one queue and all victims share another queue.

The 500-node benchmark and live-heap capture measured:

| Metric | Before | After | Change |
|---|---:|---:|---:|
| Reachable live heap after reclaim, with the session retained | 940.09 MB | 94.88 MB | -845.21 MB / -89.9% |
| Total benchmark allocations | 76.16 GB | 74.45 GB | -1.71 GB / -2.25% |
| Benchmark runtime | 172.48 s | 165.61 s | -4.0% |

Before the refactor, `minruntimePlugin.cacheReclaimProtection` retained 856.87 MB, or 91.15% of the benchmark's reachable live heap. After removing the cache, neither the allocation profile nor the live-heap profile contains any `minruntime` samples.

The boolean cache also freezes the protection decision at the first evaluation of each job pair. Recomputing from the cached duration and victim timestamp avoids that per-pair time dependency while removing the scheduler's largest retained allocation in this reclaim workload.

enoodle:
## Description

- Adds a scalable reclaim benchmark matching the `reclaims many single GPU Jobs` scale-test shape: eight running and eight pending one-GPU jobs per node, with 10, 50, 100, 200, and 500-node entry points. A dedicated 500-node variant configures a 30-second reclaim min-runtime while keeping one-minute-old victims reclaimable.
- Removes the quadratic pending-job/victim-job min-runtime protection cache while retaining the existing queue-pair duration cache.
- Re-evaluates reclaim protection from the victim's last-start timestamp and adds a regression test for repeated checks of the same job pair.

The 500-node benchmark contains 4,000 running jobs and 4,000 pending jobs. The previous cache could retain approximately 16 million boolean entries during one scheduler session.

| Metric | Before | After | Change |
|---|---:|---:|---:|
| Reachable live heap after reclaim, with the session retained | 940.09 MB | 94.88 MB | -845.21 MB / -89.9% |
| Total benchmark allocations | 76.16 GB | 74.45 GB | -1.71 GB / -2.25% |
| Benchmark runtime | 172.48 s | 165.61 s | -4.0% |

After the change, neither the allocation profile nor the live-heap profile contains any `minruntime` samples.

### Nonzero min-runtime benchmark

`BenchmarkReclaimManySingleGPUJobsWithMinRuntime_500Node` configures a 30-second reclaim min-runtime on the victim queue. Running jobs are one minute old, so the duration has expired and the benchmark still verifies the complete replacement of all 4,000 running jobs by 4,000 pending jobs.

| Version | Runtime | Allocated | Allocations |
|---|---:|---:|---:|
| Job-pair protection cache | 154.56 s | 76.15 GB | 792,516,512 |
| Protection cache removed | 157.79 s | 74.44 GB | 792,336,359 |
| Difference | +2.1% | -1.71 GB / -2.25% | -180,153 |

The allocation reduction remains stable with a nonzero min-runtime. The 2.1% single-run timing difference is smaller than the observed variance across 500-node runs, so repeated samples are required for a statistically meaningful runtime comparison.

## Related Issues

Fixes #1808

## Checklist

- [x] Self-reviewed

## Breaking Changes

None.

### DIFF
--- a/pkg/scheduler/plugins/minruntime/minruntime.go
+++ b/pkg/scheduler/plugins/minruntime/minruntime.go
@@ -30,8 +30,6 @@ type minruntimePlugin struct {
 	reclaimResolveMethod     string
 	queues                   map[common_info.QueueID]*queue_info.QueueInfo
 
-	preemptProtectionCache map[common_info.PodGroupID]bool
-
 	resolver *resolver
 }
 
@@ -63,9 +61,6 @@ func New(arguments framework.PluginArguments) framework.Plugin {
 		log.InfraLogger.Errorf("Invalid reclaim resolve method %v, using default value %v", plugin.reclaimResolveMethod, resolveMethodLCA)
 		plugin.reclaimResolveMethod = resolveMethodLCA
 	}
-	// setup caches on plugin init, but they will be reset on session open anyway
-	plugin.preemptProtectionCache = make(map[common_info.PodGroupID]bool)
-
 	return plugin
 }
 
@@ -79,13 +74,11 @@ func (mr *minruntimePlugin) OnSessionOpen(ssn *framework.Session) {
 	ssn.AddReclaimScenarioValidatorFn(mr.reclaimScenarioValidatorFn)
 	ssn.AddPreemptScenarioValidatorFn(mr.preemptScenarioValidatorFn)
 	mr.queues = ssn.ClusterInfo.Queues
-	mr.preemptProtectionCache = make(map[common_info.PodGroupID]bool)
 	mr.resolver = NewResolver(mr.queues, mr.defaultPreemptMinRuntime, mr.defaultReclaimMinRuntime)
 }
 
 func (mr *minruntimePlugin) OnSessionClose(ssn *framework.Session) {
 	mr.queues = nil
-	mr.preemptProtectionCache = nil
 	mr.resolver = nil
 }
 
@@ -161,9 +154,6 @@ func (mr *minruntimePlugin) isReclaimMinRuntimeProtected(pendingJob *podgroup_in
 }
 
 func (mr *minruntimePlugin) isPreemptMinRuntimeProtected(_ *podgroup_info.PodGroupInfo, victim *podgroup_info.PodGroupInfo) bool {
-	if cached, ok := mr.preemptProtectionCache[victim.UID]; ok {
-		return cached
-	}
 	victimQueue := mr.queues[victim.Queue]
 
 	minRuntime, err := mr.resolver.getPreemptMinRuntime(victimQueue)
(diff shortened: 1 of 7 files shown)


pkg/scheduler/plugins/minruntime/minruntime.go
```diff
@@ -31,7 +31,6 @@ type minruntimePlugin struct {
 	queues                   map[common_info.QueueID]*queue_info.QueueInfo
 
 	preemptProtectionCache map[common_info.PodGroupID]bool
-	reclaimProtectionCache map[common_info.PodGroupID]map[common_info.PodGroupID]bool
```
davidLif, line 34: If you remove the reclaimProtectionCache, I think you should remove preemptProtectionCache as well. otherwise, we have 2 different behaviors - for preempt, we look at the "now" time only once, and keep it for the whole cycle, and for reclaim, Update the "now" base for each protected query.

enoodle: This is essentially the same because the plugin is being initialized in every session, so the behavior is the same. The preempt cache is not exploding in size so I didn't want to remove it but it is possible as well.

gshaibi: It is indeed initialized in every session, but during the session the behavior drifted, right?
Before we calculated protection once and now we recalculate in real time.
In case a session is very long, do you think @davidLif 's concern might be significant?

enoodle: I removed the preempt cache as well

gshaibi (dismissed):

davidLif (approved):

Issue #1783, opened 2026-06-28
feat(scheduler): design graceful resolution of dynamic queue quota changes (scale-in/out over-quota)


SiorMeir:
### What you would like to be added?

A design for how KAI-Scheduler should prevent and gracefully resolve the over-quota / over-limit states that arise when a Queue's `spec.resources` change on a running cluster (scale-in, scale-out, hierarchy restructuring).

***Problem scenarios***

1. Deserved (quota) scale-in — non-preemptible allocations exceed the reduced quota; they can't be reclaimed. What is the contract?
2. limit scale-in — allocated exceeds the new limit; only future allocations are blocked, existing ones run over limit indefinitely.
3. Hierarchy restructuring — parent quota dropped below sum of children, or re-parenting, oversubscribes a parent.
4. overQuotaWeight / scale-out — likely safe; confirm as no-op and document.

### Why is this needed?

Operators or users can edit `Queue.spec.resources` (quota/deserved, limit, overQuotaWeight) at any time. A reduction or restructuring can instantly leave allocated workloads over their deserved quota or hard limit, and the system has no defined convergence behavior:

- The validating webhook is opt-in and warning-only, never blocks, has inconsistent CPU/GPU/Memory coverage, and never validates limit or limit >= quota.
- Runtime checks only gate new allocations — nothing acts on workloads that became over-quota/over-limit after a spec change.
- Non-preemptible workloads over a reduced deserved quota cannot be reclaimed and run over quota until completion, silently — no Queue condition, event, or metric reflects it.

enoodle:
There is an option to enable the queue validator (added here: https://github.com/kai-scheduler/KAI-Scheduler/pull/857  - we should add docs on how to use it) it will prevent the first point, but can be extended to guard the others as well

davidLif:
#857 relates only to the third point

davidLif:
This validations should be enabled/disabled by a configuration flag

thc1006:
If I have a few thoughts about this issue (Yes I really have).
Then what will be recommend as the best way to contribute or provide feedback? Would it be better to open a PR for review, since that would allow me to explain the details more clearly, or should I continue the discussion by commenting on the issue page?

thc1006:
[...]

PR #857, opened 2026-01-07
feat(queue-controller): add queue validator

AdheipSingh:
## Description
  Summary of Working Validation:

  1. CREATE: Child queues cannot exceed parent quotas
  2. UPDATE: Warnings when parent quota is reduced below children
  3. DELETE: Parent queues cannot be deleted if they have children

## Related Issues

Fixes #81 

## Checklist

> **Note:** Ensure your PR title follows the [Conventional Commits format](https://github.com/NVIDIA/KAI-Scheduler/blob/main/CONTRIBUTING.md#pr-title-guidelines) (e.g., `feat(scheduler): add new feature`)

- [ ] Self-reviewed
- [ ] Added/updated tests (if needed)
- [ ] Updated documentation (if needed)

## Additional Notes

Screenshot 2026-01-08 at 3 37 49 AM

1. test-parent-queue.yaml
```
apiVersion: scheduling.run.ai/v2
kind: Queue
metadata:
  name: parent-queue
spec:
  resources:
    cpu:
      quota: 2000
    gpu:
      quota: 2
    memory:
      quota: 4096
  ```
  
  2. test-valid-child.yaml 
  ```
  apiVersion: scheduling.run.ai/v2
kind: Queue
metadata:
  name: valid-child-queue
spec:
  parentQueue: parent-queue
  resources:
    cpu:
      quota: 1000
    gpu:
      quota: 1
    memory:
      quota: 2048
  ```
  
  3. test-invalid-child.yaml 
  ```
  apiVersion: scheduling.run.ai/v2
kind: Queue
metadata:
  name: invalid-child-queue
spec:
  parentQueue: parent-queue
  resources:
    cpu:
      quota: 3000  # This exceeds parent's 2000
    gpu:
      quota: 1
    memory:
      quota: 2048
  ```
  
  cc @enoodle 

## Release Notes

* **New Features**
  * Added queue validation webhook with optional parent-child quota relationship checks via new `--enable-quota-validation` flag.
  * Validates resource definitions and prevents deletion of queues with child queues.

* **Tests**
  * Added comprehensive validation tests for queue webhook operations.

PR #1810, opened 2026-07-01
feat(queue-controller): reject queue limit/quota reductions below allocation

thc1006:
## Description

Reworks the queue validating webhook per the design agreed in #1783. Adds a separate opt-in `--enforce-quota-violation` flag (values `None` / `Warning` / `Block`, default `None`). On queue update, per resource (GPU / CPU / memory), the webhook checks that:

- a `limit` is not lowered below the amount currently allocated to the queue (`status.Allocated`), and
- a `quota` is not lowered below the amount allocated to non-preemptible workloads (`status.AllocatedNonPreemptible`), which the scheduler cannot reclaim at runtime.

Flag behavior:

- `--enforce-quota-violation=Block`: the offending update is **rejected**.
- `--enforce-quota-violation=Warning`: the same conditions are reported as admission **warnings**.
- `--enforce-quota-violation=None` (default): the check is **skipped**, so a default install behaves exactly as before.

The existing `--enable-quota-validation` flag (released in v0.14.0) is a **separate flag, kept unchanged**: it independently gates the parent/child quota-relationship warnings.

Behavior details:

- Only genuine reductions are flagged. An unchanged value, or raising a bound, on an already over-limit queue is left alone, so unrelated edits (e.g. `displayName`) are never blocked.
- Only `-1` (unlimited) is treated as unbounded, matching the scheduler. `0` is a real bound (a hard zero cap / zero quota), so lowering a positive bound to `0` below the allocation **is** flagged.
- Allocation is read from the queue's own status (last value persisted by the controller); a missing or empty status counts as no allocation, so freshly created or unreconciled queues are never blocked.
- Spec values (millicpu, MB, GPU fraction) are converted into the status units (cores, bytes, GPU fraction) and both sides are rounded to the queue-metrics precision, so a value set exactly equal to the current allocation is not tripped by float rounding.
- The webhook rule matches the main `queues` resource only (not the `queues/status` subresource), so the scheduler's own status writes do not trigger it.
- This is an admission-time, best-effort check against the last observed allocation. It does not guarantee a queue can never end up over its limit (the scheduler may allocate between the status read and the update); it blocks updates that are already below the observed allocation. Conversely, a stale-high allocation (for example a finished workload whose status has not reconciled down yet) may briefly reject an otherwise-safe reduction until the status catches up.

This is stricter than core ResourceQuota / Kueue / Volcano, which allow the reduction and converge at runtime. The strongest justification is the quota-vs-non-preemptible check: non-preemptible work over quota cannot be reclaimed, so blocking that reduction protects a state the runtime cannot fix. The check is therefore opt-in and off by default.

This revision supersedes the earlier version of the PR (static `limit >= quota` / non-negativity / parent-child checks), which was dropped per the discussion in #1783.

## Related Issues

Part of #1783

## Checklist

- [x] Self-reviewed
- [x] Added/updated tests
- [x] Updated CHANGELOG

## Breaking Changes

None (user-facing). `--enforce-quota-violation` is a new flag defaulting to `None`, so a default install behaves exactly as before. The released `--enable-quota-validation` flag is kept, unchanged. The `--strict-queue-validation` flag from an earlier revision of this PR was never released, so replacing it with `--enforce-quota-violation` breaks nothing.

Internal refactor, no behavior change: the MIG resource-name parser `ExtractGpuAndMemoryFromMigResourceName` moved from the internal `pkg/scheduler/api/common_info/resources` package to `pkg/common/resources`, so the webhook, the queue metric and the scheduler share a single definition. All in-repo callers are updated. External code importing that internal scheduler package path (not a supported public API) would need to switch to the new import.

## Additional Notes

- **GPU accounting.** The allocation-reduction check and the `kai_queue_allocated_gpus` metric now use a shared helper (`pkg/common/resources.SumGpuAllocation`) that counts GPU allocation the way the scheduler does: standard extended GPUs across vendors (any `*gpu` resource), explicit DRA GPU DeviceClass device counts (e.g. `gpu.nvidia.com`), and NVIDIA MIG profiles (`nvidia.com/mig-<slices>g.<mem>gb`, counted as slice portions). GPU memory and other domain-qualified `gpu`-adjacent extended resources (e.g. `nvidia.com/gpu.memory`, `run.ai/gpu.memory`, `volcano.sh/vgpu-memory`) are not counted. This also removes the metric's previous non-deterministic first-match behavior.
- **DRA `AllocationModeAll` is a conservative estimate.** For a DRA request using `AllocationModeAll`, the Queue status producer records one device (the request does not carry the final allocated count), so the "currently allocated" GPU figure for such a request is a lower bound rather than the exact count. This is an existing producer limitation, not introduced here; the enforcement check inherits it.
- **Allocation accounting inherited from the queue status.** The check reads `status.Allocated` / `status.AllocatedNonPreemptible` as written by the PodGroup status producer, which sums running pods' regular container requests plus KAI's GPU-specific resources (GPU-sharing fractions and DRA device counts). It does not yet include init-container peaks, native sidecars (`restartPolicy: Always` init containers), or Pod `Overhead`, which the scheduler does reserve against a queue. For pods that use those, the observed allocation can be lower than the scheduler's, so a reduction can go undetected. This primarily affects CPU and memory. Standard extended-resource GPU requests (`nvidia.com/gpu`, `amd.com/gpu`) and container-level MIG requests (`nvidia.com/mig-*`) placed in init containers or native sidecars are undercounted the same way, since the scheduler folds those into the pod request while the status producer sums only `pod.Spec.Containers`. DRA GPU counts, KAI GPU-sharing fractions and memory, and legacy annotation-driven MIG are added through separate producer paths and are not affected by this omission. This is a pre-existing limitation of the queue-status accounting, shared with the `kai_queue_allocated_*` metrics; aligning the status producer with the scheduler's pod-request accounting is tracked as a follow-up ([#1880](https://github.com/kai-scheduler/KAI-Scheduler/issues/1880)).
- **Flag wiring / Operator scope.** `--enforce-quota-violation` is wired through the queuecontroller options only, matching the existing `--enable-quota-validation` (neither is exposed through the Operator Config / Helm values today, per the v0.14 migration guide). So under a standard Operator install the mode stays `None` unless the deployment args are set. Wiring the mode into the Operator Config CRD can follow as a separate change if maintainers want it plumbed through.
- **`0` / `-1` semantics.** Only `-1` is unlimited (matching the scheduler); `0` is a real bound. Reducing a positive `limit`/`quota` to `0` below the allocation is flagged. This corrects the earlier revision, which skipped `0`.
- **Value validity.** The value-validity check (rejecting `< -1`) remains a separate follow-up; the reduction check no longer treats `< -1` as unbounded, so such values are not silently allowed.
- **Out of scope.** The pre-existing `--enable-quota-validation` parent/child checks retain their baseline `-1` handling and are not touched here. One such gap (children-quota summation ignoring the `-1` unlimited sentinel) is tracked separately in #1881.


thc1006:
Thanks for the review. Addressed all of it in 95c47344:

- `exceedsParent` now treats an unlimited (`-1`) child as exceeding a bounded parent; previously only the parent side was handled.
- The children-sum checks treat any unlimited child as making the whole sum unlimited, rather than raw addition where a `-1` could cancel real usage.
- Reused `constants.UnlimitedResourceQuantity` instead of the local sentinel.
- Added the missing per-child GPU/Memory warnings in `validateChildrenQuotaSum` (it only covered CPU before).
- Messages now render `-1` as `unlimited`, and I added tests for an unlimited child and an unlimited existing sibling under a bounded parent.

enoodle:
@coderabbitai review

pkg/admission/webhook/queuehooks/queue_validator.go
```diff
@@ -92,6 +120,106 @@ func (v *queueValidator) ValidateDelete(ctx context.Context, queue *v2.Queue) (a
...
+}
+
+func gpuAllocated(list v1.ResourceList) float64 {
+	for name, quantity := range list {
+		if strings.HasSuffix(string(name), gpuResourceSuffix) {
+			return quantity.AsApproximateFloat64()
+		}
+	}
+	return 0
+}
```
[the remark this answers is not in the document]
thc1006: Fixed in 8958dd8d — `gpuAllocated` now sums all `/gpu`-suffixed resources instead of returning the first match, so the strict/warn decision is deterministic on multi-vendor queues. Added a regression test covering the `nvidia.com/gpu` + `amd.com/gpu` case.

pkg/admission/webhook/queuehooks/queue_validator.go
```diff
@@ -171,25 +263,25 @@ func (v *queueValidator) validateChildrenQuotaSum(ctx context.Context, parentQue
 		totalChildrenGPU += child.Spec.Resources.GPU.Quota
 		totalChildrenMemory += child.Spec.Resources.Memory.Quota
```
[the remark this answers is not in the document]
thc1006: Good catch. This raw-addition (`-1` unlimited) behavior predates this PR — it's part of the existing #857 warning-only parent/child checks and isn't modified here, so I kept it out of scope to keep this change focused on the allocation-reduction validator. Happy to open a separate issue/PR for it if that's preferred; no strong opinion either way.

pkg/admission/webhook/queuehooks/queue_validator.go
```diff
@@ -6,17 +6,30 @@ package queuehooks
...
 
-const missingResourcesError = "resources must be specified"
+const (
+	missingResourcesError = "resources must be specified"
+
+	// Spec quotas/limits use millicpu for CPU and megabytes for Memory, while the queue status stores
+	// allocation in cores and bytes; these convert spec values into the status units for comparison.
+	milliCPUToCPU     = 1000
+	megabytesToBytes  = 1000000
+	gpuResourceSuffix = "/gpu"
```
enoodle: This should fit the logic from the [scheduler](https://github.com/kai-scheduler/KAI-Scheduler/blob/main/pkg/scheduler/api/resource_info/resource_vector.go#L192-L194) so maybe we can remove the "/" and use the constant?

thc1006: Good call — will switch to `constants.GpuResource` and drop the local suffix, aligning with `isGpuResource` in the scheduler. Folding it into the flag rework so it lands in one update.

thc1006:
Rebased on main and force pushed the review rework.

tldr: the two flags are now a single `--enforce-quota-violation` taking None, Warning or Block.

None leaves validation off (default), Warning surfaces the parent/child quota issues and the reductions below allocation as warnings, Block also rejects those reductions. `--enable-quota-validation` already shipped in v0.14.0 so I replaced it with the new flag plus a migration note rather than keeping a shim. Also moved `gpuAllocated` onto the shared `constants.GpuResource` so it matches `isGpuResource` in the scheduler. Changelog and tests updated, 27 specs green.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ## [Unreleased]
 
 ### Added
+- Added an opt-in `--enforce-quota-violation` flag to the queue-controller with three modes: `None` (default, no validation), `Warning` (report parent/child quota-relationship issues and updates that reduce a resource limit below the amount currently allocated to the queue — or a quota below the amount allocated to non-preemptible workloads, both read from the queue status — as admission warnings), and `Block` (additionally reject those allocation-reducing updates). This replaces the previous `--enable-quota-validation` flag; `--enforce-quota-violation=Warning` is its equivalent. [#1783](https://github.com/kai-scheduler/KAI-Scheduler/issues/1783) [thc1006](https://github.com/thc1006)
```
davidLif: No need to create a migration guide or changelog for --enable-quota-validation. we are handeling it separately.

docs/migrationguides/v0.14.0/README.md
```diff
@@ -33,6 +33,10 @@ When enabled, it adds warnings when child queue quotas exceed their parent's
 quota. This flag is not exposed as a Helm value and defaults to `false`; no
 action is required unless you want to opt in.
 
+> **Note:** In a later release this flag was superseded by `--enforce-quota-violation`,
```
davidLif: No need to create a migration guide for --enable-quota-validation. we are handeling it separately.

thc1006:
TLDR of the recent pushes, in case it helps whoever reviews next:

- GPU accounting is now shared between the webhook and the `kai_queue_allocated_gpus` metric via `SumGpuAllocation` in `pkg/common/resources`. It counts standard extended GPUs across vendors, DRA GPU DeviceClass device counts, and MIG profiles, so `Block` no longer misses DRA or MIG allocation. The metric also stops returning a single GPU resource picked at random.
- Fixed a false positive in that helper: only slash-free DeviceClass keys are treated as DRA GPUs now, so `gpu`-named extended resources that are not GPU counts (for example `volcano.sh/vgpu-memory`) are no longer summed as GPUs. Added tests for DRA, MIG, mixed, and the vGPU-memory case.

One limitation I want to be upfront about: the check reads the queue status allocation, which today sums regular container requests plus KAI's GPU-specific resources, but not init-container peaks, native sidecars, or Pod overhead (which the scheduler does reserve). So a CPU/memory reduction can be under-detected for pods that use those; GPU checks are unaffected. That is a pre-existing queue-status accounting gap, so rather than grow this PR I opened #1880 to track unifying it with the scheduler. It is noted in the PR description too.

Still opt-in and off by default. Happy to adjust anything.

pkg/admission/webhook/queuehooks/queue_validator.go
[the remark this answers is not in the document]
thc1006: This is a real, pre-existing issue, but it lives in the `--enable-quota-validation` parent/child path (`validateChildrenQuotaSum` / `validateParentChildQuota`), not in the allocation-reduction check this PR adds. To keep this PR scoped I filed it separately as #1881, with the failing example and a suggested fix (mirror the `-1`-aware add and compare already in the scheduler's `resource_share/resource_quantities.go`). Resolving this thread since it is outdated and tracked there.

thc1006:
@davidLif one naming question before this releases. #1819 is moving `--enable-quota-validation` to `limitDescendentsOverSubscription` with `None` / `Warning` / `Block`. This PR adds a separate `--enforce-quota-violation` (also `None` / `Warning` / `Block`) for the allocation-reduction check. Do you want the two tri-state flags named consistently, or is `--enforce-quota-violation` the name to keep? Happy to rename now to avoid a post-release compatibility change.

cmd/queuecontroller/app/app.go
```diff
@@ -69,9 +69,15 @@ func Run(opts *Options, clientConfig *rest.Config, ctx context.Context) error {
 		return nil
 	}
 
+	enforcementMode, err := queuehooks.ParseEnforcementMode(opts.EnforceQuotaViolation)
```
davidLif: enforcementMode as a variable name is not very clear. Something like `quotavailoationMode` would be better

pkg/common/resources/gpu.go
```diff
@@ -0,0 +1,45 @@
...
+//   - DRA GPUs: device counts written under the GPU DeviceClass name (e.g. gpu.nvidia.com). DeviceClass
+//     names are Kubernetes object names and never contain "/", which distinguishes them from ordinary
+//     domain-qualified extended resources;
+//   - MIG: nvidia.com/mig-<slices>g.<mem>gb, counted as <slices> GPU portions each, matching the
+//     scheduler's MIG accounting.
+//
+// Domain-qualified extended resources that merely contain "gpu" in their name but are not a GPU count
+// (e.g. nvidia.com/gpu.memory, run.ai/gpu.memory, volcano.sh/vgpu-memory) are not counted: they carry a
+// "/" and match neither the "gpu" suffix nor the slash-free DeviceClass rule.
+func SumGpuAllocation(list v1.ResourceList) float64 {
```
davidLif, line 17: I do not want any changes to code in the services of the scheduler or queuecontroller metrics.  
Where di you take the original calulation from?

pkg/admission/webhook/queuehooks/queue_validator.go
```diff
@@ -57,26 +108,38 @@ func (v *queueValidator) ValidateUpdate(ctx context.Context, oldQueue, newQueue
...
+			}
+			warnings = append(warnings, parentWarnings...)
+		}
+
+		if len(oldQueue.Status.ChildQueues) > 0 {
+			childWarnings, err := v.validateChildrenQuotaSum(ctx, newQueue)
+			if err != nil {
+				return append(warnings, childWarnings...), err
+			}
+			warnings = append(warnings, childWarnings...)
```
davidLif: I would consider moving it to an external function

pkg/admission/webhook/queuehooks/queue_validator.go
```diff
@@ -57,26 +108,38 @@ func (v *queueValidator) ValidateUpdate(ctx context.Context, oldQueue, newQueue
...
 	var warnings admission.Warnings
 
-	if newQueue.Spec.ParentQueue != "" {
-		parentWarnings, err := v.validateParentChildQuota(ctx, newQueue)
-		if err != nil {
-			return parentWarnings, err
+	// Enforce updates that reduce a limit below what the queue already has allocated, or a quota below its
+	// non-preemptible allocation (which the scheduler cannot reclaim), governed by --enforce-quota-violation.
+	// Block rejects such an update; Warning surfaces it as an admission warning.
+	if v.mode.enabled() {
```
davidLif: It would be more readable as `quotaVailoationMode != none`

pkg/admission/webhook/queuehooks/queue_validator.go
```diff
@@ -92,6 +155,106 @@ func (v *queueValidator) ValidateDelete(ctx context.Context, queue *v2.Queue) (a
...
 }
 
+// allocationReductionViolations reports where the update lowers a resource limit below the amount already
+// allocated to the queue, or lowers a quota below the amount allocated to non-preemptible workloads (which the
+// scheduler cannot reclaim). Only genuine reductions are flagged: an unchanged or increased value never trips
+// the check, so edits to unrelated fields on an already-over-limit queue are left alone. Allocation is read
+// from the queue's own status (last value persisted by the controller); a missing entry counts as zero, so
+// freshly created or unreconciled queues are never flagged. Spec values (millicpu, megabytes, GPU fraction) are
+// converted into the status units (cores, bytes, GPU fraction) and both sides are rounded to the queue-metrics
+// precision before comparison, so a value set equal to the current allocation is not tripped by float rounding.
```
davidLif: Excessive comment size

pkg/admission/webhook/queuehooks/queue_validator.go
```diff
@@ -92,6 +155,106 @@ func (v *queueValidator) ValidateDelete(ctx context.Context, queue *v2.Queue) (a
...
+func allocationReductionViolations(oldQueue, newQueue *v2.Queue) admission.Warnings {
+	newSpec := newQueue.Spec.Resources
+	oldSpec := oldQueue.Spec.Resources
+	if newSpec == nil || oldSpec == nil {
+		return nil
+	}
+	status := oldQueue.Status
+
+	perResource := []struct {
```
davidLif: Why hold it in a struct, if several lines later you just check the values directly with if? Both of them can be driven by the same for loop

thc1006:
Dear @davidLif ,

Short version: I'll revert the scheduler and metrics changes, do the five refactors, and rebase. The one thing I'd like your call on is where the GPU helper should live.

**Scoping**

Reverting `pkg/queuecontroller/metrics/metrics.go` to its original form and restoring the scheduler files (the MIG parser package and the four import sites). After that the PR only touches the admission webhook, `cmd/queuecontroller`, `pkg/common/resources` and the CHANGELOG.

**Where the calculation came from**

It isn't a new formula. It follows what's already in the tree: the vendor-agnostic `*gpu` suffix behavior of the existing `kai_queue_allocated_gpus` metric, and the scheduler's own GPU accounting in `RequirementsFromResourceList` (`pkg/scheduler/api/resource_info/resource_requirment.go`), which recognizes `nvidia.com/gpu`, `amd.com/gpu` and MIG and counts MIG as slice portions. The reason it exists at all is that without it `Block` silently passes on a DRA or MIG queue, where the allocation is recorded under a DeviceClass name or an `nvidia.com/mig-*` key rather than `nvidia.com/gpu`.

**The call I'd like from you**

With the scheduler left as it is, that MIG parse function is only reachable from `pkg/scheduler/api/common_info/resources`. I could have the webhook import that package directly, but I'd rather not add an admission to scheduler dependency, so my plan is to keep a copy of the small parse function in `pkg/common/resources/mig.go`, next to the `IsMigResource` that already lives there. That does leave it duplicated for now, and I'd fold the two back together in a follow-up. If you'd rather the helper not sit in `pkg/common/resources` at all, I'll keep it inside `queuehooks` instead.

Worth saying plainly rather than burying: leaving the metric alone means `kai_queue_allocated_gpus` keeps the non-deterministic first-match behavior CodeRabbit flagged, and it will count GPUs differently from the webhook. I'll open a follow-up for that and for the duplicated parser.

**The review comments**

- `enforcementMode` becomes `quotaViolationMode` (the variable and the validator field). Say the word if you want the `EnforcementMode` type and its constants renamed to match.
- The gate becomes a plain `quotaViolationMode != EnforcementNone` instead of the helper method.
- I read the note on `:141` as the parent/child block in `ValidateUpdate`, so I'll pull that into its own function. Tell me if you meant something narrower.
- The per-resource struct collapses into a single check list, so one loop drives both the limit and the quota checks.
- The comment above `allocationReductionViolations` gets cut down, it got away from me.

**The two older threads**

Neither applies any more. The branch doesn't touch `docs/migrationguides/`, and no CHANGELOG line is about `--enable-quota-validation` (the metric line goes away with the revert, leaving just the new flag). I'll resolve both.

Rebasing onto main as well, the branch has fallen behind. I'll note each thread as the rework lands.

CHANGELOG.md
thc1006: Yep, dropped it. The only changelog line left is for the new `--enforce-quota-violation` flag.

docs/migrationguides/v0.14.0/README.md
thc1006: Gone, the branch doesn't touch the migration guides anymore.

cmd/queuecontroller/app/app.go
thc1006: Renamed to `quotaViolationMode`, both the var and the struct field.

pkg/common/resources/gpu.go
thc1006: Reverted all of it, the scheduler and metrics files match main again, so this is back to just the webhook, its cmd wiring and common. The count isn't new either: it follows the `*gpu` suffix the metric already used, plus the scheduler's MIG-as-slices logic in `RequirementsFromResourceList`. To leave the scheduler alone the MIG name parse is a small copy in `pkg/common/resources` now instead of importing the scheduler package. Left a note on the PR about a follow-up for the duplicated parser and the metric.

pkg/admission/webhook/queuehooks/queue_validator.go
thc1006: Pulled it out into `parentChildQuotaWarnings`.

pkg/admission/webhook/queuehooks/queue_validator.go
thc1006: Done, and the `enabled()` helper is gone.

pkg/admission/webhook/queuehooks/queue_validator.go
thc1006: Trimmed it, plus a couple of the other long ones near it.

pkg/admission/webhook/queuehooks/queue_validator.go
thc1006: Flattened it into one table driving a single loop, no more struct plus two ifs.

davidLif:
> Dear @davidLif ,
> 
> Short version: I'll revert the scheduler and metrics changes, do the five refactors, and rebase. The one thing I'd like your call on is where the GPU helper should live.
> 
> **Scoping**
> 
> Reverting `pkg/queuecontroller/metrics/metrics.go` to its original form and restoring the scheduler files (the MIG parser package and the four import sites). After that the PR only touches the admission webhook, `cmd/queuecontroller`, `pkg/common/resources` and the CHANGELOG.
> 
> **Where the calculation came from**
> 
> It isn't a new formula. It follows what's already in the tree: the vendor-agnostic `*gpu` suffix behavior of the existing `kai_queue_allocated_gpus` metric, and the scheduler's own GPU accounting in `RequirementsFromResourceList` (`pkg/scheduler/api/resource_info/resource_requirment.go`), which recognizes `nvidia.com/gpu`, `amd.com/gpu` and MIG and counts MIG as slice portions. The reason it exists at all is that without it `Block` silently passes on a DRA or MIG queue, where the allocation is recorded under a DeviceClass name or an `nvidia.com/mig-*` key rather than `nvidia.com/gpu`.
> 
> **The call I'd like from you**
> 
> With the scheduler left as it is, that MIG parse function is only reachable from `pkg/scheduler/api/common_info/resources`. I could have the webhook import that package directly, but I'd rather not add an admission to scheduler dependency, so my plan is to keep a copy of the small parse function in `pkg/common/resources/mig.go`, next to the `IsMigResource` that already lives there. That does leave it duplicated for now, and I'd fold the two back together in a follow-up. If you'd rather the helper not sit in `pkg/common/resources` at all, I'll keep it inside `queuehooks` instead.
> 
> Worth saying plainly rather than burying: leaving the metric alone means `kai_queue_allocated_gpus` keeps the non-deterministic first-match behavior CodeRabbit flagged, and it will count GPUs differently from the webhook. I'll open a follow-up for that and for the duplicated parser.
> 
> **The review comments**
> 
> * `enforcementMode` becomes `quotaViolationMode` (the variable and the validator field). Say the word if you want the `EnforcementMode` type and its constants renamed to match.
> * The gate becomes a plain `quotaViolationMode != EnforcementNone` instead of the helper method.
> * I read the note on `:141` as the parent/child block in `ValidateUpdate`, so I'll pull that into its own function. Tell me if you meant something narrower.
> * The per-resource struct collapses into a single check list, so one loop drives both the limit and the quota checks.
> * The comment above `allocationReductionViolations` gets cut down, it got away from me.
> 
> **The two older threads**
> 
> Neither applies any more. The branch doesn't touch `docs/migrationguides/`, and no CHANGELOG line is about `--enable-quota-validation` (the metric line goes away with the revert, leaving just the new flag). I'll resolve both.
> 
> Rebasing onto main as well, the branch has fallen behind. I'll note each thread as the rework lands.

Hello @thc1006 ,

Thank you very much for the detailed answer. I will discuss these points with my teammates to answer the points you raised here.

thc1006:
Dear @davidLif , just checking in, no rush. tldr: the only thread still open is where the MIG-parse helper should live; everything else is resolved.

The open call is whether that small parse stays as a copy in `pkg/common/resources/mig.go`, with a follow-up to dedupe it against the scheduler's version, or moves into `queuehooks`. Both leave the scheduler and metrics exactly as they are on main, so I'm easy either way. If the team hasn't had a chance, I can go ahead with the `pkg/common/resources` copy I described and open the dedupe follow-up, or wait for their call, whichever is less work for you. I've kept the branch un-rebased on purpose so I can fold the placement and the rebase into one push instead of force-pushing twice. Thanks again for the thorough review.

Issue #1751, opened 2026-06-23
kai-config Config CR missing after v0.14.6 GitOps/ArgoCD upgrade → operator crashloops "kai-config" not found and schedulers hit 401 Unauthorized on leader lease


david-gang:
## What happened?

After upgrading the Helm chart from v0.14.5 to v0.14.6 via ArgoCD (GitOps, automated sync), the cluster could no longer schedule any workloads. Two distinct, compounding failures occurred from the single upgrade, and ArgoCD reported the application **Synced / Healthy** the entire time.

### 1. Operator crashloops reconciling SchedulingShards — `kai-config` Config CR is missing

Both shards (`default`, `dra-shard`) failed to reconcile:

```
INFO  Failed to get the singleton KAI Config instance  {"controller":"schedulingshard",...}
ERROR Reconciler error {"controller":"schedulingshard",...,"error":"Config.kai.scheduler \"kai-config\" not found"}
```

- `kubectl get config.kai.scheduler -A` → `No resources found` (the singleton was gone).
- The CRD itself was present: `configs.kai.scheduler` (created 2026-04-23).

Observed chart mechanism (from the v0.14.6 chart):
- `templates/kai-config.yaml` renders empty in production (the Config is only emitted under the test-only flag `_test.renderKaiConfig=true`), so `kai-config` is not a Helm/ArgoCD-tracked resource.
- `kai-config` is instead created by a Helm `post-install,post-upgrade` hook Job (`kai-config-deployer`) that mounts ConfigMap `kai-config-manifest` and runs:
  ```
  kubectl apply --server-side --force-conflicts --field-manager=kai-config-deployer -f /manifest/kai-config.yaml
  ```
- All hook resources (Job, ConfigMap, SA, ClusterRole/Binding) carry `helm.sh/hook-delete-policy: hook-succeeded,before-hook-creation`.
- The deployer also runs `kubectl annotate config kai-config helm.sh/hook- helm.sh/hook-weight-` (consistent with `kai-config` having been a pre-install/pre-upgrade hook in earlier versions; the move to the post-hook Job appears in the v0.14.6 changes).

How this interacts with ArgoCD (observed):
- ArgoCD maps `helm.sh/hook: post-*` → PostSync hooks, which run only on a full sync operation — not during `selfHeal` of individually drifted resources.
[...]

PR #1815, opened 2026-07-02
feat(chart): support ArgoCD/GitOps deployment

### LINKED ISSUE #1794: Support kai installation using argoCD

* See parent issue
* Use the same chart for regular and argoCD installation (add flags if necessary).
* Nice to have: make sure all kai components (deployments, etc) are visible in the argo deployment

### Why is this needed?

support argoCD installation

gshaibi:
## Description

Support installing KAI Scheduler with ArgoCD/GitOps using the same Helm chart.

By default the `kai-config` Config CR is applied out-of-band by the `kai-config-deployer` hook Job, so GitOps engines never track it — no drift detection, no selfHeal, invisible in the app tree. In incident #1751 an upgrade deleted the CR and ArgoCD kept reporting Synced/Healthy while the operator crashlooped.

Changes:
- **`kaiConfig.render`** (default `false`): renders the Config CR inline as a regular, tracked release resource. Mutually exclusive with `kaiConfigDeployer.enabled` — rendering fails fast when both are set. Since the operator sets ownerReferences from the CR to everything it creates, the whole KAI component tree becomes visible in the ArgoCD UI.

  | `kaiConfigDeployer.enabled` | `kaiConfig.render` | Result |
  |---|---|---|
  | `true` (default) | `false` (default) | hook Job applies the CR out-of-band (unchanged) |
  | `false` | `true` | inline tracked release resource — **GitOps/ArgoCD mode** |
  | `false` | `false` | chart never creates kai-config (unchanged) |
  | `true` | `true` | template fails with a clear message |

- **`openshift`** (default `false`): forces OpenShift mode. The existing `lookup`-based autodetection returns nil under offline rendering (ArgoCD repo-server), which meant no SCC and wrong hook securityContexts on OpenShift+ArgoCD. All 11 lookup sites now go through a shared `kai-scheduler.openshift` helper (`or .Values.openshift (lookup ...)`); plain-Helm autodetection is unchanged.
- ArgoCD `PostDelete` hook annotations on the post-delete cleanup Job and its SA/RBAC (requires ArgoCD >= 2.10). In render mode the CR is a tracked release resource pruned by Helm/ArgoCD itself, so the cleanup script's CR deletion stays gated on `kaiConfigDeployer.enabled` only.
- `argocd.argoproj.io/sync-options: Prune=false` next to existing `helm.sh/resource-policy: keep` annotations (default Queues, default SchedulingShard, resource-reservation namespace).
- Inline CR carries `SkipDryRunOnMissingResource=true,ServerSideApply=true` sync-options (fresh-install CRD establishment; SSA adoption of a previously hook-managed CR).
- New [GitOps installation guide](docs/gitops/README.md): install-mode matrix, Application example, OpenShift notes.
- Removed the internal `_test.renderKaiConfig` test knob in favor of `kaiConfig.render`; new `kai_config_render` and `openshift` chart test suites.

## Related Issues

Fixes #1794
Related: #1751

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. All new values default to existing behavior; the removed `_test.renderKaiConfig` was an internal test-only knob.

## Additional Notes

- **New e2e suite** `test/e2e/suites/gitops` (dedicated `e2e-gitops-tests` CI job): installs KAI through a live ArgoCD Application and verifies the full lifecycle — first sync on a fresh cluster, healthy Config CR + operands, workload scheduling, **selfHeal recreating a deleted kai-config CR (the #1751 regression)**, and uninstall via the PostDelete hook. Run locally with `hack/run-e2e-gitops-kind.sh --local-images-build`. All 5 specs pass locally against kind + ArgoCD v3.4.4 (83s).
- The e2e suite caught a real bug before merge: the inline Config CR rendered without `apiVersion` (template comment chomped the newline after the license header), which ArgoCD rejects with `ComparisonError`. Fixed + guarded by a chart-test assert.
- Verified: `make test-chart` (19 suites, 142 tests) and a local `helm template` matrix across all four install modes, plus OpenShift on/off and the embedding regression (`kaiConfigDeployer.enabled=false` only → unchanged output).
- Follow-ups to file separately: custom ArgoCD health check (Lua) for the Config CR. (The scheduler 401-retry half of #1751 was fixed separately in #1823.)


.github/workflows/on-pr.yaml
```diff
@@ -357,6 +348,51 @@ jobs:
         run: |
           ginkgo -r --keep-going --trace -vv --label-filter 'upgrade' --output-dir=. --json-report=e2e-upgrade-report.json ./test/e2e/suites/upgrade
 
+  e2e-gitops-tests:
```
SiorMeir, line 351: Do we need this on all pr runs? might make the CI longer

SiorMeir (dismissed):

.github/workflows/on-pr.yaml
gshaibi: It does not extend CI wall time - it runs in parallel and finishes ~19 min before the main e2e job (8 min vs 27). It is also currently the only check covering the selfHeal/ownerReferences recovery path. If it turns out noisy we can path-filter it in a follow-up.

SiorMeir (approved):

Issue #1653, opened 2026-05-31
perf(scheduler): excessive allocation churn in reclaim scenario search inflates GC pressure


david-gang:
## Punchline

The reclaim path allocates **~35 GB to produce ~287 MB of live state (a ~120× transient-allocation overhead)**, and the rate grows as `nodes × scenarios` — so GC starts stealing the very CPU cores doing scheduling, exactly when the cluster is large and preemption is busiest.

The core observation: several of the expensive allocations in the inner loop depend only on **per-session-invariant** data (queue resource shares, the podAffinity node list), yet are rebuilt per node per scenario. This is a recomputation-in-an-N×M-loop problem, not a styling nit.

## Summary

Profiling a snapshot replay shows **~35 GB allocated over a ~40 s run** while only **~287 MB is in-use** — almost entirely short-lived churn. GC accounts for **~13% of CPU samples** (`runtime.gcDrain` cum). The allocations concentrate in the reclaim/preemption scenario search, which re-runs the full fit/score pipeline over every node for every probed scenario.

## Reproduction

Replay a captured cluster snapshot with the built-in heap profiler:

```bash
go build -o ./bin/snapshot-tool ./cmd/snapshot-tool/
./bin/snapshot-tool --filename <snapshot>.gzip --verbosity 2 --memprofile mem.prof
go tool pprof -sample_index=alloc_space -top -focus="KAI-scheduler" mem.prof
```

## Allocation hot path

```
JobSolver.Solve → probeAtK → searchMaxSolvableK → tryProbeAndDiscard
  → byPodSolver.solve → runSimulation → tryScenarioWithEvictedVictims
  → TryToVirtuallyAllocatePreemptorAndGetVictims → AllocateJob
  → allocateSubGroupSet → FittingNode → OrderedNodesByTask → {NodeOrderFn, NodePreOrderFn plugins}
```

## Top allocators (`alloc_space`, flat)

| Bytes | Function | Location |
|---|---|---|
| 2.40 GB | `podAffinity.nodePreOrderFn.func1` — builds a `[]ksf.NodeInfo` of empty `&NodeInfo{}` (one per node, every call); `PrePodAffinity` adds 1.45 GB cum | `pkg/scheduler/plugins/podaffinity/podaffinity.go:56,59` |
| 2.32 GB | `resource_share.NewResourceQuantities` | `pkg/scheduler/plugins/proportion/resource_share/resource_quantities.go:22` |
[...]

PR #1822, opened 2026-07-04
perf(scheduler): reduce proportion reclaim allocation churn

### LINKED ISSUE #1824: perf(scheduler): avoid rebuilding proportion resource maps during reclaim search
## Problem

The proportion plugin repeatedly materializes temporary `ResourceQuantities` maps while ordering queues and evaluating reclaim victims. These operations run inside the reclaim scenario-search loop even though the underlying `QueueResourceShare` and job `ResourceVector` values already exist.

In the profile from #1653:

- `NewResourceQuantities`: 2.32 GB allocated
- `QueueResourceShare.buildResourceQuantities`: 1.89 GB allocated

The 500-node reclaim benchmark additionally showed these conversions dominating allocation throughput.

## Scope

- Compare persistent queue-share state and cached resource vectors directly.
- Avoid temporary `ResourceQuantities` maps in queue-order and reclaim predicates.
- Avoid allocating `ReclaimerInfo` for every victim check.
- Preserve existing CPU, memory, and aggregate-GPU proportion semantics.

## Acceptance criteria

- Existing proportion and reclaim tests continue to pass.
- The 500-node benchmark demonstrates reduced runtime, allocated bytes, and allocation count.
- A heap profile confirms that the affected constructors no longer dominate allocation throughput.

Parent issue: #1653

enoodle:
## Description

Large reclaim searches repeatedly materialized `ResourceQuantities` maps for queue ordering and victim filtering even though the underlying queue state and job resource vectors were already available. At 500 nodes, those short-lived maps dominated allocation throughput and GC pressure.

This change:

- compares persistent `QueueResourceShare` state and cached `ResourceVector` values directly in the hot queue-order and reclaim predicates;
- returns `ReclaimerInfo` by value instead of allocating one object per victim check;
- keeps mutable scenario quantities owned by their existing callers rather than introducing shared scratch state;
- adds lazy `V(N).Do(...)` logging for expensive verbose messages and removes the redundant `IsVerbose` API;
- preserves the current CPU, memory, and aggregate-GPU proportion semantics while keeping vector conversion behind one seam that can later evolve with AccountedResource.

## Related Issues

Fixes #1824.
Part of #1653.

## Checklist

- [x] Self-reviewed

## Breaking Changes

None.

## Additional Notes

The benchmark reproduces the scale test that fills 500 eight-GPU nodes with 4,000 single-GPU jobs, submits a full replacement set from another queue, and reclaims all running jobs. Results are one operation per run (`-benchtime=1x`):

| Metric | Before | After | Change |
|---|---:|---:|---:|
| Runtime | 164.72 s/op | 103.60 s/op | -37.11% |
| Allocated bytes | 74.44 GB/op | 11.40 GB/op | -84.68% |
| Allocations | 792.34 M/op | 312.12 M/op | -60.61% |

Profile details:

- `buildResourceQuantities`, `NewResourceQuantities`, and `getInvolvedResourcesNames`: approximately 57.4 GB sampled allocation before, 50.5 MB after.
- `buildReclaimerInfo`: approximately 3.0 GB / 32.2 million objects before, absent as an allocator after.
- Post-GC live heap stayed roughly flat at 42–44 MiB, confirming this is transient allocation churn rather than retained state.
- Disabled verbose strategy and node-set logs allocate zero objects in focused tests.
- Disabled node-availability logging allocates four objects per score before this change and zero after.

Validation:

```text
go test ./pkg/scheduler/log ./pkg/scheduler/framework ./pkg/scheduler/plugins/proportion/... -count=1
go test ./pkg/scheduler/actions/integration_tests/reclaim -run '^TestManySingleGPUJobsReclaimTopology$' -count=1
go vet ./...
go test ./pkg/scheduler/actions/integration_tests/reclaim -run '^$' -bench '^BenchmarkReclaimManySingleGPUJobsWithMinRuntime_500Node$' -benchtime=1x -count=1 -benchmem -memprofile=pr-head.mem.pprof
```


pkg/scheduler/log/log.go
```diff
@@ -34,6 +33,19 @@ type SchedulerLogger interface {
...
+	enabled bool
+}
+
+// Do invokes logFn only when the verbosity level is enabled.
+func (logger VerboseLogger) Do(logFn func(*zap.SugaredLogger)) {
+	if logger.enabled {
+		logFn(logger.SugaredLogger)
+	}
+}
+
```
SiorMeir: I feel this makes the log statements a bit harder to read. 

What do you think about doing: 
```
func (logger VerboseLogger) Do(logFn func()) {
    if logger.enabled {
        logFn()
    }
}
```
 instead?

then on call we do:

```vl := log.InfraLogger.V(7)
vl.Do(func() {
    vl.Infof("Node sets: %v", names)
})```

enoodle: :+1: Thank you good idea - your suggestion is a simpler interface

SiorMeir (approved):

PR #1823, opened 2026-07-04
fix(scheduler): exit on 401 Unauthorized API responses

### LINKED ISSUE #1817: Scheduler retries 401 Unauthorized forever on stale ServiceAccount token instead of exiting
## Context

Second half of #1751 (follow-up from #1794 / PR #1815).

During the #1751 incident, after the `kai-config` Config CR was deleted, the operator recreated scheduler pods whose projected ServiceAccount tokens became invalid. The scheduler kept retrying API requests that failed with `401 Unauthorized` indefinitely, staying `Running`/`Ready` while doing no work — so nothing (probes, restart policy, alerts) surfaced the failure.

## Expected behavior

On persistent `401 Unauthorized` from the API server (an unrecoverable auth state for the current token), the scheduler should fail fast — exit (or fail its health probe) so the kubelet restarts it and it picks up a fresh token, and the failure becomes visible to operators.

## Scope

- Scheduler API client / informer error handling for persistent 401s.
- Consider the same audit for the other long-running components (binder, controllers).

reyshazni:
## Description

When the scheduler's projected ServiceAccount token becomes invalid (e.g. after a Config CR deletion triggers pod recreation), all API requests fail with 401 Unauthorized. The default client-go behavior retries indefinitely, so the scheduler stays Running while doing no useful work.

This PR wraps the scheduler's `rest.Config` transport to exit the process on any 401 response. Since a 401 on a projected ServiceAccount token is unrecoverable without a pod restart, exiting lets the kubelet restart the pod with a fresh token.

The transport wrapper covers all API calls (watches, lists, gets, creates) uniformly through a single `config.Wrap` call.

## Related Issues

Fixes #1817

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. No behavioral change in the happy path.

## Additional Notes

Reworked from per-informer WatchErrorHandler to transport-level interception per @gshaibi's suggestion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Podgrouper now preserves an existing PodGroup's topology constraint when the workload does not specify one, so an externally-assigned topology is not overwritten. Workload topology annotations still take precedence when present.
 
 ### Fixed
+- Scheduler now exits on 401 Unauthorized API responses instead of retrying indefinitely with a stale ServiceAccount token. [#1817](https://github.com/kai-scheduler/KAI-Scheduler/issues/1817)
 - Reduced scheduler memory use during large reclaim operations by removing redundant per-job-pair min-runtime protection caching; effective min-runtime durations remain cached per queue pair. [#1808](https://github.com/kai-scheduler/KAI-Scheduler/issues/1808)
 - Fixed reclaim abandoning valid over-quota victims when an unrelated under-deserved queue appeared earlier in victim ordering. [#1750](https://github.com/kai-scheduler/KAI-Scheduler/issues/1750)
 - Restricted Helm post-delete cleanup to KAI operator-managed Deployments and preserved externally managed `kai-config` resources when `kaiConfigDeployer.enabled=false`.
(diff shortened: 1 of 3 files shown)


gshaibi (commented):
Thanks @reyshazni  for the fix!

Suggestion: instead of setting a handler on a few chosen informers, we can catch the 401 once at the HTTP client level.

If we wrap the shared rest.Config with a transport that exits on any 401 response, every client and informer is covered, the exit happens on the first failed request, and the binder follow-up becomes the same one-liner on its config.

reyshazni:
Thanks for the review! @gshaibi 

I initially scoped it to watch reconnects since a transport-level exit on any 401 felt more aggressive, but agreed the transport wrapper is more comprehensive and simpler to extend.

I can rework this to wrap rest.Config with a transport that exits on 401. Should I force-push to this branch or open a fresh PR?

gshaibi:
> Thanks for the review! @gshaibi
> 
> I initially scoped it to watch reconnects since a transport-level exit on any 401 felt more aggressive, but agreed the transport wrapper is more comprehensive and simpler to extend.
> 
> I can rework this to wrap rest.Config with a transport that exits on 401. Should I force-push to this branch or open a fresh PR?

@reyshazni whatever is more convenient for you, I don't mind.

reyshazni:
Reworked to transport-level wrapping via config.Wrap in RunApp(). Single wrapper covers all clients and informers. Ready for another look @gshaibi

gshaibi (approved):
Thanks, this looks good!

reyshazni:
thanks! @gshaibi 

looking forward for the merge!

SiorMeir (approved):

PR #1838, opened 2026-07-06
perf(scheduler): deduplicate equivalent scenario candidates before simulation

### LINKED ISSUE #1719: Deduplicate equivalent reclaim scenario candidates before simulation

gshaibi:
## Description

Adds a scenario deduplication cache to the bounded scenario search, implementing the follow-up deferred by the generator portfolio design ("Scenario Deduplication Cache"). Generators can rediscover the same effective victim set — within one generator across accumulation steps, and across generators (NodeLocalGreedy, MultiNodeGang) — and each duplicate previously paid for a full simulation.

**How it works:**
- The solver computes a canonical sha256 fingerprint for every scenario the portfolio emits (pending task UIDs, recorded-victim UIDs, potential-victim UIDs, preemptor UID — each section sorted, so equivalent victim sets emitted in different orders or batch groupings collapse to one key). Task UIDs stand in for node placements, which are fixed within a session; the code comment records that a cache outliving a session must add node assignments to the key.
- Candidates whose fingerprint already **failed** simulation in the same job solve are skipped before simulation and counted as `state="duplicate"` in `scenario_search_scenarios_total` (plus a `duplicate` result in the duration histogram).
- Only failed simulations (unsolved / validator-rejected) are recorded, via `ObserveCurrentAttempt`; solved scenarios stay re-emittable so the final probe can rebuild the winning statement.
- The cache spans one `JobSolver.SolveWithResult` call — all probes and all generators of one pending job — which is what enables cross-generator dedup. Applies to reclaim, preempt, and consolidation.

**Supporting fixes** (surfaced by review of this change):
- `byPodSolver.solve` now rolls back its feasible-node additions after validator-rejected and errored simulations too (previously only after cleanly unsolved ones), so the probe's feasible-node set stays derived from the recorded victims covered by the fingerprint — the determinism premise the cache relies on. **Also submitted standalone as #1841 for focused review** (identical patch plus a dedicated regression test; merges cleanly in either order).
- The multi-node-gang generator returns an untyped nil on exhaustion instead of a typed-nil `*ByNodeScenario` through the interface; the portfolio keeps a defensive typed-nil guard.

**Results** — `BenchmarkReclaimUnschedulableDistributedJob_100Node`: 1,020 of 1,398 candidates skipped before simulation (matching the duplicate count measured in the issue), simulations reduced to 378/op, wall clock ~22.8s → ~7.9s per reclaim cycle (~2.9x).

## Related Issues

Fixes #1719

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. Metrics note: `state="duplicate"` is a new value in `scenario_search_scenarios_total`; `emitted` now counts only candidates handed to the solver (duplicates and exhaustion sentinels excluded), so `emitted == simulated`. Raw generator productivity is recoverable as `emitted + duplicate`.

## Additional Notes

- Test coverage: fingerprint canonicality/discrimination unit tests; portfolio tests for skip, failures-only recording (solved scenarios stay re-emittable — the critical invariant), cross-generator and cross-portfolio dedup, nil-cache off-switch; a job-solver end-to-end test with repeated failing scenarios; a real reclaim-action test asserting duplicates are skipped on a 50-node topology.
- Benchmarks now report `simulated/op` and `duplicate/op` via prometheus counter deltas, so a future dedup regression is visible in benchmark output even before wall clock moves.
- Design doc section "Scenario Deduplication Cache" rewritten from "to be specified" to the implemented design; `METRICS.md` and `CHANGELOG.md` updated.
- Cache cost: one 32-byte map entry per failed simulation; fingerprint hashing streams into sha256 (no payload copies) and is microseconds against millisecond-scale simulations.


CHANGELOG.md
SiorMeir, line 1: rebase CHANGELOG?

pkg/scheduler/actions/common/solvers/scenario_fingerprint_test.go
```diff
@@ -0,0 +1,197 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/actions/common/solvers/scenario"
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/api"
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/api/pod_info"
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/api/podgroup_info"
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/framework"
+)
+
+func TestFingerprintScenarioIsOrderIndependent(t *testing.T) {
```
SiorMeir, line 21: SUGGESTION: add a test that does a "E2E" on this 
inserting a scenario and comparing it to a hard coded string with all of the suffixes and prefixes you expect

It'll help visualize what's happening

CHANGELOG.md
gshaibi: Rebased the branch onto latest main.

pkg/scheduler/actions/common/solvers/scenario_fingerprint_test.go
gshaibi: Added `TestFingerprintScenarioEncodesCanonicalSections`: builds a scenario with known task UIDs and compares its fingerprint against the hard-coded payload `"10" \x1f "20" \x00 "21" \x1f "30" \x1f "40" \x00 "41"` (sha256'd), with each section labeled.

SiorMeir (dismissed):

SiorMeir (approved):

PR #1852, opened 2026-07-08
fix(scheduler): Track extended resources present on only a subset of nodes in MaxNodeResourcesPredicate

### LINKED ISSUE #1851: Extended resources present on only a subset of nodes are reported as unavailable cluster-wide
### Issue
Pods requesting an extended resource (advertised by a device plugin on a subset of nodes) were persistently rejected by the scheduler as `UnschedulableAndUnresolvable` with:

`The pod <ns>/<name> requires ..., <resource>: 1000. No node in the default node-pool has <resource> resources`

even though several nodes in the pool advertised ample capacity for that resource in `.status.allocatable`.

### Root cause
Resource vector indices are assigned in node-scan order, so nodes scanned before an extended resource is first seen get an `AllocatableVector` with no slot for it. `NewMaxNodeResourcesPredicate` seeds its max-resources accumulator from the first node returned by Go map iteration, and `ResourceVector.SetMax` only compares indices up to `min(len(v), len(other))` without growing the receiver. When the randomly-chosen seed node predates the resource's discovery, the accumulator permanently reads 0 for that resource. Any resource not present on every node is affected; in our cluster (resource on 24 of 75 nodes), roughly 2/3 of scheduling sessions rejected such pods, so they were effectively unschedulable.

The predicate should report the true per-resource maximum across all nodes in the pool, independent of map iteration order, and admit pods whose requests fit on at least one node.

### Environment

- Kubernetes version: v1.34.2
- KAI Scheduler version: v0.16.1 (bug also present on main)
- Cloud provider or hardware configuration: OKE / OCI
- Tools that you are using KAI together with: a device plugin exposing an extended resource (any such plugin reproduces it)

ayyub-abdulrezak-ai:
## Description

`MaxNodeResourcesPredicate` builds its per-resource maximum by seeding an accumulator from the first node returned by Go map iteration, then folding in the rest with `ResourceVector.SetMax`. Vector indices are assigned in node-scan order, so nodes scanned before an extended resource is first seen have no slot for it. `SetMax` only compared indices up to `min(len(v), len(other))` and never grew the receiver — so whenever the randomly-chosen seed node lacked a slot for an extended resource, the accumulator read 0 for it regardless of what other nodes advertise, and every pod requesting it was rejected with `No node in the <pool> node-pool has <resource> resources`. This affects any resource not present on every node; e.g. with the resource on 24 of 75 nodes, roughly 2/3 of scheduling sessions reject such pods.

Fix: `SetMax` grows the receiver to the longer vector's length before taking the element-wise max, matching the existing `Add`/`Sub` pattern (receiver changed from value to pointer). This also covers the topology plugin's `SetMax` call site, which had the same latent issue.

Tests:
- `resource_vector_test.go`: `SetMax` with equal lengths, shorter receiver, longer receiver, smaller argument values, and empty vectors.
- `maxNodeResources_test.go`: with an extended resource on one of five nodes — a fitting pod is accepted, an oversized request is rejected, and the accumulator tracks the true max (repeated 20× since the seed is random); a resource on no node is still rejected with the "No node" message.

Verified the new tests fail without the fix and the full suite passes with it.

## Related Issues

Fixes [issue #1851](https://github.com/kai-scheduler/KAI-Scheduler/issues/1851)

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. Both call sites of `SetMax` use addressable values; behavior changes only in the previously-broken mismatched-length case.

## Additional Notes

CHANGELOG.md updated under `[Unreleased] > Fixed`.


enoodle (approved):

enoodle:
@ayyub-abdulrezak-ai Thanks for the fix!

SiorMeir:
@ayyub-abdulrezak-ai Hi!

DCO checks fail - please follow [CONTRIBUTING.md](https://github.com/kai-scheduler/KAI-Scheduler/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-dco) guide to fix and we'll approve this

gshaibi (approved):

ayyub-abdulrezak-ai:
> @ayyub-abdulrezak-ai Hi!
> 
> DCO checks fail - please follow [CONTRIBUTING.md](https://github.com/kai-scheduler/KAI-Scheduler/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-dco) guide to fix and we'll approve this

Signed off on both commits, thanks for the tip @SiorMeir! And thanks for the speedy approvals @enoodle @gshaibi!

enoodle (approved):

Issue #1846, opened 2026-07-07
Upgrade e2e tests fail for ~40 minutes after a release is cut


gshaibi:
The `resolve_version` step in on-pr.yaml picks the upgrade-from version with `gh release list`, but the release workflow takes ~37 minutes to publish the helm chart to ghcr. Any PR whose upgrade test runs in that window resolves the new tag and fails on:

```
Error: failed to perform "FetchReference" on source: ghcr.io/kai-scheduler/kai-scheduler/kai-scheduler:v0.16.3: not found
```

Happened on #1845, which ran 8 minutes after the v0.16.3 release was created.

Suggested fix: after resolving the tag, check the chart actually exists (`helm show chart oci://... --version`) and fall back to the previous release if not.

PR #1854, opened 2026-07-08
fix(ci): resolve upgrade-from version by checking OCI chart availability

maishivamhoo123:
## Fix upgrade-from version race when chart isn’t published yet

**Problem:** #1846   
A new GitHub release may be created before its Helm chart is pushed to GHCR (~40 min delay). The upgrade job would pick that tag and fail with `not found`.

**Fix:**  
Walk release tags newest‑first, check each chart’s existence with `helm show chart`, and pick the first one that’s actually published.
before fix :- 
Screenshot 2026-07-08 122003
After fix :-   
Screenshot 2026-07-08 122339

### DIFF
--- a/.github/workflows/on-pr.yaml
+++ b/.github/workflows/on-pr.yaml
@@ -311,7 +311,7 @@ jobs:
         run: |
           TARGET_BRANCH="${{ github.base_ref }}"
           if [[ "$TARGET_BRANCH" =~ v([0-9]+)\.([0-9]+) ]]; then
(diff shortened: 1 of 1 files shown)


maishivamhoo123:
@gshaibi @enoodle @itsomri and @team can you please review this pr ?

gshaibi (approved):

gshaibi:
Looks great @maishivamhoo123 , thank you for contributing.

SiorMeir (approved):

PR #1864, opened 2026-07-08
perf(scheduler): reduce task fit-error memory usage

### LINKED ISSUE #1827: perf(scheduler): bound per-task per-node fit-error retention during allocation
## Problem

When the allocate action tries a pending task against candidate nodes, `Session.FittingNode` records node-specific failures when `writeFittingDelta` is enabled.

For every rejected task/node pair, the scheduler can:

- construct a `TasksFitError` containing short and detailed reason slices;
- eagerly format requested, used, and capacity values;
- insert the error into a node-keyed `TasksFitErrors` map;
- retain that map through `PodGroupInfo.TasksFitErrors`.

This scales as `pending tasks × candidate nodes`. In the 500-node scale test, 4,000 pending single-pod jobs can produce up to two million retained node-specific failures.

## Evidence

The end-to-end 500-node scale-test profile attributed approximately 1.47 GB of retained heap directly to fit-error state:

- `NewFitErrorWithDetailedMessage`: 531.7 MB
- `fmt.Sprintf` below insufficient-resource error construction: 446.7 MB
- `NewFitErrorInsufficientResource`, excluding callees: 166.7 MB
- map backing arrays below `TasksFitErrors.AddNodeErrors`: 329.3 MB

The complete `NewFitErrorInsufficientResource` subtree retained approximately 1.15 GB.

This state was created by the allocate action before reclaim ran. It therefore does not appear in the reclaim-only benchmark used for the other #1653 allocation findings.

## Scope

- Define which fit-error information must be retained for:
  - normal unschedulable pod messages;
  - aggregated reason counts;
  - detailed fit-error reporting;
  - scheduler debug logging.
- Avoid retaining a complete independently formatted error for every task/node pair.
- Evaluate approaches such as:
  - aggregating identical reasons with counts;
  - retaining a bounded sample of representative nodes;
  - storing structured shared reason data;
  - formatting detailed messages only when consumed.
- Ensure old fit errors are released or replaced when a job is evaluated again.
- Add a focused allocate or complete-cycle benchmark that reproduces many pending tasks failing across many nodes.

## Acceptance criteria

- Retained detailed error state no longer grows as one complete error object per task/node pair.
- Default unschedulable messages preserve their reason counts and useful diagnostic information.
- Detailed-fit-error behavior is explicitly defined and covered by tests.
- Scheduling, node-fit decisions, and task status transitions remain unchanged.
- A heap profile of the scale reproduction shows a material reduction in:
  - `NewFitErrorInsufficientResource`;
  - `NewFitErrorWithDetailedMessage`;
  - insufficient-resource string formatting;
  - `TasksFitErrors` node-map storage.
- The benchmark reports retained heap and allocation volume separately.

## Non-goals

- Changing node-fit or resource-availability semantics.
- Addressing scenario `PodInfo`/`ResourceVector` cloning.
- Addressing proportion resource maps or pod-affinity allocations.

## Related issues

Parent issue: #1653.

This finding comes from the end-to-end scale-test profile rather than the reclaim-only snapshot described directly in #1653.

enoodle:
## Description

This PR bounds scheduler memory retained for node-fit errors when many pending tasks fail across many candidate nodes.

Previously, each pending task could retain a complete, independently formatted error for every rejected node. Retained memory therefore scaled as `pending tasks × candidate nodes`.

This PR:

- Runs the scale-test-inspired `ReclaimManySingleGPUJobs` benchmark through the complete scheduling cycle while measuring retained memory after Allocate and after the full cycle.
- Stores aggregated reason counts per task instead of retaining per-node error objects.
- Preserves compact unschedulable messages and their reason counts.
- Recomputes detailed per-node errors only when detailed reporting or V(6) logging requests them.
- Makes diagnostics local to a complete allocation attempt.
- Publishes only diagnostics from the failed alternative that allocated the most tasks; the first alternative wins ties.
- Prevents reclaim, preempt, and consolidation simulations from mutating user-facing fit errors.
- Makes topology filtering return structured diagnostics instead of mutating `PodGroupInfo`.
- Avoids constructing topology diagnostics when fit-error collection is disabled.
- Safely skips typed-nil fit errors during detailed recomputation.
- Replaces stale allocation-owned diagnostics when a job is evaluated again.

Scheduling decisions, node-fit behavior, task transitions, and `isPipelineOnly` semantics remain unchanged.

### 500-node full-cycle reclaim benchmark results

The benchmark models the scale-test scenario: running single-GPU jobs fill the cluster, another queue submits a full set of pending single-GPU jobs, and reclaim evaluates the running jobs. Each iteration executes Allocate, Consolidate, Reclaim, Preempt, and StaleGangEviction.

| Metric | `origin/main` | This PR | Change |
|---|---:|---:|---:|
| Live-heap delta after Allocate | 598 MB | 73 MB | -87.7% |
| Live-heap delta after full cycle | 612 MB | 86 MB | -86.0% |
| Bytes/op | 14.13 GB | 13.37 GB | -5.4% |
| Allocations/op | 397.5 M | 391.4 M | -1.5% |
| Tasks with fit errors after Allocate | 4,000 | 4,000 | unchanged |

The unchanged fit-error task count verifies that the retained-memory reduction does not come from skipping failed-task diagnostics.

## Related Issues

Fixes #1827

Related to #1653

## Checklist

- [x] Self-reviewed
- [x] Added/updated tests
- [x] Updated documentation

## Breaking Changes

No user-facing breaking changes.

The internal `SubsetNodesFn` interface now returns structured node-set diagnostics instead of allowing topology implementations to write fit errors directly into `PodGroupInfo`.

## Additional Notes

Validation performed:

- `go test ./pkg/scheduler/... -timeout=30m -count=1`
- `make validate`
- Direct before/after runs of `BenchmarkReclaimManySingleGPUJobs_500Node` with `-benchmem -benchtime=1x`

The focused fit-error microbenchmarks used during investigation were removed after the full-cycle reclaim benchmark reproduced the retained-memory improvement.

Detailed fit errors are transient. They are recomputed against the current scheduling session when requested and are not retained per task/node pair.


enoodle:
split into 3 PRs for review

PR #1868, opened 2026-07-08
feat(build): publish FIPS-enabled images and add global.fips helm option

### LINKED ISSUE #1867: Publish Fips Enabled docker images

* publish fips enabled images major.minor.fix-fips
* add an option in Helm to install/update with fips enabled images

### Why is this needed?

some orginazations require FIPS enabled images, that way they can enjoy KAI too :)

gshaibi:
## Description

Publishes a FIPS-enabled variant of every image on release tags, tagged `<version>-fips`, and adds a `global.fips` Helm value to install with them.

- **Build**: `make build FIPS=1` compiles all services with the Go toolchain's native FIPS 140-3 mode (`GOFIPS140=v1.0.0`, embeds the CMVP-validated Go Cryptographic Module, FIPS mode on by default at runtime) and suffixes image tags with `-fips`. No toolchain or base image change. `crd-upgrader` (no Go code) is republished under the `-fips` tag as well, since all chart hooks use it and the suffix is chart-wide.
- **CI**: a new `build-and-push-fips` job in `push-artifacts.yaml` runs on release tags only, in parallel with the regular job, and asserts every built binary records `GOFIPS140` in its build info before the workflow can succeed.
- **Helm**: `global.fips` (default `false`) appends `-fips` to every resolved image tag via a new `kai-scheduler.imageTag` helper, covering all 16 tag references (components in the kai-config CR, operator Deployment, `MS_TAG` env fallback, and all hook jobs). The suffix applies to whatever tag resolves (per-service tag, `global.tag`, or chart version), so FIPS is orthogonal to version pinning. No operator code changes needed.

## Related Issues

Fixes #1867

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. Default rendering is byte-identical (`global.fips` defaults to `false`; all 145 chart tests pass unchanged).

## Additional Notes

Verified locally:
- `make build-go-amd SERVICE_NAME=scheduler FIPS=1` → `go version -m` shows `build GOFIPS140=v1.0.0-c2097c7c`; absent without `FIPS=1`.
- `make docker-build-crd-upgrader VERSION=1.2.3 FIPS=1 -n` → single `-fips` suffix through make recursion.
- `helm template --set global.fips=true` → every image reference ends in `-fips`; default render contains none.
- New `tests/fips_test.yaml` helm-unittest suite; full chart suite passes (145/145).

### DIFF
--- a/.github/workflows/push-artifacts.yaml
+++ b/.github/workflows/push-artifacts.yaml
@@ -64,3 +64,41 @@ jobs:
       - name: Push Helm Chart
         run: |
           helm push ./charts/kai-scheduler-$PACKAGE_VERSION.tgz oci://${{ env.DOCKER_REGISTRY }}
+
+  build-and-push-fips:
+    name: Build & Push FIPS
+    if: startsWith(github.ref, 'refs/tags/')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+
+      - name: Extract package version from tag
+        run: |
+          PACKAGE_VERSION=${GITHUB_REF_NAME}
+          echo "PACKAGE_VERSION=$PACKAGE_VERSION" >> $GITHUB_ENV
+          echo $PACKAGE_VERSION
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: '1.26.3'
+
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v4
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+
+      - name: Docker build & push FIPS images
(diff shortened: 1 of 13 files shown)


gshaibi:
Smoke-tested on kind: installed the chart with global.fips=true using locally built -fips images. All components came up clean (no restarts, no panics in logs), and a test pod was grouped, scheduled and bound end to end.

.github/workflows/push-artifacts.yaml
```diff
@@ -64,3 +64,41 @@ jobs:
       - name: Push Helm Chart
         run: |
           helm push ./charts/kai-scheduler-$PACKAGE_VERSION.tgz oci://${{ env.DOCKER_REGISTRY }}
+
+  build-and-push-fips:
+    name: Build & Push FIPS
+    if: startsWith(github.ref, 'refs/tags/')
```
SiorMeir, line 70: shouldn't this be if fips_flag == true ?

SiorMeir (approved):

.github/workflows/push-artifacts.yaml
gshaibi: This is the build job in the CI, it should always run for any release.
The FIPS flag is relevant during deployment only.

PR #1870, opened 2026-07-08
podgrouper: register nvidia.com/v1beta1 DynamoGraphDeployment for gang scheduling

HungLee-nvidia:
## Summary
The podgrouper hub only registers `nvidia.com/v1alpha1/DynamoGraphDeployment` (`pkg/podgrouper/podgrouper/hub/hub.go`). Dynamo 1.2.0 promoted the DGD API and now serves/owns the resource as **v1beta1**, so on clusters running Dynamo >= 1.2.0 the DGD-owned pods' top-owner GVK is `nvidia.com/v1beta1/DynamoGraphDeployment`, which does not match the registered `v1alpha1` key. The podgrouper then falls back to per-pod PodGroups (`minMember=1`), so DGD components are **not gang-scheduled**.

This registers the `v1beta1` GVK alongside `v1alpha1`, routed to the same `skipTopOwnerGrouper`. In non-Grove mode this yields one PodGroup per component Deployment (`minMember = replicas`); in Grove mode it propagates metadata to PodGang/PodClique as before. Backward compatible (v1alpha1 retained).

## Evidence (KAI v0.15.3 + Dynamo 1.2.0, real cluster)
- A v1beta1 DGD with a 2-replica worker component produced two separate PodGroups (`minMember=1`, owner `Pod/v1`) -> no gang.
- `hub.go` on `main` registers only `nvidia.com/v1alpha1 DynamoGraphDeployment`.
- Setting `pod-group-name` on the pod template is overwritten by the podgrouper, so it is not a viable workaround.

## Change
```go
table[metav1.GroupVersionKind{
    Group:   "nvidia.com",
    Version: "v1beta1",
    Kind:    "DynamoGraphDeployment",
}] = skipTopOwnerGrouper
```

## Test plan
- Deploy a v1beta1 DGD with a multi-replica worker component and assert a single worker PodGroup with `minMember = replicas` and all-or-nothing binding.
- Happy to add a unit case to the hub grouping tests if preferred.

### DIFF
--- a/pkg/podgrouper/podgrouper/hub/hub.go
+++ b/pkg/podgrouper/podgrouper/hub/hub.go
@@ -319,10 +319,14 @@ func NewDefaultPluginsHub(kubeClient client.Client, searchForLegacyPodGroups,
 		Kind:    "TrainJob",
 	}] = skipTopOwnerGrouper
 
-	// Dynamo uses Grove Grouper and needs to propagate metadata from DynamoGraphDeployment to PodGang and PodClique.
+	// Dynamo uses the Grove grouper and needs to propagate metadata from
+	// DynamoGraphDeployment to PodGang and PodClique. Match every served version
+	// via the wildcard-version fallback in GetPodGrouperPlugin: v1alpha1 (Dynamo
+	// <=1.1.x), v1beta1 (1.2.0+, which serves/owns the DGD), and any future
+	// version, so gang grouping keeps working across DGD API promotions.
 	table[metav1.GroupVersionKind{
 		Group:   "nvidia.com",
-		Version: "v1alpha1",
+		Version: "*",
 		Kind:    "DynamoGraphDeployment",
 	}] = skipTopOwnerGrouper
 


pkg/podgrouper/podgrouper/hub/hub.go
```diff
@@ -326,6 +326,15 @@ func NewDefaultPluginsHub(kubeClient client.Client, searchForLegacyPodGroups,
...
 
+	// Dynamo 1.2.0+ serves and owns DynamoGraphDeployment as v1beta1; register it so the
+	// same skip-top-owner grouping applies (per-component gang in non-Grove mode,
+	// metadata propagation to PodGang/PodClique in Grove mode).
+	table[metav1.GroupVersionKind{
+		Group:   "nvidia.com",
+		Version: "v1beta1",
+		Kind:    "DynamoGraphDeployment",
+	}] = skipTopOwnerGrouper
+
```
enoodle:
```suggestion
Version: "*",
Kind:    "DynamoGraphDeployment",
}] = skipTopOwnerGrouper

```

Should start at line 325.. github can't show this ..

@HungLee-nvidia What do you think about this ? it will be more future proof

HungLee-nvidia: Good call — done. Switched to `Version: "*"` and collapsed the separate `v1alpha1` + `v1beta1` entries into a single wildcard registration (matching the existing run.ai workload entries just above). The wildcard is picked up by the `gvk.Version = "*"` fallback in `GetPodGrouperPlugin`, so it now covers `v1alpha1`, `v1beta1`, and any future DGD API promotion without another change here. Pushed as c93ce45.

enoodle:
@HungLee-nvidia Grove is moving soon to create the pod group on the grove side, but in the mean time it would be good to have this - Thank you.

enoodle (approved):

SiorMeir (approved):

PR #1839, opened 2026-07-06
ci: switch to CHANGELOG fragments and automate publish proccess

SiorMeir:
## Description

Adopt [changie](https://changie.dev) to manage changelog entries as **per-PR fragments** under `.changes/unreleased/` instead of hand-editing the shared `## [Unreleased]` block in `CHANGELOG.md`, and add **release automation** on top. This eliminates the CHANGELOG merge/backport conflicts that motivated the change and turns cutting a release into two clicks.

We keep the following conventions:
- **`CHANGELOG.md` is the single source of truth** for released versions. There is no parallel per-version archive; at release the pending fragments are **folded into `CHANGELOG.md` and cleared**.
- Released format is unchanged (Keep a Changelog). The committed `CHANGELOG.md` normalizes two accidental double-blank gaps to single blanks (invisible in rendered markdown).

### How it works

**Per PR (contributor)** — unchanged, one command:
- `make changelog` writes `.changes/unreleased/<...>.yaml` (kind → body → optional issue/author). Two PRs never touch the same file; backports cherry-pick the fragment with no conflict.
- CI (`validate-changelog`) requires a fragment (or a `skip-changelog` / `dependencies` label).
- After merge, `changelog-comment` has KAIBOT post a comment linking the merged fragment(s) — comment-only perms, works for forks via `pull_request_target`, never checks out/executes PR code or pushes.

**At release (maintainer) — two new workflows:**
1. **`Release — Prepare Changelog`** (`release-prepare.yaml`): folds the pending fragments into `CHANGELOG.md` as a new `## [version]` section, clears them, and opens a `release/prepare-<version>` PR against the dispatched branch (labeled `skip-changelog` + `release`). Run it from `main` for a minor/major, from a `v*.*` branch for a patch. Triggered by the discretion of the maintaining team.
 
2. **`Release — Tag & Publish`** (`release-publish.yaml`, automatic): when that PR merges, it reads the version + notes from the top of `CHANGELOG.md`, creates the git tag, and creates the GitHub Release.

### Mechanics
- `hack/changelog-fold.sh <version>` renders pending fragments via `changie batch --dry-run` and prepends the section into `CHANGELOG.md` (Keep-a-Changelog spacing), then deletes the fragments.
- `Makefile`: `changelog-release VERSION=` now folds in place; `changelog-regen` removed; `changelog-preview VERSION=` added.
- `main` describes non-patch (minor/major) releases; `v*.*` branches describe their patch releases (just the version passed at dispatch).
[...]

PR #1874, opened 2026-07-09
fix(scheduler): prevent DRA GPU count overflow from understating queue demand

### LINKED ISSUE #1873: DRA GPU count overflow can understate queue demand and prevent eligible GPU reclaim
### What happened?

KAI Scheduler aggregates DRA GPU device counts from pending workloads into queue-level GPU demand and fair-share accounting. These counts are positive `int64` values accepted by the Kubernetes API, but their aggregate can exceed `math.MaxInt64`.

Unchecked signed addition can therefore wrap a positive aggregate into a negative value. This violates two required accounting invariants: aggregate GPU demand must remain non-negative, and adding a positive workload demand must never decrease the queue's existing demand.

The user-visible problem is not that the oversized workload remains Pending. That is the expected outcome. The problem is the collateral effect on legitimate workloads in the same queue. A wrapped negative DRA count can partially or completely cancel their GPU demand in queue accounting. Those workloads may still be considered when GPUs are idle, but the understated or negative fair share can prevent them from reclaiming GPUs currently allocated to another queue, so otherwise eligible workloads are left Pending.

The fix must ensure that all DRA GPU count aggregation paths preserve non-negative, monotonic accounting. An oversized workload may remain unschedulable, but it must not reduce another workload's recorded demand, corrupt parent-queue accounting, or alter legitimate reclaim eligibility through integer wraparound.

KAI Scheduler's DRA GPU bookkeeping sums user-controlled device counts with a plain `+=` on `int64`, with no overflow guard. A ResourceClaim's `spec.devices.requests[].exactly.count` is an `int64` that the apiserver accepts up to `math.MaxInt64` (validation only rejects `count <= 0`, there is no upper bound). A pod that references two GPU ResourceClaims, each with `count: 9223372036854775807`, drives the aggregated device count past `math.MaxInt64`, wrapping it negative.

The negative count flows from `ExtractDRAGPUResourcesFromClaims` (`pkg/common/resources/dra.go`) through `SetDraGpus` / `GetDraGpusCount` into `GetGpusQuota` (`pkg/scheduler/api/resource_info/gpu_resource_requirment.go`), so the proportion plugin computes a negative GPU request for the queue.

Reproduced on a live cluster (Kubernetes v1.36.1, `gpu.nvidia.com` DeviceClass, KAI Scheduler v0.16.3). With a queue `kai-dra-test-q` (gpu quota 8) and a pod referencing two `count: 9223372036854775807` claims, the scheduler logged:

```
Resource division result for queue <kai-dra-test-q>: GPU: deserved: <8>, requested: <-2>, ... fairShare: <-2>
allocate ... resources: <[0 0 -2 1 0 0 0]>
```

The negative value also propagated up to the parent queue `default-parent-queue` (`requested: -2`, `fairShare: -2`), corrupting the fair-share accounting across the queue hierarchy.

The aggregated device count should never wrap negative. An oversized request should saturate at a large positive (and simply unschedulable) value, leaving the queue's GPU fair-share accounting correct and non-negative.

### Environment

- KAI Scheduler: v0.16.3 (code paths also present on `main`)
- Kubernetes: v1.36.1
- DRA: `resource.k8s.io/v1`, `gpu.nvidia.com` DeviceClass

### Additional context

Root cause is the raw `+=` accumulation of user-controlled counts in `countGPUDevicesFromClaim` and `ExtractDRAGPUResourcesFromClaims`, plus the downstream `draGpuCounts` accumulators in `GpuResourceRequirement`. Fix is to accumulate with a saturating add. A PR follows.

thc1006:
### Problem

An oversized but API-valid DRA GPU request from a pending workload can cancel the legitimate GPU demand of other workloads in the same queue. The Kubernetes API accepts any positive `int64` for a `ResourceClaim`'s `...Exactly.Count` (validation only rejects `count <= 0`), so a workload can drive KAI's aggregated device count past `math.MaxInt64`, where a plain `+=` wraps it negative.

### Impact

The oversized workload itself never obtains GPUs. It stays `Pending`, which is the expected outcome. The damage lands on the other workloads sharing its queue: the negative count is subtracted from the queue's aggregate GPU demand, so their honest demand is understated or cancelled outright. Those workloads can still allocate while GPUs are idle, but the understated fair share makes the proportion reclaim-eligibility gate reject them, so an otherwise eligible workload is left `Pending` once the GPUs are held by another queue.

### Root cause

Four aggregation boundaries accumulated user-controlled counts with raw signed addition:

- requests within a single `ResourceClaim` (`countGPUDevicesFromClaim`), including the `AllocationModeAll` bookkeeping `+1`
- claims grouped by device class (`ExtractDRAGPUResourcesFromClaims`)
- requirements across workloads (`GpuResourceRequirement.Add`)
- device classes summed into the total DRA count (`GetDraGpusCount`)

### Fix

Every one of those boundaries now accumulates with a saturating add, so an oversized request stays a large positive (and simply unschedulable) value instead of wrapping negative. The helper lives in the dependency-light `pkg/common/math`, so the scheduler's resource model does not pull in the Kubernetes and DRA client packages for the sake of one integer addition.

## Reproduction

Observed on a live cluster (Kubernetes v1.36.1, `gpu.nvidia.com` DeviceClass, KAI Scheduler v0.16.3). With a queue `kai-dra-test-q` (gpu quota 8) and a pod referencing two `count: 9223372036854775807` GPU claims, the scheduler logged the queue's GPU `requested` and `fairShare` as `-2`, and the value propagated up to `default-parent-queue`:

```
Resource division result for queue <kai-dra-test-q>: GPU: deserved: <8>, requested: <-2>, ... fairShare: <-2>
allocate ... resources: <[0 0 -2 1 0 0 0]>
```

## Tests

All of these fail on the pre-fix code and pass with the change:

- `MaxInt64 + MaxInt64` across two claims of the same device class
- `ExactCount(MaxInt64)` alongside an `AllocationModeAll` request in a single claim
- multi-device-class aggregation in `GetDraGpusCount` and `GpuResourceRequirement.Add`
- a proportion regression that places the oversized workloads and a legitimate 1-GPU workload in one queue, with another queue holding the GPUs, and asserts that the queue ledger stays non-negative, that the oversized demand actually reaches the ledger, and that the reclaim-eligibility gate still admits the legitimate workload

The proportion test is a queue-ledger and reclaim-gate test. It does not drive the full reclaim action (victim selection and eviction), which is a separate stage.

## Related Issues

Fixes #1873

## Checklist

- [x] Self-reviewed

## Breaking Changes

None.

### DIFF
--- a/pkg/common/math/saturating.go
+++ b/pkg/common/math/saturating.go
@@ -7,9 +7,9 @@ package math
 
 import stdmath "math"
 
-// SaturatingAddInt64 returns a + b, clamped to [math.MinInt64, math.MaxInt64]
+// SaturatingAdd returns a + b, clamped to [math.MinInt64, math.MaxInt64]
 // on overflow instead of wrapping around to the opposite sign.
-func SaturatingAddInt64(a, b int64) int64 {
+func SaturatingAdd(a, b int64) int64 {
 	sum := a + b
 	// Overflow can only happen when both operands share a sign and the sign of
 	// the result flips.
--- a/pkg/common/math/saturating.go
+++ b/pkg/common/math/saturating.go
@@ -1,0 +1,23 @@
+// Copyright 2025 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+// Package math holds dependency-light arithmetic helpers that the scheduler's
+// resource model can use without pulling in Kubernetes or DRA client packages.
+package math
+
+import stdmath "math"
+
+// SaturatingAddInt64 returns a + b, clamped to [math.MinInt64, math.MaxInt64]
+// on overflow instead of wrapping around to the opposite sign.
+func SaturatingAddInt64(a, b int64) int64 {
+	sum := a + b
+	// Overflow can only happen when both operands share a sign and the sign of
+	// the result flips.
+	if a > 0 && b > 0 && sum < 0 {
+		return stdmath.MaxInt64
+	}
+	if a < 0 && b < 0 && sum >= 0 {
+		return stdmath.MinInt64
+	}
+	return sum
+}
--- a/.changes/unreleased/Fixed-20260715-153936.yaml
+++ b/.changes/unreleased/Fixed-20260715-153936.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: |-
+  Saturated the DRA GPU device-count accumulation so a ResourceClaim requesting an oversized device count can no longer overflow the queue controller's int64 GPU total to a negative value. [#1873](https://github.com/kai-scheduler/KAI-Scheduler/issues/1873) [thc1006](https://github.com/thc1006)
--- a/pkg/common/math/saturating_test.go
+++ b/pkg/common/math/saturating_test.go
@@ -1,0 +1,30 @@
+// Copyright 2025 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+package math
+
+import (
+	stdmath "math"
+	"testing"
+)
+
+func TestSaturatingAddInt64(t *testing.T) {
+	cases := []struct {
+		name string
+		a, b int64
+		want int64
+	}{
+		{"normal sum", 2, 3, 5},
+		{"positive overflow saturates to MaxInt64", stdmath.MaxInt64, stdmath.MaxInt64, stdmath.MaxInt64},
+		{"positive overflow by one", stdmath.MaxInt64, 1, stdmath.MaxInt64},
+		{"negative overflow saturates to MinInt64", stdmath.MinInt64, stdmath.MinInt64, stdmath.MinInt64},
+		{"mixed signs do not overflow", stdmath.MaxInt64, stdmath.MinInt64, -1},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			if got := SaturatingAddInt64(tc.a, tc.b); got != tc.want {
+				t.Fatalf("SaturatingAddInt64(%d, %d) = %d, want %d", tc.a, tc.b, got, tc.want)
+			}
+		})
+	}
+}
--- a/pkg/common/math/saturating_test.go
+++ b/pkg/common/math/saturating_test.go
@@ -8,7 +8,7 @@ import (
 	"testing"
 )
 
-func TestSaturatingAddInt64(t *testing.T) {
+func TestSaturatingAdd(t *testing.T) {
 	cases := []struct {
 		name string
 		a, b int64
@@ -22,8 +22,8 @@ func TestSaturatingAddInt64(t *testing.T) {
 	}
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
-			if got := SaturatingAddInt64(tc.a, tc.b); got != tc.want {
-				t.Fatalf("SaturatingAddInt64(%d, %d) = %d, want %d", tc.a, tc.b, got, tc.want)
+			if got := SaturatingAdd(tc.a, tc.b); got != tc.want {
+				t.Fatalf("SaturatingAdd(%d, %d) = %d, want %d", tc.a, tc.b, got, tc.want)
 			}
 		})
 	}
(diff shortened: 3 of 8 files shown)


enoodle:
@thc1006 That is a lot of code for a problem that can't really happen currently, why do you need it? Does it worth the extra complications?

thc1006:
You're right to push on the cost, and I don't want to hand you weight you'll have to carry.

On whether it can happen: the pod never binds, but the count still wraps the queue's GPU accounting negative while it's Pending (I put the detail on the issue). Left alone, that tends to resurface later as a confusing fair-share bug that looks like a scheduler fault rather than a bad request.

The fix itself can be small: the real change is a few lines at the two dra.go sites, the rest is a helper and tests, and it matches the guard kueue (kubernetes-sigs/kueue#12897) and Volcano (volcano-sh/volcano#5621) already added for the same field. I'm glad to trim to just those sites, and equally glad to close it if you'd still rather not carry it. Your call, and thanks for taking the time.

pkg/common/math/saturating.go
```diff
@@ -0,0 +1,33 @@
...
+//
+// DRA device counts (ResourceClaim ...Exactly.Count) are user-controlled and
+// effectively unbounded: the apiserver only rejects count <= 0, so a single
+// claim may request up to math.MaxInt64 devices, and a pod may reference
+// several such claims. Accumulating these counts with a plain += can wrap a
+// large positive sum into a negative value, which then flows into GPU quota /
+// fair-share accounting as a negative GPU request. Saturating keeps the running
+// total a meaningful (if very large) positive number, so such a request is
+// simply treated as unschedulable rather than silently corrupting a queue's
+// accounting.
```
enoodle: This is not really a description of the function but a description of the PR / issue, it does not belong here.

enoodle (commented):
Can you remove the test from pkg/scheduler/plugins/proportion/proportion_dra_overflow_test.go and add a test in the actions integration tests that will go through all the parts of the scheduler (You can take this commit if you want : https://github.com/enoodle/KAI-Scheduler/commit/51719592af78e9090be31c0fac44623512c6884a)

pkg/common/math/saturating.go
thc1006: Done in 1945221: trimmed the godoc to just the function's contract, and moved a short version of the rationale to countGPUDevicesFromClaim, where the user-controlled Exactly.Count is accumulated. Happy to drop that from the code too and keep the full rationale in the PR and #1873 if you'd prefer.

enoodle:
@thc1006 Can you please remove the test from pkg/scheduler/plugins/proportion/proportion_dra_overflow_test.go and add a test in the actions integration tests that will go through all the parts of the scheduler. Example test is here: https://github.com/enoodle/KAI-Scheduler/commit/51719592af78e9090be31c0fac44623512c6884a

enoodle:
Also we merged https://github.com/kai-scheduler/KAI-Scheduler/pull/1839 and now the CHANGELOG.md file is updated only by automation.

thc1006:
Hi @enoodle,

Thanks, both done. I moved the overflow coverage out of the proportion plugin and into a reclaim integration test based on your example, so it now goes through the full scheduler cycle: a valid GPU-DRA job, an oversized job whose two claims together request more than int64 can hold, and a victim in the same queue. I kept you as co-author on that commit since the test is yours. It passes locally.

I also moved the changelog to a `.changes/unreleased/` fragment and dropped the manual CHANGELOG.md edit now that #1839 automates publishing, and rebased onto the latest main.

enoodle (dismissed):

pkg/common/math/saturating.go
```diff
@@ -0,0 +1,23 @@
...
+
+// Package math holds dependency-light arithmetic helpers that the scheduler's
+// resource model can use without pulling in Kubernetes or DRA client packages.
+package math
+
+import stdmath "math"
+
+// SaturatingAddInt64 returns a + b, clamped to [math.MinInt64, math.MaxInt64]
+// on overflow instead of wrapping around to the opposite sign.
+func SaturatingAddInt64(a, b int64) int64 {
```
SiorMeir: Let's align this with the similar contributions you mentioned and rename this to just `SaturatingAdd`

SiorMeir (approved):

enoodle (approved):

Issue #1818, opened 2026-07-03
Custom ArgoCD health check (Lua) for the kai-config Config CR


gshaibi:
## Context

Follow-up from #1794 / PR #1815, which added GitOps/ArgoCD installs with the `kai-config` Config CR rendered as a tracked release resource.

ArgoCD has no built-in health assessment for the `kai.scheduler/v1` Config CRD, so the Application can report `Healthy` while the operator is still reconciling (or failing to reconcile) the CR. The gap is documented in `docs/gitops/README.md` ("Known limitations") and the e2e suite compensates by gating on the CR's status conditions instead of Application health.

## Proposal

Publish a Lua health check that maps the Config CR's status conditions (`Deployed`, `Available`, `DependenciesFulfilled`, checked against `observedGeneration`) to ArgoCD health:

- all three True for the current generation → `Healthy`
- otherwise → `Progressing` (or `Degraded` when a condition is False with a terminal reason)

Deliverables:
- `resource.customizations.health.kai.scheduler_Config` snippet in `docs/gitops/README.md` for users to add to `argocd-cm`.
- Optionally upstream it to the argo-cd built-in health library (`resource_customizations/kai.scheduler/Config/health.lua`).
- Same treatment for `SchedulingShard` if useful.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

mateenali66:
I'd like to take this one.

PR #1875, opened 2026-07-09
docs: add ArgoCD custom health check for Config and SchedulingShard

mateenali66:
Adds an argocd-cm custom health check for the `kai.scheduler/v1` Config and SchedulingShard resources, as described in #1818.

ArgoCD has no built-in health for these CRs, so an Application can show `Healthy` while the operator is still reconciling (the e2e suite already gates on the CR status conditions instead of Application health). The health check mirrors that gating: it maps `Deployed`, `Available` and `DependenciesFulfilled` (checked against `observedGeneration`) to `Healthy` when all three are True for the current generation, `Degraded` when one is False for it, and `Progressing` otherwise.

The doc shows the full Lua under the `Config` key and notes that the same block works for `SchedulingShard` (identical conditions).

Testing: ran the Lua (extracted back out of the YAML block) through a small harness covering healthy, missing-condition, stale-generation, false-condition (Degraded), unknown-status, and no-status inputs. Also validated the argocd-cm YAML parses. The script uses only Lua 5.1-compatible constructs, matching ArgoCD's gopher-lua.

Left the optional argo-cd built-in health-library upstream (resource_customizations/) out of this PR since it lives in the argoproj/argo-cd repo; happy to follow up there if you'd like.

Part of #1818.

## Summary by CodeRabbit

* **Documentation**
  * Added an ArgoCD health-check guide for scheduler resources.
  * Clarified how to configure custom health rules so ArgoCD reflects `Progressing`, `Degraded`, and `Healthy` states more accurately.
  * Included guidance for applying the same health configuration to both supported resource types.

### DIFF
--- a/docs/gitops/README.md
+++ b/docs/gitops/README.md
@@ -68,6 +68,61 @@ Notes:
 - `retry` and the CR's `SkipDryRunOnMissingResource` sync-option cover the first sync, where the Config CRD is not established yet.
 - Default Queues, the default SchedulingShard and the resource-reservation namespace carry `Prune=false` (matching their `helm.sh/resource-policy: keep`), so user-modified scheduling configuration is not pruned on sync.
 
+## ArgoCD health check
+
+ArgoCD has no built-in health assessment for the `kai.scheduler/v1` `Config` and `SchedulingShard` resources, so the Application can report `Healthy` while the operator is still reconciling the CR (or failing to). The operator records progress on the CR's status conditions (`Deployed`, `Available`, `DependenciesFulfilled`), which the e2e suite already gates on instead of Application health (see [#1751](https://github.com/kai-scheduler/KAI-scheduler/issues/1751)).
+
+Add a custom health check to `argocd-cm` so ArgoCD reflects the operator's own status. It reports `Healthy` only when all three conditions are `True` for the current `metadata.generation`, `Degraded` when one is `False` for the current generation, and `Progressing` otherwise (conditions not yet observed for the latest generation):
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: argocd-cm
+  namespace: argocd
+data:
+  resource.customizations.health.kai.scheduler_Config: |
+    local hs = {}
+    hs.status = "Progressing"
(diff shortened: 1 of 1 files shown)


mateenali66:
@coderabbitai review

enoodle (dismissed):

mateenali66:
Retitled to drop the invalid `gitops` scope. This is docs-only (docs/gitops/README.md), so it needs the `skip-changelog` label like the other docs PRs; could a maintainer add it?

somanythingstodo:
just my two cents (I am not the author of this fix).
I tested this on kind cluster from gitops-e2e script, and even with no failures, the app goes "Degraded" before turning "Healthy". (it also depends on ArgoCD configuration, but with default config it spends 3 minutes in Degraded status)
And this could result in health-degraded notifications on every install/upgrade.

I think Available=False is a normal transient state while the pods roll out, but the check maps any "False" to "Degraded". So we should consider the status "Degraded" only if DependenciesFulfilled=False, regardless of the other conditions (and any other False should map to "Progressing").

mateenali66:
Confirmed against the operator's `status_reconciler`: `Available` and `Deployed` carry the same reason (`available`/`deployed`) whether the pods are still rolling out or genuinely stuck, so a transient `False` is indistinguishable from a broken one and mapping either to Degraded fires on every install/upgrade. `DependenciesFulfilled=False` is the one condition with a distinct, actionable reason (`dependencies_missing`), so it's the only one that should be Degraded.

Updated the check accordingly: `DependenciesFulfilled=False` for the current generation maps to Degraded (surfacing the missing-dependency message), and any other `False` (`Deployed`/`Available`) stays Progressing until all three reach `True`. Reran the Lua through the harness, including the transient `Available=False` rollout case, which now reports Progressing instead of Degraded. Appreciate you exercising it on a live kind cluster.

enoodle (approved):

gshaibi (approved):

gshaibi:
I'll go ahead and merge.
Thanks @mateenali66 for your contribution and @somanythingstodo for your test and insight 🙏

PR #1877, opened 2026-07-09
Add Karta pod grouper fallback plugin

### LINKED ISSUE #1527: podgrouper - declarative pod grouping for arbitrary CRDs via Karta
## Summary

right now KAI's `pod-grouper` needs a Go plugin per workload type. every new CRD (PyTorchJob, RayCluster, JobSet, MPIJob, LWS, Grove, etc.) gets wired manually under `pkg/podgrouper/podgrouper/plugins/...` and registered in `hub.go`. for unknown CRDs we fall back to the default grouper which treats every pod as its own group (no gang scheduling, `MinAvailable=1`).

idea - add an optional declarative path so a CRD can be supported without writing or merging Go code. concretely, integrate [Karta](https://github.com/run-ai/karta) (Apache-2.0, OSS) as a generic fallback plugin. Karta is a CRD + Go library that describes the structure of any k8s workload (root + child components, pod templates, scaling, status, gang-scheduling instructions) using jq path expressions, all expressed as YAML.

when the new plugin is enabled, an admin or workload vendor can ship a `Karta` resource that describes a CRD, and KAI's pod-grouper produces correct pod groups, `MinAvailable` and grouping keys with no code change.

## Motivation

the k8s AI / batch ecosystem keeps shipping new CRDs (Kubeflow Trainer v2, JobSet variants, NVIDIA Dynamo, Milvus, Grove, plus tons of internal / proprietary ones). with the current model:

- every new CRD = new in-tree plugin + RBAC + tests + a KAI release to consume it
- operators of internal / proprietary CRDs cant use KAI gang-scheduling without forking
- the hub table in `pkg/podgrouper/podgrouper/hub/hub.go` keeps growing and concentrates per-CRD knowledge inside KAI

a declarative extension point pushes that knowledge out of the scheduler and into per-workload manifests, so KAI can support arbitrary CRDs without releases gated on KAI's roadmap.

## What is Karta

Karta is an open source library + CRD that describes the structure of a workload type:

- `rootComponent` - top-level CRD GVK + status mapping
- `childComponents` - owned resources (Deployments, StatefulSets, Jobs, custom kinds, ...) with pod template paths, scale paths, instance selectors, pod selectors
- `optimizationInstructions.gangScheduling.podGroups[]` - which components form a pod group, which keys to group pods by, which filters apply

it already ships pre-built definitions for JobSet, PyTorchJob, RayCluster, RayJob, KServe InferenceService, Knative, MPIJob, NIM Service, LWS, Milvus, NVIDIA Dynamo. consumed via Go modules (`github.com/run-ai/karta/pkg/{api,instructions,resource}`).

the Karta API itself doesnt depend on KAI, so its reusable by any controller / scheduler / platform.

## Mapping Karta components to KAI SubGroups

KAI's `PodGroup` CRD already supports a hierarchical `SubGroups[]` field with per-subgroup `MinMember`, an optional `Parent` (for nested trees), and per-subgroup `TopologyConstraint`. existing in-tree plugins like `grove` and `kubeflow/pytorch` already populate `Metadata.SubGroups` rather than just a flat `MinAvailable`.

Karta's structure model lines up almost 1:1 with this:

- a Karta `rootComponent` + `childComponents` tree maps to a SubGroup tree (`childComponent.ownerRef` -> `SubGroup.Parent`)
- each `gangScheduling.podGroups[].members[].componentName` becomes a SubGroup
- the per-component `scaleDefinition` (replicas / minReplicas / maxReplicas) provides the SubGroup's `MinMember`
- multi-instance components (`instanceIdPath`) can be expanded into one SubGroup per instance

so the proposed plugin shouldnt just compute a flat `MinAvailable` for every pod, it should produce a full `Metadata.SubGroups` slice on the first pod it sees for a top owner, the same way `grove` / `pytorch` do today. that way an admin who ships a Karta gets the same scheduling expressiveness as a hand-written plugin (per-subgroup gang semantics, per-subgroup topology, etc.).

worth calling out - Karta on its own only describes structure + grouping, it doesnt describe scheduler-level concepts like topology levels or `MinSubGroup`. those would need to either be added to Karta's `optimizationInstructions` (upstream, in the Karta repo) or layered on via KAI-side annotations on the Karta CR. open question below.

## References

- Karta repo - https://github.com/run-ai/karta
- Karta technical guide - https://github.com/run-ai/karta/blob/main/docs/Technical%20Guide.md
- Karta example library - https://github.com/run-ai/karta/tree/main/docs/examples
- current KAI pod-grouper hub - `pkg/podgrouper/podgrouper/hub/hub.go`

davidLif:
## Description

Yaml based grouping

## Related Issues

Fixes #1527

## Checklist

- [x] Self-reviewed


pkg/podgrouper/podgrouper/plugins/karta/grouper.go
```diff
@@ -0,0 +1,218 @@
...
+) (*podgroup.SubGroupMetadata, error) {
+	if err := validateSubGroupMapping(componentFactory, subGroupMapping); err != nil {
+		return nil, err
+	}
+
+	minAvailable, err := instructions.CalculateSubtreeScale(ctx, subGroupMapping.ComponentName, nil, componentFactory, g.kartaSummary)
+	if err != nil {
+		return nil, err
+	}
+	if minAvailable == 0 {
```
gshaibi, line 130: Is that always invalid? Don't we allow scaled-to-zero groups?

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func (g *KartaGrouper) Name() string {
+	return "Karta Grouper"
+}
+
+func (g *KartaGrouper) GetPodGroupMetadata(topOwner *unstructured.Unstructured, pod *v1.Pod, _ ...*metav1.PartialObjectMetadata) (*podgroup.Metadata, error) {
+	ctx := context.Background()
+
+	gangScheduling := g.kartaSummary.GetKarta().Spec.Instructions.GangScheduling
```
gshaibi: Consider nil-checking

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
```diff
@@ -0,0 +1,218 @@
...
+func (g *KartaGrouper) getPodGroupMetadataFromPod(pod *v1.Pod) (*podgroup.Metadata, error) {
+	podAsOwner, err := g.convertPodToTopOwner(pod)
+	if err != nil {
+		return nil, err
+	}
+	return g.defaultGrouper.GetPodGroupMetadata(podAsOwner, pod)
+}
+
+func (g *KartaGrouper) calcPodGroupName(topOwner *unstructured.Unstructured, groupingKeys []string) string {
+	baseName := fmt.Sprintf("%s-%s", constants.PodGroupNamePrefix, topOwner.GetUID())
```
gshaibi, line 211: `groupingKeys` should be sanitized imo

pkg/podgrouper/podgrouper/plugins/karta/hub.go
```diff
@@ -0,0 +1,148 @@
...
+	}, nil
+}
+
+func getKartaForGvk(ctx context.Context, kubeClient client.Client, gvk metav1.GroupVersionKind) (*kartav1alpha1.Karta, error) {
+	kartas := &kartav1alpha1.KartaList{}
+	if err := kubeClient.List(ctx, kartas, getKartaPerGvkLabelSelectors(gvk)); err != nil {
+		return nil, err
+	}
+
+	for index := range kartas.Items {
```
gshaibi, line 85: I think that in case of multiple Kartas for the same GVK, we should pick one deterministicly - consider a deterministic tie-break (such as `creationTimestamp`, `name` etc.)

cmd/podgrouper/app/options.go
```diff
@@ -39,6 +40,7 @@ func (o *Options) AddFlags(fs *flag.FlagSet) {
 	fs.IntVar(&o.MaxConcurrentReconciles, "max-concurrent-reconciles", 10, "Max concurrent reconciles")
 	fs.BoolVar(&o.SearchForLegacyPodGroups, "search-legacy-pg", true, "If this flag is enabled, try to find pod groups with legacy name format. If they exist, use the found pod groups instead of creating new once with current name format")
 	fs.BoolVar(&o.KnativeGangSchedule, "knative-gang-schedule", true, "Schedule knative revision as a gang. Defaults to true")
+	fs.BoolVar(&o.GenericKartaFallback, "generic-karta-fallback", true, "Enable Karta-backed generic pod grouping fallback for workload GVKs without native plugins")
```
gshaibi, line 43: Consider exposing it in the KaiConfig or even on the chart's values

pkg/podgrouper/podgrouper/plugins/karta/hub.go
```diff
@@ -0,0 +1,148 @@
...
+	client            client.Client
+	defaultGrouper    grouper.Grouper
+	isKartaCrdMissing *missingCrd
+}
+
+var logger = log.FromContext(context.Background())
+
+func NewKartaHub(client client.Client, defaultGrouper grouper.Grouper) *KartaHub {
+	isKartaCrdMissing := missingCrd{}
+	isKartaCrdMissing.load()
```
gshaibi: Why?

pkg/podgrouper/podgrouper/plugins/karta/hub.go
```diff
@@ -0,0 +1,148 @@
...
+		logger.Info("Karta CRD is not installed")
+		g.isKartaCrdMissing.mark()
+		return nil
+	}
+
+	logger.Error(err, "Failed to create Karta grouper for gvk", "gvk", gvk.String())
+	return nil
+}
+
+func (g *KartaHub) getKartaGrouperForGvk(ctx context.Context, gvk metav1.GroupVersionKind) (*KartaGrouper, error) {
```
gshaibi, line 61: Consider caching it so we won't have to recalculate it on every pod's event

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
```diff
@@ -0,0 +1,218 @@
...
+		return baseName
+	}
+	return fmt.Sprintf("%s-%s", baseName, strings.Join(groupingKeys, "-"))
+}
+
+func (g *KartaGrouper) convertPodToTopOwner(pod *v1.Pod) (*unstructured.Unstructured, error) {
+	obj := &metav1.PartialObjectMetadata{}
+	obj.SetGroupVersionKind(pod.GroupVersionKind())
+	obj.TypeMeta = pod.TypeMeta
+	obj.ObjectMeta = pod.ObjectMeta
```
gshaibi: Can this and https://github.com/kai-scheduler/KAI-Scheduler/blob/7f5c3fc9400759a36e9ba3e1ffd0ac07628d48b0/pkg/podgrouper/podgrouper/podgrouper.go#L146 be extracted to a shared location to avoid duplication?

pkg/podgrouper/podgrouper/hub/hub.go
```diff
@@ -56,50 +57,56 @@ const (
...
+	return ph.defaultGroupingHandler
 }
 
 func (ph *DefaultPluginsHub) HasMatchingPlugin(gvk metav1.GroupVersionKind) bool {
+	_, found := ph.getRegularPlugin(gvk)
+	return found
+}
+
+func (ph *DefaultPluginsHub) getRegularPlugin(gvk metav1.GroupVersionKind) (grouper.Grouper, bool) {
 	if _, found := ph.customPlugins[gvk]; found {
```
gshaibi:
```suggestion
	if f, found := ph.customPlugins[gvk]; found {
		return f, true
	}
```

pkg/podgrouper/podgrouper/hub/hub.go
```diff
@@ -56,50 +57,56 @@ const (
...
-		return true
+		return ph.customPlugins[gvk], true
 	}
 
 	// search using wildcard version
 	gvk.Version = "*"
 	if _, found := ph.customPlugins[gvk]; found {
```
gshaibi:
```suggestion
	if f, found := ph.customPlugins[gvk]; found {
		return f, true
	}
```

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
```diff
@@ -0,0 +1,218 @@
...
+) (*podgroup.Metadata, error) {
+	podGroupMetadata, err := g.defaultGrouper.GetPodGroupMetadata(topOwner, pod)
+	if err != nil {
+		return nil, err
+	}
+	podGroupMetadata.Name = g.calcPodGroupName(topOwner, []string{podGroupDefinition.Name})
+	podGroupMetadata.MinSubGroup = nil
+	podGroupMetadata.SubGroups = nil
+
+	if podGroupDefinition.Topology != nil {
```
gshaibi, line 80: In case it is nil, should we nullify the podGroupMetadata's Topology? Since it can be populated earlier by the default grouper.

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
```diff
@@ -0,0 +1,218 @@
...
+	podQuerier := resource.NewPodQuerier(pod)
+	for _, subGroupMapping := range podGroupDefinition.SubGroups {
+		subGroupMetadata, err := g.constructSubgroup(ctx, componentFactory, subGroupMapping, podQuerier, pod)
+		if err != nil {
+			return nil, err
+		}
+		podGroupMetadata.SubGroups = append(podGroupMetadata.SubGroups, subGroupMetadata)
+	}
+
+	if len(podGroupMetadata.SubGroups) > 0 {
```
gshaibi, line 99: What about the case where no subgroups exist? What should the `MinAvailable` be? (currently I think it is 1)

gshaibi (commented):
Looks great, left some comments.

pkg/podgrouper/podgrouper/plugins/karta/hub.go
```diff
@@ -0,0 +1,148 @@
...
+	return gangScheduling != nil &&
+		(gangScheduling.PodGroup != nil || len(gangScheduling.PodGroups) > 0)
+}
+
+type missingCrd struct {
+	retryAfterUnixNano atomic.Int64 // Unix timestamp in nanoseconds
+}
+
+func (b *missingCrd) load() bool {
+	retryAfter := b.retryAfterUnixNano.Load()
```
gshaibi: Since retryAfter is 0 at initialization, and `time.Now() > 0` always, this can be reduced to:
```suggestion
	return time.Now().UnixNano() < b.retryAfterUnixNano.Load()
```

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
davidLif: It is sanatized in the Karta object Validate function called earlier.

pkg/podgrouper/podgrouper/plugins/karta/hub.go
davidLif: Why what? the initial load is to get a read for the first time if the crd is missing or not.

pkg/podgrouper/podgrouper/plugins/karta/hub.go
davidLif: The Karta object is cached by the client, so it won't actually read the object each time, once it found the matching object

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
davidLif: We don't want to support elastic subgroups in the Karta integration.

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
davidLif: Technically, we check for this when we are looking for a Karta matching the GVK, but I will add some checks here

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
gshaibi: Got it. So consider having an appropriate error log in that case (since the component did expose a replica scale definition)

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
gshaibi: Sanitized to a kubernetes resource name? (regarding length, special character, etc.)

pkg/podgrouper/podgrouper/plugins/karta/hub.go
gshaibi: But you don't do anything with the load return value...

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
davidLif: this was a bug, removed behavior related to that

pkg/podgrouper/podgrouper/hub/hub.go
davidLif: Applied

pkg/podgrouper/podgrouper/hub/hub.go
davidLif: Applied

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
davidLif: Here we don't nulify it - we change it if not null

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
davidLif: Currently it's the number of pods. Will create unitest

pkg/podgrouper/podgrouper/plugins/karta/hub.go
davidLif: You are right. will change

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
davidLif: Ok. I understand better now

pkg/podgrouper/podgrouper/plugins/karta/hub.go
davidLif: Right. Will remove

pkg/podgrouper/podgrouper/plugins/karta/hub.go
davidLif: Will fail if 2 kartas for the same gvk will have grouping instructions - I should choose one with internal tie-breakerts.

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
davidLif: TestGetPodGroupMetadata_PodGroupV2WithoutSubGroups

pkg/podgrouper/podgrouper/plugins/karta/hub.go
davidLif: removed

cmd/podgrouper/app/options.go
davidLif: Exposed

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
gshaibi: But if it is null on the podGroupDefinition, don't you think we should nullify it on the podGroupMetadata?

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
davidLif: Added validation

pkg/podgrouper/podgrouper/plugins/karta/grouper.go
davidLif: No

gshaibi (approved):

PR #1879, opened 2026-07-11
fix(e2e): restore scale test job creation

enoodle:
## Description

Restore scalable Job creation in the KWOK scale tests after the distributed Job refactor.

- keep bulk one-pod Job creation asynchronous and preserve the existing retry behavior
- split distributed Job submission from controller-generated resource readiness
- wait once for the expected Pod and PodGroup counts in distributed scale batches
- return Job, PodGroup, and Pods through `JobResult`
- propagate concurrent creation failures through the parent Ginkgo goroutine
- delete the owning Job between unschedulable-delay samples

## Related Issues

N/A

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. Changes affect test infrastructure helpers and their in-repository callers.

## Additional Notes

Validated with compile-only checks for the resource helpers, scale suite, and topology suite, focused `go vet`, and `git diff --check`.

The KWOK scale suite was not run locally because it requires the dedicated scale-test environment. `make validate` reached a pre-existing generation-order failure where generated clients are removed before `controller-gen` imports them; its working-tree side effects were restored.

### DIFF
--- a/test/e2e/modules/resources/rd/object.go
+++ b/test/e2e/modules/resources/rd/object.go
@@ -1,0 +1,38 @@
+// Copyright 2026 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+package rd
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	runtimeClient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	operationAttemptsRetries = 10
+	retryInterval            = 100 * time.Microsecond
+)
+
+// CreateObjectWithRetries creates an object using the scale-test retry behavior.
+func CreateObjectWithRetries(
+	ctx context.Context, kubeClient runtimeClient.Client, obj runtimeClient.Object,
+) error {
+	key := runtimeClient.ObjectKeyFromObject(obj)
+	err := kubeClient.Get(ctx, key, obj)
+	if err == nil {
+		return fmt.Errorf("object %v already exists in the cluster", key)
+	}
+
+	for i := 0; i < operationAttemptsRetries; i++ {
+		err = kubeClient.Create(ctx, obj)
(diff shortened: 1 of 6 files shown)


test/e2e/modules/resources/rd/object.go
```diff
@@ -0,0 +1,38 @@
...
+	"fmt"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	runtimeClient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	operationAttemptsRetries = 10
+	retryInterval            = 100 * time.Microsecond
```
SiorMeir, line 16: every 0.1 milli second?

enoodle: that is the interval we had before in test/e2e/scale/kwok_test_utils.go

test/e2e/modules/resources/rd/object.go
```diff
@@ -0,0 +1,100 @@
...
+		if err := waitForCreateRetry(ctx, createRetryDelay(attempt, err)); err != nil {
+			return err
+		}
+	}
+
+	var err error
+	for i := 0; i < operationAttemptsRetries; i++ {
+		err = kubeClient.Create(ctx, obj)
+		if err == nil || apierrors.IsAlreadyExists(err) {
+			return nil
```
gshaibi, line 27: Why here its fine and on the previous loop not? Can you please explain the retry behaviour?

enoodle: that is a mistake - should be allowed on both. I will add a description to the PR for this. Or maybe I will split it into another PR for easier review.

test/e2e/modules/resources/rd/object.go
enoodle: I will fix this in a future PR - currently I want to get some of the functionality that we had before.

SiorMeir (approved):

Issue #1881, opened 2026-07-12
Parent/child quota validation sums children quotas without handling the -1 unlimited sentinel


thc1006:
### Summary

The `--enable-quota-validation` parent/child quota warnings sum children queues' quota with raw addition and never treat `-1` (`constants.UnlimitedResourceQuantity`) as unlimited. An unlimited child (`-1`) lowers the running total and masks a real over-allocation.

Two functions in `pkg/admission/webhook/queuehooks/queue_validator.go`:

- `validateChildrenQuotaSum` (around lines 314-359): `totalChildrenCPU += child.Spec.Resources.CPU.Quota` (and GPU / Memory), then `if totalChildrenCPU > parentQueue.Spec.Resources.CPU.Quota` with no `-1` guard. The per-child compare on the same values is also unguarded.
- `validateParentChildQuota` (around lines 258-312): same pattern (`totalChildrenCPU += existingChild.Spec.Resources.CPU.Quota`, plain `>` comparisons).

The sentinel is defined in `pkg/common/constants/constants.go`: `UnlimitedResourceQuantity = float64(-1)`.

### Example (false negative)

Parent `CPU.Quota = 10`, child A `CPU.Quota = -1` (unlimited), child B `CPU.Quota = 5`.

The loop computes `0 + (-1) + 5 = 4`; `4 > 10` is false, so no warning is emitted. An unlimited child makes the aggregate effectively unbounded, which exceeds the finite parent, so a warning should fire. The per-child check has the same flaw (`-1 > 10` is false), so an individually-unlimited child never trips its own comparison either. In a borderline config where the finite remainder alone still exceeds the parent, a warning does fire but reports a meaningless number instead of "unlimited".

### Impact

- False-negative parent/child quota warnings (and a misleading number in the borderline case) whenever any child uses `-1`.
- Gated by `--enable-quota-validation`. `-1` is the documented unlimited sentinel a user can set on any queue's CPU / GPU / Memory quota, so it is user-reachable.

### Scope

Pre-existing. Both functions were introduced in #857 (`4c7b05c9`) and are untouched by the `--enforce-quota-violation` work in #1810. Found during the #1810 review; #1810 deliberately leaves this path alone.

### Suggested fix

[...]

PR #1883, opened 2026-07-12
feat(queue-controller): expose descendent quota over-subscription via Helm and Config CR

### LINKED ISSUE #1819: feat(chart): expose queue controller --enable-quota-validation via Helm and Config CR

The queue-controller supports opt-in hierarchical quota validation via the CLI flag `--enable-quota-validation` (default `false`). When enabled, the queue validation webhook emits warnings (not errors) when:

- a child queue's quota exceeds its parent's quota (CPU, GPU, Memory)
- the sum of sibling quotas exceeds the parent's quota

The flag exists in code but is only reachable by manually patching the queue-controller Deployment. It is not exposed through the operator Config CR or Helm chart.

## Current state

| Layer | Status |
|-------|--------|
| CLI | `--enable-quota-validation` in `cmd/queuecontroller/app/options.go` (default `false`) |
| Validator | `pkg/admission/webhook/queuehooks/queue_validator.go` |
| Operator args | `pkg/operator/operands/queue_controller/resources.go` — `buildArgsList()` does **not** pass the flag |
| Config CR | `webhooks.enableValidation` exists (webhook on/off) but no quota-validation toggle |
| Helm | no `queuecontroller.enableQuotaValidation` in `values.yaml` |

## Expected behavior

Enable from Helm:

```yaml
queuecontroller:
  enableQuotaValidation: true
```

Wired through:

1. `values.yaml` → `kai-config` Config CR (`spec.queueController.enableQuotaValidation` or under `webhooks`)
2. Operator `buildArgsList()` → `--enable-quota-validation=true` when set
3. Helm unit test (same pattern as `enabled_test.yaml`)

Default remains `false` for backward compatibility.

### Why is this needed?

- chart-exposed validation helps operators catch bad configs at admission time without patching deployments

## Proposed scope

- [ ] Add field to `QueueController` API (Config CR)
- [ ] Pass flag in operator `buildArgsList()`
- [ ] Helm `values.yaml` + `_helpers.tpl` wiring
- [ ] Regenerate CRD (`make manifests`)
- [ ] Helm test asserting value reaches Config CR / deployment args
- [ ] CHANGELOG entry

kshitizlohia1994:
## Description

The queue-controller's `--enable-quota-validation` flag (opt-in descendent quota checks in the queue validation webhook) was only reachable by manually patching the queue-controller Deployment — it wasn't exposed through the operator Config CR or the Helm chart.

This PR:

- Renames `--enable-quota-validation` (bool) to `--limit-descendents-over-subscription` with three enforcement modes: `none` (default, unchanged behavior), `warning` (admission warnings), `block` (rejects the request).
- Adds `spec.queueController.limitDescendentsOverSubscription` to the Config CR (kubebuilder enum, default `none`) and wires it through the operator's `buildArgsList()`.
- Exposes `queuecontroller.limitDescendentsOverSubscription` in `values.yaml` / `_helpers.tpl`.
- Regenerates the CRD and deepcopy code.

## Related Issues

Closes #1819

## Testing

- **Unit tests**: expanded `queue_validator_test.go` to cover all three modes on both `ValidateCreate` (parent/child) and `ValidateUpdate` (children-sum) paths, including GPU/Memory-only violations, multi-violation aggregation into a single `block` error, and `ParseOverSubscriptionMode` parsing/validation. Added `buildArgsList()` tests in the operator package and Helm chart assertions in `enabled_test.yaml` for the default/`warning`/`block` values.
- **Integration tests (envtest)**: new suite under `pkg/admission/webhook/queuehooks/integration_tests/` that boots a real API server + etcd via envtest, registers the actual `ValidatingWebhookConfiguration`, and drives real `Queue` create/update calls through it — verifying `none` admits silently, `warning` admits with a warning surfaced by the API server, and `block` rejects the request (and that the rejected object is never persisted), for both the create (parent/child) and update (children-sum) paths.
- **kind cluster (live Helm install)**: built the `operator` and `queue-controller` images locally, loaded them into a kind cluster, and did a real `helm install`/`helm upgrade` of this chart. Confirmed the operator correctly reads `spec.queueController.limitDescendentsOverSubscription` from the Config CR and injects `--limit-descendents-over-subscription <mode>` into the live queue-controller Deployment, and that the real webhook rejects an over-subscribing child queue in `block` mode with the expected aggregated error message (and admits a within-quota child).

## Checklist

- [x] Self-reviewed
- [x] Added/updated tests
- [x] Updated documentation (CHANGELOG.md)

## Breaking Changes

`--enable-quota-validation` is renamed to `--limit-descendents-over-subscription`. Anyone who was manually patching the queue-controller Deployment with the old flag needs to switch to the new one. Default behavior (`none`/off) is unchanged.

## Additional Notes

Scope intentionally limited to the existing quota-validation surface; a related open PR (#1810) adds a second, separate flag for a different check and can be wired through the same Config CR field pattern in a follow-up once merged.


kshitizlohia1994:
Ran this end-to-end on a local kind cluster (`kai-quota-test`, `kindest/node:v1.35.0`) to confirm the operator/chart wiring.

Setup: cross-compiled `operator` and `queuecontroller` locally, built their `prod`-stage images, loaded them into kind, then `helm install ./deployments/kai-scheduler --no-hooks --set kaiConfig.render=true --set queuecontroller.limitDescendentsOverSubscription=block` (skipping the crd-upgrader hook Job; CRDs still install from the chart's `crds/` dir).

Verified:
- kubectl get config kai-config -o jsonpath='{.spec.queueController.limitDescendentsOverSubscription}' → `block`
- Operator (the locally-built image) reconciled that into the live Deployment: `kubectl get deploy queue-controller -o jsonpath='{.spec.template.spec.containers[0].args}'` → ["--metrics-listen-address",":8080","--limit-descendents-over-subscription","block"]
- `ValidatingWebhookConfiguration kai-queue-validation-v2` registered and reachable
- Parent queue (cpu quota 100) admitted normally
- Over-subscribing child (cpu 200, gpu 2, memory 2048) rejected by the live webhook:
  `Forbidden: admission webhook "queue-validation.kai.scheduler" denied the request: queue quota over-subscription: child queue CPU quota (200) exceeds parent queue kt-parent CPU quota (100); total children CPU quota (200) exceeds parent queue kt-parent CPU quota (100); child queue GPU quota (2.00) exceeds parent queue kt-parent GPU quota (1.00); child queue Memory quota (2048) exceeds parent queue kt-parent Memory quota (1024)`
  and confirmed it was never persisted (`kubectl get queue` → NotFound)
- Within-quota child admitted normally in `block` mode
- `helm upgrade --reuse-values --set queuecontroller.limitDescendentsOverSubscription=warning` → operator reconciled the Deployment args to `warning` without a restart of the release

pkg/admission/webhook/queuehooks/integration_tests/suite_test.go
enoodle, line 1: Lets use the already setup `pkg/env-tests/queuecontroller` for this instead

cmd/queuecontroller/app/options.go
```diff
@@ -18,7 +18,10 @@ type Options struct {
 	EnableLeaderElection         bool
 	EnableWebhook                bool
 	SkipControllerNameValidation bool // Set true for env tests
-	EnableQuotaValidation        bool // Enable validation warnings for queue quota relationships
+
+	// LimitDescendentsOverSubscription controls admission enforcement when a child
+	// queue's quota exceeds its parent's quota (one of none, warning, block).
```
enoodle: No need for those comments here, it is clear enough from the arg doc below. Can you leave it as a small comment in the saem line?

deployments/kai-scheduler/templates/_helpers.tpl
```diff
@@ -188,6 +188,9 @@ spec:
       {{- end }}
 
   queueController:
+    {{- if .Values.queuecontroller.limitDescendentsOverSubscription }}
+    limitDescendentsOverSubscription: {{ .Values.queuecontroller.limitDescendentsOverSubscription }}
+    {{- end }}
```
enoodle, line 204: Is it used only in the test? we don't really have to test every parameter that we pass along in the helm chart tests.

pkg/admission/webhook/queuehooks/integration_tests/suite_test.go
kshitizlohia1994: done. moved the validator coverage into the shared pkg/env-tests harness and dropped the standalone suite. initially added a separate envtest bootstrap to prove the webhook was reachable end-to-end, but that duplicated the existing work so removed it and reused the shared apiserver/client instead.

cmd/queuecontroller/app/options.go
kshitizlohia1994: done. collapsed it to a single trailing // one of none, warning, block comment on the field.

deployments/kai-scheduler/templates/_helpers.tpl
kshitizlohia1994: right. dropped the helm chart assertions. kept only the _helpers.tpl wiring.

enoodle:
@kshitizlohia1994 looks good,
Can you fix the changelog merge issue (we are working on a system that will prevent those merge conflicts) and sign off on the commits?

thc1006:
Heads up on something that gets worse once `block` exists here.

`validateChildrenQuotaSum` adds the child quotas with a raw `+=`, and `-1` means unlimited. Parent CPU quota 10, children with quotas `-1` and `5`: the total comes out as 4, `4 > 10` is false, and nothing fires. Right now that's just a missed warning. Under `block` it's an over-subscribing hierarchy that gets admitted, which is backwards. `validateParentChildQuota` has the same thing. I filed it as #1881 with the repro. The scheduler already has `-1`-aware add and compare in `resource_share/resource_quantities.go` if you want something to copy.

I'm reading from your description that `block` rejects on the children-sum path too, so tell me if I've got that wrong.

Your call whether it goes in here, since you're the one in that code now. If you'd rather keep this tight, I'll send it as a small PR after yours lands so we're not both editing the same two functions.

On the #1810 overlap you already called out: the part that'll actually bite is that both PRs add their own tri-state parser to `queue_validator.go` (`ParseOverSubscriptionMode` here, `ParseEnforcementMode` there), on top of the same `app.go`/`options.go` changes. Don't sequence around me though. Land yours, and I'll rebase #1810 on top and use your mode type for `--enforce-quota-violation` rather than shipping a second one.

kshitizlohia1994:
> Heads up on something that gets worse once `block` exists here.
> 
> `validateChildrenQuotaSum` adds the child quotas with a raw `+=`, and `-1` means unlimited. Parent CPU quota 10, children with quotas `-1` and `5`: the total comes out as 4, `4 > 10` is false, and nothing fires. Right now that's just a missed warning. Under `block` it's an over-subscribing hierarchy that gets admitted, which is backwards. `validateParentChildQuota` has the same thing. I filed it as #1881 with the repro. The scheduler already has `-1`-aware add and compare in `resource_share/resource_quantities.go` if you want something to copy.
> 
> I'm reading from your description that `block` rejects on the children-sum path too, so tell me if I've got that wrong.
> 
> Your call whether it goes in here, since you're the one in that code now. If you'd rather keep this tight, I'll send it as a small PR after yours lands so we're not both editing the same two functions.
> 
> On the #1810 overlap you already called out: the part that'll actually bite is that both PRs add their own tri-state parser to `queue_validator.go` (`ParseOverSubscriptionMode` here, `ParseEnforcementMode` there), on top of the same `app.go`/`options.go` changes. Don't sequence around me though. Land yours, and I'll rebase #1810 on top and use your mode type for `--enforce-quota-violation` rather than shipping a second one.

thank you @thc1006 for highlighting this issue. i can take care of this in the same pr. post the fix, verified the -1 fix live on a kind cluster in block mode using real queue yamls.

**case 1** finite parent, unlimited (-1) child. rejected by the webhook.

```yaml
apiVersion: scheduling.run.ai/v2
kind: Queue
metadata: {name: fix-parent-a}
spec:
  resources:
    cpu: {quota: 10, limit: -1, overQuotaWeight: 1}
    gpu: {quota: 10, limit: -1, overQuotaWeight: 1}
    memory: {quota: 10, limit: -1, overQuotaWeight: 1}
---
apiVersion: scheduling.run.ai/v2
kind: Queue
metadata: {name: fix-child-unlimited}
spec:
  parentQueue: fix-parent-a
  resources:
    cpu: {quota: -1, limit: -1, overQuotaWeight: 1}
    gpu: {quota: 0, limit: -1, overQuotaWeight: 1}
    memory: {quota: 0, limit: -1, overQuotaWeight: 1}
```

```
Error from server (Forbidden): admission webhook "queue-validation.kai.scheduler" denied the request: queue quota over-subscription: child queue cpu quota (unlimited) exceeds parent queue fix-parent-a cpu quota (10); total children cpu quota (unlimited) exceeds parent queue fix-parent-a cpu quota (10)
```

before the fix this was silently admitted.

**case 2** unlimited (-1) parent, finite child. admitted.

```yaml
apiVersion: scheduling.run.ai/v2
kind: Queue
metadata: {name: fix-parent-unlimited}
spec:
  resources:
    cpu: {quota: -1, limit: -1, overQuotaWeight: 1}
    gpu: {quota: -1, limit: -1, overQuotaWeight: 1}
    memory: {quota: -1, limit: -1, overQuotaWeight: 1}
---
apiVersion: scheduling.run.ai/v2
kind: Queue
metadata: {name: fix-child-finite}
spec:
  parentQueue: fix-parent-unlimited
  resources:
    cpu: {quota: 1000, limit: -1, overQuotaWeight: 1}
    gpu: {quota: 8, limit: -1, overQuotaWeight: 1}
    memory: {quota: 8192, limit: -1, overQuotaWeight: 1}
```

before the fix this was wrongly rejected.

both match the expected behavior, on top of the unit and env test coverage.

kshitizlohia1994:
> @kshitizlohia1994 looks good, Can you fix the changelog merge issue (we are working on a system that will prevent those merge conflicts) and sign off on the commits?

done. removed the duplicated preemption-delay changelog entry from the rebase and moved this pr's entry to the top of added (latest first), and signed off all three commits. rebased onto latest main, now mergeable.

pkg/admission/webhook/queuehooks/queue_validator.go
```diff
@@ -6,33 +6,65 @@ package queuehooks
...
+	case "", OverSubscriptionModeNone:
+		return OverSubscriptionModeNone, nil
+	case OverSubscriptionModeWarning:
+		return OverSubscriptionModeWarning, nil
+	case OverSubscriptionModeBlock:
+		return OverSubscriptionModeBlock, nil
+	default:
+		return "", fmt.Errorf("invalid over-subscription mode %q: must be one of none, warning, block", value)
+	}
+}
```
enoodle, line 54:
```suggestion
func ParseOverSubscriptionMode(value string) (OverSubscriptionMode, error) {
    mode := OverSubscriptionMode(value)
    if mode == "" {
            mode = OverSubscriptionModeNone
    }

    switch mode {
    case OverSubscriptionModeNone,
            OverSubscriptionModeWarning,
            OverSubscriptionModeBlock:
            return mode, nil
    default:
            return "", fmt.Errorf("invalid over-subscription mode %q: must be one of none, warning, block", value)
    }
}

```

enoodle: Just a small nit, take it or leave it..

enoodle (commented):
validateParentChildQuota still only validates the CPU resource, Do we want to fix it in this PR or create a separate issue for that?

kshitizlohia1994:
> validateParentChildQuota still only validates the CPU resource, Do we want to fix it in this PR or create a separate issue for that?

@enoodle if it's alright with you, would it be okay to keep the scope of this pr limited and let me raise a separate small pr for that fix? here's the issue for reference: https://github.com/kai-scheduler/kai-scheduler/issues/1923. happy to take care of it in a follow-up pr.

kshitizlohia1994:
Validated the following cases on a live kind cluster (real webhook, built images loaded from this branch, tested against `queue-controller`'s `--limit-descendents-over-subscription` flag):

| Case | Mode | Result | Valid |
|---|---|---|---|
| Child CPU/GPU/Memory individually exceeding parent quota | block | Rejected, correct dimension named | ✅ |
| Child within all parent quotas | block | Admitted | ✅ |
| Sibling GPU sum exceeding parent, each child individually within quota | block | Rejected (`total children GPU quota (6) exceeds parent... (4)`) | ✅ |
| Sibling Memory sum exceeding parent, each child individually within quota | block | Rejected | ✅ |
| Unlimited (`-1`) child under a finite parent | block | Rejected, rendered as "unlimited" in the message | ✅ |
| Finite child under an unlimited (`-1`) parent | block | Admitted | ✅ |
| Update path: shrinking parent CPU/GPU/Memory below existing children's combined quota | block | Rejected for all three dimensions | ✅ |
| Same violation as above | warning | Admitted, violation surfaced as an admission warning instead of a rejection | ✅ |

kshitizlohia1994:
@enoodle should i continue forward with this pr changes or this seems not relevant for the time being?

Issue #1872, opened 2026-07-09
Make deferred resize a scheduling event KAI considers when pre-empting and reclaiming


sam-huang1223:
### What you would like to be added?

When a node is full, and an in-place resize is performed for a pod within the node, KAI scheduler should pre-empt/reclaim other pods in the node to allow resizing to complete

### Why is this needed?

So that we can automatically scale resource requests of important workloads to make room for other less important workloads while also protecting the important workload to ensure that if it needs to scale up it can scale up.

SiorMeir:
Hi @sam-huang1223 ! 👋🏻 

As your issue title suggests, in order to support this we'll need to trigger a scheduling cycle on a `Deferred` state of a pod, and use the currently implemented preemption logic.
Note that this may result in the pod not being able to resize, based on priority and preemptibility of other workloads running on the node.

Would you like to contribute a PR to add this missing functionality?

sam-huang1223:
@SiorMeir not super familiar with go but happy to burn some tokens! will have a PR up soon

gshaibi:
I'm not sure we want deferred resizes to trigger preemption/reclaim the same way pending jobs do. A pending pod has no resources and is fully blocked while a resized pod is already running, it just wants more.

The Kubernetes KEP ([1287](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/1287-in-place-update-pod-resources/README.md#non-goals)) explicitly excludes scheduler preemption for deferred resizes from scope.

gshaibi:
> I'm not sure we want deferred resizes to trigger preemption/reclaim the same way pending jobs do. A pending pod has no resources and is fully blocked while a resized pod is already running, it just wants more.
> 
> The Kubernetes KEP ([1287](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/1287-in-place-update-pod-resources/README.md?rgh-link-date=2026-07-28T11%3A41%3A07.000Z#non-goals)) explicitly excludes scheduler preemption for deferred resizes from scope.

After a maintainers discussion we decided we do want to trigger fair evictions for deferred resizes.

Hagay-RunAI:
Hi @sam-huang1223 We have started to explore and design the implementation this issue. Would you be able to help us with more details on the use-cases your organization uses resource resize for, and what types of workload frameworks (controllers) do you use ? afaik not all frameworks support resource resize.

davidLif:
Hello @sam-huang1223,

[...]

PR #1892, opened 2026-07-13
feat(scheduler): free room for kubelet-deferred in-place pod resizes

sam-huang1223:
## What

Makes a **kubelet-deferred in-place pod resize** (KEP-1287) a scheduling event KAI acts on. When a node is full, the kubelet leaves the resize `Deferred` instead of evicting anything, so an important workload that is scaled back up silently stays at its old size. KAI frees room on the resizing pod's **own node** so the kubelet can actuate the resize on its next attempt.

The growth is scheduled **exactly like any other pending pod in the queue** — it carries the resizing pod's queue/priority/preemptibility and flows through the *unchanged* `allocate`/`preempt`/`reclaim`, so there is **no resize-specific gate**. KAI's existing rules apply as-is: it preempts only strictly-lower-priority same-queue work and reclaims from other queues only within fair share; if the queue is not entitled (or only higher-priority/non-preemptible neighbours exist) the resize stays `Deferred`, precisely as such a pending pod would stay `Pending`. **No new action, config, CRD, or Helm change**, and KAI never calls the `/resize` subresource.

## How

The resize is turned into a pending demand the existing `allocate → preempt → reclaim` machinery already satisfies — so quota, fairness, victim selection, and scenario validation are all reused for free:

- **Detection** (`pkg/scheduler/api/pod_info/resize.go`): `PodResizePending` + reason `Deferred`; the per-resource positive delta of desired (spec) − actual (`status.containerStatuses[].resources`), diffed over raw `ResourceList`s so an explicit `nvidia.com/gpu: 0` cancels instead of being misread as a fractional device. `Infeasible` resizes ignored.
- **Truthful accounting** (`regularContainerRequests`): a deferred-resize pod is charged at `desired − delta` (its actually-granted size), keeping the node view truthful and avoiding double-counting. Invariant: `charge (actual) + reservation (delta) = desired`.
- **Node-pinned reservation** (`cluster_info.InjectResizeReservations`, in `Snapshot()`): a synthetic **pending** `PodGroupInfo` for the delta, pinned to the pod's node via a required `kubernetes.io/hostname` affinity, in the pod's queue/priority/preemptibility, flagged `IsResizeReservation`, given a minimal PodGroup, charged the delta only (no `pods:1`).
- **No-bind guard** (`Session.BindPod`): a reservation has no real pod, so it is never bound — it only reserves capacity and drives eviction, reaching `Pipelined` at most. The kubelet actuates the real resize once room exists; next snapshot the delta is zero and no reservation is produced.

## Testing

- **Unit** — detection & accounting (`pod_info/resize_test.go`), no-bind guard (`framework/bind_resize_test.go`), injection (`cluster_info/resize_injection_test.go`).
- **Integration** — the real `Snapshot()` path charges the resizer at actual **and** injects the reservation (`cluster_info/resize_snapshot_test.go`).
- **Composition** — with the reservation injected, `preempt` evicts a lower-priority same-queue victim and pipelines the reservation (never binding it) **within quota**, and evicts nothing **over quota** (`actions/preempt/composition_resize_test.go`).
- **e2e (CI/Kind, k8s 1.35)** — fills a node, defers a resize, asserts room is freed within quota and not over quota (`test/e2e/suites/resize/`). Guarded on `InPlacePodVerticalScaling`; cannot run in the unit environment.

Design: `docs/developer/designs/deferred-resize-preemption/`. Example: `examples/deferred-resize-preemption/`.

## Compatibility

Additive and behavior-preserving. The only change to existing accounting is that a pod **with a deferred resize** is charged at its actual size instead of desired; all other pods are unchanged. `allocate` / `preempt` / `reclaim` / `consolidation` and the proportion/fairness plugin are otherwise untouched. No new action, action-priority, config, CRD, or Helm change.

Refs #1872


enoodle:
Hi @sam-huang1223 
We usually prefer to create a design doc in its own PR for major changes so that we can more easily discuss the direction, but this also works.
Do you think maybe this could have been implemented inside the current actions? for example by marking those pods as pending (but they already have spec.NodeName set so they have to go to that node) 

I am not sure what exactly happened there but it seems like it didn't really implement adding a new action but added it inside the current Preempt action.

Maybe it should start by adding e2e / env-test (this will probably be easier) tests and work in TDD fashion to implement it.

sam-huang1223:
> Hi @sam-huang1223 We usually prefer to create a design doc in its own PR for major changes so that we can more easily discuss the direction, but this also works. Do you think maybe this could have been implemented inside the current actions? for example by marking those pods as pending (but they already have spec.NodeName set so they have to go to that node)
> 
> I am not sure what exactly happened there but it seems like it didn't really implement adding a new action but added it inside the current Preempt action.
> 
> Maybe it should start by adding e2e / env-test (this will probably be easier) tests and work in TDD fashion to implement it.

hi sorry this is not ready for review yet! much iteration to come. I should have left this in draft state

gshaibi:
Hi @sam-huang1223, thanks for the work you’ve put into this.

This PR is still marked as a draft and hasn’t been updated for about a month. Are you planning to continue working on this?

Issue #1880, opened 2026-07-12
podgroupcontroller: align ResourcesStatus accounting with scheduler pod requests


thc1006:
### Summary

`Queue.status.allocated` / `allocatedNonPreemptible` (and the corresponding PodGroup `ResourcesStatus`) are computed by `calculatedAllocatedResources` / `calculateRequestedResources` in `pkg/podgroupcontroller/controllers/metadata/pod.go`, which sum only `pod.Spec.Containers[*].Resources.Requests` plus KAI's GPU-sharing and DRA resources. They do not account for:

- `pod.Spec.InitContainers` (init-phase peak)
- native sidecars (`initContainers[].restartPolicy: Always`)
- `pod.Spec.Overhead`

The scheduler's own per-pod request (`getPodResourceRequest` in `pkg/scheduler/api/pod_info/pod_info.go`) does include all three, mirroring Kubernetes `PodRequests` / kubelet `AggregateContainerRequests`. The queue's allocated share in the proportion plugin is built from `AcceptedResourceVector`, which for base resources is a clone of that peak `ResReqVector`. So the scheduler reserves these against a queue while the status does not report them.

### Impact

- `kai_queue_allocated_*` metrics undercount CPU/memory for pods that use init containers, native sidecars, or Pod overhead.
- The opt-in `--enforce-quota-violation` admission check (added in #1810) reads the same status, so for such pods it can under-detect a limit/quota reduction below the scheduler's actual allocation.

GPU accounting is mostly unaffected, since GPUs are usually requested by the main container. The exception: standard extended-resource GPU (`nvidia.com/gpu`, `amd.com/gpu`) and container-level MIG (`nvidia.com/mig-*`) requested in an init container or native sidecar are also undercounted here, because the scheduler folds them into the pod request while the producer sums only `pod.Spec.Containers`. DRA GPU counts, GPU-sharing fractions and memory, and legacy annotation-driven MIG are on separate paths and are not affected.

### Decision needed

`Queue.status.allocated` is documented as "current allocated ... for all running jobs" (steady-state usage), while the scheduler treats a queue's allocation as the reserved peak (including finished init-container peaks). These are two different models. Aligning them requires deciding which semantics the status field should carry (peak vs steady-state), updating the API field comment accordingly, and, if moving to the scheduler's model, factoring the pod-request aggregation into a shared helper reused by both the scheduler and the podgroupcontroller, taking care not to double-count GPU-sharing / DRA GPU resources.

[...]

PR #1893, opened 2026-07-13
fix(pod-group-controller): count native sidecars and pod overhead in queue status

thc1006:
## Description

The PodGroup status producer summed only `pod.Spec.Containers`, so native sidecars (`initContainers` with `restartPolicy: Always`) and Pod `Overhead` never reached `PodGroup.status.resourcesStatus` or `Queue.status.allocated`. Both are held for as long as the pod runs and the scheduler charges both, so a queue running Istio-style sidecars or a Kata/gVisor runtime reported less than it was actually using.

`calculatedAllocatedResources` and `calculateRequestedResources` now share `podSteadyStateResources`: regular containers, native sidecars, and Pod overhead. The API field comments and the generated CRD descriptions now say what the field actually holds.

Stage 1 of #1880. A GPU asked for by a sidecar or set in an overhead, and the peak of a non-restartable init container, are deliberately left out, see below.

## Related Issues

Part of #1880

## Checklist

- [x] Self-reviewed
- [x] Added/updated tests
- [x] Updated CHANGELOG

## Breaking Changes

None to the API. One behavior change to know about before upgrading: `kai_queue_allocated_cpu_cores` and `kai_queue_allocated_memory_bytes` rise for queues running sidecar or sandboxed workloads, and they are the default input to the scheduler's time-based fairshare usage database (`pkg/scheduler/cache/usagedb/prometheus`). Where that is enabled they step once on upgrade for pods that are already running.

`kai_queue_allocated_gpus` does not move, and that is enforced rather than asserted: `withoutGpuResources` drops exactly what `getAllocatedGpus` counts as a GPU, so no new `*/gpu` key can enter the status from a sidecar or an overhead. `TestPodSteadyStateResources/a_sidecar's_vendor_gpu_never_reaches_the_status` fails if that stops being true.

## Additional Notes

- **Why a GPU from a sidecar or an overhead is left out.** Two independent reasons, and both had to be closed.

  First, correctness. For a pod carrying a `gpu-fraction`, `gpu-memory` or legacy MIG annotation, `updatePodAdditionalFields` rebuilds the pod's GPU requirement from that annotation and throws the container GPU request away. The producer has no such step, so counting a sidecar's GPU on top of the annotation-derived value would report a GPU that nothing reserved. Two of those pods reach a live cluster: `getFirstGPULimit` only checks `nvidia.com/gpu` limits, so an `amd.com/gpu` sidecar is admitted, and it never looks at MIG annotations at all. `TestCalculateRequestedResources_GpuAnnotations` pins them.

  Second, the metric. `getAllocatedGpus` returns the first key with a `/gpu` suffix in map order, and the repo's own test asserts that `somethingelse/gpu` counts as a GPU, so the suffix rule is deliberate. Letting *any* `*/gpu` name into the status from a sidecar would therefore move `kai_queue_allocated_gpus`, and a pod with two such keys would make it non-deterministic. So the filter uses the metric's own rule (`*/gpu`, plus MIG) rather than the three names the scheduler routes into its GPU requirement.

- **What that costs, stated rather than buried.**
  - On a pod with *no* GPU annotation the scheduler really does reserve a sidecar's GPU, and the producer now does not. The drop is unconditional while the correctness reason above is not. Counting it on some pods and not others looked worse than leaving it out until #1880 shares the scheduler's normalization. It isn't a regression: `main` counts no sidecar resource at all.
  - Matching the metric's `*/gpu` rule also drops a `*/gpu`-suffixed resource that the scheduler treats as an ordinary scalar. That is two wrong classifications cancelling, and I would rather say so than pretend otherwise. The real fix is `getAllocatedGpus`'s suffix heuristic. I wrote that fix in #1810 and reverted it on request, so it is deliberately out of scope here; say the word and I will send it as its own PR.
  - The same unsoundness is **already live through regular containers** and is untouched. A `gpu-fraction` pod with `amd.com/gpu` on a regular container, or a legacy MIG annotation with `nvidia.com/gpu` on one, are both admitted today and both report a GPU the scheduler discards. Fixing that changes what an existing whole-GPU pod reports, which is beyond stage 1.

- **Why the init peak is left out.** The scheduler charges `max(regular + sidecars, init peak) + overhead`. Sidecars and overhead are held for the pod's whole life, so they belong in the status under either reading of `status.allocated` (steady-state usage, or the scheduler's reserved peak). The init peak is the one part that depends on which reading wins, and that's the open question in #1880. This has nothing to do with the GPU question: it needs the scheduler's max semantics, not its annotation handling.

- **Why not `resourcehelper.PodRequests`.** Fair question, since it's a direct dependency and the numa plugin already uses it. With default options it adds the init peak and there's no flag to drop just that. You can get a steady-state sum by handing it a pod copy whose init containers are filtered to sidecars, so the max term degenerates, but that reads as a trick. It also honours pod-level resources and adds Pod overhead unfiltered, and `getPodResourceRequest` does neither, so a straight swap would move the numbers in directions the scheduler wouldn't follow. This isn't a claim that the two implementations are in lockstep: they plainly aren't, and the shared aggregation path #1880 asks for is still the right end state. It's a claim that a stage-1 change shouldn't widen the gap.

- **Pod overhead and GPUs.** `getPodResourceRequest` adds only the `BaseResource` half of `pod.Spec.Overhead`, so a GPU in a RuntimeClass overhead is reserved as zero. This PR leaves it out of the status too, so the two agree. Upstream `PodRequests` and upstream ResourceQuota both count it, which suggests the scheduler side is the odd one; raised in #1894 rather than changed here.

- **No e2e test.** Covered by focused tests on the helper and on `calculateRequestedResources`, plus an end-to-end case through `handlePodGroupStatus` that also pins that a plain init container is *not* counted. Happy to add an e2e case asserting `Queue.status.allocated` for a sidecar pod if you'd like one.


pkg/apis/scheduling/v2/queue_types.go
enoodle, line 1: Please remove the updates to the doc string for all the fields. It is correct but I think that instead of repeating it 12 times we should write it once in a doc file.

pkg/podgroupcontroller/controllers/metadata/pod.go
```diff
@@ -16,6 +17,10 @@ import (
 	"github.com/kai-scheduler/KAI-scheduler/pkg/podgroupcontroller/controllers/resources"
 )
 
+// gpuResourceSuffix is how the queue accounting recognizes a GPU, matching getAllocatedGpus in the
+// queue-controller metrics.
+const gpuResourceSuffix = "/gpu"
```
enoodle: This is already defined somewhere in the project for this same exact purpose, please use that. If you need you can move this const to a central location.

pkg/podgroupcontroller/controllers/metadata/pod_test.go
```diff
@@ -9,14 +9,339 @@ import (
...
+		})
+	}
+}
+
+// TestCalculateRequestedResources_GpuAnnotations pins why a sidecar's GPU is left out. The scheduler rebuilds a
+// pod's GPU requirement from a GPU-sharing or legacy MIG annotation and discards the container request, so
+// adding a sidecar GPU on top would report a GPU that nothing reserved. Two of these pods reach a live cluster:
+// getFirstGPULimit only looks for nvidia.com/gpu limits, so an amd.com/gpu sidecar is admitted, and it never
+// looks at MIG annotations at all. The nvidia-on-fraction pods are rejected by admission today and are here as
+// defense in depth.
```
enoodle: We don't need to document current bugs in the test doc string.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
 
 ### Fixed
+- The PodGroup and Queue status now count native sidecars (`initContainers` with `restartPolicy: Always`) and Pod `Overhead`; only regular containers were summed before, so a sidecar's requests and a RuntimeClass overhead were missing from `PodGroup.status.resourcesStatus` and from `Queue.status.allocated`, `allocatedNonPreemptible` and `requested`. `kai_queue_allocated_cpu_cores` and `kai_queue_allocated_memory_bytes` rise accordingly for queues running sidecar or sandboxed workloads, and since they are the default input to the scheduler's time-based fairshare usage database they step once on upgrade for pods that are already running. `kai_queue_allocated_gpus` does not move: a resource the queue counts as a GPU is left out when it comes from a sidecar or an overhead. That, and the peak of a non-restartable init container, are tracked in #1880. [#1880](https://github.com/kai-scheduler/KAI-Scheduler/issues/1880) [thc1006](https://github.com/thc1006)
```
enoodle: Please make this one line a bit more readable for a human that wants to understand what was fixed.

pkg/podgroupcontroller/controllers/metadata/pod.go
```diff
@@ -86,13 +91,48 @@ func isPodScheduled(pod *v1.Pod) bool {
 	return false
 }
 
+// podSteadyStateResources sums what a pod holds concurrently while it runs: regular containers, native
+// sidecars (init containers with restartPolicy Always) and Pod overhead. The peak of a plain init container is
+// not included, and neither is a GPU asked for by a sidecar or by an overhead, see withoutGpuRequirements.
+func podSteadyStateResources(pod *v1.Pod) v1.ResourceList {
```
enoodle: The scheduler is not using exactly this algorithm to count the resources. Maybe we should import and reuse the code from the scheduler.

enoodle (changes requested):
Thank you for the fix!

pkg/apis/scheduling/v2/queue_types.go
thc1006: Done. Reverted the field comments in both `queue_types.go` and `podgroup_types.go`, and the regenerated CRD descriptions with them, so `pkg/apis` and the CRDs are back to being untouched by this PR.

The content now lives once under [Queue Status](https://github.com/NVIDIA/KAI-Scheduler/blob/fix/podgroup-status-sidecar-overhead/docs/queues/README.md#queue-status) in `docs/queues/README.md`, with a section on what the status counts and what it still leaves out.

pkg/podgroupcontroller/controllers/metadata/pod.go
thc1006: Done. It was `gpuResourceNameSuffix` in `pkg/queuecontroller/metrics`, private, which is why I could not reach it. Moved it to `constants.GpuResourceSuffix` and pointed both `getAllocatedGpus` and this filter at it, so the two share one definition of what counts as a GPU rather than each keeping its own.

`getAllocatedGpus` behaves exactly as before. If you flip the constant, the metrics tests and the status tests now fail together, which is the property I wanted out of the move.

pkg/podgroupcontroller/controllers/metadata/pod_test.go
thc1006: Fair. Trimmed to what the test actually pins: the annotation alone decides the pod GPU in the status, and a sidecar GPU request never adds to it.

CHANGELOG.md
thc1006: Rewrote it. It now says what changed, who was under-reporting, which metrics step on upgrade, and links to the doc section for the details. Let me know if it is still too dense.

pkg/podgroupcontroller/controllers/metadata/pod.go
thc1006: You are right that they do not match, and I should have said so in the code and not only in the PR description.

The scheduler charges `max(regular + sidecars, init peak) + overhead`, plus `pods: 1`. This charges `regular + sidecars + overhead` and drops a GPU coming from a sidecar or an overhead, so it is the steady-state part only.

I did look at importing `getPodResourceRequest` and backed off, for three reasons:

- It is not the whole algorithm. `NewTaskInfo` calls it and then `updatePodAdditionalFields` rewrites the GPU requirement from the `gpu-fraction` / `gpu-memory` / legacy MIG annotations and throws the container request away. Importing just the first half would make GPU-sharing pods worse, not better.
- It returns `*resource_info.ResourceRequirements`, which lifts GPUs into their own requirement and carries `pods: 1`. The status is a plain `v1.ResourceList` whose GPU arrives on separate paths (`ExtractGPUSharingReceivedResources`, DRA claims), so there is no straight conversion back.
- `podgroupcontroller` imports nothing from `pkg/scheduler` today, so this would be the first edge between them.

The two things reuse would actually change, the init peak and a sidecar GPU, are the two questions #1880 is still open on. That is why I kept this PR to the part that holds under either answer.

What I would rather do, and I suspect it is what you are really after, is lift the aggregation into `pkg/common/resources` and have both the scheduler and this controller call it. That is the end state #1880 describes. It does move the status numbers, so it is your call: I can do it here if you want it in one PR, or land this and send the extraction next.

enoodle: I now noticed that you droping all the GPUs from this status, was it intentional? Where do you plan to add them back?

Ideally this status will show the same numbers that the scheduler uses internally to expose to the user what KAI is using for the job, even if after the init containers finished running the job is using less resources.

thc1006: TL;DR: not all of them, only a GPU asked for by a native sidecar or set in the pod overhead. A regular container's GPU, a GPU-sharing request, and a DRA GPU all still show up in the status.

Here is where each one comes from in the change:

- a regular container's GPU stays in `podSteadyStateResources`, since the `withoutGpuResources` filter runs on the sidecar and overhead sums only, not on `pod.Spec.Containers`;
- a GPU-sharing request is added back by `ExtractGPUSharingReceivedResources` and `ExtractGPUSharingRequestedResources`;
- a DRA GPU is added back by `DRAGPUResourceListFromClaims`.

The one thing I drop on purpose is a sidecar's or an overhead's GPU. For a GPU-sharing pod the scheduler rebuilds the GPU requirement from the `gpu-fraction` / `gpu-memory` / legacy MIG annotation and discards the container request, so passing a sidecar GPU through would report a GPU that nothing actually reserved, and it would also shift the metric that feeds the fairshare usage database.

You are right that this still does not match the scheduler in two spots: the init-container peak, and that sidecar GPU. Those are the two open questions on #1880, so I kept this PR to the part that stays correct whichever way they are settled.

If you would rather the status track the scheduler exactly now, I think the right move is the one you raised first: lift the aggregation into `pkg/common/resources` and have both the scheduler and this controller call it. I am glad to do that in this PR, or to land this and send the extraction as the #1880 follow-up. Whichever you prefer.

davidLif: And why do you drop the gpus from overhead and sidecars for pod requesting non-shared gpus?

For shared gpus, you can look at the webhooks validations and see if we allow sidecar gpus for pods requesting fractions (I believe that this option is blocked, and then you don't need to handle it).

enoodle: @thc1006 - Those AI generated Issues and PRs are hard to go through, I couldn't go into 1880 yet. I think that in this PR and also generally for KAI it would be better to keep a uniform definition of allocated, and it should be similar to what is implemented in the scheduler service because all the decisions that it makes derive from that. The main goal of the status and metrics on the queues is to help explain the scheduler.

thc1006: Dear @enoodle,

You're both right. The status should mirror the scheduler's allocated, not a separate version of it, so I'll put the sidecar GPU back (the scheduler does count it) and stop rebuilding the number in the controller.

@davidLif, agreed on that one: `getPodResourceRequest` adds native sidecars, so dropping their GPU here under-reports a non-shared pod. Overhead GPU is the part the scheduler leaves out, so that stays dropped.

Quick context in case #1880 is still on your list: the controller was summing only regular containers, so a job with native sidecars or Pod overhead showed less than the scheduler actually reserved.

The real fix is a single shared calculation that both the scheduler and this controller call, which is the uniform definition you're describing. Where it should live, and whether this controller may depend on scheduler code, is your and the team's call. Let me know the shape you'd prefer and I'll build it, and I'm glad to hold this PR until then so it's done once.

And you're right about the length. I'll keep these shorter from here.

davidLif: Hello @thc1006 ,

You are right that we will need a shared calculation to reflect the pod requested/allocated resources. I would work on it separatly. The bug you found in the counting difference is important, and thank you for that. Because it is part of a more delicate point (expected VS received resources and quota consumption) i didn't want to do it as a part of #1880 . This PR is already imporing the situation, and if we can go with it under the current "separate functions" assumption, it would be great.

thc1006: Hi @davidLif,

Thank you, and thanks for taking on the shared calculation. Done: a native sidecar's GPU is counted now, since the scheduler counts it. I kept the drop only for pods whose GPU is rebuilt from an annotation (a fraction, gpu-memory, or a legacy MIG annotation), where the annotation decides.

One wrinkle on leaning on the webhook: admission rejects a whole nvidia.com/gpu next to a fraction, but it doesn't inspect MIG annotations, so a whole-GPU sidecar on a legacy-MIG pod is reachable. Removing the drop outright would double-count that one, so I scoped it instead. Overhead GPU stays out, since the scheduler adds only the base half of the overhead. Tests and the README are updated.

thc1006:
Hi @enoodle,

Fixed. The #1880 note now says it is about aligning the status with what the scheduler reserves, since the shared aggregation is a separate effort. I also switched the changelog to a `.changes/unreleased/` fragment and dropped the manual CHANGELOG.md edit, so the branch is no longer in conflict.

pkg/podgroupcontroller/controllers/metadata/pod.go
enoodle: It is still only counting the native sidecar init containers. should be `max(sum of init containers, normal container + native sidecars)`

thc1006: Hi @enoodle,

Done. `podReservedResources` now takes the per-resource max of the steady state and the init phase, plus overhead, matching `getPodResourceRequest`. I ported `initContainerEffects`, so the init phase is a running max (each non-restartable init plus the sidecars declared before it), not a flat sum. A GPU on such an init follows the sidecar rule, dropped only when the pod's GPU is rebuilt from an annotation. Tests and the README are updated.

thc1006:
@enoodle following up on the last thread: `podReservedResources` now returns `max(steady, init phase) + overhead`, matching `getPodResourceRequest`. The init phase is the running max of each non-restartable init plus the sidecars declared before it, not a flat sum. I rebased onto main and the podgroupcontroller tests are green. Could you take another look when you have time? Thanks for pushing on the scheduler alignment.

PR #1908, opened 2026-07-14
fix(scheduler): clear PodGroup Unschedulable condition after scheduling

### LINKED ISSUE #1888: PodGroup schedulingConditions not cleared after the workload schedules
## Description

When a podgroup pends and gets an `Unschedulable` scheduling condition, the condition stays in `status.schedulingConditions` forever - it is not cleared or updated when the workload later schedules successfully. Anyone reading the status API sees a running workload that still looks unschedulable.

The scheduler only touches scheduling conditions while the job has pending tasks (`RecordJobStatusEvent` in `pkg/scheduler/cache/status_updater/default_status_updater.go` - the unschedulable path is gated on `GetNumPendingTasks() > 0`). There is no code path that removes or resolves the condition once the job is scheduled.

## Reproduction

1. Create a leaf queue with CPU quota+limit of 500m.
2. Submit pod A (400m) to the queue - it schedules.
3. Submit pod B (400m) - it pends, and its podgroup gets an `Unschedulable` scheduling condition with quota details.
4. Delete pod A - pod B schedules and runs.
5. `kubectl get podgroup <pg-of-B> -o jsonpath='{.status.schedulingConditions}'` still shows the `Unschedulable` condition.

## Expected

The condition is removed (or marked resolved, e.g. `status: False`) once the podgroup schedules.

## Environment

Reproduced on current main (kind, default install).

Thezone-1:
## Description

A PodGroup's `UnschedulableOnNodePool` scheduling condition was never cleared once the workload actually scheduled. The condition is only written while the group has pending/gated tasks (in `RecordJobStatusEvent`), and there was no code path that removed it afterwards — so `status.schedulingConditions` kept an `Unschedulable` entry forever. Anyone reading the status API saw a running workload that still looked unschedulable.

This adds the missing path: when a pod group has no pending or gated tasks, `RecordJobStatusEvent` now clears the `UnschedulableOnNodePool` condition for the group's node pool. If there is nothing to clear, no status update is triggered, so normally-running pod groups are unaffected.

**Why remove the condition instead of setting `Status: False`?**
The set path already writes `Status: v1.ConditionTrue`, but the current consumers treat the *presence* of the condition as "unschedulable". Removing it is unambiguous for both presence-based and status-based readers and keeps the change minimal. Happy to switch to a `Status: False` "resolved" marker instead if maintainers prefer to preserve transition history — that would also mean keeping the `Status: True` on the set path for symmetry.

## Related Issues

Fixes #1888

## Checklist

- [x] Self-reviewed
- [ ] Updated documentation (if needed) — n/a, no user-facing API change

## Breaking Changes

None. Behavior only changes for pod groups that previously retained a stale `Unschedulable` condition after scheduling.

## Additional Notes
- New helper `removePodGroupSchedulingCondition` drops conditions for a given node pool and returns whether anything changed; `clearPodGroupSchedulingCondition` wires it into the scheduled path.
- Unit test `TestRemovePodGroupSchedulingCondition` covers: nothing to remove, removal on schedule, a different node pool left untouched, and only the matching node pool removed.
- Reproduction from the issue (submit A+B to a quota-limited queue, delete A so B schedules) is covered logically by the unit test; happy to add an envtest/e2e case for the full flow if you'd like it in this PR.


enoodle:
Maybe the binder should be responsible to remove this condition? The reason is that if the binding fails the pod will be left with no status to explain why it was pending for some time. @davidLif What do you think?

Thezone-1:
@enoodle, thanks — you're right that clearing it in the `else` branch happened at allocation time, before the bind was confirmed. If the bind then failed, the pod lost the explanation for why it was pending.

Thezone-1:
I've kept it in `RecordJobStatusEvent,` but now it only clears once the tasks are actually `Bound`/`Running`. While anything is still `Allocated`/`Pipelined`/`Binding` the condition stays, so a failed bind just drops back to `Pending` and gets re-reported — no gap, thus,  a pipelined group no longer loses its condition while it's still waiting.

I looked at moving it into the binder, but the binder doesn't touch PodGroup status at all today, so that's a much bigger change. Happy to go that route if you and @davidLif prefer it — just let me know.

pkg/scheduler/cache/status_updater/default_status_updater.go
```diff
@@ -597,6 +610,38 @@ func equalSchedulingConditions(a, b *enginev2alpha2.SchedulingCondition) bool {
...
+func hasTasksAwaitingBind(job *podgroup_info.PodGroupInfo) bool {
+	for _, status := range bindInFlightStatuses {
+		if len(job.PodStatusIndex[status]) > 0 {
+			return true
+		}
+	}
+	return false
+}
+
+func removePodGroupSchedulingCondition(podGroup *enginev2alpha2.PodGroup, nodePool string) bool {
```
davidLif: I think the function should clean all unscheduleable conditions, not just for a given nodepool

davidLif:
I think that the scheduler is the right place to change the status for the podgroup. The binder should only act on pods.

pkg/scheduler/cache/status_updater/default_status_updater.go
Thezone-1: fixed and pushed

Thezone-1:
> I think that the scheduler is the right place to change the status for the podgroup. The binder should only act on pods.

Agreed, I kept the status update in the scheduler. Pushed the follow-up, so it now clears all of the PodGroup's scheduling conditions once its tasks are bound, rather than only the current node pool's.

Thanks for the review

davidLif (dismissed):

davidLif:
@Thezone-1 can you please fix the changelog? When changed the way our changelog works, and this blocks the merge.

Thezone-1:
@davidLif  I have resolved the changelog. Thank you for reviewing the PR

davidLif (approved):

davidLif:
@Thezone-1 the commits are missing DCO signature

Thezone-1:
Rebased with --signoff - all four commits now carry DCO signature. Could you check and approve for the workflow to run and re-verify?

FYI @davidLif @enoodle

PR #1922, opened 2026-07-15
perf(scheduler): retain compact task fit errors

### LINKED ISSUE #1827: perf(scheduler): bound per-task per-node fit-error retention during allocation

enoodle:
## Description

Benchmark harness and baseline evidence come from the independent benchmark PR #1920.

This PR bounds scheduler memory retained for node-fit errors when many pending tasks fail across many candidate nodes.

Previously, each pending task retained a complete fit error for every rejected node, so retained memory scaled as `pending tasks × candidate nodes`. This PR retains aggregated reason counts per task instead, making retained task-fit-error memory scale as `pending tasks × unique reasons`.

It:

- Stores compact reason counts instead of per-node error objects.
- Preserves compact unschedulable messages, reason counts, ordering, and default messages.
- Recomputes complete per-node details from current-session pre-filters and filters only when detailed reporting or enabled V(6) logging requests them.
- Passes the detailed resolver as a normal function during status recording; no resolver state is retained.
- Preserves direct pre-filter errors without unnecessary node recomputation.
- Falls back to the compact message if detailed recomputation cannot produce diagnostics.
- Safely ignores empty and typed-nil fit diagnostics.
- Removes obsolete per-node fit-error storage APIs.

This standalone implementation is based directly on `main`. It preserves the existing `Session.FittingNode` API and immediate fit-error publication semantics; it does not depend on the allocation ownership changes in #1921.

Scheduling decisions, node-fit behavior, task transitions, and user-facing compact and detailed error output remain unchanged.

### 500-node full-cycle reclaim benchmark

| Metric | `origin/main` | This PR | Change |
|---|---:|---:|---:|
| Live heap after Allocate | 597.6 MB | 73.2 MB | -87.8% |
| Live heap after full cycle | 612.2 MB | 87.8 MB | -85.7% |
| Bytes/op | 14.13 GB | 13.93 GB | -1.4% |
| Allocations/op | 397.5 M | 397.4 M | effectively unchanged |
| Tasks with fit errors after Allocate | 4,000 | 4,000 | unchanged |

The unchanged fit-error task count verifies that the retained-memory reduction does not come from skipping failed-task diagnostics. Total allocation churn remains near `main`; the improvement is the amount of fit-error data retained across the scheduling cycle.

## Related Issues

Fixes #1827

Related to #1653

## Checklist

- [x] Self-reviewed
- [x] Added/updated tests
- [x] Added changie fragment

## Breaking Changes

No user-facing breaking changes.

The internal cache/status-updater fit-error path accepts a function that recomputes transient detailed node errors from the current scheduling session.

## Additional Notes

Validation performed:

- Focused common-info, podgroup-info, framework, and cache tests
- `go test ./pkg/scheduler/... -count=1`
- `make validate`
- `BenchmarkReclaimManySingleGPUJobsFullCycle_500Node` with `-benchmem -benchtime=1x -count=1`

Latest benchmark result:

- `73,165,728` bytes live after Allocate
- `87,770,120` bytes live after full cycle
- `13,930,262,256 B/op`
- `397,448,456 allocs/op`

The expensive 500-node benchmark remains outside the default CI benchmark regex.

### DIFF
--- a/pkg/scheduler/api/common_info/pod_errors.go
+++ b/pkg/scheduler/api/common_info/pod_errors.go
@@ -170,45 +170,73 @@ func (f *TasksFitError) Error() string {
 }
 
 type TasksFitErrors struct {
-	nodes map[string]*TasksFitError
-	err   string
+	reasonCounts map[string]int
+	err          string
 }
 
 func NewFitErrors() *TasksFitErrors {
-	f := new(TasksFitErrors)
-	f.nodes = make(map[string]*TasksFitError)
-	return f
+	return &TasksFitErrors{reasonCounts: make(map[string]int)}
 }
 
 func (f *TasksFitErrors) SetError(err string) {
 	f.err = err
 }
 
-func (f *TasksFitErrors) SetNodeError(nodeName string, err error) {
-	var fe *TasksFitError
-	switch obj := err.(type) {
-	case *TasksFitError:
-		obj.NodeName = nodeName
-		fe = obj
-	default:
-		fe = NewFitError("", "", nodeName, err.Error())
+func (f *TasksFitErrors) AddReasonCounts(errors *TasksFitErrors) {
+	if errors == nil {
+		return
 	}
-
-	f.nodes[nodeName] = fe
-}
-
-func (f *TasksFitErrors) AddNodeErrors(errors *TasksFitErrors) {
-	for nodeName, fitError := range errors.nodes {
-		f.nodes[nodeName] = fitError
+	if f.reasonCounts == nil {
+		f.reasonCounts = make(map[string]int)
+	}
+	for reason, count := range errors.reasonCounts {
+		f.reasonCounts[reason] += count
 	}
 }
 
-func (f *TasksFitErrors) DetailedError() string {
-	if f.err == "" {
-		f.err = ResourcesWereNotFoundMsg
+func (f *TasksFitErrors) AddNodeError(err error) {
+	if err == nil {
+		return
 	}
-	reasonMessages := []string{"\n" + f.err + "."}
-	for _, node := range f.nodes {
+	if f.reasonCounts == nil {
+		f.reasonCounts = make(map[string]int)
+	}
+	var reasons []string
+	if fitError, ok := err.(*TasksFitError); ok {
+		if fitError == nil {
+			return
+		}
+		reasons = fitError.Reasons
+	} else {
+		reasons = []string{err.Error()}
+	}
+	for _, reason := range reasons {
+		f.reasonCounts[reason]++
+	}
+}
+
+func (f *TasksFitErrors) HasNodeErrors() bool {
+	return len(f.reasonCounts) != 0
+}
+
+func (f *TasksFitErrors) ReasonCount(reason string) int {
+	return f.reasonCounts[reason]
+}
+
+func (f *TasksFitErrors) UniqueReasonCount() int {
+	return len(f.reasonCounts)
+}
+
+func (f *TasksFitErrors) DetailedError(nodeErrors []*TasksFitError) string {
+	baseError := f.err
+	if baseError == "" {
+		baseError = ResourcesWereNotFoundMsg
+	}
+	reasonMessages := []string{"\n" + baseError + "."}
+	if len(nodeErrors) == 0 {
+		return strings.Join(reasonMessages, "")
+	}
+	for _, node := range nodeErrors {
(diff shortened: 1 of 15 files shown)


pkg/scheduler/api/common_info/pod_errors.go
```diff
@@ -170,45 +170,73 @@ func (f *TasksFitError) Error() string {
...
-		f.nodes[nodeName] = fitError
+func (f *TasksFitErrors) AddNodeError(err error) {
+	if err == nil {
+		return
 	}
+	if f.reasonCounts == nil {
+		f.reasonCounts = make(map[string]int)
+	}
+	var reasons []string
+	if fitError, ok := err.(*TasksFitError); ok {
```
itsomri, line 205: Feels a bit weird to cast the err to TasksFitError and then take the reason from it - can't we just cast the err directly to string? Or even make the reasonCounts `map[err]int`?

itsomri: Or just continue to exposing the method arg as TasksFitErrors?

SiorMeir (approved):

pkg/scheduler/api/common_info/pod_errors.go
enoodle: I think that the map key can't be `error` because it is not an immutable type.

enoodle: I will look into your suggestion of removing this and always calling `Error`

enoodle: The problem with err.Error() is that it will add the node names, which will go against the idea of aggregating the errors together.
We also get `error` types from the predicates that we want to also conserve here so we can't change the function signature to TasksFitErrors

PR #1925, opened 2026-07-15
docs(gpu-sharing): fix hami-core install flags to match chart values

### LINKED ISSUE #1912: Cannot find parameter scheduler.gpuSharing.hamicoreEnabled in values.yaml file
I want to try HAMi (hami-core) with KAI Scheduler, does the documentation still valid https://github.com/kai-scheduler/KAI-Scheduler/tree/v0.16.4/docs/gpu-sharing/hami ? I cannot find this parameter 'scheduler.gpuSharing.hamicoreEnabled inside the values.yaml'

mesutoezdil:
the docs enable hami-core with `--set scheduler.gpuSharing.enabled=true --set scheduler.gpuSharing.hamicoreEnabled=true`, but those keys don't exist in the kai-scheduler chart (checked v0.16.0 through main), so helm silently ignores them and nothing gets enabled.

fix: use the real values `--set global.gpuSharing=true --set binder.plugins.hamicore.enabled=true`. verified with `helm template`.

fixes #1912

### DIFF
--- a/docs/gpu-sharing/hami/README.md
+++ b/docs/gpu-sharing/hami/README.md
@@ -41,8 +41,8 @@ KAI Scheduler's GPU sharing feature allows multiple pods to share a single GPU, 
 
   ```bash
   helm install kai-scheduler oci://ghcr.io/nvidia/kai-scheduler \
-    --set scheduler.gpuSharing.enabled=true \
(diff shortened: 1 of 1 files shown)


mesutoezdil:
cc @archlitchi @enoodle

enoodle (approved):

PR #1931, opened 2026-07-16
fix(scheduler): count shared DRA GPU device once per node

### LINKED ISSUE #1930: shared DRA ResourceClaim double-counted per pod makes node unschedulable
## What happened?

When two or more pods share a single physical GPU through one DRA `ResourceClaim`
(the claim's `status.reservedFor` lists more than one consumer — a normal
time-slicing / MPS setup), the scheduler counts that one physical device once
**per consuming pod** instead of once per device.

On a node whose shared device pushes the counted GPU usage above the node's
physical GPU count, the node's idle GPU count goes negative and the node becomes
unschedulable for **every** task — including pods that request no GPU at all. The
rejection surfaces as a misleading message, e.g.:

```
<node>: Node didn't have enough resources: GPUs, requested: 0, used: 2, capacity: 1
no nodes with enough resources were found: 1 node(s) didn't have enough resources: GPUs.
```

Concretely, on a single-GPU node with one claim shared by two pods, the scheduler
reports `used: 2, capacity: 1`, idle `-1`. A 0-GPU pod pinned to that node stays
`Pending` with the message above.

Root cause is in the node resource accounting: `addTaskResources` adds each
task's DRA GPU count from its allocated `ResourceClaimInfo` to `UsedVector`, and
nothing dedups a device that is already contributed by another pod on the same
node sharing the same claim. (`pkg/scheduler/api/node_info/node_info.go` →
`addTaskResources`; DRA GPU count comes from
`pkg/common/resources/dra.go` → `ExtractDRAGPUResourcesFromClaims`, which counts
per requested device with no cross-pod awareness.)

A physical GPU shared by several pods through one ResourceClaim should count
towards the node's used GPUs **once**, so idle never goes below zero and the node
stays schedulable for other workloads.

## How can we reproduce it?

1. On a single-GPU node, create one `ResourceClaim` requesting that GPU with a
   sharing strategy that allows multiple consumers (DRA time-slicing or MPS).
2. Schedule two pods that both reference the claim, so `status.reservedFor` holds
   both. Verify via scheduler logs (`schedulingshard` `spec.args.v: "7"`) that the
   node shows `used: 2, capacity: 1`, idle GPU `-1`.
3. Schedule any third pod (even one requesting 0 GPUs) with a nodeSelector onto
   that node — it stays `Pending` with "didn't have enough resources: GPUs".

## Environment

- Kubernetes version: 1.36
- KAI Scheduler version: v0.16.4
- DRA: NVIDIA DRA driver (`gpu.nvidia.com`), device-plugin disabled
  (`devicePlugin.enabled=false`); scheduling driven entirely by DRA
- Hardware: NVIDIA RTX PRO GPUs, one physical GPU shared by two pods via MPS

TensorRaya:
## Description

`addTaskResources` adds each task's DRA GPU count — taken from its allocated
`ResourceClaimInfo` — to the node `UsedVector`. When several pods share one
physical GPU through a single `ResourceClaim` (its `status.reservedFor` lists
more than one consumer, as in DRA time-slicing or MPS), each consuming pod
carries the same allocated device, so the device is counted once per pod. On a
node whose shared device pushes the counted usage above the node's physical GPU
count, `IdleVector` goes negative and the node is rejected for every task —
including GPU-less ones — via `MaxNodeResourcesPredicate`, surfacing as a
misleading "didn't have enough resources: GPUs" message (e.g. on a 1-GPU node
shared by two pods: `used: 2, capacity: 1`, idle `-1`).

Fix: track a per-device reference count on the node, keyed by the allocated
device's `driver/pool/device` tuple. `addTaskResources` subtracts from the
tracked GPU vector any device already referenced by another pod on the node, so
a shared physical device contributes to `UsedVector` exactly once;
`removeTaskResources` mirrors this, keeping the device counted while any consumer
remains and releasing it only when the last one leaves. The reference count
lives on `GpuSharingNodeInfo` (cloned with the rest of the node's sharing state).
Only NVIDIA GPU DRA devices (`driver == gpu.nvidia.com`) are tracked; other DRA
devices are untouched.

Non-DRA and single-consumer DRA GPU accounting is unchanged: the dedup only ever
skips a device that a prior pod on the same node already contributed.

Tests (`pkg/scheduler/api/node_info/dra_shared_device_info_test.go`), exercising
the real `AddTask`/`RemoveTask` accounting path:
- Two pods sharing one claim on a 1-GPU node → used GPUs `1`, idle `0` (not
  `2`/`-1`).
- Add/remove symmetry: used stays `1` while any consumer of the shared device
  remains, returns to `0` when the last leaves.
- Two pods on two distinct physical GPUs (own claim each) → both counted (`2`),
  guarding against over-dedup.

Verified the shared-claim tests fail without the fix (reproducing `used: 2`,
idle `-1`) and pass with it; the distinct-devices test stays green either way.
Full `./pkg/scheduler/...` suite passes.

## Related Issues

Fixes #1930

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. Accounting changes only in the previously-broken case of a physical DRA
device shared by more than one pod on a node; exclusive claims and non-DRA GPUs
are counted exactly as before.

## Additional Notes

CHANGELOG.md updated under `[Unreleased] > Fixed`.

Validated with:
`go test ./pkg/scheduler/api/node_info/ ./pkg/common/resources/` and the full
`go test ./pkg/scheduler/...` suite (77 packages ok). `gofmt` and `go vet` clean
on all touched files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected GPU resource accounting when multiple pods share the same physical GPU through a single resource claim.
  * Prevented shared GPUs from being counted multiple times, which could incorrectly make nodes appear unschedulable.
  * Improved GPU usage tracking as shared GPU consumers are added and removed.
* **Changelog**
  * Added release notes documenting the GPU accounting fix and recent version updates.

### DIFF
--- a/pkg/scheduler/api/node_info/dra_shared_device_info.go
+++ b/pkg/scheduler/api/node_info/dra_shared_device_info.go
@@ -1,0 +1,120 @@
+// Copyright 2025 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+package node_info
+
+import (
+	resourceapi "k8s.io/api/resource/v1"
+
+	"github.com/kai-scheduler/KAI-scheduler/pkg/common/resources"
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/api/pod_info"
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/api/resource_info"
+)
+
+// draDeviceKey uniquely identifies a physical DRA device on the node.
+func draDeviceKey(result resourceapi.DeviceRequestAllocationResult) string {
+	return result.Driver + "/" + result.Pool + "/" + result.Device
+}
+
+// allocatedGPUDeviceKeys returns the keys of all GPU devices allocated to the
+// task via DRA ResourceClaims. Non-GPU devices are ignored: they are not part
+// of the GPU accounting that this dedup protects.
+func (ni *NodeInfo) allocatedGPUDeviceKeys(task *pod_info.PodInfo) []string {
+	var keys []string
+	for _, claimAllocation := range task.ResourceClaimInfo {
+		if claimAllocation == nil || claimAllocation.Allocation == nil {
+			continue
+		}
+		for _, result := range claimAllocation.Allocation.Devices.Results {
+			if !resources.IsGPUDeviceClass(result.Driver) {
+				continue
+			}
+			keys = append(keys, draDeviceKey(result))
+		}
+	}
+	return keys
+}
+
+// sharedDRAGpuDiscount returns the number of GPU devices the task requests via
+// DRA claims that are already counted on this node for other pods. A task that
+// shares an allocated device with a running pod does not need additional GPU
+// capacity for that device.
+func (ni *NodeInfo) sharedDRAGpuDiscount(task *pod_info.PodInfo) float64 {
+	discount := 0.0
+	for _, key := range ni.allocatedGPUDeviceKeys(task) {
+		if ni.DRASharedDeviceRefCount[key] > 0 {
+			discount++
+		}
+	}
+	return discount
+}
+
+// dedupSharedDRAGpus removes from resourcesToTrack the GPU count that would
+// double-count physical DRA devices already referenced by other pods on the
+// node. It also updates the node's per-device reference count. It must be
+// called once per addTaskResources, before the vector is added to UsedVector.
+func (ni *NodeInfo) dedupSharedDRAGpus(task *pod_info.PodInfo, resourcesToTrack resource_info.ResourceVector) {
+	current := resourcesToTrack.Get(resource_info.GPUIndex)
+	if current <= 0 {
+		// The task contributes no GPUs to the used vector (e.g. a resource
+		// reservation task whose GPU index was zeroed). Tracking its devices
+		// would both risk a negative deduction below and mask the reference
+		// count of the real consuming pods, so leave the accounting untouched.
+		return
+	}
+
+	alreadyCounted := 0.0
+	for _, key := range ni.allocatedGPUDeviceKeys(task) {
+		if ni.DRASharedDeviceRefCount[key] > 0 {
+			// Another pod on this node already contributed this physical
+			// device to the used vector: do not count it again.
+			alreadyCounted++
+		}
+		ni.DRASharedDeviceRefCount[key]++
+	}
+
+	if alreadyCounted > current {
+		// Never deduct more than the task's own GPU contribution.
+		alreadyCounted = current
+	}
+	if alreadyCounted > 0 {
+		resourcesToTrack.Set(resource_info.GPUIndex, current-alreadyCounted)
+	}
+}
+
+// releaseSharedDRAGpus is the inverse of dedupSharedDRAGpus: it decrements the
+// per-device reference count and adds back the GPU count for devices that
+// remain referenced by other pods (and were therefore never subtracted on this
+// task's removal path). It must be called once per removeTaskResources.
+func (ni *NodeInfo) releaseSharedDRAGpus(task *pod_info.PodInfo, resourcesToTrack resource_info.ResourceVector) {
+	current := resourcesToTrack.Get(resource_info.GPUIndex)
+	if current <= 0 {
+		// Mirror of dedupSharedDRAGpus: a task that contributed no GPUs never
+		// incremented the reference count, so it must not decrement it here.
+		return
+	}
+
+	stillShared := 0.0
+	for _, key := range ni.allocatedGPUDeviceKeys(task) {
+		if ni.DRASharedDeviceRefCount[key] > 1 {
+			// The device stays referenced by another pod after this removal:
+			// it must remain in the used vector, so this task's removal must
+			// not subtract it.
+			stillShared++
+		}
+		if ni.DRASharedDeviceRefCount[key] > 0 {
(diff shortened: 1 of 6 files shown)


TensorRaya:
@coderabbitai review

.changes/unreleased/0029-fixed.yaml
enoodle, line 1: please generate this using `make changelog` so that the file name will have a timestamp which will prevent collisions.

pkg/scheduler/api/node_info/dra_shared_device_info.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	return keys
+}
+
+// isGPUDRADriver reports whether the DRA driver name belongs to an NVIDIA GPU
+// device. The device-class name is not present on the allocation result, so the
+// driver name is used instead.
+func isGPUDRADriver(driver string) bool {
+	return driver == commonconstants.NvidiaGpuDraDriver
+}
```
enoodle: There is already a function for that, we don't need this constant either: https://github.com/kai-scheduler/KAI-Scheduler/blob/main/pkg/common/resources/dra.go#L149

pkg/scheduler/api/node_info/dra_shared_device_info.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+// A single physical DRA device may be shared by several pods through one
+// ResourceClaim with more than one entry in status.reservedFor. Each such pod
+// carries the same allocated device in its ResourceClaimInfo, so counting the
+// device once per pod (the naive per-task accounting) inflates the node's used
+// GPU count above physical capacity and drives IdleVector negative, making the
+// whole node unschedulable. draSharedDeviceRefCount tracks how many pods on the
+// node currently reference each allocated device (keyed by driver/pool/device)
+// so a shared device contributes to UsedVector exactly once.
```
enoodle: please remove this - we don't need to the describe the issue in the code.

enoodle (commented):
I created 2 allocate action tests to verify this PR. one of them fails because of the mismatch between the test GPU driver name and the one tested here (see comment) But I suspect that the one where one of the pods sharing the DRA claims is still pending will fail after your PR is fixed. Can you take a look at this after you fix the gpu driver detection ?
The tests for reference: https://github.com/kai-scheduler/KAI-Scheduler/compare/main...enoodle:KAI-Scheduler:review-1931-test-cases

TensorRaya:
Thanks for writing the test cases, that was a big help. I cherry-picked your commit into the branch (kept your authorship) and addressed everything:

- **GPU driver detection**: dropped the hardcoded constant and switched to the existing `resources.IsGPUDeviceClass`, so the detection matches the rest of the codebase. Your first test passes now.
- **Pending pod sharing an allocated device**: you were right, it failed after the driver fix. The accounting fix alone wasn't enough — a later pod sharing an already-counted device still requested a full GPU against an idle count of 0 and was rejected by `lessEqualTaskToNodeResources`. Added `sharedDRAGpuDiscount`: devices already referenced on the node are discounted from the task's GPU requirement during the fit check. Both of your tests pass, and I verified they go red with the fix neutralized.
- Removed the issue-description comment block from the code.
- Regenerated the changelog fragment via `make changelog` (timestamped filename).

`make validate`-equivalent ran clean locally (gofmt, go vet, golangci-lint, full `./pkg/scheduler/...` test suite).

enoodle:
Hi @TensorRaya sorry for the delay,
I want to merge it now but it has a minor conflict (I think should just be both with the dedup first). Do you want to fix it of should I ?

TensorRaya:
I did it — rebased on current main, PR is mergeable again.

The conflict was in `addTaskResources`/`removeTaskResources`, against the new DRA-backed extended resources work (#1943). Resolved with the dedup first, then the allocatable-zeroing loop, as you suggested: `dedupSharedDRAGpus` needs the real per-device counts to decide whether another pod on the node already contributed the device — running it after the zeroing would hand it a vector where DRA resources are already 0. `removeTaskResources` mirrors the same order.

Since #1943 touches the same accounting path, I re-ran the counter-proof instead of trusting a green suite: with `dedupSharedDRAGpus`/`releaseSharedDRAGpus` neutralized, `TestAddTask_SharedDRAClaimCountedOnce`, `TestAddRemoveTask_SharedDRAClaimSymmetry` and your two allocate-action tests all fail; restored, all pass. So the fix still does work on top of main, it isn't subsumed by the extended-resources change.

Gates run locally against main: `go build ./pkg/scheduler/...`, `go vet`, `gofmt`, golangci-lint 2.11.3 (0 issues), and the node_info + allocate integration suites. Your test commit is untouched and still attributed to you.

pkg/scheduler/api/node_info/dra_shared_device_info.go
```diff
@@ -0,0 +1,100 @@
...
+		if ni.DRASharedDeviceRefCount[key] == 0 {
+			delete(ni.DRASharedDeviceRefCount, key)
+		}
+	}
+
+	if stillShared > 0 {
+		current := resourcesToTrack.Get(resource_info.GPUIndex)
+		resourcesToTrack.Set(resource_info.GPUIndex, current-stillShared)
+	}
+}
```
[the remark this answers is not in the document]
enoodle: we currently don't create resource reservation pods with DRA claims.

enoodle (approved):

gshaibi (approved):
LGTM. Thanks!

Issue #1615, opened 2026-05-23
Add scheduling gate on Pods to avoid Karpenter scaling up for out-of-limit Pods


RRethy:
### What you would like to be added?

Scheduling gate support for queue admission — when pods cannot be scheduled due to queue quota/capacity limits (as opposed to insufficient node resources), they should be held with a scheduling gate rather than marked as Unschedulable.

This is the same pattern implemented by Volcano and Kueue.

### Why is this needed?

KAI currently marks all unallocated pods as Unschedulable regardless of the reason they can't be scheduled. Cluster autoscalers (Karpenter) watch for this to trigger node provisioning.

This means pods blocked by queue limits will trigger scale ups even if they will never be admitted.

With scheduling gates, Karpenter won't scale up unnecessarily.

RRethy:
Volcano, https://github.com/volcano-sh/volcano/blob/master/docs/design/scheduling-gates-queue-admission.md, does this right now with a sane implementation. The only downside is the head of line blocking for gated pods due to the reservation mechanism, but it's a reasonable tradeoff and could be improved in the future.

RRethy:
I spent some time thinking about a solution that doesn't have the same drawbacks as the Volcano approach. This is what I came up with that correctly support scheduling gates and has Karpenter only scale up when it needs to.

1. Admission webhook adds a `kai.scheduler/queue-admission` scheduling gate.
2. In the scheduling loop, KAI evaluates podgroups normally if the pods have either no scheduling gate or only the KAI scheduling gate. Any other scheduling gates and they're treated as gated (existing behaviour).
3. At the end of the scheduling loop in session close, KAI looks at unscheduled podgroups that are quota-admissible (`IsJobOverQueueCapacityFn` if I understand correctly) and greedily fills up the quota by priority. These podgroups are never scheduled to any nodes though (since Karpenter needs to create nodes for them still).
4. KAI creates dummy pods to mimic the pods in the podgroups that are quota-admissible. These pods go into a special namespace `kai-reservation-pods` and are marked as unschedulable. Any dummy pods from previous loops that are no longer relevant are cleaned up. These dummy pods are solely there for Karpenter to see unschedulable pods; they never get bound to a node.
5. Karpenter scales up solely based on the podgroups that are quota-admissible.
6. In a future scheduling loop, KAI assigns the pods to the new nodes (and the old dummy pods get cleaned up).

The scheduling gates are only removed by the binder asynchronously.

[...]

PR #1932, opened 2026-07-16
docs(docs): add queue admission gating design proposal

shaq918:
## Description

Design proposal for holding queue-quota-blocked pods with a Kubernetes scheduling gate (`kai.scheduler/queue-admission`), so autoscalers stop provisioning nodes for pods that will never be admitted, and scheduling sessions scale with cluster capacity rather than submission volume.

Highlights of the proposed design:

- **The gate caches the existing quota verdict — enforcement never moves.** A pod stays gated iff `IsJobOverQueueCapacityFn` would reject its podgroup. Webhook outage, manual gate removal, or disabling the feature all degrade to today's exact behavior.
- **Gate addition is dumb, gate removal is smart.** An admission plugin adds the gate unconditionally for opted-in queues (no quota computation at admission time — queue usage isn't reliably knowable there). A new terminal `ungate` scheduler action removes gates for podgroups the existing `capacity_policy` oracle admits, greedily in queue/job fairness order with cumulative accounting.
- **No reservation/dummy pods.** Ungating is decided on quota admissibility, not node fit — an ungated, capacity-starved pod becomes a real Unschedulable pending pod, which is the autoscaler's native signal with true resources and affinity. `node-scale-adjuster` is untouched.
- **Cross-session soundness via a Reserved bucket**: pending ungated demand reserves quota, so admissions across sessions never collectively exceed quota. Fully reconstructible from cluster state.
- **Fair-share parity**: gated demand counts into queue `Request` like pending demand, so fair-share division is identical with and without gating.
- Strictly opt-in (global Config toggle fanned out by the operator + per-queue label), gang-atomic admission decisions with guaranteed convergence for torn gangs, and defined failure-mode behavior (see table in the doc).

Open questions are listed at the end of the doc — feedback welcome on those in particular (opt-in surface, naming, ungate hysteresis, orphaned-pod escape hatch).

## Related Issues

Relates to #1615 (design for it; implementation to follow in separate PRs)

## Checklist

- [x] Self-reviewed
- [ ] Added/updated tests (if needed) — N/A, design doc only
- [ ] Changelog fragment — N/A, `skip-changelog` (docs only)

## Breaking Changes

None — design doc only. The proposed feature is opt-in, default off.

## Additional Notes

Happy to bring this to the community call for discussion.

## Summary by CodeRabbit

* **Documentation**
  * Added a design document for opt-in queue admission gating in KAI Scheduler.
  * Documented scheduling gate behavior, quota reservation, autoscaler signaling, gang and elastic workloads, preemption, and fair-share interactions.
  * Outlined observability expectations, failure handling, rollback behavior, and open design questions.

### DIFF
--- a/docs/developer/designs/queue-admission-gating/README.md
+++ b/docs/developer/designs/queue-admission-gating/README.md
@@ -1,0 +1,149 @@
+# Queue Admission Gating
+
+*Status: Proposed*
+
+Related issues: [#1615](https://github.com/kai-scheduler/KAI-Scheduler/issues/1615) (request), [#1783](https://github.com/kai-scheduler/KAI-Scheduler/issues/1783) (quota-change semantics — interacts, see below)
+
+## Motivation
+
+Queue quota and limits are enforced only at allocation time (`capacity_policy`). Pods blocked purely by queue policy are still created, marked `Unschedulable`, and re-examined every session. Two costs follow:
+
+1. **Autoscaler waste.** Cluster autoscalers (Karpenter, CA) treat `Unschedulable` pods as demand and provision nodes for workloads that will never be admitted. On GPU instance types this is expensive, and the nodes sit idle until scale-down. Volcano and Kueue both hold quota-blocked pods back from autoscaler visibility; KAI does not.
+2. **Session action cost scales with submissions, not capacity.** Fully gated podgroups are excluded from the input of every action, so job ordering, allocation attempts, and — most significantly — preempt/reclaim scenario generation shrink to the schedulable set. Scope of this claim: **snapshot construction is unchanged** — PodInfos/PodGroupInfos are still built for all pods, gated or not, because the `ungate` action evaluates gated podgroups from the same session snapshot. Snapshot cost stays O(pods submitted), identical to today; the reduction applies to what the actions do with the snapshot, where the expensive work lives. Scheduling signatures already deduplicate repeated *scheduling attempts*, but do not remove over-quota podgroups from job ordering or scenario search, and do not hide pods from autoscalers. (A lighter snapshot representation for gated pods — quota math needs only their resource requests and gate state, not node-fit data — is a possible future optimization, out of scope here.)
+
+The proposal: hold quota-blocked pods with a Kubernetes scheduling gate. Gated pods are invisible to autoscalers (no `Unschedulable` condition is ever written) and are already excluded from session readiness by existing machinery — both improvements reuse mechanisms Kubernetes and KAI already have.
+
+**Core principle: the gate is a cache of the existing quota verdict, never a new enforcement mechanism.** Allocation-time enforcement is unchanged and remains authoritative. Two one-way guarantees define the feature:
+
+1. **Safety**: the gate is removed only after an affirmative admission decision by the quota oracle.
+2. **Liveness**: every gated podgroup is re-evaluated each session and ungated once admissible.
+
+The absence of a gate implies nothing — fail-open paths exist by design (webhook outage, manual gate removal, feature disablement), and all of them degrade to exactly today's behavior, never to a quota violation.
+
+## API
+
+Gate on the pod:
+
+```yaml
+spec:
+  schedulingGates:
+    - name: kai.scheduler/queue-admission
+```
+
+Opt-in per queue via a field on the Queue spec:
+
+```go
+type QueueSpec struct {
+    // AdmissionGating enables queue-admission scheduling gates for pods in this queue.
+    // +optional
+    AdmissionGating *bool `json:"admissionGating,omitempty"`
+}
+```
+
+Global feature toggle in the Config CR, default off. The operator fans one knob out to both sides — the admission flag and the scheduler flag plus the `ungate` action — so webhook and scheduler cannot disagree across upgrades:
+
+```yaml
+spec:
+  global:
+    queueAdmissionGating:
+      enabled: false
+```
+
+**The two switches compose as master-switch + per-queue opt-in, not as a default/override pair.** The global toggle is the prerequisite: it installs and activates the admission plugin and the `ungate` action, so nothing is gated anywhere while it is off, regardless of per-queue settings. When it is on, `QueueSpec.AdmissionGating` selects which queues are gated — a queue is gated iff *(global enabled ∧ that queue opts in)*, with one deliberate exception for pods whose queue does not exist at admission (see below). This keeps the webhook out of the pod-create path entirely until an operator turns the feature on cluster-wide, and avoids a per-queue field that silently does nothing when the machinery is not installed.
+
+## Semantics
+
+### Gate addition — deliberately dumb
+
+A new mutating-admission plugin in the existing pod-mutator chain appends the gate at pod CREATE (gates are create-only in Kubernetes) to pods with `schedulerName: kai-scheduler` whose queue opts in. The webhook performs **no quota computation** — queue usage is not reliably knowable at admission time (pods may be created before anything in the queue has been scheduled, and the PodGroup may not exist yet). All quota intelligence stays in the scheduler, evaluated on a consistent session snapshot. Cost: admissible pods in opted-in queues pay one extra scheduling cycle before allocation.
+
+The opt-in rule has one exception: if the pod's queue does not exist at admission time, the opt-in field cannot be read, and the gate is added anyway — a pod referencing no real queue can never schedule, and leaving it visible to autoscalers would recreate the over-provisioning problem. Removal happens in `ungate` if/when the queue exists and opts in. Symmetrically, pods found gated for an existing queue that has **not** opted in are ungated unconditionally by the action — the same self-healing path as feature disablement.
+
+### Gate removal — the `ungate` action
+
+A new scheduler action, appended last in the action list so it observes end-of-session queue state (including this session's allocations and evictions). Per session:
+
+1. Collect podgroups with gated tasks whose **only** gate is the KAI gate. Any foreign gate present → leave the pod untouched (its gate owner keeps control).
+2. Order podgroups with the existing queue/job ordering, so fairness ordering is preserved. Podgroups with partially removed gates (see gang semantics) are admitted first.
+3. Greedily admit against the quota oracle (below), committing each admission into simulated queue shares before evaluating the next — N admissions never collectively exceed quota.
+4. Admitted → enqueue gate-removal patches for the admitted pod set through the scheduler's existing async pod-patch queue (remove only the KAI gate entry, conflict-retried). Not admitted → record the quota reason on the podgroup (conditions below).
+
+Gated pods already flow correctly through the scheduler: they map to the `Gated` task status, are excluded from readiness (`alive − gated ≥ minAvailable`) and from allocation candidates, and all-gated podgroups are filtered out of the input of allocate/preempt/reclaim/consolidation. The `ungate` action is exempt from that readiness filter — it builds its own input by iterating podgroups with KAI-gated tasks directly from the session snapshot; otherwise fully gated podgroups would never reach it. This is also why snapshot construction cannot shrink (see Motivation): gated pods must be present in the snapshot for the ungate decision. The action-cost reduction for the other actions reuses the existing readiness path; no new scheduler fast-path is introduced.
+
+### Quota oracle and the Reserved bucket
+
+Admissibility mirrors the two existing capacity checks (non-preemptible over deserved quota; anyone over limit), with one addition. Quota is consumed only at allocation, so a podgroup ungated in session N that cannot yet allocate (e.g. waiting for node provisioning) would consume nothing in session N+1, and a second podgroup could be admitted against the same quota. To close this, pending **ungated** demand is counted as a `Reserved` bucket in queue attributes:
+
+- non-preemptible: admissible iff `Deserved ≥ AllocatedNotPreemptible + ReservedNotPreemptible + requested`
+- limit: admissible iff `MaxAllowed ≥ Allocated + Reserved + requested`
+
+Both checks walk the queue hierarchy exactly as `capacity_policy` does today: `Reserved` aggregates up parent queues the same way `Allocated` does, and the greedy pass uses the same hierarchical queue ordering — admission ordering and accounting see precisely what the capacity checks see.
+
+Reserved is recomputed from cluster state each session ("admitted" ≡ pending without the KAI gate) — no persisted or in-memory-only state, so a scheduler restart loses nothing.
+
+Reserved is an autoscaler-fidelity optimization, not a quota-correctness mechanism — allocation-time enforcement remains authoritative. This permits **bounded reservation**: a pod that remains pending releases its reservation after an expiry. Without it, a single ungated workload that can never schedule for a non-quota reason (unsatisfiable affinity, pending PVC) would hold its reservation forever and head-of-line-block every other gated workload in the queue — a regression against today, where a broken pending pod blocks nothing. After expiry the next podgroup in order can be admitted; if both later become runnable they arbitrate at the allocation-time capacity check, which is exactly today's behavior.
+
+The expiry is a single duration configured in global settings (`spec.global.queueAdmissionGating.reservationTimeout`), not a per-queue knob — it is a coarse "how long to trust a reservation while a node is provisioning" value, and the tradeoff (too short → reservations lapse before scale-up completes and quota is briefly double-offered, arbitrated correctly at allocation; too long → a stuck workload blocks its queue for that duration) does not vary meaningfully per queue. The expiry anchor is a `kai.scheduler/queue-admitted-timestamp` annotation stamped on the pod in the same patch that removes the gate — reconstructible from cluster state, mirroring the existing `last-start-timestamp` pattern.
+
+### Autoscaler signal — no reservation pods
+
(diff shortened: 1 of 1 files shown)


shaq918:
@coderabbitai review

docs/developer/designs/queue-admission-gating/README.md
```diff
@@ -0,0 +1,130 @@
...
+| D6 | Pending ungated demand reserves quota (Reserved bucket) so cross-session admissions never collectively exceed quota |
+| D7 | Gated demand counts into queue `Request` — fair-share division identical with and without gating |
+| D8 | Strictly opt-in: global Config toggle (default off) fanned out by the operator to webhook and scheduler together, plus per-queue label |
+| D9 | Foreign gates are never touched; pods with any non-KAI gate are ignored by the ungate action |
+
+## Open Questions
+
+1. Queue opt-in surface: label (no CRD churn) vs `QueueSpec` field?
+2. Gate and action naming: `kai.scheduler/queue-admission` / `ungate` vs alternatives (`admit`)?
+3. Behavior when the pod's queue does not exist at admission: gate (safe for autoscalers, but hides typo'd queues) or skip (proposed default)?
```
enoodle: add gated - it will only be removed in "ungate" if/when the queue exists.

docs/developer/designs/queue-admission-gating/README.md
```diff
@@ -0,0 +1,130 @@
...
+4. Ungate hysteresis or rate-limiting after a large quota increase (thundering herd of ungates)?
+5. Max-gate-age escape hatch for orphaned gated pods (queue deleted, podgroup never created)?
+6. Should hierarchical (parent-queue) limits factor into the greedy admission order differently than leaf quota?
```
enoodle: if `IsJobOverQueueCapacityFn` is looking at these, when so should the action order  / accounting.

docs/developer/designs/queue-admission-gating/README.md
```diff
@@ -0,0 +1,130 @@
...
+7. Interaction with usage-based / time-based fairness: should gated time be visible to usage history?
```
enoodle: its not relevant because we are ungating based on limits, and this effects fair share.

docs/developer/designs/queue-admission-gating/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Gate removal — the `ungate` action
+
+A new scheduler action, appended last in the action list so it observes end-of-session queue state (including this session's allocations and evictions). Per session:
+
+1. Collect podgroups with gated tasks whose **only** gate is the KAI gate. Any foreign gate present → leave the pod untouched (its gate owner keeps control).
+2. Order podgroups with the existing queue/job ordering, so fairness ordering is preserved. Podgroups with partially removed gates (see gang semantics) are admitted first.
+3. Greedily admit against the quota oracle (below), committing each admission into simulated queue shares before evaluating the next — N admissions never collectively exceed quota.
```
enoodle, line 68: What if we ungate a job that can't be scheduled because of some unrelated reason (missing config map for example) - Will this job then break the queue and block it from using its full limit?

enoodle (commented):
Super interesting, I think the whole thing should be opt it (one flag that the operator will use to enable this in the  admission webhook and enable the ungate action in the scheduler)
I have one question about ungate decision and suggestions to the open questions.

docs/developer/designs/queue-admission-gating/README.md
shaq918: Good catch — as originally written, that pod's reservation would hold forever and head-of-line-block every other gated workload in the queue, which is a regression against today (a broken pending pod currently blocks nobody). Addressed with **bounded reservation** (now D12): Reserved is an autoscaler-fidelity optimization, not a quota-correctness mechanism — the allocation-time capacity check stays authoritative — so a pod that remains pending releases its reservation after a configurable expiry. After expiry the next podgroup in order can be admitted; if both later become runnable they arbitrate at the allocation-time check, which is exactly today's behavior. The expiry anchor is a `kai.scheduler/queue-admitted-timestamp` annotation stamped in the same patch that removes the gate (reconstructible from cluster state, mirroring the `last-start-timestamp` pattern). Default value/config surface is a new open question.

(Aside: the specific missing-ConfigMap case actually binds and gets stuck in ContainerCreating, so it consumes Allocated — same as today. The general class — unsatisfiable affinity, pending PVC — is the real one, and the expiry covers it.)

docs/developer/designs/queue-admission-gating/README.md
shaq918: Adopted — now D10: the gate is added even when the queue doesn't exist at admission; `ungate` releases it if/when the queue exists and the podgroup is admissible. Since opt-in can't be checked without a Queue CR, the complement is also specified: pods found gated for an existing queue that has *not* opted in are ungated unconditionally (same self-healing path as feature disablement).

docs/developer/designs/queue-admission-gating/README.md
shaq918: Agreed — made explicit (D11): Reserved aggregates up parent queues exactly as Allocated does, and the greedy pass uses the same hierarchical queue ordering, so admission ordering and accounting see precisely what the capacity checks see.

docs/developer/designs/queue-admission-gating/README.md
shaq918: Dropped the open question.

docs/developer/designs/queue-admission-gating/README.md
```diff
@@ -0,0 +1,130 @@
...
+2. **Scheduler load scales with submissions, not capacity.** Every session's snapshot, podgroup construction, and job ordering carry all submitted pods — O(pods submitted), not O(pods schedulable). A single tenant submitting far beyond quota degrades session latency for every queue. Scheduling signatures deduplicate repeated *scheduling attempts* within a session, but do not shrink the snapshot or the job-order heap, and do not hide pods from autoscalers.
+
+The proposal: hold quota-blocked pods with a Kubernetes scheduling gate. Gated pods are invisible to autoscalers (no `Unschedulable` condition is ever written) and are already excluded from session readiness by existing machinery, so both costs disappear using mechanisms Kubernetes and KAI already have.
+
+**Core principle: the gate is a cache of the existing quota verdict, never a new enforcement mechanism.** Allocation-time enforcement is unchanged. A pod stays gated iff `IsJobOverQueueCapacityFn` would reject its podgroup today. Webhook outage, manual gate removal, or disabling the feature all degrade to exactly today's behavior — never to a quota violation.
```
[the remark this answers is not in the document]
shaq918: Fixed — restated as two one-way guarantees (safety: gate removed only after an affirmative admission decision; liveness: every gated podgroup re-evaluated each session) with an explicit note that absence of a gate implies nothing and allocation-time enforcement is authoritative. D1 updated to match.

docs/developer/designs/queue-admission-gating/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+1. Collect podgroups with gated tasks whose **only** gate is the KAI gate. Any foreign gate present → leave the pod untouched (its gate owner keeps control).
+2. Order podgroups with the existing queue/job ordering, so fairness ordering is preserved. Podgroups with partially removed gates (see gang semantics) are admitted first.
+3. Greedily admit against the quota oracle (below), committing each admission into simulated queue shares before evaluating the next — N admissions never collectively exceed quota.
+4. Admitted → enqueue gate-removal patches for the admitted pod set through the scheduler's existing async pod-patch queue (remove only the KAI gate entry, conflict-retried). Not admitted → record the quota reason on the podgroup (conditions below).
+
+Gated pods already flow correctly through the scheduler: they map to the `Gated` task status, are excluded from readiness (`alive − gated ≥ minAvailable`) and from allocation candidates, and all-gated podgroups are filtered before actions run. The O(schedulable) win reuses this path; no new scheduler fast-path is introduced.
```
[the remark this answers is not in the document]
shaq918: Fixed — the doc now states `ungate` is exempt from the readiness filter and builds its own input by iterating podgroups with KAI-gated tasks directly from the session snapshot.

docs/developer/designs/queue-admission-gating/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Failure modes
+
+| Failure | Behavior |
+|---|---|
+| Webhook unavailable / plugin disabled | Pods created ungated → exactly today's behavior (enforcement is unchanged at allocation) |
+| Scheduler crash mid-gang ungate | Torn gang re-decided next session; partial-first ordering guarantees convergence |
+| Manual gate removal by user | Harmless bypass of noise suppression only; allocation-time capacity check still enforces quota |
+| Quota reduced after ungating | Gates cannot be re-added (Kubernetes one-way ratchet); pods degrade to today's allocation-time blocking. Accepted semantics; composes with #1783 |
+| Queue deleted / never existed | Pods stay gated with a distinct condition — ungating them would recreate the autoscaler over-provisioning for pods that can never schedule. Recovery on queue (re)creation or pod deletion |
```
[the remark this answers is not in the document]
shaq918: Resolved per @enoodle's direction: always gate; removal happens in `ungate` when the queue exists (D10). The failure table and the former open question now agree.

shaq918:
@enoodle addressed all review comments — doc updated in fdad5144 (one-way safety/liveness invariant, ungate exempt from readiness filter, missing-queue → gate per your direction, hierarchical Reserved accounting, and bounded reservation with expiry for the stuck-ungated-job concern). Replies threaded inline. PTAL when you get a chance.

docs/developer/designs/queue-admission-gating/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Motivation
+
+Queue quota and limits are enforced only at allocation time (`capacity_policy`). Pods blocked purely by queue policy are still created, marked `Unschedulable`, and re-examined every session. Two costs follow:
+
+1. **Autoscaler waste.** Cluster autoscalers (Karpenter, CA) treat `Unschedulable` pods as demand and provision nodes for workloads that will never be admitted. On GPU instance types this is expensive, and the nodes sit idle until scale-down. Volcano and Kueue both hold quota-blocked pods back from autoscaler visibility; KAI does not.
+2. **Scheduler load scales with submissions, not capacity.** Every session's snapshot, podgroup construction, and job ordering carry all submitted pods — O(pods submitted), not O(pods schedulable). A single tenant submitting far beyond quota degrades session latency for every queue. Scheduling signatures deduplicate repeated *scheduling attempts* within a session, but do not shrink the snapshot or the job-order heap, and do not hide pods from autoscalers.
```
enoodle: We will still build pod groups / pod infos for all pods because the ungate action only comes after the sessions snapshot. We also have to build them so that the ungate action could use that information to consider them to "ungate"

enoodle (commented):
Looks good. I have one more comment.

docs/developer/designs/queue-admission-gating/README.md
shaq918: You're right — since `ungate` decides from the session snapshot, PodInfos/PodGroupInfos are still built for all pods, and snapshot construction stays O(pods submitted). Re-scoped the motivation accordingly: the reduction applies to what the actions do with the snapshot — job ordering, allocation attempts, and preempt/reclaim scenario generation exclude fully gated podgroups — while snapshot cost is unchanged from today. Also noted (as out-of-scope future work) that gated pods only need resource requests + gate state for the quota math, so a lighter snapshot representation is possible if snapshot construction ever measures as a bottleneck.

enoodle (approved):

docs/developer/designs/queue-admission-gating/README.md
```diff
@@ -0,0 +1,142 @@
...
+| D12 | Reservations are bounded: a pod that remains pending releases its reservation after a configurable expiry, preventing head-of-line blocking by workloads unschedulable for non-quota reasons |
+
+## Open Questions
+
+1. Queue opt-in surface: label (no CRD churn) vs `QueueSpec` field?
+2. Gate and action naming: `kai.scheduler/queue-admission` / `ungate` vs alternatives (`admit`)?
+3. Ungate hysteresis or rate-limiting after a large quota increase (thundering herd of ungates)?
+4. Max-gate-age escape hatch for orphaned gated pods (queue deleted, podgroup never created)?
+5. Reservation expiry: default value and configuration surface (global vs per-queue)?
+6. Scale evidence to attach: session duration and API write volume with a large over-quota backlog (e.g. 50k pods), gated vs ungated, per the scale-test methodology.
```
enoodle: 1. QueueSpec
2. :+1: for the suggestions
3. I don't think that we need this as part of this step.
4. If the queue/podgroup is deleted then the pod can't be scheduled anyway, why does it matter?
5. This is the tricky part - should be some timeout that is configurable by the user in the global settings.
6. :+1:

enoodle (changes requested):
We should clear those open questions first!

docs/developer/designs/queue-admission-gating/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+Global feature toggle in the Config CR, default off. The operator fans one knob out to both sides — the admission flag and the scheduler flag plus the `ungate` action — so webhook and scheduler cannot disagree across upgrades:
+
+```yaml
+spec:
+  global:
+    queueAdmissionGating:
+      enabled: false
+```
+
```
SiorMeir: @shaq918 Just to clarify - these can co-exist, right?

I.e: in the scenario where I'm setting false on the global CR but gating a specific queue, Only this queue will benefit from admission gating, and vice versa (opting out on a specific queue)

Worth adding a line clarifying this

docs/developer/designs/queue-admission-gating/README.md
SiorMeir: @shaq918 

Agreeing with @enoodle on all questions

Regarding scale consideration: this should translate to single scenario (We don't want to overcrowd the tests) that is interesting to check. Your suggestion of a large number of pods waiting at the gate sounds good.

shaq918: @enoodle thanks — folded all six of your answers into Decided Points (commit d4a9abe2):

1. Opt-in is now a `QueueSpec.AdmissionGating` field (D8), not a label.
2. Names locked: gate `kai.scheduler/queue-admission`, action `ungate` (D13).
3. Dropped ungate hysteresis from this iteration — deferred until a real thundering-herd shows up (D14).
4. Agreed — a deleted queue leaves its pods unschedulable regardless, so no escape hatch; D10 updated to say this explicitly.
5. Reservation expiry is now a single global `spec.global.queueAdmissionGating.reservationTimeout` (D12) — kept it global rather than per-queue since it's a coarse "how long to trust a reservation while a node provisions" value that doesn't vary meaningfully per queue.
6. :+1: — scale scenario written up in a Validation section (one narrow case per @SiorMeir's note in this thread).

Open Questions section is now empty. PTAL.

docs/developer/designs/queue-admission-gating/README.md
shaq918: Good question — clarified in the doc (D8 + the API section). The two compose as **master-switch + per-queue opt-in**, not a default/override pair:

- The global toggle is the prerequisite — it installs and activates the admission plugin + `ungate` action. While it's off, nothing is gated anywhere, regardless of per-queue settings (the webhook stays out of the pod-create path entirely).
- When it's on, `QueueSpec.AdmissionGating` selects which queues are gated. A queue is gated iff *(global enabled and that queue opts in)*.

So the specific scenario — global false, one queue opted in — gates nothing, because the machinery isn't installed. I went this way rather than "global default + per-queue override" to avoid an always-on mutating webhook in every pod-create path and a per-queue field that silently no-ops when the feature isn't installed. If you'd prefer the default/override semantics (global sets a cluster-wide default, per-queue overrides both directions) that's a straightforward change — let me know which you'd rather have.

docs/developer/designs/queue-admission-gating/README.md
shaq918: @SiorMeir on the scale scenario — :+1:, kept it to a single case (large gated backlog over quota vs the same backlog ungated, measuring session duration + API writes) so it doesn't overcrowd the suite. Written up under Validation. (Your co-existence question is answered in the separate thread on the API section.)

docs/developer/designs/queue-admission-gating/README.md
SiorMeir: Got it, 
I'm ok with it as long as it is described in the design & later in the docs.

SiorMeir:
LGTM

SiorMeir (approved):

enoodle (approved):

shaq918:
Thanks for the thorough review @enoodle @SiorMeir — happy with where the design landed. Ready to merge whenever suits you.

I'd like to take the implementation as well. My plan is to split it into the phases outlined in the design — starting with the scheduler-side `ungate` action, then the admission plugin, operator/config wiring, and observability + the scale scenario — each as its own PR.

shaq918:
@enoodle rectified the DCO check (force-pushed to add `Signed-off-by`; squashed to a single commit, content unchanged from what was approved). Could you add the `skip-changelog` label so the workflows can go green? It's a docs-only PR so there's no changelog fragment. Thanks!

PR #1940, opened 2026-07-19
docs(agents): non-interactive make changelog for agents

SiorMeir:
## Description

`make changelog` runs `changie new`, an interactive TUI. Agents skip it and write fragment YAML by hand instead, which bypasses validation and produces inconsistent results.

This PR makes `make changelog KIND=<kind> BODY="..."` write the fragment directly without any prompts, while keeping the interactive fallback for humans (`make changelog` with no args).

Also wires up the pre-existing `hack/changelog-verify.sh` into the Makefile as `make changelog-verify`, and updates AGENTS.md so agents see the non-interactive form first.

## Related Issues

Fixes #

## Checklist

- [x] Self-reviewed

## Breaking Changes

None — `make changelog` with no args still calls `changie new` interactively.

## Additional Notes

Fragment is created under `.changes/unreleased/` with a timestamped filename matching the changie config format.


Makefile
```diff
@@ -116,8 +116,18 @@ $(CHANGIE): $(LOCALBIN)
 	test -s $(LOCALBIN)/changie || GOBIN=$(LOCALBIN) go install github.com/miniscruff/changie@$(CHANGIE_VERSION)
 
 .PHONY: changelog
-changelog: changie ## Add a changelog entry as a fragment (interactive). Use instead of editing CHANGELOG.md.
-	$(CHANGIE) new
+changelog: changie ## Add a changelog entry. Agents: make changelog KIND=Fixed BODY="...". Humans: make changelog (interactive).
```
gshaibi, line 119: Isn't the AGENTS.md doc enough?

gshaibi (approved):

Makefile
SiorMeir: I would like the agents to use a static hard coded command and not "invent" ways to create fragments so I think this is ok

PR #1942, opened 2026-07-19
feat(scheduler): override stale gang eviction

### LINKED ISSUE #1913: feat(scheduler): allow jobs to override stale gang eviction grace period
## What you would like to be added?

Add a job-level stale gang eviction configuration that overrides the scheduler-wide `default-staleness-grace-period`.

The setting should allow a job controller to specify a longer grace period or opt the job out of KAI stale gang eviction, while jobs without the setting continue to use the global default.

## Why is this needed?

Some job controllers own their workloads lifecycle and need to handle partial gang loss themselves. For example, Grove may need time to detect a partially available gang and apply controller-specific recovery or termination logic.

Today, the grace period is scheduler-wide. A short value can cause KAI to evict controller-managed jobs before that logic completes, while a long or infinite value affects every workload. A job-level override would let controller-managed jobs use their own lifecycle policy without changing stale-gang handling cluster-wide.

thebhdn:
## Description

Add a job-level stale gang eviction configuration that overrides the scheduler-wide default-staleness-grace-period.

The grace period is scheduler-wide. A short value can cause KAI to evict controller-managed jobs before that logic completes, while a long or infinite value affects every workload. A job-level override would let controller-managed jobs use their own lifecycle policy without changing stale-gang handling cluster-wide.

## Related Issues

https://github.com/kai-scheduler/KAI-Scheduler/issues/1913

## Checklist

- [x] Self-reviewed

## Breaking Changes

None - new feature.

## Additional Notes

🤔

## Summary by CodeRabbit

* **New Features**
  * Added an optional per-PodGroup staleness grace period.
  * Configure the duration using standard Kubernetes duration values; negative values disable stale workload eviction.
  * PodGroups now use their configured grace period, or the scheduler-wide default when unspecified.

* **Documentation**
  * Added configuration guidance and runtime behavior documentation.

* **Examples**
  * Added runnable manifests demonstrating delayed eviction of stale gang workloads.


SiorMeir:
@coderabbitai review

thebhdn:
Folks, just to clarify - we should also handle annotations on workloads for staleness grace period(e.g. same as preemtion delay)? Cuz I thought that it this case it should be explicitly configured only via `spec.`

gshaibi:
@thebhdn Most PodGroup features can be configured either via a workload annotation or directly in the PodGroup spec. I think we should follow that same pattern here.

Workload annotations are useful when the user doesn't create the PodGroup itself and relies on KAI's PodGrouper to create one for him.

pkg/podgrouper/podgroup/handler.go
```diff
@@ -102,11 +102,12 @@ func (h *Handler) createPodGroupForMetadata(podGroupMetadata Metadata) *scheduli
...
-			PriorityClassName: podGroupMetadata.PriorityClassName,
-			SubGroups:         []schedulingv2alpha2.SubGroup{},
-			Preemptibility:    podGroupMetadata.Preemptibility,
-			PreemptionDelay:   podGroupMetadata.PreemptionDelay,
+			Queue:                podGroupMetadata.Queue,
+			PriorityClassName:    podGroupMetadata.PriorityClassName,
+			SubGroups:            []schedulingv2alpha2.SubGroup{},
+			Preemptibility:       podGroupMetadata.Preemptibility,
+			PreemptionDelay:      podGroupMetadata.PreemptionDelay,
+			StalenessGracePeriod: podGroupMetadata.StalenessGracePeriod,
```
gshaibi, line 110: What sets `podGroupMetadata.StalenessGracePeriod`?

thebhdn: Yeah, currently its `nil`. I am going to update code to handle `annotations` so then it will make sense

pkg/apis/scheduling/v2alpha2/podgroup_types.go
```diff
@@ -82,6 +82,12 @@ type PodGroupSpec struct {
 	// allocation into free capacity, nor the PodGroup's own evictability.
 	// +optional
 	PreemptionDelay *metav1.Duration `json:"preemptionDelay,omitempty" protobuf:"bytes,9,opt,name=preemptionDelay"`
+
+	// StalenessGracePeriod is the minimum duration a stale PodGroup must remain in stale
```
itsomri:
```suggestion
	// StalenessGracePeriod is the minimum duration a stale PodGroup is allowed to remain in stale
```

pkg/apis/scheduling/v2alpha2/podgroup_types.go
```diff
@@ -82,6 +82,12 @@ type PodGroupSpec struct {
...
+	// status before stale workloads may be evicted to make room. Negative values disable
+	// eviction for this PodGroup. Defaults to the scheduler's global staleness grace period.
```
itsomri, line 88:
```suggestion
	// stale gang eviction for this PodGroup. Defaults to the scheduler's global staleness grace period.
```

pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
```diff
@@ -295,6 +297,28 @@ func (dg *DefaultGrouper) calcPodGroupPreemptionDelay(allOwners []*metav1.Partia
 	return nil
 }
 
+// calcPodGroupPreemptionDelay reads the preemption-delay annotation from owners then the pod.
+// First valid value wins; invalid or negative durations are ignored with a warning.
```
itsomri: Aren't negative values valid - and mean "exempt me from stale gang eviction"?

pkg/podgrouper/podgrouper/plugins/defaultgrouper/default_grouper.go
```diff
@@ -295,6 +297,28 @@ func (dg *DefaultGrouper) calcPodGroupPreemptionDelay(allOwners []*metav1.Partia
 	return nil
 }
 
+// calcPodGroupPreemptionDelay reads the preemption-delay annotation from owners then the pod.
```
gshaibi: Please fix this comment

pkg/apis/scheduling/v2alpha2/podgroup_types.go
```diff
@@ -128,6 +134,16 @@ func ParsePreemptionDelay(value string) (*metav1.Duration, error) {
...
+// ParseStalenessGracePeriod parses a staleness grace period string (e.g. "-10s" "30s", "5m").
+// Returns an error for invalid values.
+func ParseStalenessGracePeriod(value string) (*metav1.Duration, error) {
+	stale, err := time.ParseDuration(value)
+	if err != nil {
+		return nil, err
+	}
+	return &metav1.Duration{Duration: stale}, nil
+}
+
```
SiorMeir, line 146: @thebhdn  This may allow negative values that don't makes sense in a grace period

Maybe we should add a guard here?

itsomri (dismissed):
This looks really good! Added some minor comments, feel free to ping me on slack for re-review. Thanks for the contribution!

pkg/apis/scheduling/v2alpha2/podgroup_types.go
itsomri: negative = no stale gang eviction

gshaibi (commented):
Thanks for contributing!

itsomri (approved):

SiorMeir (approved):

PR #1943, opened 2026-07-20
feat(scheduler,binder): implement KEP-5004 DRA-backed extended resources

gshaibi:
## Description

Implements KEP-5004 (DRA Extended Resources) support for KAI Scheduler. When a `DeviceClass` declares an `ExtendedResourceName`, pods can use classic `nvidia.com/gpu: N` syntax on DRA-only nodes without changing their spec.

Design doc: [`docs/developer/designs/dra-extended-resources/README.md`](docs/developer/designs/dra-extended-resources/README.md)

### Changes

**API** (`pkg/apis/scheduling/v1alpha2/`)
- Added `ExtendedResourceClaimAllocation` struct to `BindRequestSpec` carrying the allocation result, device requests, and container mappings from the scheduler to the binder

**Scheduler**
- `ClusterInfo` and `NodeInfo` now hold a `DeviceClassByResource` cache (built from `DeviceClass` objects each session) to look up which extended resources are DRA-backed
- Vector fit checks and topology domain scoring skip DRA-backed extended resource dimensions so they don't falsely block scheduling
- New file `extended_resources.go` in the DRA scheduler plugin: helpers to detect DRA-backed resources, build the synthetic in-memory `ResourceClaim`, and generate `DeviceRequest`/`ContainerMapping` entries (ported from upstream `extendeddynamicresources.go`)
- DRA scheduler plugin `preFilter` detects pods with DRA-backed extended resources; `filter` builds and allocates the synthetic claim against the node's DRA allocator; `allocate` stores the result in `PodInfo.ExtendedResourceClaimAllocation`

**Binder**
- `K8sPlugin.IsRelevant` now takes the `BindRequest` so the DRA plugin can activate for pods with extended resource allocations even if they have no `pod.spec.resourceClaims`
- Binder DRA plugin `Bind` creates the `ResourceClaim` in the API, patches its status with the allocation result, and patches `pod.Status.ExtendedResourceClaimStatus` so the kubelet injects devices
- Binder DRA plugin `UnAllocate` deletes the created claim on rollback

**Common**
- `ExtractDRAGPUResourcesFromClaims` now skips synthetic extended-resource claims (annotated with `resource.kubernetes.io/extended-resource-claim: true`) to avoid double-counting GPU quota

## Related Issues

Fixes #

## Checklist

- [x] Self-reviewed

## Breaking Changes

`K8sPlugin.IsRelevant` signature changed to `IsRelevant(pod *v1.Pod, request *v1alpha2.BindRequest) bool`. Any external implementations of `K8sPlugin` will need to be updated.

## Additional Notes

- The temporary guard that rejected all device-plugin GPU requests on DRA nodes (`HasDRAGPUs` check) is removed; fit checking is now delegated to the DRA plugin filter
- Lint timed out locally (known infra issue); code passes `go vet` and `errcheck`/`unused` linters
- Envtest-based integration tests (`pkg/env-tests`, `pkg/binder/controllers/integration_tests`) require `setup-envtest` binaries not present locally; these are pre-existing infra constraints


pkg/scheduler/plugins/dynamicresources/extended_resources_upstream.go
```diff
@@ -0,0 +1,171 @@
...
+
+// Functions in this file are ported from:
+// k8s.io/kubernetes/pkg/scheduler/framework/plugins/dynamicresources/extendeddynamicresources.go
+//
+// Signatures are adapted to use *extendedresourcecache.ExtendedResourceCache in place of
+// fwk.DeviceClassResolver, and to drop klog.Logger and *stateData parameters that are
+// not needed in KAI's scheduling context.
+//
+// When upgrading Kubernetes, replace this file wholesale with the updated upstream logic.
+
```
SiorMeir, line 11: Why not use the upstream API and just const the params that are not needed?

This may create overhead

.changes/unreleased/Added-20260723-150619.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Added
+body: |
+Added support for DRA-backed extended resources (KEP-5004). Pods can request a `DeviceClass`'s `extendedResourceName` (e.g. `nvidia.com/gpu: N`) without a `ResourceClaim`, and the scheduler routes it through DRA. See the design [DRA-backed extended resources](docs/developer/designs/dra-extended-resources/README.md).
```
SiorMeir: this renders weird in github...

maybe shorten this out a bit?

pkg/scheduler/plugins/dynamicresources/extended_resources_upstream.go
gshaibi: How do you intend to use the upstream API?
These specific functions are private, and the `dynamicresources` plugin can't be used as a whole without putting a lot of effort to adjust it to our use, as it uses a state object that we do not (I guess that is one of the main reasons we didn't use it as a whole in the first place and wrote our own dra plugin).

.changes/unreleased/Added-20260723-150619.yaml
gshaibi: Sure

pkg/scheduler/plugins/dynamicresources/extended_resources_upstream.go
SiorMeir: Got it - if that's the case this sounds like the better solution

Then maybe we just need to remove the k8s upgrade comment, because I don't see a near future where we remove this.

SiorMeir (dismissed):

enoodle (commented):
@gshaibi 
Could we add focused unit coverage for rollback after a synthetic extended-resource claim is created? Please force a later bind failure and assert that rollback deletes the annotated ResourceClaim and clears pod.Status.ExtendedResourceClaimStatus, including an idempotency case.

enoodle (approved):

PR #1946, opened 2026-07-21
fix(scheduler): Fixed numa behavior for non-QoS-guaranteed pods with device request

itsomri:
## Description

This PR fixes the NUMA plugin behavior for non-QoS-Guaranteed pods that request devices such as GPU. Our initial wrong understanding of NUMA in k8s assumed that only Guaranteed pods are NUMA aligned, but extended resources such as GPUs can be aligned for Burstable or BestEffort pods.

## Checklist

- [x] Self-reviewed

### DIFF
--- a/pkg/scheduler/plugins/numa/numa.go
+++ b/pkg/scheduler/plugins/numa/numa.go
@@ -236,15 +236,46 @@ func parseIgnoreList(arguments framework.PluginArguments) sets.Set[v1.ResourceNa
 	return ignoreList
 }
 
-// shouldHandle engages the plugin for any Guaranteed task on a rejecting-policy node: the
-// kubelet aligns every Guaranteed pod (fractional/MIG included, on cpu/memory). The request
-// intersection in the evaluator decides which resources actually constrain each task.
+// shouldHandle engages the plugin on a rejecting-policy node for any task the kubelet would
+// NUMA-align. Guaranteed tasks are always handled. A non-Guaranteed task is handled if it requests a
+// topology-aware device.
 func (pp *numaPlugin) shouldHandle(task *pod_info.PodInfo, topo *node_info.NumaTopology) bool {
-	if topo == nil || !isModeledPolicy(topo.Policy) {
+	if topo == nil || !isModeledPolicy(topo.Policy) || task.Pod == nil {
 		return false
 	}
+	if isGuaranteed(task) {
+		return true
+	}
+	return pp.requestsAlignedDevice(task, topo)
+}
 
+// isGuaranteed reports whether the task's pod is Guaranteed QoS.
+func isGuaranteed(task *pod_info.PodInfo) bool {
 	return task.Pod != nil && task.Pod.Status.QOSClass == v1.PodQOSGuaranteed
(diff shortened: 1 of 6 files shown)


pkg/scheduler/plugins/numa/numa.go
```diff
@@ -236,17 +236,48 @@ func parseIgnoreList(arguments framework.PluginArguments) sets.Set[v1.ResourceNa
...
 
-// shouldHandle engages the plugin for any Guaranteed task on a rejecting-policy node: the
-// kubelet aligns every Guaranteed pod (fractional/MIG included, on cpu/memory). The request
-// intersection in the evaluator decides which resources actually constrain each task.
+// shouldHandle engages the plugin on a rejecting-policy node for any task the kubelet would
+// NUMA-align. Guaranteed tasks are always handled. A non-Guaranteed task is handled if it requests a
+// topology-aware device.
 func (pp *numaPlugin) shouldHandle(task *pod_info.PodInfo, topo *node_info.NumaTopology) bool {
-	if topo == nil || !isModeledPolicy(topo.Policy) {
+	if topo == nil || !isModeledPolicy(topo.Policy) || task.Pod == nil {
```
SiorMeir, line 243: NIT: ` || task.Pod == nil` is redundant with `isGuaranteed`'s check , no?

pkg/scheduler/plugins/numa/numa_test.go
```diff
@@ -90,20 +90,48 @@ func TestParseIgnoreList(t *testing.T) {
...
+		ResReqVector:        resource_info.NewResourceVectorFromResourceList(requests, testVectorMap),
+		Pod: &v1.Pod{
+			Status: v1.PodStatus{QOSClass: qos},
+			Spec:   v1.PodSpec{Containers: []v1.Container{{Resources: v1.ResourceRequirements{Requests: requests, Limits: requests}}}},
+		},
+	}
+}
+
+// burstableTask builds a Burstable pod (requests only) with the given container requests.
+func burstableTask(requests v1.ResourceList) *pod_info.PodInfo {
```
SiorMeir: NIT: maybe rename to `buildBurstableTask` for clearer intentions?
or even use QoS in the name as burstable is a wide term we may use elsewhere

pkg/scheduler/plugins/numa/numa.go
itsomri: No: right now, if pod==nil, we return false right away. If we remove the guard, `isGuaranteed` will return false, and the method will continue to the `requestsAlignedDevice` fn, which could return true.
In practice, Pod will never be nil in real world scheduling cycles

pkg/scheduler/plugins/numa/numa_test.go
itsomri: The usage of this fn is really only in the NUMA test suite so I think Burstable is pretty clear in this context. Changed to makeBurstableTask though (aligned with the existing makeTask) (and also changed gPod > makeGuaranteedTask)

SiorMeir (approved):

PR #1949, opened 2026-07-22
fix(scheduler): stop status-updater retry loop for deleted PodGroups

### LINKED ISSUE #1945: status-updater floods logs and API server retrying status/patch updates for deleted PodGroups
### Problem
The scheduler's status-updater repeatedly retries PodGroup status/annotation updates for
PodGroups that no longer exist, producing a high-rate, sustained flood of error logs and an
equal rate of failed API-server writes. Two log lines are involved, both from
`pkg/scheduler/cache/status_updater/concurrency.go`:

- `Failed to update pod group status <ns>/<pg>: podgroups.scheduling.run.ai "<pg>" not found`
- `Failed to patch pod group <ns>/<pg>: podgroups.scheduling.run.ai "<pg>" not found`

Version: v0.16.4.

### Indication (observed)
- Combined rate of the two lines above held a flat **~50 lines/s (~3,000/min) for 2+ hours**,
  cycling across different PodGroups and across both lines. At times this was **~44% of all
  scheduler log volume**.
- One PodGroup flooded via the status path for **~6 hours**, then stopped **without a restart**
  (self-resolved).
- A PodGroup actively flooding via the patch path was **absent from the scheduler snapshot**
  (`/get-snapshot` -> `rawObjects.podGroups`, i.e. the informer lister) at the same time it was
  flooding — so the flood is not explained by the lister retaining a deleted object.
- Affected PodGroups show **no other scheduler activity** (no allocate/reclaim/preempt/
  consolidation); they are workloads that were deleted (environment with frequent
  delete/recreate churn).
- Each log line corresponds to a failed `UpdateStatus`/`Patch` call, so the API server takes an
  equal rate of failing writes.

### Confirmation
Restarting the scheduler deployment cleared the flood immediately. It held ~3,000 lines/min
through 17:59 UTC; the deployment was restarted at ~17:59:35 UTC; from 18:01 UTC onward the two
error lines dropped to **0** while the scheduler kept logging normally (total scheduler log
volume stayed ~4,000-5,700 lines/min). This indicates the retry state is entirely **in-memory**
(the in-flight update maps / queue), not a persistent external condition. The flood is expected
to recur as PodGroups are deleted.

### Analysis (hypothesis)
In `updatePodGroup` (~L105-134), on any error the update is re-enqueued via `pushToUpdateQueue`.
Only `statusErr` `Conflict` is treated as terminal; `NotFound` on the status path — and **all**
errors on the patch path, which has no `Conflict`/`NotFound` handling — fall through to the
re-enqueue. For a deleted PodGroup the call returns `NotFound` on every attempt, so the update
loops. The backstop `cleanUpdatesForNonSeenPodGroups` (`pod_group_status_sync.go:57`) does not
stop it, apparently because the retry re-stores the in-flight entry faster than cleanup removes
it. This would explain why the loop persists even for PodGroups absent from the lister, and why
different PodGroups flood for anywhere from seconds to hours.

### Suggested direction (tentative)
Classify errors as terminal vs retryable in `updatePodGroup`: treat `NotFound` (and likely
`Conflict`) as terminal for **both** `statusErr` and `patchErr` — drop the in-flight entry and
return without re-enqueue, logging at low verbosity (a PodGroup deleted mid-flight is expected).
Retry only on genuinely transient errors.

Thezone-1:
## Description

The status-updater can get stuck retrying status/patch writes for PodGroups that no longer exist. In `updatePodGroup`, only a `Conflict` on the status update was treated as terminal — every other error, including `NotFound`, fell through to `pushToUpdateQueue`, which re-stored the in-flight entry and re-enqueued it.

For a deleted PodGroup this becomes a tight retry loop where every iteration is a failed API call and an error log line. As reported in the issue, this held at ~50 lines/sec for hours (up to ~44% of scheduler log volume), and the `cleanUpdatesForNonSeenPodGroups` backstop couldn't win the race because the retry re-stored the entry faster than cleanup removed it.

This change classifies `NotFound` and `Conflict` as terminal for **both** the status update and the patch. Terminal errors drop the in-flight entry (logged at low verbosity) instead of re-enqueuing; only genuinely transient errors are retried. If a status update is still required, the next scheduling cycle enqueues a fresh one.

## Related Issues

Fixes #1945

## Checklist

- [x] Self-reviewed
- [x] Added a changelog fragment

## Breaking Changes

None.

## Additional Notes

- Added a regression test for the deleted-PodGroup (`NotFound`) case: the in-flight entry is dropped and nothing is re-queued.
- Updated `TestDefaultStatusUpdater_RetryAfterError`, which previously relied on a non-existent PodGroup returning `NotFound` to prove "retry after error". Since `NotFound` is now terminal, the test injects a genuinely transient error so it still verifies the retry path.


gshaibi (dismissed):
That is a great fix, thank you @Thezone-1 !

pkg/scheduler/cache/status_updater/concurrency.go
```diff
@@ -103,17 +103,27 @@ func (su *defaultStatusUpdater) updatePodGroup(
...
-		if statusErr != nil {
-			if apierrors.IsConflict(statusErr) {
-				// Don't retry this update if the resource version is outdated - The status update cannot be updated with the given object.
-				// If a pod group status update is required (e.g. a scheduling condition) a new status update with an updated object
-				//  will be enqueued in the next scheduling cycle.
-				log.StatusUpdaterLogger.V(5).Infof("Conflict updating pod group status %s/%s, will retry in next cycle: %v",
+		// A NotFound (the pod group was deleted) or Conflict (the write targets a stale
+		// resource version) can never succeed on retry. Re-enqueuing such updates spins a
+		// tight retry loop that floods logs and the API server (#1945). Drop them; if a
+		// status update is still required, a fresh one is enqueued next scheduling cycle.
```
enoodle: Please remove the part of the comment about the issue and explaining the current failure - it is not relevant to the code.

pkg/scheduler/cache/status_updater/concurrency.go
```diff
@@ -141,6 +151,13 @@ func (su *defaultStatusUpdater) updatePodGroup(
 	}
 }
 
+// isTerminalUpdateError reports whether an error from a pod group status/patch write can
+// never succeed on retry: NotFound (the pod group was deleted) or Conflict (the write
+// targets a stale resource version).
```
enoodle: The comment is mostly redundant - one simple line is enough

pkg/scheduler/cache/status_updater/default_status_updater_test.go
```diff
@@ -863,11 +863,12 @@ func TestDefaultStatusUpdater_RetryAfterError(t *testing.T) {
 	statusUpdater := New(kubeClient, kubeAiSchedClient, recorder, 1, false, nodePoolLabelKey)
 
 	updateCalls := 0
-	// wait with pod groups update until signal is given.
+	// Return a transient (non-terminal) error so the update is retried. NotFound/Conflict
+	// are terminal and intentionally not retried, so this must be a generic error.
```
enoodle:
```suggestion
	// Return a transient error so the update is retried.
```

enoodle (changes requested):
Comments should be less obvious and only describe the code. It is in the Agents.md file.

Thezone-1:
@enoodle Done — trimmed the comments to describe only the code and applied your suggestion in the test. Thanks for the review!

gshaibi (approved):

enoodle (approved):

enoodle:
Thanks for the fix @Thezone-1

PR #1950, opened 2026-07-22
fix(scheduler): don't leak pipeline-only fit errors onto live jobs

### LINKED ISSUE #1948: Pipeline-only allocation leaks speculative fit errors
# Pipeline-only allocation leaks speculative fit errors

`AllocateJob(..., true)` runs during solver simulations. On pre-predicate
failure, `allocateTask` writes task fit errors to live session job even though
simulation later rolls back. `Statement.Rollback` does not restore fit errors.

Result: person inspecting pending job's fit errors can see reason from failed
Reclaim simulation, not reason why scheduler could not place job. This can send
them debugging pre-predicate or victim state that did not block authoritative
`Allocate` attempt.

Reproduces on `origin/main`.

## Reproducer

[`TestSchedulingCyclePreservesAllocateFitErrors`](https://github.com/enoodle/KAI-Scheduler/blob/fit-error-action-loop-regression/pkg/scheduler/actions/integration_tests/reclaim/reclaim_fit_errors_test.go):

```sh
env GOCACHE=/tmp/kai-fit-error-main-go-cache \
  go test ./pkg/scheduler/actions/integration_tests/reclaim \
  -run TestSchedulingCyclePreservesAllocateFitErrors -count=1 -v
```

Test runs normal action loop. `Allocate` leaves only `pending-gang-1` with GPU
fit error. Reclaim's pipeline-only simulation fails injected state-dependent
pre-predicate for `pending-gang-0`, then rollback leaves extra error:

```
task fit errors produced by Allocate were first changed by reclaim
```

## Expected

Fit errors must explain why job is currently pending. Pipeline-only allocation
must not add simulation-only task or job errors. Preserve authoritative Allocate
diagnostics.

Test uses injected pre-predicate to prove ownership bug; does not claim known
production plugin reproduces exact condition.

Thezone-1:
## Description

`AllocateJob(..., isPipelineOnly=true)` runs inside solver simulations (Reclaim, preempt, consolidation) that are rolled back with `Statement.Rollback`. Two paths recorded fit errors directly on the live session job during those simulations:

- `allocateTask` wrote a task fit error when `PrePredicateFn` failed.
- `allocateTasksOnNodeSet` → `handleFailedTaskAllocation` wrote job fit errors on any task failure.

`Statement.Rollback` restores allocation state but **not** fit errors, so these simulation-only diagnostics leaked onto the pending job and could obscure the authoritative reason produced by the real `Allocate` pass. Someone inspecting a pending job's fit errors could then see a reason from a failed Reclaim simulation rather than why the scheduler actually couldn't place the job.

This change skips task and job fit-error writes when `isPipelineOnly`, mirroring the existing `FittingNode(..., writeFittingDelta=!isPipelineOnly)` behavior that already suppresses per-node fit deltas during simulation. The authoritative `Allocate` pass (`isPipelineOnly=false`) records fit errors exactly as before.

## Related Issues

Fixes #1948

## Checklist

- [x] Self-reviewed
- [x] Added a changelog fragment

## Breaking Changes

None.

## Additional Notes

Includes the reproducer from the issue (`TestSchedulingCyclePreservesAllocateFitErrors`, authored by @enoodle): it runs a full action loop where Reclaim's pipeline-only simulation hits an injected pre-predicate failure, and asserts the job's task/job fit errors after all actions equal those produced by the initial `Allocate`.


Thezone-1:
@
Hi @enoodle — friendly ping on this one. CI is green and it's mergeable whenever you get a chance.

Since you filed #1948, I wanted to flag the overlap with your own #1921 ("scope fit errors to authoritative allocation attempts") — that looks like it's tackling the same leak from the allocation-attempt side, whereas this PR guards the fit-error writes in `allocate.go` on `!isPipelineOnly` (mirroring `FittingNode`'s `writeFittingDelta`) plus adds your reproducer as `reclaim_fit_errors_test.go`.

Happy to go whichever way makes sense: land this as the targeted fix, rebase it on top of #1921, or close it out if #1921 already covers the case. Just let me know what you'd prefer.
@

enoodle:
@Thezone-1 I haven't had time to look into this enough to form a solid opinion because this is a relatively minor issue. I hope to look into this next week.

enoodle (commented):
I think that the topology plugin would still add JobFitErrors to the job and other plugins can do that too. (see pkg/scheduler/plugins/topology/job_filtering.go:40 for example)

Thezone-1:
yeah you're right, and it's not only topology. job_filtering.go writes in 5 places and RecordVictimInvariantPrePredicateFailure in action_eligibility.go does too. FittingNode is fine since it already checks writeFittingDelta, but nothing stops the next plugin from adding one more.

so guarding call sites is a bit pointless, it only covers whatever I found today. the general version would be muting the writes on the PodGroupInfo for the duration of the pipeline-only attempt so it doesn't matter who writes them. I have that working locally but honestly I don't love it, it's a mutable flag on a shared object restored by defer, so an early return that skips the restore silently drops real fit errors, which is this bug backwards.

you said this one is fairly minor anyway, and your #1921 gets at the same leak from the allocation-attempt side without needing any of that. so I'm fine just closing this if #1921 covers it.

let me know which you'd rather, happy either way.

Thezone-1:
closing this one out. you were right that guarding the call sites in allocate.go only covers the writers I happened to find, and the general version (muting fit-error writes on the PodGroupInfo for the duration of the pipeline-only attempt) trades a narrow gap for a mutable flag on a shared object, which I don't think is a good trade for an issue you called minor.

#1921 gets at the same leak from the allocation-attempt side and doesn't need any of that, so it's the better home for this.

the reproducer test in here (TestSchedulingCyclePreservesAllocateFitErrors in pkg/scheduler/actions/integration_tests/reclaim/reclaim_fit_errors_test.go) is yours from #1948 and still passes on main. happy to send it as a standalone test-only PR if you want the coverage while #1921 lands. thanks for the review.

Issue #1756, opened 2026-06-24
[Numa awareness] Support best effort topology manager


davidLif:
### What you would like to be added?

Add scoring in the numa scheduler plugin, to try and point the pod to the node where he has a better numa allocation (less zones).

Decide later the exact scoring factor for numa.

### Why is this needed?

Better support topology manager

keinanDaniel:
1. Bin pack < Numa score < Topology (need to later add to user requested)
2. Need to think about the scoring mechanisms (Maybe # of numa nodes)

PR #1965, opened 2026-07-23
feat(scheduler): Implement node scoring in NUMA plugin

itsomri:
## Description

This PR adds node scoring to the NUMA plugin, thus extending support to `BestEffort` mode.

Notes:
- The plugin strictly prefers minimizing the zones a pod will allocate on
- Only `LeastNUMANodes` strategy was implemented for now - more can be added according to user feedback
- The plugin score is lower than network topology and gpu sharing
- A significant performance improvement was observed when testing the preempt scenario mentioned in #1911 - presumably because the lower scoring of unfeasible nodes increases the chances of finding a fit in FittingNodes. The scenario now finishes in ~65s vs ~187s (a ~64% improvement)

## Related Issues

Implements #1756 

## Checklist

- [x] Self-reviewed

### DIFF
--- a/pkg/scheduler/plugins/numa/evaluator.go
+++ b/pkg/scheduler/plugins/numa/evaluator.go
@@ -4,6 +4,7 @@
 package numa
 
 import (
+	"math"
 	"sort"
 
 	v1 "k8s.io/api/core/v1"
@@ -14,8 +15,11 @@ import (
 	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/api/resource_info"
 )
 
-// stackZones bounds the mask/scratch stack buffers; nodes with more NUMA zones fall back to heap.
-const stackZones = 16
+// stackZones and stackAware bound the mask/scratch stack buffers; larger nodes fall back to heap.
+const (
+	stackZones = 16
+	stackAware = 16
+)
 
 // zoneAllocation accumulates, per zone index, the amounts to place there (as a ResourceVector delta).
 // placementFromAllocation materializes it into a pod_info.NUMAPlacement.
@@ -49,15 +53,21 @@ func (pp *numaPlugin) alignedAware(task *pod_info.PodInfo, node *node_info.NodeI
 	return out
 }
 
-// allocatable reports whether the kubelet Topology Manager would align the task on the node. A task
-// the plugin does not constrain passes through as true.
+// allocatable reports whether the kubelet Topology Manager would align the task on the node (the
+// predicate). A task the plugin does not filter passes through as true.
 func (pp *numaPlugin) allocatable(task *pod_info.PodInfo, node *node_info.NodeInfo) bool {
+	if node == nil || !pp.shouldFilter(task, node.NumaTopology) {
+		return true
+	}
 	return pp.solveTask(task, node, nil)
 }
 
 // evaluate returns the task's expected per-zone allocation on the node (nil for a task the plugin
-// does not constrain). Used by the placement path; the predicate uses allocatable (feasibility only).
+// does not account for). Used by the placement and scoring paths.
 func (pp *numaPlugin) evaluate(task *pod_info.PodInfo, node *node_info.NodeInfo) (zoneAllocation, bool) {
+	if node == nil || !pp.shouldScore(task, node.NumaTopology) {
+		return nil, true
+	}
 	alloc := zoneAllocation{}
 	if !pp.solveTask(task, node, alloc) {
 		return nil, false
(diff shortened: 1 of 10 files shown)


pkg/scheduler/plugins/numa/evaluator.go
```diff
@@ -137,6 +147,72 @@ func (singleNUMAEvaluator) fit(topo *node_info.NumaTopology, aware []int, req re
...
+	var remArr [stackAware]float64
+	remaining := remArr[:0]
+	if len(aware) > stackAware {
+		remaining = make([]float64, 0, len(aware))
+	}
+	for _, idx := range aware {
+		remaining = append(remaining, req.Get(idx))
+	}
+	var usedArr [stackZones]bool
+	used := usedArr[:len(topo.Zones)]
```
gshaibi: What if `len(topo.Zones) > stackZones` here? Will it panic? If so, do we want to handle that?

pkg/scheduler/plugins/numa/numa.go
```diff
@@ -143,6 +169,68 @@ func (pp *numaPlugin) prePredicate(task *pod_info.PodInfo, _ *podgroup_info.PodG
...
+// 1/span. Skipped for non-NUMA-sensitive tasks and when no node reports topology.
+func (pp *numaPlugin) nodeScore(task *pod_info.PodInfo, node *node_info.NodeInfo) (float64, error) {
+	if pp.maxZones == 0 || !pp.wantsNuma(task) {
+		return 0, nil
+	}
+	span, ok := pp.assumedSpan(task, node)
+	if !ok {
+		return 0, nil
+	}
+	return scores.Numa / float64(span), nil
```
SiorMeir, line 192: why not round this up?

pkg/scheduler/plugins/numa/evaluator.go
```diff
@@ -14,8 +15,11 @@ import (
 	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/api/resource_info"
 )
 
-// stackZones bounds the mask/scratch stack buffers; nodes with more NUMA zones fall back to heap.
-const stackZones = 16
+// stackZones and stackAware bound the mask/scratch stack buffers; larger nodes fall back to heap.
+const (
+	stackZones = 16
+	stackAware = 16
```
SiorMeir, line 21: Configurable?

pkg/scheduler/plugins/numa/evaluator.go
```diff
@@ -137,6 +147,72 @@ func (singleNUMAEvaluator) fit(topo *node_info.NumaTopology, aware []int, req re
 	return nil, false
 }
 
+// bestEffortEvaluator (best-effort) greedily builds the narrowest zone mask whose summed Available
+// covers the request, adding at each step the zone that covers the most unmet demand. best-effort
+// never rejects, so the result feeds scoring and the in-cycle ledger, not an admit decision; the
+// greedy width approximates the kubelet's best-effort span (exact when one resource dominates).
```
SiorMeir: having a bit of a hard time understanding the algo here

Can we maybe simplify the comment to explain this?

pkg/scheduler/plugins/numa/evaluator.go
itsomri: It's right on the next statement.
Having a const (known at compile time) len for the array lets the compiler allocate memory more efficiently.
16 should suffice in most cases. If it's longer, it will revert to dynamic memory allocation

pkg/scheduler/plugins/numa/evaluator.go
itsomri: That would defeat the whole purpose
Having a const (known at compile time) len for the array lets the compiler allocate memory more efficiently.
16 should suffice in most cases. If it's longer, it will revert to dynamic memory allocation (less efficient, but will not break the code)

pkg/scheduler/plugins/numa/numa.go
itsomri: Why round this up?
scores.Numa is 100 (subject to future change), span is an integer usually in range $[1, 16]$
Rounding to any direction will lose us granularity, we could score two spans identically even though they're different

pkg/scheduler/plugins/numa/evaluator.go
gshaibi: But won't it panic here and won't get to the next statement?

gshaibi: Oh I see you fixed it, great.

gshaibi (dismissed):
LGTM

pkg/scheduler/plugins/numa/evaluator.go
itsomri: Basically it evaluates approximately how the kubelet would place the pod.
I'm saying "Approximately" because I already observed cases where the placement of the same pod spec on the same node condition gave different results.
The kubelet itself actually calculates many more permutations for placements (each "hint provider" evaluates a set of possible placements, the kubelet merges those to find the optimal fit), like we do in the `restricted` evaluator. The difference is, that the restricted evaluator only accepts a constant width per pod*node - so it's search space is narrower - while the BestEffort will need to evaluate all possible widths
So instead of re-calculating these permutations for every possible width, I went for a pretty simple greedy algorithm that will work in most cases where there's one domniant resource: 
- Start with an empty mask
- For a given request, iterate Z (zone) times (outer loop)
  - Iterate over the zones, find the one that "helps" us the most (covers the most remaining request for any resource)
  - Allocate resources for it, continue iterating if there's a remaining request

This should work well in cases where there's a specific resource that's usually in contention, which I think is the usual case. You could find adversarial cases where this returns the wrong score - but it's still just a non-optimal allocation, this evaluator does not reject pods.
The greedy implementation is simpler when compared to the more exhaustive one for Restricted, and should be more performant (although that wasn't the main concern, and I haven't validated this specifically)

SiorMeir (approved):

PR #1988, opened 2026-07-28
fix(scheduler): set Go memory limit from cgroups

enoodle:
## Description

Derive the scheduler Go runtime memory limit from its effective cgroup memory limit to prevent the runtime heap target from outgrowing the container limit.

- Uses `automemlimit` (Pinned to `v0.7.5`) only for its cgroup v1/v2 detector.
- Apply `runtime/debug.SetMemoryLimit` at startup and refresh it every 15 seconds with a scheduler-owned, cancellable ticker.
- Add per-`SchedulingShard` `goMemLimitRatio` (default `0.9`) and optional `goMemLimit` override.
- Preserve explicit `GOMEMLIMIT`, avoid system-memory fallback, and document the VPA downsize race.

## Related Issues

None.

## Checklist

- [x] Self-reviewed
- [x] Added/updated tests
- [x] Updated documentation

## Breaking Changes

None.


.changes/unreleased/fixed-20260727-234051.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Fixed
+body: |-
+  Prevent some scheduler OOM kills by setting Go's memory limit
```
SiorMeir, line 3: NIT: you added an opt in option, not technically a fix

cmd/scheduler/app/memory_limit.go
```diff
@@ -0,0 +1,155 @@
...
+	Stop()
+}
+
+type runtimeMemoryLimitTicker struct {
+	*time.Ticker
+}
+
+func (t runtimeMemoryLimitTicker) Chan() <-chan time.Time {
+	return t.C
+}
```
SiorMeir: Why the timer? Can you explain the use? 🙏🏻

.changes/unreleased/fixed-20260727-234051.yaml
enoodle: It is always on - you have to opt out, hopefully no one will

enoodle: By opt out I mean - someone will have to set their specific GOMEMLIMIT instead..

cmd/scheduler/app/memory_limit.go
enoodle: I think maybe this struct can be removed, I will look into it.
It is used so that the memoryLimitManager would check the cgroup limit every 15 seconds in case it was updated by some container resize.

.changes/unreleased/fixed-20260727-234051.yaml
SiorMeir: Got it, sounds good 
Resolving

cmd/scheduler/app/memory_limit.go
SiorMeir: Sounds like a bit of a churn but I can't think of a better solution

why 15 seconds, maybe a longer interval?

SiorMeir (dismissed):

cmd/scheduler/app/memory_limit.go
enoodle: This operation of checking the cgroup is light, there is always the fear that the container will be resized down and the GC will not get the message soon enough.

enoodle: There is no real reason to choose 15 seconds, just seemed to make sense.

SiorMeir (approved):

PR #1995, opened 2026-07-29
feat(scheduler): add gpujoborder plugin for configurable JobOrderFn tiebreak

### LINKED ISSUE #1966: preempt: JobOrderFn tiebreak has no resource-awareness, falls back to creation-time order

A resource-aware secondary comparator for JobOrderFn, to be tried only when the existing priority-based comparison ties.

Currently, priority.go's JobOrderFn compares job Priority only. On a tie, Session.JobOrderFn (in session_plugins.go) falls through to its built-in fallback: order by CreationTimestamp, then UID. For the victim queue specifically, this comparison is inverted, so among equal-priority eviction candidates, the most recently created job is currently preferred as the victim, with no consideration of GPU footprint or eviction cost.

Since AddJobOrderFn appends to a slice and the composition loop short-circuits on the first non-zero result, a new plugin could register a second comparator that only activates when priority ties, e.g. preferring to evict the candidate with the larger GPU footprint, similar in spirit to a lex-min / minimum-disruption selection. This wouldn't require changes to priority.go or the composition mechanism itself.

### Why is this needed?

Verified this behavior both in source and live: two identical-priority pods with different GPU footprints (1 GPU vs 2 GPUs), forced into the same eviction scenario twice with creation order reversed between runs. The evicted pod tracked creation order exactly both times (most-recently-created evicted), matching the fallback logic above not resource footprint.

I don't have production data on how often same-priority ties actually occur in real deployments, so I'm genuinely unsure whether this is common enough to be worth the added plugin surface, or a known/deprioritized edge case. Wanted to raise it as a possible needs-design discussion rather than assume it's worth building, curious what the maintainers think before going further.

CoolingCube:
## Description
Adds an additive `JobOrderFn` plugin that activates only when the existing
`priority` plugin's comparison ties (composition is additive per
`session_plugins.go`; no changes to `priority.go` or the composition
mechanism itself).

Configurable via `mode` argument: `prefer-larger` (default) or
`prefer-smaller`, comparing `GetAliveTasksRequestedGPUs()` between
same-priority jobs. Unrecognized `mode` values log a warning and fall back
to `prefer-larger`. Falls through unchanged to the existing
CreationTimestamp/UID tiebreak if GPU counts also match.

A `dominant-resource` mode (comparing by whichever resource represents a
job's largest relative share) was discussed in #1966 but isn't included —
`ClusterInfo` has no pre-computed total cluster capacity, so it would need
its own summation over `Nodes` mirroring what `proportion.go` does for
queue fair-share, rather than being a quick addition. Happy to take that
on as a follow-up if there's interest.

## Related Issues
Fixes #1966

## Checklist
- [x] Self-reviewed
- [x] Added/updated tests (if needed) — 5 unit tests: priority-difference
      deferral, both modes' comparison behavior, same-priority-same-GPU
      fallthrough, unrecognized-mode fallback

## Breaking Changes
None — purely additive; no existing plugin or behavior is modified.

## Additional Notes
Full design discussion and the live reversed-creation-order test that
motivated this are in #1966.


enoodle:
@CoolingCube Can you please use the PR template ? https://github.com/kai-scheduler/KAI-Scheduler/blob/main/.github/pull_request_template.md
It will help with review.

pkg/scheduler/plugins/resourceaware/resourceaware.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/log"
+)
+
+type resourceAwarePlugin struct {
+	mode string
+}
+
+func New(arguments framework.PluginArguments) framework.Plugin {
+	mode := arguments.GetString("mode", "prefer-larger")
+	if mode != "prefer-larger" && mode != "prefer-smaller" {
```
enoodle: Can you use constants for `prefer-larger` and other strings here?

pkg/scheduler/plugins/resourceaware/resourceaware.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (rp *resourceAwarePlugin) JobOrderFn(l, r interface{}) int {
+	lv := l.(*podgroup_info.PodGroupInfo)
+	rv := r.(*podgroup_info.PodGroupInfo)
+
+	if lv.Priority != rv.Priority {
+		return 0
+	}
+
+	lGPU := lv.GetAliveTasksRequestedGPUs()
```
enoodle: The plugin is called `resourceaware` but it is only looking at GPUs

CoolingCube: Renamed to `gpujoborder` to match the actual scope and the existing naming convention (`gpusharingorder`, `gpupack`, `gpuspread)`. Happy to change it again if you'd prefer something else.

pkg/scheduler/plugins/resourceaware/resourceaware.go
CoolingCube: Added` ModePreferLarger`/`ModePreferSmaller` constants and replaced the raw string comparisons.

CoolingCube:
The benchmark bot's regression above was real, traced it to the reclaim action rebuilding its entire victims queue once per candidate reclaimer job (pre-existing behavior), which made` GetAliveTasksRequestedGPUs()`'s per-call cost newly visible at scale. Cached it the same way `activeAllocatedCount `already is on` PodGroupInfo`, should resolve once the fresh benchmark run goes through. Let me know if you'd rather see something different.

CoolingCube:
Just checking in, this has been green for a while now (perf fix, license/formatting fix, and the naming/constants changes from your review are all in). No rush at all, just didn't want it to slip off the radar. Let me know if there's anything else you'd like changed.

enoodle (dismissed):
Thanks @CoolingCube ! I wanted to look for ways to allow integrating such plugins more easily without requiring review and merge into this repo, but I think we should merge it for now anyway.

gshaibi:
Hi @CoolingCube, thanks for contributing.
The JobOrder plugins apply not only to victims, but also to pending jobs order - does it match with your intention? what does larger/smaller job mean when choosing a job to allocate?

CoolingCube:
Good question @gshaibi and worth being precise: this isn't hypothetical. `job_order_by_queue.go` uses the same `ssn.JobOrderFn` for both victim selection and pending-job ordering, just inverted for the victim case, so `gpujoborder` is already active on the ordering path too, not something that would apply if extended.
Honestly, I don't think 'larger/smaller' obviously means the same thing for choosing which job to allocate as it does for choosing which job to evict, for eviction it's about disruption cost, but for allocation it's not clear preferring one size over another is right at all. I don't have real evidence either way for the ordering case specifically. Happy to think through what the right behavior there should be, or whether ordering should be scoped out of this plugin's tiebreak entirely, since it's a genuinely separate question from eviction.

gshaibi:
> Good question @gshaibi and worth being precise: this isn't hypothetical. `job_order_by_queue.go` uses the same `ssn.JobOrderFn` for both victim selection and pending-job ordering, just inverted for the victim case, so `gpujoborder` is already active on the ordering path too, not something that would apply if extended. Honestly, I don't think 'larger/smaller' obviously means the same thing for choosing which job to allocate as it does for choosing which job to evict, for eviction it's about disruption cost, but for allocation it's not clear preferring one size over another is right at all. I don't have real evidence either way for the ordering case specifically. Happy to think through what the right behavior there should be, or whether ordering should be scoped out of this plugin's tiebreak entirely, since it's a genuinely separate question from eviction.

Thanks, I agree that allocation order is a separate question. Since this change is intended to control victim selection, it shouldn’t affect the order of pending jobs.

Also, because the victim queue reverses `JobOrderFn`, `prefer-larger` currently schedules larger jobs first but selects smaller jobs as victims.

Would it be possible to scope this behavior to victim selection only and add a test confirming that pending-job ordering remains unchanged?

CoolingCube:
> > Good question @gshaibi and worth being precise: this isn't hypothetical. `job_order_by_queue.go` uses the same `ssn.JobOrderFn` for both victim selection and pending-job ordering, just inverted for the victim case, so `gpujoborder` is already active on the ordering path too, not something that would apply if extended. Honestly, I don't think 'larger/smaller' obviously means the same thing for choosing which job to allocate as it does for choosing which job to evict, for eviction it's about disruption cost, but for allocation it's not clear preferring one size over another is right at all. I don't have real evidence either way for the ordering case specifically. Happy to think through what the right behavior there should be, or whether ordering should be scoped out of this plugin's tiebreak entirely, since it's a genuinely separate question from eviction.
> 
> Thanks, I agree that allocation order is a separate question. Since this change is intended to control victim selection, it shouldn’t affect the order of pending jobs.
> 
> Also, because the victim queue reverses `JobOrderFn`, `prefer-larger` currently schedules larger jobs first but selects smaller jobs as victims.
> 
> Would it be possible to scope this behavior to victim selection only and add a test confirming that pending-job ordering remains unchanged?

Ran it against the merged code directly and it matches exactly what you flagged: `prefer-larger` evicts the smaller job, because the victim queue's inversion flips the same sign that makes larger jobs allocate first. Good catch, and it's exactly why scoping this to victim-selection only isn't just cleanup, a single shared comparator genuinely can't satisfy both paths correctly at once.
I don't see an existing way to register a comparator that only applies to the victim queue in the current framework, `JobOrderFn`'s signature has no way to know which queue it's being called from, so this would likely need a real extension point rather than a small patch, which sounds related to what enoodle mentioned wanting to explore anyway.
Given this is already approved, wanted to flag it before it merges rather than after. Happy to take a pass at either a scoped fix or a minimal sign correction in the meantime whichever's more useful.

CoolingCube:
Pushed the fix  added a dedicated `VictimOrderFn`/`AddVictimOrderFn` extension point so victim-specific comparators no longer share (or invert) the same sign used for pending-job ordering. `gpujoborder` now registers only via the new path. Added a test confirming pending-job ordering is byte-identical with or without the plugin registered. Full framework/actions suite passes

pkg/scheduler/framework/session_plugins.go
```diff
@@ -285,6 +294,21 @@ func (ssn *Session) JobOrderFn(l, r interface{}) bool {
...
 }
 
+// VictimOrderFn composes registered victim-specific comparators directly --
+// a negative result means "l is the BETTER victim", with no external
+// inversion needed or applied (unlike the old !JobOrderFn(l, r) pattern).
+// If no victim-specific comparators are registered, this falls back to the
+// existing !JobOrderFn(l, r) behavior, preserving exact current behavior
+// for any plugin that never needed the ordering/victim distinction.
+func (ssn *Session) VictimOrderFn(l, r interface{}) bool {
+	for _, vof := range ssn.VictimOrderFns {
```
gshaibi, line 317: Thanks, the scoping looks good.

One thing - the victim comparators run before the inverted JobOrderFn chain, so this still outranks elastic for same-priority jobs. e.g. a larger at-min job gets evicted before trimming an elastic job that's above min.

Can we flip it - run the inverted job order comparators first and only use victim comparators on ties? JobOrderFn never ties (timestamp fallback), so VictimOrderFn needs to go over the chain itself:

```
for _, jof := range ssn.JobOrderFns {
    if j := jof(l, r); j != 0 {
        return j > 0
    }
}
for _, vof := range ssn.VictimOrderFns {
    if v := vof(l, r); v != 0 {
        return v < 0
    }
}
return !jobOrderCreationFallback(l, r)
```

Path without victim fns stays identical. A test for the elastic case would be good too.

pkg/scheduler/plugins/gpujoborder/gpujoborder.go
```diff
@@ -0,0 +1,76 @@
...
+package gpujoborder
+
+import (
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/api/podgroup_info"
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/framework"
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/log"
+)
+
+const (
+	ModePreferLarger  = "prefer-larger"
```
gshaibi: Small thing - since this only affects eviction now, "prefer-larger" can be read as "prefer to keep larger". These are hard to rename later, what do you think about "evict-larger-first" / "evict-smaller-first"?

In any case, a short documentation would be great.

gshaibi (commented):
Looks great, left two more comments.

CoolingCube:
Pushed a fix for both 

For the composition order: you're right, `VictimOrderFn` was checking victim-specific comparators before the existing `JobOrderFns` chain, so a raw GPU-size comparison could outrank elastic's at-min/above-min protection. Fixed exactly as you outlined, the raw `JobOrderFns` are checked first (inverted), `VictimOrderFns` only apply as a tiebreak, and the `CreationTimestamp`/UID fallback is now shared via `jobOrderCreationFallback` so both paths stay consistent. Added a test with both `elastic `and `gpujoborder` registered together confirming elastic's protection wins (with a follow-up commit fixing a duplicate-UID bug in the test helper itself, so the confirmed GPU numbers are accurate too).

For the naming: agreed, renamed to `evict-larger-first`/`evict-smalle`r-first and added doc comments on the package and both constants."

pkg/scheduler/framework/session_plugins.go
```diff
@@ -268,21 +277,61 @@ func (ssn *Session) QueueAllocatedResources(queue *queue_info.QueueInfo) *resour
...
-	if lv.CreationTimestamp.Equal(&rv.CreationTimestamp) {
-		return lv.UID < rv.UID
-	} else {
-		return lv.CreationTimestamp.Before(&rv.CreationTimestamp)
+	return jobOrderCreationFallback(l, r)
+}
+
+// VictimOrderFn composes registered victim-specific comparators, but only
+// as a tiebreak AFTER the existing JobOrderFns chain (priority.go,
+// elastic.go, etc.) has had its say -- fixed per @gshaibi's review: the
```
gshaibi: Please remove excessive commenting, we prefer having the bare minimum commenting needed (preferably none if possible)

CoolingCube: Removed the comments, happy to explain the constructor choice in the test helper here if it ever comes up, rather than baking it into the source

pkg/scheduler/api/podgroup_info/job_info.go
```diff
@@ -93,6 +93,7 @@ type PodGroupInfo struct {
 	tasksToAllocateInitResourceVector resource_info.ResourceVector
 	PodStatusIndex                    map[pod_status.PodStatus]pod_info.PodsMap
 	activeAllocatedCount              *int
+	aliveTasksRequestedGPUs           *float64
```
davidLif, line 96: I do not see any code that invalidates this number if the number of "alive" pods is changed

gshaibi (commented):
LGTM. One thing I am missing is documentation, especially regarding the new extension point.
@davidLif you had some concerns - feel free to point them out.

pkg/scheduler/framework/session_plugins.go
```diff
@@ -268,21 +272,36 @@ func (ssn *Session) QueueAllocatedResources(queue *queue_info.QueueInfo) *resour
...
+}
 
-	// If no job order funcs, order job by CreationTimestamp first, then by UID.
-	lv := l.(*podgroup_info.PodGroupInfo)
-	rv := r.(*podgroup_info.PodGroupInfo)
-	if lv.CreationTimestamp.Equal(&rv.CreationTimestamp) {
-		return lv.UID < rv.UID
-	} else {
-		return lv.CreationTimestamp.Before(&rv.CreationTimestamp)
+func (ssn *Session) VictimOrderFn(l, r interface{}) bool {
```
davidLif, line 311: Up until now, plugins that wanted to effect victims order registered on JobOrderFn. Even after this change, they should still have the same effect on the victims order as before

davidLif (commented):
With the addition of VictimOrderFn, this PR needs to add some integration tests making sure that plugins with a JobOrderFn still effect the victims order, preserving exisitng logic.

pkg/scheduler/api/podgroup_info/job_info.go
gshaibi: It should happen [here](https://github.com/kai-scheduler/KAI-Scheduler/pull/1995/changes#diff-d44d6abc0654dfccf32f4ec2f072b0c7dab4bd45ad699e94d3ea59251980662eR369)

CoolingCube: Confirmed, `invalidateTasksCache()` (which nils `aliveTasksRequestedGPUs` along with the other cached fields) is called from `addTaskIndex` and `deleteTaskIndex`. Both add and delete are covered directly (`AddTaskInfo` → `addTaskIndex`, and the delete path → `deleteTaskIndex`). Status changes go through `UpdateTaskStatus`, which calls` resetTaskState` (→ `deleteTaskIndex`) and then `AddTaskInfo` (→ `addTaskIndex`), so a status change invalidates the cache twice, once on each side of the transition. I don't see a path that mutates `PodStatusIndex `without going through one of these two functions. Let me know if there's a case I'm missing.

pkg/scheduler/framework/session_plugins.go
CoolingCube: pushed `TestVictimOrderFn_PreservesExistingJobOrderFnBehavior`, directly comparing the old` !ssn.JobOrderFn(l, r) `against the new` ssn.VictimOrderFn(l, r)` across many priority combinations with no` VictimOrderFns `registered. Confirms existing plugins keep identical effect on victim ordering

gshaibi (approved):
LGTM

CoolingCube:
@davidLif, I went ahead and resolved the two threads above since the fixes/tests covered them. tell me if you want another look at anything  ?

davidLif (approved):

PR #2012, opened 2026-08-04
fix(scheduler): keep succeeded pods visible so completing gangs are not evicted

### LINKED ISSUE #1968: stalegangeviction evicts pods of successfully completing jobs
## Symptom

For gang-scheduled workloads with `minMember` equal to the replica count (JobSet / LeaderWorkerSet / PyTorchJob, or a plain PodGroup), when pods complete successfully one by one, the `stalegangeviction` action evicts all remaining running pods of the gang ~60s (the default staleness grace period) after the first pod completes Eviction event: `Workload doesn't meet the minimum required number of pods (N), evicting remaining pod: <ns>/<name>`.

## Root cause

`PodGroupInfo.IsStale()` (`pkg/scheduler/api/podgroup_info/job_info.go`) contains a guard: a podgroup with any Succeeded pod is never considered stale. 
```go
if pgi.PodStatusIndex[pod_status.Succeeded] != nil {
    return false
}
```

Since #1647 the scheduler's pod informer filters terminal pods at watch time with field selector `status.phase!=Succeeded,status.phase!=Failed` (`pkg/scheduler/cache/cache.go`). Succeeded pods therefore never enter the scheduler cache, `PodStatusIndex[Succeeded]` is populated, and the guard is dead code. When a pod completes, it simply disappears from the scheduler's view, then the podgroup becomes stale, and after the grace period every remaining active pod is deleted — exactly while the job is finishing successfully.

## Affected versions

v0.16.2 or later.

## Steps to reproduce

1. Install KAI v0.16.4 with default configuration (`stalegangeviction` enabled, 60s default staleness grace period).
2. Submit a gang workload `minMember: 2` whose two pods exit 0 at different times (e.g. pod-0 sleeps 60s, pod-1 sleeps 600s).
3. Observe: ~60s after pod-0 completes, pod- is deleted by the scheduler; the podgroup carries the `kai.scheduler/stale-podgroup-timestamp` annotation and the eviction event above is emitted.

## Proposed fix

Stop filtering `Succeeded` pods at the scheduler pod informer (keep filtering `Failed only), restoring the visibility the `IsStale()` guard relies on.

## Workarounds on affected versions

Downgrade to v0.16.1, or set `--default-staleness-grace-period=-1s` (or a large value), or remove `stalegangeviction` from the scheduler config `actions` list.

Pods that complete successfully must not cause the remaining pods of the gang to be evicted.

### Environment

- Kubernetes version: v1.35.2
- KAI Scheduler version: v0.16.4
- Cloud provider or hardware configuration
- Tools that you are using KAI together with: JobSet, Helm
- Anything else that is relevant: default scheduler configuration (stalegangeviction enabled, 60s grace period)

Thezone-1:
Fixes #1968

## Problem

For a gang-scheduled workload whose pods exit 0 at different times, `stalegangeviction` deletes every remaining running pod roughly one staleness grace period after the first pod completes, while the job is finishing successfully:

```
Workload doesn't meet the minimum required number of pods (N), evicting remaining pod: <ns>/<name>
```

## Root cause

`PodGroupInfo.IsStale()` guards against exactly this: a podgroup with any Succeeded pod is never considered stale.

```go
if pgi.PodStatusIndex[pod_status.Succeeded] != nil {
    return false
}
```

Since #1647 the scheduler pod informer filters both `Succeeded` and `Failed` at watch time via a field selector, so Succeeded pods never enter the cache, `PodStatusIndex[Succeeded]` is never populated, and the guard became dead code. A completing pod simply disappears from the scheduler's view, the podgroup is judged stale, and the survivors are evicted.

## Fix

Two commits.

**1. Keep succeeded pods visible.** The watch filter and the notion of terminality are decoupled rather than sharing one list:

- `watchFilteredPodPhases = {Failed}` feeds the informer's field selector, via `filterFailedPods` (renamed from `filterTerminalPods`).
- `terminalPodPhases = {Succeeded, Failed}` is left alone, because it also backs `isTerminated()`, the guard in `Evict()` against evicting an already terminated pod. Dropping Succeeded from the shared list would have restored `IsStale()` at the cost of breaking that guard.

Failed pods stay filtered, which preserves most of #1647's watch load reduction: failures are the high churn class in AI workloads (OOM, crashloop), while succeeded pods are bounded by gang size and are garbage collected with the workload.

**2. Strip cached succeeded pods.** Per the review note on the issue, a succeeded pod is never placed on a node and never counted towards node resources, so `compactSchedulerPod` now routes it to `compactSucceededPod`, which retains only what snapshot assembly reads: name, namespace, UID, phase, the podgroup annotation, and the subgroup and app labels. Everything else, including the whole pod spec, is dropped. This goes further than the trimming in #1648 and keeps the memory cost of re-admitting these pods small.

## Testing

- `pkg/scheduler/actions/stalegangeviction`: three regression cases added to `TestStaleGangEviction`, covering a gang with a succeeded pod, a gang down to one running pod, and a succeeded pod inside one sub group. All three fail without commit 1.
- `pkg/scheduler/cache`: six specs for `compactSchedulerPod` covering the succeeded path, and the existing field selector assertion updated to expect `status.phase!=Failed` only.
- Full package runs are green: `pkg/scheduler/cache`, `pkg/scheduler/cache/cluster_info`, `pkg/scheduler/cache/status_updater`, `pkg/scheduler/api/podgroup_info`, `pkg/scheduler/actions/stalegangeviction`. `go vet` is clean.

Reported by @lixiang233.

### DIFF
--- a/.changes/unreleased/Changed-20260801-235945.yaml
+++ b/.changes/unreleased/Changed-20260801-235945.yaml
@@ -1,6 +1,0 @@
-kind: Changed
-body: Succeeded pods cached by the scheduler are stripped to the fields podgroup accounting reads, limiting the memory cost
-time: 2026-08-01T23:59:45.8835399+05:30
-custom:
-    Author: Thezone-1
-    Issue: "1968"
--- a/.changes/unreleased/Changed-20260801-235945.yaml
+++ b/.changes/unreleased/Changed-20260801-235945.yaml
@@ -1,0 +1,6 @@
+kind: Changed
+body: Succeeded pods cached by the scheduler are stripped to the fields podgroup accounting reads, limiting the memory cost
+time: 2026-08-01T23:59:45.8835399+05:30
+custom:
+    Author: Thezone-1
+    Issue: "1968"
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -83,7 +83,6 @@ var terminalPodPhases = []v1.PodPhase{
 	v1.PodFailed,
 }
 
-// Succeeded pods are watched so PodGroupInfo.IsStale can tell a completing gang from a stalled one.
 var watchFilteredPodPhases = []v1.PodPhase{
 	v1.PodFailed,
 }
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -83,9 +83,14 @@ var terminalPodPhases = []v1.PodPhase{
 	v1.PodFailed,
 }
 
-func filterTerminalPods(options *metav1.ListOptions) {
-	selectors := make([]string, 0, len(terminalPodPhases))
-	for _, phase := range terminalPodPhases {
+// Succeeded pods are watched so PodGroupInfo.IsStale can tell a completing gang from a stalled one.
+var watchFilteredPodPhases = []v1.PodPhase{
+	v1.PodFailed,
+}
+
+func filterFailedPods(options *metav1.ListOptions) {
+	selectors := make([]string, 0, len(watchFilteredPodPhases))
+	for _, phase := range watchFilteredPodPhases {
 		selectors = append(selectors, fmt.Sprintf("status.phase!=%s", phase))
 	}
 	selector := strings.Join(selectors, ",")
@@ -103,7 +108,7 @@ func registerSchedulerPodInformer(informerFactory informers.SharedInformerFactor
 			metav1.NamespaceAll,
 			resyncPeriod,
 			k8scache.Indexers{k8scache.NamespaceIndex: k8scache.MetaNamespaceIndexFunc},
-			filterTerminalPods,
+			filterFailedPods,
 		)
 	})
 }
--- a/pkg/scheduler/cache/pod_transform.go
+++ b/pkg/scheduler/cache/pod_transform.go
@@ -33,8 +33,6 @@ func compactSchedulerPod(obj any) (any, error) {
 	return compact, nil
 }
 
-// A succeeded pod is never placed on a node or counted towards node resources, so only the
-// identity, podgroup, subgroup, app and phase that snapshot assembly reads are retained.
 func compactSucceededPod(pod *v1.Pod) *v1.Pod {
 	compact := &v1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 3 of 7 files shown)


.changes/unreleased/Changed-20260801-235945.yaml
enoodle, line 1: That doesn't really interesting for someone that uses  KAI, we don't need this.

pkg/scheduler/cache/cache.go
```diff
@@ -83,9 +83,14 @@ var terminalPodPhases = []v1.PodPhase{
 	v1.PodFailed,
 }
 
-func filterTerminalPods(options *metav1.ListOptions) {
-	selectors := make([]string, 0, len(terminalPodPhases))
-	for _, phase := range terminalPodPhases {
+// Succeeded pods are watched so PodGroupInfo.IsStale can tell a completing gang from a stalled one.
```
enoodle: please remove this line, we can't explain every bug in the code. it is usually just LLM leftovers.

enoodle (changes requested):
Thanks for the fix, I have 2 minor comments.

pkg/scheduler/cache/pod_transform.go
```diff
@@ -26,6 +33,37 @@ func compactSchedulerPod(obj any) (any, error) {
...
+		},
+		Status: v1.PodStatus{
+			Phase: pod.Status.Phase,
+		},
+	}
+
+	if podGroup, found := pod.Annotations[commonconstants.PodGroupAnnotationForPod]; found {
+		compact.Annotations = map[string]string{commonconstants.PodGroupAnnotationForPod: podGroup}
+	}
+	for _, key := range []string{commonconstants.SubGroupLabelKey, commonconstants.AppLabelName} {
```
enoodle: Why do we need the AppLabelName ?

.changes/unreleased/Changed-20260801-235945.yaml
Thezone-1: Removed. The Fixed fragment for the eviction bug is the only one left.

pkg/scheduler/cache/cache.go
Thezone-1: Removed. Also dropped the equivalent comment above compactSucceededPod in pod_transform.go.

pkg/scheduler/cache/pod_transform.go
Thezone-1: Good catch, we do not. Removed.

The label is only read by IsResourceReservationTask / IsScaleAdjustTask, and neither applies to a succeeded pod:

- node_info.addTaskResources never runs, since AddTasksToNode gates on IsActiveUsedStatus and Succeeded is not in activeUsedStatuses
- the proportion plugin gates on the same thing
- the last caller, the reservation split in getNodeToPodInfosMap, was actually harmful here: the compacted pod has no Spec, so NodeName is empty, and only nodePodInfosMap[noNodeName] gets drained into existingPodsMap. nodeReservationPodInfosMap under an empty node name is never consumed, so keeping the label silently dropped a succeeded reservation pod from the snapshot.

Only SubGroupLabelKey is retained now, and the loop collapses to a single lookup.

gshaibi (approved):
Looks good. Thank you.

enoodle (approved):

Issue #1906, opened 2026-07-14
Research & design support for resize requests in KAI


keinanDaniel:
### What you would like to be added?

Research & design support for resize requests in KAI

### Why is this needed?

Research & design support for resize requests in KAI

gshaibi:
Assumed Requirements:
1. **Forbid Quota Violation due to resize** - Alice's team has a queue with a 10 GPU quota. Their job is running with 8 GPUs. Alice submits a resize to 12 GPUs — KAI rejects it because it would exceed the queue's quota.
2. **Perform fair evictions for Deferred pod resizes** - Bob's pod was resized from 4 to 8 CPUs and the kubelet marked it Deferred — the node doesn't have 4 free CPUs. KAI evicts a lower-priority preemptible pod on the same node, freeing enough capacity for the kubelet to actuate the resize.
3. **Account resize allocations correctly**
   a. Carol resizes her pod's memory from 8 GiB to 16 GiB. The kubelet marks the resize Infeasible — the node can never fit 16 GiB for that container. KAI ignores the new spec and continues accounting for 8 GiB, so the queue isn't charged for resources that will never be allocated.
   b. Carol resizes her pod's memory from 16 GiB to 8 GiB. Before the kubelet shrinks the container, KAI still accounts for 16 GiB — it won't phantom-release capacity and schedule another pod into resources that aren't free yet.

Issue #1894, opened 2026-07-13
Pod request accounting is fragmented: resize status, pod-level resources, and GPU overhead


thc1006:
# Pod request accounting is fragmented: resize status, pod-level resources, and GPU overhead

Found while working on #1880 (the PodGroup status producer only summed regular containers). Before changing the producer I checked what upstream actually does, against the versions this repo effectively builds with. Note the `replace` directives, `go list -m` is the thing to read, not the `require` block:

```
k8s.io/api               v0.35.4   (required at v0.36.2, replaced down)
k8s.io/apimachinery      v0.35.4   (required at v0.36.2, replaced down)
k8s.io/component-helpers v0.35.4
k8s.io/kubernetes        v1.35.4
```

So this is a Kubernetes 1.35 library line throughout, and every source link below is pinned to the `v1.35.4` tag rather than the moving `release-1.35` branch.

## Scope and decision requested

This is an umbrella/design issue, not a request to land three changes in one PR. Each one can be fixed, or documented as intentional, on its own.

The decisions I am asking for:

1. For a pod already placed on a node, should KAI's allocated accounting use the status-aware effective request that kube-scheduler and ResourceQuota use, or is desired-spec accounting a deliberate choice?
2. Should the scheduler's and the status producer's accounting honour pod-level resource overrides (cpu, memory, hugepages)?
3. Is dropping NVIDIA and AMD GPUs and MIG devices from Pod overhead intentional, or an artifact of the `BaseResource` / `GpuResourceRequirement` split?
4. Should these paths sit behind shared accounting primitives with explicit modes, rather than three hand-written aggregations?

This can be closed once each divergence is either corrected or written down as intentional.

## Minimal counterexamples

| case | input | Kubernetes 1.35.4 | KAI today |
|---|---|---:|---:|
| pending shrink | `spec.cpu=1`, actuated `cpu=4`, allocated `cpu=4` | 4 CPU | 1 CPU |
| pod-level override | container sum `cpu=1`, `pod.spec.resources.requests.cpu=4` | 4 CPU | 1 CPU |
| GPU in overhead | container `nvidia.com/gpu=1`, overhead `nvidia.com/gpu=1` | 2 GPUs | 1 GPU |

Each of these can be pinned as a focused unit test by constructing the Pod object directly. The resize case does not need a live kubelet to show the calculator divergence, only a pod whose status carries the actuated and allocated values.

## What is already correct

[...]

PR #2021, opened 2026-08-06
feat(scheduler,admission): implement in-place pod resize accounting and queue admission

gshaibi:
## Description

Implements the design from #1997 (research issue #1906).

**Effective-request accounting (KEP-1287 / in-place pod vertical scaling)**

When Kubernetes performs an in-place pod resize, the pod may temporarily be running with different resources than its spec declares. This PR makes the scheduler charge for the _effective_ resource consumption rather than the raw spec:

- `pkg/scheduler/cache/pod_transform.go` — Preserves `ContainerStatus.AllocatedResources` and `ContainerStatus.Resources` (enacted resources) in the informer compact transform so resize state is visible to the scheduler without increasing informer memory footprint unnecessarily. Also retains the `PodResizePending` condition.
- `pkg/scheduler/api/pod_info/effective_requests.go` — Implements the effective-request model: `max(spec, enacted, allocated)` per container/resource for normal/Deferred/InProgress resizes; `max(enacted, allocated)` when `PodResizePending=Infeasible` (so an infeasible target does not inflate accounting).
- `pkg/scheduler/api/pod_info/pod_info.go` — `getPodResourceWithoutInitContainers` now uses the effective-request model.

**Best-effort `pods/resize` admission webhook**

A new validating webhook rejects resize requests that would push a queue over its configured limits, without blocking legitimate resizes when the webhook is unavailable:

- `pkg/admission/webhook/v1alpha2/podhooks/pod_resize_validator.go` — Computes `delta = max(proposed - old spec, 0)` and walks the queue hierarchy. Rejects if any queue would exceed its CPU/memory `Limit` (all workloads) or `Quota` (non-preemptible workloads only). Best-effort: lookup failures (PodGroup, Queue, PriorityClass) allow the resize.
- `cmd/admission/app/app.go` — Registers the webhook at `/validate--v1-pod-resize` with `failurePolicy=ignore` so webhook unavailability never blocks a legitimate resize.

## Related Issues

Implements the design from #1997 (research issue #1906).
Related: #1894 (addresses the resize-status fragment), #1872 (deferred resize eviction — separate track, #2051).

## Checklist

- [x] Self-reviewed

## Breaking Changes

None.

## Additional Notes

The design document added by the first commit in this branch (`docs: propose in-place pod resize accounting and queue admission`) describes the design rationale in full. The commits in this PR implement that design.

**Test coverage:**
- `pkg/scheduler/api/pod_info/effective_requests_test.go` — 7 tests covering: no-status fallback, normal upsize uses max, downsize-in-progress uses enacted, infeasible excludes spec, multiple containers, `isPodResizeInfeasible`, `maxResourceList`
- `pkg/admission/webhook/v1alpha2/podhooks/pod_resize_validator_test.go` — 9 tests covering: non-KAI pod allowed, no PodGroup allowed, downsize allowed, CPU limit exceeded denied, CPU limit not exceeded allowed, memory limit exceeded denied, unlimited limit allowed, non-preemptible quota exceeded denied, preemptible-only quota exceeded allowed


.changes/unreleased/changed-20260810-114327.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Changed
+body: |-
+  Queue allocated status now includes init-container peak and sidecar requests
```
davidLif, line 3: Why both in the same PR?

cmd/admission/app/app.go
```diff
@@ -159,6 +167,21 @@ func (app *App) Run() error {
 		return err
 	}
 
+	// The canonical /validate--v1-pod path is taken by the pod validator above,
```
itsomri: Would this mean that the apiserver <> admission will have 2 api calls per pod? Why not just add this validation to the existing endpoint?

cmd/admission/app/options.go
```diff
@@ -99,6 +101,16 @@ func InitOptions() *Options {
 		fmt.Sprintf("Runtime class to be set for GPU fraction pods (defaults to %s). "+
 			"Whole-GPU pods are not affected. Set to empty string to disable.",
 			constants.DefaultRuntimeClassName))
+	fs.BoolVar(&options.ValidatePodResizeQuota,
+		"validate-pod-resize-quota", true,
+		"Enable best-effort hierarchical queue quota checks on pods/resize requests. "+
```
itsomri: What does "best-effort" mean in this context?

docs/developer/designs/in-place-pod-resize/README.md
```diff
@@ -0,0 +1,161 @@
+<!--
+Copyright 2026 NVIDIA CORPORATION
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# In-Place Pod Resize Accounting and Queue Admission
+
+*Status: Proposed*
```
itsomri: Status here should be done - once we merge this PR the feature is implemented

cmd/admission/app/app.go
gshaibi: No — the rules are disjoint, so no request is ever validated twice: this endpoint matches only UPDATE on the `pods/resize` subresource, which the existing `pods` rule doesn't cover. They also can't share a webhook entry because the failure policies differ (resize is fail-open `Ignore`, pod validation is `Fail`). Expanded the comment here to say so (59b52464).

cmd/admission/app/options.go
gshaibi: Fail-open: a resize is admitted rather than blocked when the webhook or its queue/podgroup lookups are unavailable, and concurrent activity can transiently cross a limit (details in docs/in-place-resize). Spelled this out in the flag help (59b52464).

docs/developer/designs/in-place-pod-resize/README.md
gshaibi: Updated to *Status: Implemented* (b44d055d).

README.md
```diff
@@ -43,6 +43,7 @@ It can run alongside other schedulers installed on the cluster.
 - [Min-guaranteed-runtime](https://github.com/kai-scheduler/KAI-scheduler/tree/main/docs/developer/designs/min-runtime): ensures a time period in which the scheduler must not preempt or reclaim a running workload, even if preemptible.
 - Workload Consolidation: Reallocate running workloads intelligently to reduce fragmentation and increase cluster utilization.
 - [Elastic Workloads](docs/elastic/README.md): Dynamically scale workloads within defined minimum and maximum pod or SubGroup thresholds.
+- [In-Place Pod Resize](docs/in-place-resize/README.md): Accurate queue accounting for in-place CPU/memory resizes (KEP-1287), with admission checks against queue limits and quota.
```
davidLif: Is this feature important enough to mention it in the main readme?

test/e2e/suites/resize/resize_test.go
```diff
@@ -0,0 +1,183 @@
...
+const (
+	resizeTimeout = 2 * time.Minute
+	resizePoll    = 2 * time.Second
+)
+
+var _ = Describe("In-place pod resize", func() {
+	// setupTest must run inside each It: GetConnectivity captures the node's
+	// context, which ginkgo cancels when the node exits, so a TestContext
+	// created in BeforeEach is unusable by the time the It body runs.
+	setupTest := func(ctx context.Context) *testcontext.TestContext {
```
davidLif: Please use our standard way for setting these things. For an example, you can look at `DescribeAllocateElasticSpecs`

davidLif: This should include cleanup

test/e2e/suites/resize/resize_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+		parentQueue := queue.CreateQueueObject(utils.GenerateRandomK8sName(10), "")
+		childQueue := queue.CreateQueueObject(utils.GenerateRandomK8sName(10), parentQueue.Name)
+		childQueue.Spec.Resources.CPU = v2.QueueResource{Quota: 500, OverQuotaWeight: 1, Limit: 2000}
+		testCtx.InitQueues([]*v2.Queue{childQueue, parentQueue})
+
+		pod := runPodWithCPU(ctx, testCtx, "500m")
+
+		err := resizeCPU(ctx, testCtx, pod, "3")
+		Expect(err).To(HaveOccurred(), "upsize past the queue limit should be denied")
```
davidLif: If there is a config option to enable/disable this part, please add a function that makes sure to patch the kai config accordingly and undo the change as a defer. For an ecxample, see `SetPlacementStrategy` and it's usgae. This applies to the other tests here as well.

pkg/operator/operands/admission/admission_test.go
```diff
@@ -83,8 +83,8 @@ var _ = Describe("Admission", func() {
 				Expect(mutatingWebhook.Webhooks).To(HaveLen(1))
 				Expect(mutatingWebhook.Webhooks[0].ClientConfig.CABundle).To(Equal(secret.Data[certKey]))
 
-				// One webhook for pods, one for the Topology one-to-one alias validation.
-				Expect(validatingWebhook.Webhooks).To(HaveLen(2))
+				// Webhooks: pods (Fail), pods/resize (Ignore), topology (Fail).
```
davidLif: What do you mean by "ignore"?

davidLif (commented):
ReclaimUnschedulableDistributedJob - degredation. This might be due to AggregateContainerRequests

.changes/unreleased/changed-20260810-114327.yaml
gshaibi: No special reason, the webhook is dependent on the new accounting, but not the other way around.
I can split it if you prefer.

README.md
gshaibi: You're right, I'll drop it.

pkg/operator/operands/admission/admission_test.go
gshaibi: That's the webhook failurePolicy. Refined the comment to denote that.

pkg/operator/operands/admission/admission_test.go
```diff
@@ -83,8 +83,9 @@ var _ = Describe("Admission", func() {
 				Expect(mutatingWebhook.Webhooks).To(HaveLen(1))
 				Expect(mutatingWebhook.Webhooks[0].ClientConfig.CABundle).To(Equal(secret.Data[certKey]))
 
-				// One webhook for pods, one for the Topology one-to-one alias validation.
-				Expect(validatingWebhook.Webhooks).To(HaveLen(2))
+				// failurePolicy per webhook: pods=Fail, topology=Fail, pods/resize=Ignore
```
itsomri: What does this comment have to do with anything, the failPolicy is not even validated here, looks like claude got fixated on that

pkg/operator/operands/admission/resources.go
```diff
@@ -318,6 +318,29 @@ func (a *Admission) validatingWCForKAIConfig(
 				},
 			},
 		},
+		{
+			// pods/resize is best-effort: failurePolicy=Ignore so webhook unavailability
+			// never blocks a legitimate in-place pod resize.
```
itsomri: this comment, if needed at all, should be above the FailurePolicy config. And I think it's really redundant at this point

cmd/admission/app/app.go
```diff
@@ -159,6 +167,24 @@ func (app *App) Run() error {
 		return err
 	}
 
+	// Separate endpoint, not an addition to the pod validator above: the rules
```
itsomri: This can be significantly simplified, something like "Add a new webhook for the resize subresource with Ignore failPolicy"

cmd/admission/app/options.go
```diff
@@ -99,6 +101,18 @@ func InitOptions() *Options {
 		fmt.Sprintf("Runtime class to be set for GPU fraction pods (defaults to %s). "+
 			"Whole-GPU pods are not affected. Set to empty string to disable.",
 			constants.DefaultRuntimeClassName))
+	fs.BoolVar(&options.ValidatePodResizeQuota,
+		"validate-pod-resize-quota", true,
```
itsomri, line 105: These are also very verbose. Use a short description like the rest of the cli args and document more deeply in the new doc

pkg/operator/operands/admission/admission_test.go
gshaibi: fixed

pkg/operator/operands/admission/resources.go
gshaibi: removed

docs/in-place-resize/README.md
```diff
@@ -0,0 +1,93 @@
...
+  admission:
+    inPlacePodResize:
+      validateQuota: true              # default
+      blockUpsizeOnBoundedQueues: false # default
+```
+
+| Field | Default | Effect |
+|---|---|---|
+| `validateQuota` | `true` | Master switch. When `false`, the webhook admits all resizes without any checks, and `blockUpsizeOnBoundedQueues` is ignored. |
+| `blockUpsizeOnBoundedQueues` | `false` | Strict mode: reject **any** upsize on a queue (or ancestor) that has a finite CPU or memory limit — regardless of free headroom — and, for non-preemptible workloads, on any queue with a finite quota. Closes the concurrent-resize race at the cost of disallowing upsizes on bounded queues entirely. |
```
itsomri, line 87: Why is this implemented only for limits and there’s nothing similar for quota and non-preemptibles? Was this discussed in the design?

pkg/admission/webhook/v1alpha2/podhooks/pod_resize_validator.go
```diff
@@ -0,0 +1,245 @@
...
+		isPreemptible = true
+	}
+
+	delta := podResizeDelta(oldPod, newPod)
+	if len(delta) == 0 {
+		return nil
+	}
+
+	queueName := pg.Spec.Queue
+	for queueName != "" {
```
itsomri, line 98: `validateResize` now does D (depth of queues) API requests per pod update. Maybe it’s nothing, but maybe it’s worth testing in a real, busy cluster? A pod can update several times in a normal lifecycle. Maybe add some more early exits - no sense in checking a pod that doesn’t have a nodeName, or is being deleted, or is still pending, etc. Or is there?
Could be a premature optimization maybe. IDK

itsomri (dismissed):

docs/in-place-resize/README.md
gshaibi: It is not only for limits, its for quota too

pkg/admission/webhook/v1alpha2/podhooks/pod_resize_validator.go
gshaibi: Good point, I'll add an early exit.

itsomri (approved):

PR #2028, opened 2026-08-07
fix(binder): avoid cache-lag race between ConfigMap create and update in GPU sharing PreBind

### LINKED ISSUE #2016: Flaky E2E: Fraction GPU request - missing GPU-sharing ConfigMap
## Failing test

`Schedule pod with resource request > GPU Resources > Fraction GPU request - fill all the GPUs`

`test/e2e/suites/allocate/resources/resources_specs.go:95`

## Failure message

```
Expected pod to be scheduled
...
BindingError Failed to bind pod kai-zgytixqyjg/uabomgfcqu to node kind-worker4: plugin gpusharing failed in PreBind: failed to update gpu sharing configmap uabomgfcqu-dh8gw9t-shared-gpu-0 for pod <kai-zgytixqyjg/uabomgfcqu>: ConfigMap "uabomgfcqu-dh8gw9t-shared-gpu-0" not found
```

The same pod was subsequently reported unschedulable because no node had enough GPU resources.

## CI run

https://github.com/kai-scheduler/KAI-Scheduler/actions/runs/31009536709/job/92325039326?pr=2015

This occurred on PR #2015 (`fix(ci): preserve DCO signatures in backports`), which changes only `.github/workflows/backport.yaml` and does not touch E2E, scheduling, binding, or GPU-sharing code.

## Related history

This is the same test covered by closed #1188, but with a different symptom: #1188 observed an unexpected allocated-pod count, whereas this run failed when the GPU-sharing ConfigMap was missing during `PreBind`.

srujanreddya:
For pods admitted by the current webhook (`NVIDIA_VISIBLE_DEVICES` via `ConfigMapKeyRef`), write all capabilities ConfigMap data in a single `UpsertJobConfigMap` call instead of creating an empty ConfigMap and then updating it via separate Get+Patch calls. This eliminates the race where the informer cache has not yet reflected the newly created ConfigMap by the time `UpdateConfigMapEnvironmentVariable` calls Get, causing the bind to fail with "ConfigMap not found".

Also handle AlreadyExists in UpsertJobConfigMap for concurrent reconcile safety.

Fixes #2016


pkg/binder/plugins/gpusharing/gpu_sharing.go
enoodle, line 1: The changes here seem to be from a different issue, lets remove from this PR.

pkg/binder/common/gpusharingconfigmap/config_map.go
```diff
@@ -61,12 +61,19 @@ func UpsertJobConfigMap(ctx context.Context,
...
+		createErr := kubeClient.Create(ctx, desiredConfigMap)
+		if createErr == nil {
+			return nil
+		}
+		if !errors.IsAlreadyExists(createErr) {
 			return fmt.Errorf("failed to create configmap %s/%s for pod %s/%s, error: %v",
+				pod.Namespace, desiredConfigMap.Name, pod.Namespace, pod.Name, createErr)
+		}
+		// Concurrent create: another goroutine created it first; re-fetch for patching.
+		if err = kubeClient.Get(ctx, client.ObjectKey{Name: desiredConfigMap.Name, Namespace: pod.Namespace}, existingCm); err != nil {
```
enoodle: If the previous Get failed, why would this one succeed?

enoodle (changes requested):

pkg/binder/plugins/gpusharing/gpu_sharing.go
enoodle: I now understand that this is the real fix and the AlreadyExists retry is not really doing much. I will review this again.

pkg/binder/plugins/gpusharing/gpu_sharing.go
```diff
@@ -60,33 +61,55 @@ func (p *GPUSharing) PreBind(
...
+		capabilitiesConfigMapName, err := gpusharingconfigmap.ExtractCapabilitiesConfigMapName(pod, containerRef)
+		if err != nil {
+			return fmt.Errorf("failed to get capabilities configmap name: %w", err)
+		}
+		data := map[string]string{
+			constants.NvidiaVisibleDevices: nVisibleDevicesStr,
+			common.NumOfGpusEnvVarBC:       bindRequest.Spec.ReceivedGPU.Portion,
+			common.GPUPortion:              bindRequest.Spec.ReceivedGPU.Portion,
+		}
+		return gpusharingconfigmap.UpsertJobConfigMap(ctx, p.kubeClient, pod, capabilitiesConfigMapName, data)
```
enoodle: This should be implemented in pkg/binder/common/gpu_access.go - like the similar code from SetNvidiaVisibleDevices

pkg/binder/plugins/gpusharing/gpu_sharing.go
```diff
@@ -60,33 +61,55 @@ func (p *GPUSharing) PreBind(
...
+
+	// Backward compat: NVIDIA_VISIBLE_DEVICES is served via envFrom from the direct env vars ConfigMap.
+	capabilitiesConfigMapName, err := gpusharingconfigmap.ExtractCapabilitiesConfigMapName(pod, containerRef)
 	if err != nil {
+		return fmt.Errorf("failed to get capabilities configmap name: %w", err)
+	}
+	if err = gpusharingconfigmap.UpsertJobConfigMap(ctx, p.kubeClient, pod, capabilitiesConfigMapName, map[string]string{}); err != nil {
+		return fmt.Errorf("failed to create capabilities configmap: %w", err)
+	}
+	if err = common.SetNvidiaVisibleDevices(ctx, p.kubeClient, pod, containerRef, nVisibleDevicesStr); err != nil {
```
enoodle: thats basically the same as the old implementation. But now SetNvidiaVisibleDevices will only be called for the backwards compatibility case, so some if it is dead code.

enoodle (changes requested):
I think we should implement this in pkg/binder/common/gpu_access.go combining SetNvidiaVisibleDevices and SetGPUPortion into one function that will create the config map with all needed data in one go. This way we will not have multiple functions doing the same thing here and there.

pkg/binder/plugins/gpusharing/gpu_sharing.go
srujanreddya: Thanks. Addressed.

pkg/binder/plugins/gpusharing/gpu_sharing.go
srujanreddya: Addressed.

pkg/binder/common/gpusharingconfigmap/config_map.go
```diff
@@ -61,11 +61,15 @@ func UpsertJobConfigMap(ctx context.Context,
...
+		createErr := kubeClient.Create(ctx, desiredConfigMap)
+		if createErr == nil {
+			return nil
+		}
+		if !errors.IsAlreadyExists(createErr) {
 			return fmt.Errorf("failed to create configmap %s/%s for pod %s/%s, error: %v",
-				pod.Namespace, desiredConfigMap.Name, pod.Namespace, pod.Name, err)
+				pod.Namespace, desiredConfigMap.Name, pod.Namespace, pod.Name, createErr)
 		}
+		// Concurrent create: the other creator wrote the same data; nothing more to do.
```
enoodle: who is the other creator? The config map is only created once per Pod. I think this shouldn't be part of this PR.

srujanreddya: Fixed.

pkg/binder/plugins/gpusharing/gpu_sharing.go
enoodle: I still think that `SetNvidiaVisibleDevices` + `SetGPUPortion` is very similar to `UpsertCapabilitiesConfigMapData` - we can have their "logic" merged into `UpsertCapabilitiesConfigMapData` so that we will only have one map write. Not sure which map, and what exact values, but only one write.

To be honest I don't know how important that backwards compatibility is still, but I think that if we have it at least wrap it nicely like the rest of the code.

srujanreddya: You're right that the backward compat path is inconsistent, it does a create-empty + Get+Patch pair for each ConfigMap instead of a single-shot upsert. It can't collapse to one write because `NVIDIA_VISIBLE_DEVICES` goes to the direct env vars ConfigMap while `GPU_PORTION` goes to the capabilities ConfigMap. We could reduce it to two single-shot upserts for consistency, or drop the backward compact path entirely if old (pre-webhook) pods are no longer a concern. What's your preference?

srujanreddya: **The actual options:**

  1. Reduce to 2 upserts — replace createDirectEnvMapIfMissing + SetNvidiaVisibleDevices with one UpsertJobConfigMap(directEnvMap, {NVIDIA_VISIBLE_DEVICES: val}),   
  and replace UpsertJobConfigMap(empty) + SetGPUPortion with one UpsertJobConfigMap(capMap, {GPU_PORTION: portion, RUNAI_NUM_OF_GPUS: portion}). Same behavior,
  cleaner.                                                                                                                                                           
  2. Drop backward compat entirely — if there are no pre-webhook pods still running in any cluster KAI targets, just delete the old path. That leaves a single
  UpsertCapabilitiesConfigMapData call and no SetNvidiaVisibleDevices/SetGPUPortion at all.

srujanreddya: @enoodle I have made changes as per your request

Added `UpsertBackwardCompatConfigMapData` in `gpu_access.go` that combines the old `SetNvidiaVisibleDevices` + `SetGPUPortion` logic into two single-shot `UpsertJobConfigMap` calls one for the direct env vars ConfigMap `(NVIDIA_VISIBLE_DEVICES) `and one for the capabilities ConfigMap
  `(GPU_PORTION/RUNAI_NUM_OF_GPUS)`. No more create-empty + Get+Patch pattern in the backward compapath.
  **Also removed:**
  - `SetNvidiaVisibleDevices` and `SetGPUPortion` (dead code now)
  - `createDirectEnvMapIfMissing` (was creating an empty CM just to patch it immediately after;
  `UpsertJobConfigMap` handles the create case atomically)
  - The `RUNAI-VISIBLE-DEVICES` conditional update (agreed this was dead code)

PR #2049, opened 2026-08-10
fix(binder): preserve reservations during bind cache lag

davidLif:
## Description

Fixes reservation pod cleanup during BindRequest/pod cache lag.

When `SyncForGpuGroup` sees a reservation pod but does not see any fraction pods for the GPU group, it uses BindRequests as a guard before deleting the reservation. The previous guard ignored succeeded BindRequests. In practice, the BindRequest status can advance before the pod GPU-group label is visible through the client cache, so cleanup could delete the reservation pod in the middle of a successful bind.

This keeps succeeded BindRequests protective while their target pod is still live. Failed BindRequests and succeeded BindRequests whose pod is gone or terminal do not protect the reservation.

## Related Issues

N/A

## Checklist

- [x] Self-reviewed

## Breaking Changes

None.

## Additional Notes

Validation run:

```bash
go test ./pkg/binder/binding/resourcereservation -count=1
KUBEBUILDER_ASSETS="$(pwd)/$(bin/setup-envtest use 1.34.0 -p path --bin-dir bin)" go test ./pkg/binder/controllers/integration_tests -run TestAPIs -count=1
git diff --check
```

### DIFF
--- a/pkg/binder/binding/resourcereservation/resource_reservation.go
+++ b/pkg/binder/binding/resourcereservation/resource_reservation.go
@@ -226,9 +226,9 @@ func (rsc *service) syncForPods(ctx context.Context, pods []*v1.Pod, gpuGroupToS
 	return nil
 }
 
-// hasActiveBindRequestsForGpuGroup checks if any non-terminal BindRequests reference
-// the given GPU group. This prevents premature reservation pod deletion when the
-// informer cache has not yet propagated GPU group labels on recently-bound fraction pods.
+// hasActiveBindRequestsForGpuGroup checks if BindRequests still protect the GPU group.
+// A succeeded BindRequest also protects the reservation while its pod is still alive:
+// the pod label can lag behind the BindRequest status in the controller cache.
 func (rsc *service) hasActiveBindRequestsForGpuGroup(ctx context.Context, gpuGroup string) (bool, error) {
 	bindRequestList := &schedulingv1alpha2.BindRequestList{}
 	if err := rsc.kubeClient.List(ctx, bindRequestList); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/binder/binding/resourcereservation/resource_reservation.go
```diff
@@ -226,26 +226,66 @@ func (rsc *service) syncForPods(ctx context.Context, pods []*v1.Pod, gpuGroupToS
...
+		return false, fmt.Errorf("failed to get pod for BindRequest <%s/%s>: %w",
+			bindRequest.Namespace, bindRequest.Name, err)
+	}
+
+	if slices.Contains([]v1.PodPhase{v1.PodSucceeded, v1.PodFailed}, pod.Status.Phase) {
+		return false, nil
+	}
+
+	for _, gpuGroup := range bindRequest.Spec.SelectedGPUGroups {
+		if slices.Contains(resources.GetGpuGroups(pod), gpuGroup) {
```
gshaibi, line 282: What is the effect of this loop? We return true afterwards anyway, right?

pkg/binder/binding/resourcereservation/resource_reservation.go
```diff
@@ -226,26 +226,66 @@ func (rsc *service) syncForPods(ctx context.Context, pods []*v1.Pod, gpuGroupToS
...
+		}
+
+		return true, nil
 	}
 	return false, nil
 }
+
+func (rsc *service) hasLivePodForBindRequest(ctx context.Context, bindRequest *schedulingv1alpha2.BindRequest) (bool, error) {
+	pod := &v1.Pod{}
+	err := rsc.kubeClient.Get(ctx, client.ObjectKey{
```
gshaibi, line 265: Consider also matching the UID of the pod (against the BindRequests's OwnerReference) since a pod can be recreated with the same name.

PR #2050, opened 2026-08-10
fix(scheduler): fail fast when feature gate discovery fails

### LINKED ISSUE #2038: DRA feature gate latches false permanently if the API server is briefly unreachable at scheduler startup
## What happened?

The DRA feature gate is derived from a single discovery call made once at process startup, and **any** error from that call is treated as "this cluster does not support DRA":

```go
// pkg/common/feature_gates/feature_gates.go#L50
serverVersion, err := discoveryClient.ServerVersion()
if err != nil {
	logger.Error(err, "Failed to get server version")
	return false
}
```

[`IsDynamicResourcesEnabled`](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.15.3/pkg/common/feature_gates/feature_gates.go#L46-L53) returns `false` on error, and [`SetDRAFeatureGate`](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.15.3/pkg/common/feature_gates/feature_gates.go#L26-L32) stores that result in a package-level `atomic.Bool`. It is called exactly once per process:

- scheduler — [`pkg/scheduler/cache/cache.go#L157`](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.15.3/pkg/scheduler/cache/cache.go#L157)
- binder — [`cmd/binder/app/app.go#L108`](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.15.3/cmd/binder/app/app.go#L108)

There is no retry and no re-evaluation. A transient failure — an API server restart, a rolling control-plane upgrade, a brief network partition, a CNI hiccup — is therefore indistinguishable from `resource.k8s.io` genuinely being absent, and the process stays with DRA disabled for its entire lifetime.

Once latched, every pod carrying a resource claim is rejected in preFilter at [`dynamicresources.go#L145-L152`](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.15.3/pkg/scheduler/plugins/dynamicresources/dynamicresources.go#L145-L152):

```
pod <ns>/<name> cannot be scheduled, it references resource claims <claim-name> while dynamic resource allocation is not available in the cluster
```

The message is misleading: DRA is available and healthy: the API group is served, the driver is running and publishing ResourceSlices, and ResourceClaims are created normally. Only this one process believes otherwise.

## Reproduction

1. Make the API server unreachable from the scheduler pod at the moment it starts (kill/restart the API server, or block egress to it briefly).
2. Let the scheduler container start inside that window. Its startup discovery call fails and the gate latches `false`.
3. Restore the API server. The scheduler connects normally, passes all health checks, and acquires the leader lease.
4. Submit any pod that references a ResourceClaim.

Expected: the pod schedules. Actual: it is rejected with the message above, indefinitely, on a cluster where DRA is fully functional. Restarting the process is the only recovery.

With multiple replicas and leader election, impact depends on which replica holds the lease: a replica that started during the outage will serve as a healthy-looking leader while rejecting 100% of DRA workloads, and failing over to a replica that started at a normal time restores DRA immediately. ResourceClaims move from `pending` to `allocated,reserved` within seconds of the failover, with no other change.

## Why it is hard to diagnose

Two properties turn a brief blip into a silent, long-lived outage:

1. **The error is never logged.** [`cmd/scheduler/main.go`](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.15.3/cmd/scheduler/main.go) (18 lines) is the only KAI binary that never calls `ctrl.SetLogger` — [binder#L33](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.15.3/cmd/binder/main.go#L33), [admission#L71](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.15.3/cmd/admission/app/app.go#L71), [queue-controller#L45](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.15.3/cmd/queuecontroller/main.go#L45), [podgrouper#L169](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.15.3/cmd/podgrouper/app/app.go#L169), [podgroup-controller#L29](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.15.3/cmd/podgroupcontroller/main.go#L29), [operator#L66](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.15.3/cmd/operator/app/app.go#L66), [resourcereservation#L94](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.15.3/cmd/resourcereservation/app/app.go#L94) and [nodescaleadjuster#L50](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.15.3/cmd/nodescaleadjuster/app/app.go#L50) all do. So `logger.Error(err, "Failed to get server version")` resolves to a discarding sink and nothing appears in the logs.

2. **Nothing can detect the degraded state.** The scheduler Deployment gets no liveness, readiness, or startup probe — [`operands/scheduler/resources_for_shard.go`](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.15.3/pkg/operator/operands/scheduler/resources_for_shard.go) sets none, unlike [`admission/resources.go#L60-L66`](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.15.3/pkg/operator/operands/admission/resources.go#L60-L66). The affected process stays Ready and keeps the lease indefinitely. There is also no metric exposing the gate, so the state is not observable or alertable from outside.

The net effect is a scheduler that reports healthy by every available signal while refusing all DRA work, with no log line and no metric pointing at the cause.

## Affected versions

Observed on v0.15.3. The code is identical on v0.16.x and [v0.17.0](https://github.com/kai-scheduler/KAI-Scheduler/blob/v0.17.0/pkg/common/feature_gates/feature_gates.go#L46-L53), so upgrading does not help.

## Relation to #1481

This is a follow-up to #1481, which wired this detection into the binder init path and describes `dynamicResourcesEnabled` as the authoritative DRA source for both scheduler and binder. That issue notes the upstream gate being GA-locked-to-true in K8s v1.35+ made this "not an urgent correctness issue today" — the failure above is the case where it does become one: on a DRA-capable cluster the authoritative source can be wrong, and once wrong it stays wrong.

---

A transient API server error at startup should not permanently disable DRA. Specifically:

1. **Distinguish "error" from "definitively absent."** A failed `ServerVersion()` / `ServerGroups()` call means *unknown*, not *unsupported*.
2. **Fail fast rather than degrade silently.** If discovery cannot be completed at startup, returning an error and exiting non-zero would surface the problem as a CrashLoopBackOff — visible, alertable, and self-correcting once the API server recovers. Alternatively, retry with backoff until a definitive answer is obtained.
3. **Re-evaluate rather than latch once**, so a process can recover without a restart.

Two smaller changes would cut diagnosis time from hours to minutes:

- Call `ctrl.SetLogger` in `cmd/scheduler`, so this and any other controller-runtime log line in the scheduler is actually emitted.
- Expose the gate as a gauge, so "this scheduler has DRA disabled" is observable and alertable.

Happy to open a PR for any of the above if the direction sounds right.

---

## Environment

- Kubernetes version: v1.34.x (`resource.k8s.io/v1` and `v1beta1` both served)
- KAI Scheduler version: v0.15.3 (code verified identical on v0.16.x and v0.17.0)
- Cloud provider or hardware configuration: not specific to any provider or hardware — the failure is in startup-time API discovery and applies to any DRA-capable cluster
- Tools that you are using KAI together with: not driver-specific; reproduces with any DRA driver, since the gate is evaluated before any driver or claim is inspected
- Anything else that is relevant: triggered by API server unavailability during scheduler pod startup. With leader election, the impact is visible only when the affected replica holds the lease.

hcnguyen-ai:
Fixes #2038.

## What

Both feature gates derive their value from a single discovery call at process startup, and treat *any* error from that call as "the cluster does not support this feature":

```go
serverVersion, err := discoveryClient.ServerVersion()
if err != nil {
	logger.Error(err, "Failed to get server version")
	return false
}
```

The result is stored in a package-level `atomic.Bool` with no retry and no re-evaluation, so a transient API server error — a restart, a rolling control-plane upgrade, a brief network partition — is indistinguishable from the API group genuinely being absent, and the process stays with the feature disabled for its entire lifetime.

In our case a ~5 minute API server outage caught one scheduler replica at startup. It latched DRA off, won the leader lease, and rejected every pod carrying a ResourceClaim for over two hours while reporting healthy, on a cluster where DRA was fully functional.

## Change

`IsDynamicResourcesEnabled` and `IsNodeResourceTopologyEnabled` now return `(bool, error)`. A discovery failure is returned as an error; `(false, nil)` is reserved for the feature genuinely not being served. `SetDRAFeatureGate` and `SetNodeResourceTopologyFeatureGate` return that error and **leave the stored gate untouched on failure**, so a failed call can never silently flip a gate off.

Callers treat it as fatal, per the discussion in #2038 — a component that cannot determine feature availability at startup exits non-zero and crashloops visibly, rather than running degraded. It also self-corrects once the API server recovers.

Propagated through:

- `pkg/scheduler/cache/cache.go` — `New` and `newSchedulerCache` now return an error
- `pkg/scheduler/scheduler.go` and `cmd/snapshot-tool/main.go` — the two `cache.New` callers
- `cmd/binder/app/app.go`

The NodeResourceTopology gate is included because it has the same defect, as noted in the issue discussion.

Incidentally, `newSchedulerCache` already had two failure paths that logged and returned `nil` while callers dereferenced the result. Returning an error removes those latent nil dereferences.

The `ctrl.SetLogger` fix for `cmd/scheduler` — which is why the original error was never logged — is left for a separate PR as requested.

## Tests

New coverage in `pkg/common/feature_gates/feature_gates_test.go`:

- an error is returned when `ServerVersion` or `ServerGroups` fails, for both gates
- a failed `Set*FeatureGate` leaves the previously stored value unchanged
- first direct coverage of `IsNodeResourceTopologyEnabled` for the served / not-served cases

A small `erroringDiscovery` wrapper stands in for an unreachable API server, rather than depending on fake-client internals.

Existing call sites updated for the new signatures.

## Verification

- `go build ./...`, `go vet ./pkg/... ./cmd/...`, `gofmt -l ./pkg ./cmd` — all clean
- `go test ./pkg/...` — 169 packages pass. Four envtest-based packages (`binder/controllers/integration_tests`, `env-tests`, `operator/controller/integration_tests`, `queuecontroller/controllers`) fail in `BeforeSuite` with `fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory`, i.e. `KUBEBUILDER_ASSETS` is not set up in my environment. They fail before any changed code runs.

### DIFF
--- a/pkg/common/feature_gates/feature_gates.go
+++ b/pkg/common/feature_gates/feature_gates.go
@@ -27,9 +27,6 @@ const (
 // off to reflect server-side DRA availability.
 var dynamicResourcesEnabled atomic.Bool
 
-// SetDRAFeatureGate returns an error if DRA availability could not be determined.
-// Callers must treat that as fatal: the gate is evaluated once per process, so
-// defaulting to disabled would silently reject every DRA workload until restart.
 func SetDRAFeatureGate(discoveryClient discovery.DiscoveryInterface) error {
 	enabled, err := IsDynamicResourcesEnabled(discoveryClient)
 	if err != nil {
@@ -55,8 +52,6 @@ func SetDynamicResourcesEnabledForTest(enabled bool) {
 
 var nodeResourceTopologyEnabled atomic.Bool
 
-// SetNodeResourceTopologyFeatureGate returns an error if availability could not be
-// determined; see SetDRAFeatureGate for why callers must treat that as fatal.
 func SetNodeResourceTopologyFeatureGate(discoveryClient discovery.DiscoveryInterface) error {
 	enabled, err := IsNodeResourceTopologyEnabled(discoveryClient)
 	if err != nil {
@@ -75,8 +70,7 @@ func SetNodeResourceTopologyEnabledForTest(enabled bool) {
 }
 
 // IsNodeResourceTopologyEnabled reports whether the cluster serves the
-// topology.node.k8s.io API group (the NodeResourceTopology CRD). A discovery
-// failure is returned as an error rather than reported as "not served".
+// topology.node.k8s.io API group (the NodeResourceTopology CRD).
 func IsNodeResourceTopologyEnabled(discoveryClient discovery.DiscoveryInterface) (bool, error) {
 	serverGroups, err := discoveryClient.ServerGroups()
 	if err != nil {
@@ -91,8 +85,6 @@ func IsNodeResourceTopologyEnabled(discoveryClient discovery.DiscoveryInterface)
 	return false, nil
 }
 
-// IsDynamicResourcesEnabled reports whether the cluster supports DRA. A discovery
-// failure is returned as an error rather than reported as "not supported".
 func IsDynamicResourcesEnabled(discoveryClient discovery.DiscoveryInterface) (bool, error) {
 	// Get API server version
 	serverVersion, err := discoveryClient.ServerVersion()
--- a/pkg/common/feature_gates/feature_gates.go
+++ b/pkg/common/feature_gates/feature_gates.go
@@ -4,6 +4,7 @@
 package featuregates
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 	"sync/atomic"
@@ -11,7 +12,6 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/version"
 	discovery "k8s.io/client-go/discovery"
-	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 const (
@@ -27,8 +27,16 @@ const (
 // off to reflect server-side DRA availability.
 var dynamicResourcesEnabled atomic.Bool
 
-func SetDRAFeatureGate(discoveryClient discovery.DiscoveryInterface) {
-	dynamicResourcesEnabled.Store(IsDynamicResourcesEnabled(discoveryClient))
+// SetDRAFeatureGate returns an error if DRA availability could not be determined.
+// Callers must treat that as fatal: the gate is evaluated once per process, so
+// defaulting to disabled would silently reject every DRA workload until restart.
+func SetDRAFeatureGate(discoveryClient discovery.DiscoveryInterface) error {
+	enabled, err := IsDynamicResourcesEnabled(discoveryClient)
+	if err != nil {
+		return err
+	}
+	dynamicResourcesEnabled.Store(enabled)
+	return nil
 }
 
 // DynamicResourcesEnabled reports whether DRA was determined to be usable
@@ -47,8 +55,15 @@ func SetDynamicResourcesEnabledForTest(enabled bool) {
 
 var nodeResourceTopologyEnabled atomic.Bool
 
-func SetNodeResourceTopologyFeatureGate(discoveryClient discovery.DiscoveryInterface) {
-	nodeResourceTopologyEnabled.Store(IsNodeResourceTopologyEnabled(discoveryClient))
+// SetNodeResourceTopologyFeatureGate returns an error if availability could not be
+// determined; see SetDRAFeatureGate for why callers must treat that as fatal.
+func SetNodeResourceTopologyFeatureGate(discoveryClient discovery.DiscoveryInterface) error {
+	enabled, err := IsNodeResourceTopologyEnabled(discoveryClient)
+	if err != nil {
+		return err
+	}
+	nodeResourceTopologyEnabled.Store(enabled)
+	return nil
 }
 
 func NodeResourceTopologyEnabled() bool {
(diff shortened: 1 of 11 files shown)


pkg/common/feature_gates/feature_gates.go
```diff
@@ -27,8 +27,16 @@ const (
 // off to reflect server-side DRA availability.
 var dynamicResourcesEnabled atomic.Bool
 
-func SetDRAFeatureGate(discoveryClient discovery.DiscoveryInterface) {
-	dynamicResourcesEnabled.Store(IsDynamicResourcesEnabled(discoveryClient))
+// SetDRAFeatureGate returns an error if DRA availability could not be determined.
+// Callers must treat that as fatal: the gate is evaluated once per process, so
+// defaulting to disabled would silently reject every DRA workload until restart.
```
gshaibi: I recommend leaving this to the caller's discretion - they can implement their own retry logic if needed.

pkg/common/feature_gates/feature_gates.go
```diff
@@ -60,44 +75,40 @@ func SetNodeResourceTopologyEnabledForTest(enabled bool) {
...
 		}
 	}
-	return false
+	return false, nil
 }
 
-func IsDynamicResourcesEnabled(discoveryClient discovery.DiscoveryInterface) bool {
-	logger := log.Log.WithName("feature-gates")
-
+// IsDynamicResourcesEnabled reports whether the cluster supports DRA. A discovery
```
gshaibi: Could we remove this comment? The implementation seems clear enough on its own.

gshaibi (commented):
Looks great, thank you!
Left some minor comments.

pkg/common/feature_gates/feature_gates.go
hcnguyen-ai: Agreed, removed. Also dropped the matching comment on `SetNodeResourceTopologyFeatureGate`, since it referenced this one.

pkg/common/feature_gates/feature_gates.go
hcnguyen-ai: Removed. I left the existing comment on `IsNodeResourceTopologyEnabled` in place and just reverted the sentence I'd appended to it, since that one predates this PR — happy to drop it entirely if you'd prefer.

gshaibi (approved):
LGTM, approved.

davidLif (approved):

PR #2055, opened 2026-08-11
docs: Design gang scheduling of multiple workloads

itsomri:
## Description

This PR proposes a design for supporting grouping several independent workloads with different topOwners together.

## Related Issues

Fixes #

## Checklist

- [ ] Self-reviewed

### DIFF
--- a/docs/developer/designs/multi-workload-gang/README.md
+++ b/docs/developer/designs/multi-workload-gang/README.md
@@ -184,12 +184,7 @@ Input PodGroups continue to exist, so resource and scheduling status need to be 
 
 Today, PodGroup resource status is calculated from pods whose `pod-group-name` directly names that PodGroup. Under this design, input PodGroups therefore naturally retain resource totals while the rendered PodGroup has no directly annotated pods. The queue controller currently sums every PodGroup, so copying aggregate resources to the rendered object without filtering would double-count them.
 
-Two accounting models are possible:
-
-- Input PodGroups remain authoritative for resource status and queue accounting; the rendered PodGroup is excluded from resource summation and owns aggregate scheduling conditions. If the rendered priority or preemptibility differs from the inputs, queue accounting must reclassify input resources using the effective rendered policy rather than summing each input's `allocatedNonPreemptible` value.
-- The rendered PodGroup reports aggregate resources and input PodGroups are excluded from queue accounting while grouped.
-
-The selected model must also account for the rendered priority and preemptibility. If these differ from an input PodGroup, resource classification based on the input policy can disagree with the scheduler's effective policy.
+To avoid ambiguity, resources should only be reported on one of the levels. Since the podgroups' preemptibility can differ from the leaf queues' in the merge, the resource accounting would make more sense in the top-level podgroup. To formalize: resource allocation is only reported on podgroups with no `workloadGroupName`.
 
 Regardless of which object is authoritative, every pod's resources must be counted exactly once. An inactive rendered revision must not be counted in addition to its inputs, and activating a rendered revision must switch accounting without treating both representations as independent workloads. The queue controller needs one canonical activation predicate: for example, count the rendered root only when its UID/revision is active and all recorded inputs match that revision; otherwise count the inputs and ignore the inactive root. Status propagation remains eventually consistent, but one reconcile must never deliberately sum both representations.
 
--- a/docs/developer/designs/multi-workload-gang/README.md
+++ b/docs/developer/designs/multi-workload-gang/README.md
@@ -23,41 +23,41 @@ While these patterns can sometimes be expressed as a single hierarchical workloa
 
 ## Non-Goals
 
-- KAI should not own the individual workload objects beyond scheduling them. Users create workload objects directly or through third parties, the podgrouper may create ordinary input PodGroups for them, and the `WorkloadGang` controller composes those inputs.
+- KAI should not own the individual workload objects beyond scheduling them. Users create workload objects directly or through third parties, the podgrouper may create ordinary input PodGroups for them, and the `WorkloadGroup` controller composes those inputs.
 
 ## Design
 
 The hierarchical structure of PodGroups already provides the scheduling primitive needed to compose several gangs. This design will propose a mechanism to group several independent workloads together reliably.
 
-To provide a versatile solution that will work with both "internal" (PodGrouper-created PodGroups) and "external" (PodGroups created by users or 3rd party systems), this new flow will group PodGroup objects from the cluster. To prevent race conditions, several alternative data models are proposed. A new CRD, `WorkloadGang`, will represent the coupling of these PodGroups, including potential overrides to the workloadGroup's scheduling properties like preemptiblity and priority, topology constraints, etc. It will also help prevent race conditions by specifying the number of PodGroups expected to participate in the gang.
+To provide a versatile solution that will work with both "internal" (PodGrouper-created PodGroups) and "external" (PodGroups created by users or 3rd party systems), this new flow will group PodGroup objects from the cluster. To prevent race conditions, several alternative data models are proposed. A new CRD, `WorkloadGroup`, will represent the coupling of these PodGroups, including potential overrides to the workloadGroup's scheduling properties like preemptiblity and priority, topology constraints, etc. It will also help prevent race conditions by specifying the number of PodGroups expected to participate in the gang.
 
-The controller creates a **rendered PodGroup** containing the merged hierarchy and effective scheduling policy. Issues in the grouping process will be reported in `WorkloadGang.status` - for example, if some expected PodGroups are missing. A successful grouping will result in an actual PodGroup object in the cluster which the scheduler will treat like any other PodGroup.
+The controller creates a **rendered PodGroup** containing the merged hierarchy and effective scheduling policy. Issues in the grouping process will be reported in `WorkloadGroup.status` - for example, if some expected PodGroups are missing. A successful grouping will result in an actual PodGroup object in the cluster which the scheduler will treat like any other PodGroup.
 
 ### Proposed flow
 
-1. A user creates several workloads they want to be grouped together - for example, a `RayCluster` and a `DynamoGraphDeployment`. The user creates a `WorkloadGroup` instance with the expected number of participating PodGroups (2). The individual workloads submitted by the user include an annotation `kai.scheduler/workload-gang: <WorkloadGang object name>` to let the podgrouper / Dynamo to create their podgroups with a workloadGroupRef in their spec.
-2. Two podgroups are created alongside the objects mentioned before: one by the podgrouper for the RayCluster, the other by Grove for Dynamo. Both podgroups are created with the `workloadGroupRef` in spec, so the scheduler knows not to handle them prematurely - even if the `WorkloadGroup` creation is delayed.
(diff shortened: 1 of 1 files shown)


docs/developer/designs/multi-workload-gang/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+As use cases become more complex, and different workload types get more specialized, it becomes necessary to manage the lifecycle of several workloads together. Maybe the clearest use case is reinforcement learning (RL), which can require a complex set of workloads all operating together, but there are also simpler use cases, like some model serving patterns which require auxiliary deployment, e.g for authentication, routing, caching and so on. All these use cases require coupling the lifecycle of more than one workload together, but the users would benefit from having each workload managed by its own domain-specific controller.
+
+KAI's PodGroup API provides the necessary building block to support gang-of-gangs, due to its hierarchical structure and per-hierarchy-level minima and topology constraints - but today, a user has no way to group multiple workloads with different parent CRDs together, unless they build the entire PodGroup themselves - which is usually an unreasonable requirement from most researchers or engineers, as the grouping logic itself can be quite complicated and specialized, especially in complex use cases like disaggregated inference.
```
gshaibi, line 7: nit
```suggestion
KAI's PodGroup API provides the necessary building block to support gang-of-gangs, due to its hierarchical structure and per-hierarchy-level minima and topology constraints - but today, a user has no way to group multiple workloads with different parent CRs together, unless they build the entire PodGroup themselves - which is usually an unreasonable requirement from most researchers or engineers, as the grouping logic itself can be quite complicated and specialized, especially in complex use cases like disaggregated inference.
```

docs/developer/designs/multi-workload-gang/README.md
```diff
@@ -0,0 +1,214 @@
...
+```text
+rendered PodGroup root
+├── flat input PodGroup wrapper (preserved root minMember and topology)
+└── hierarchical input PodGroup wrapper (preserved root minSubGroup and topology)
+    └── prefixed copy of the input SubGroup tree
+```
+
+For initial implementation, we will require all podgroups to be ganged together: root minSubGroups = # of podgroups. We can support finer granularity if such use cases are provided.
+
+To avoid conflicts, every podgroup in the WorkloadGang can get a short prefix/suffix for it's subgroup names (to avoid collisions like two podgroups having a "workers" subgroup). The workloadGroup status can list these prefixes so the scheduler will be able to infer pod > subgroup.
```
gshaibi, line 101: Pod reporting should contain the generated podgroup's subgroup names and not their original subgroup name, right?

docs/developer/designs/multi-workload-gang/README.md
```diff
@@ -0,0 +1,214 @@
...
+    - name: reward-service
+      uid: 0fbc89fa-5e4f-4214-bb66-af411453a7db
+      observedGeneration: 1
+      owners:
+        - apiVersion: apps/v1
+          kind: Deployment
+          name: reward-service
+      rootSubGroup: reward-service
+  conditions:
+    - type: InputsResolved
```
gshaibi, line 139: What happens if a PodGroup is presented after expectedPodGroups were already presented? Will it just be ignored? Will there be an error? Undefined bahavior?

docs/developer/designs/multi-workload-gang/README.md
```diff
@@ -0,0 +1,214 @@
...
+| Field | Proposed behavior or decision required |
+| --- | --- |
+| `metadata.name` | Derive a scheduler-global collision-resistant name from the `WorkloadGroup`, for example `wg-<name>`. The scheduler currently identifies PodGroups by name rather than namespaced name, which needs to be fixed internally. |
+| `metadata.namespace` | Use the `WorkloadGroup` namespace and require every input PodGroup and pod to use it. |
+| `metadata.ownerReferences` | Set the `WorkloadGroup` as the sole controller owner. |
+| Annotations and labels | Do not inherit anything. |
+| `queue` | All input podgroups must belong to the same queue. |
+| `priorityClassName` | Use an explicit group value when configured. If not, use the minimum value of all participating podgroups. document this behavior in the CRD. |
+| `preemptibility` | Use an explicit group value when configured. Otherwise use the strictest of all participating podgroups. Document this behavior in the CRD. |
+| `preemptionDelay` | Prefer an explicit group value when configured. Otherwise use the min of all participating podgroups. |
```
gshaibi, line 174: Not matters much - but I think we might want to default to the max value instead (maximize niceness, consisting with the `priorityClassName` default you mentioned above)

docs/developer/designs/multi-workload-gang/README.md
```diff
@@ -0,0 +1,214 @@
...
+## Status and accounting
+
+Input PodGroups continue to exist, so resource and scheduling status need to be handled accordingly.
+
+Today, PodGroup resource status is calculated from pods whose `pod-group-name` directly names that PodGroup. Under this design, input PodGroups therefore naturally retain resource totals while the rendered PodGroup has no directly annotated pods. The queue controller currently sums every PodGroup, so copying aggregate resources to the rendered object without filtering would double-count them.
+
+Two accounting models are possible:
+
+- Input PodGroups remain authoritative for resource status and queue accounting; the rendered PodGroup is excluded from resource summation and owns aggregate scheduling conditions. If the rendered priority or preemptibility differs from the inputs, queue accounting must reclassify input resources using the effective rendered policy rather than summing each input's `allocatedNonPreemptible` value.
+- The rendered PodGroup reports aggregate resources and input PodGroups are excluded from queue accounting while grouped.
```
gshaibi: I really appreciate this option's simplicity

docs/developer/designs/multi-workload-gang/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+If the rendered PodGroup owns aggregate resources, the PodGroup status controller must become relationship-aware because today it finds only pods directly annotated to the object. It should aggregate the actual descendant pods or their input statuses and patch only `resourcesStatus`, leaving scheduler-owned scheduling conditions untouched.
+
+For explainability, the rendered PodGroup should carry the scheduler's aggregate scheduling conditions, and the `WorkloadGroup` should project those conditions into per-input status. A user inspecting the group should be able to tell that one input is locally ready but blocked by another input. Existing input scheduling conditions from a previous standalone period must be cleared or explicitly marked as superseded, and the input status should identify the rendered PodGroup under which it is scheduled. Controllers must patch only the status fields they own.
```
gshaibi, line 193: "Existing input scheduling conditions from a previous standalone period" - do we want to support transition from standalone to part-of-group? what is the use case?

Also, I'm not sure I understood how will the input's status look like in case there is a condition error from another input in its group.
I wonder if we can just not publish anything on input podgroups and document that on such cases the consumers should consume the rendered podgroup's status. What do you think?

docs/developer/designs/multi-workload-gang/README.md
```diff
@@ -0,0 +1,214 @@
...
+| `subGroups` | Deep-copy each input tree, prefix names, rewrite parents, preserve minima and topology, and sort deterministically. Reject invalid trees and name collisions that cannot be resolved deterministically. |
+| Root topology | Do not infer an aggregate topology constraint from one input. Preserve each input's root constraint on its wrapper. A constraint spanning all inputs must be explicit on the `WorkloadGroup`. |
+| `markUnschedulable` and `schedulingBackoff` | Define whether these are group overrides or merged values. They apply to the rendered scheduling unit and cannot retain different per-input semantics. |
+
+## Status and accounting
+
+Input PodGroups continue to exist, so resource and scheduling status need to be handled accordingly.
+
+Today, PodGroup resource status is calculated from pods whose `pod-group-name` directly names that PodGroup. Under this design, input PodGroups therefore naturally retain resource totals while the rendered PodGroup has no directly annotated pods. The queue controller currently sums every PodGroup, so copying aggregate resources to the rendered object without filtering would double-count them.
```
gshaibi, line 185: Just want to make sure - under this design, what would be the `pod-group-name` and `kai.../sub-group` of pods belonging to a part-of-group workload? (I assume they will be the same as today and point to the input podgroup, just want to make sure)

docs/developer/designs/multi-workload-gang/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+- A `WorkloadGroup`, every input PodGroup, every participating pod, and the rendered PodGroup must be in the same namespace and queue.
+- Input PodGroups remain owned and reconciled by their original creators. The `WorkloadGroup` owns only the aggregated PodGroup.
+- A grouped input PodGroup must not be scheduled independently while the relationship is unresolved or invalid.
+- The renderer, not the scheduler, owns all merge policy. The scheduler performs only relationship and SubGroup resolution.
+- All podgroups in a `WorkloadGroup` are ganged together, meaning they all must schedule together or not at all (minSubGroup = expectedPodGroups). If a use case calls for it, we can later have a finer granularity where some podgroups are not mandatory for the workload to run. (Require subgroup-level semantics for required subgroups to work with the semi-preemptible mode).
+
+## Open questions
+
+- May an already allocated PodGroup join a workload group, and if so, how is gang atomicity restored?
```
gshaibi, line 207: What is the use case here?

docs/developer/designs/multi-workload-gang/README.md
itsomri: I don't want to actually change the annotation from the wg-gang controller, since it will compete with the regular podgrouper, so I think we'll have the prefixes in the wg-gang status and just prepend the usual subgoup name
e.g : a pod belongs to "workers" subgroup of a raycluster
the raycluster belongs to a wg-gang, so the wg-gang status declares that the RayCluster's subgroup prefix is `rc`
So the aggregated podgroup has the subgroup as rc-workers, and the scheduler does this join in-memory.

This is a suggestion, I'm not sure of the actual implementation details and how we'll decide the prefixes etc, but what do you think of the general idea?

docs/developer/designs/multi-workload-gang/README.md
itsomri: I think we'll start by explicitly forbidding this and throwing an error. If we find later use cases that require this, it's always easier to relax restrictions than to add them

docs/developer/designs/multi-workload-gang/README.md
itsomri: makes sense, will change

docs/developer/designs/multi-workload-gang/README.md
itsomri: Yeah me too, actually the doc should only have the second one, I probably forgot to push.

docs/developer/designs/multi-workload-gang/README.md
itsomri: No, it's a slop I missed, there shouldn't be any standalone period with conditions, I will fix that

docs/developer/designs/multi-workload-gang/README.md
itsomri: IDK, leaving it as an open question in case anybody has something like this, but the implementation will forbid this initially.

docs/developer/designs/multi-workload-gang/README.md
gshaibi: I agree, I don't see a use case for that, but what should be the error path - log on the podgrouper? status on the WorkloadGroup+extra input PodGroup? Anything else?

PR #2096, opened 2026-08-22
fix(scheduler): enforce PodSet gang admission floor

nvyutwu:
## Description

`GetTasksToAllocate` documents that task collection respects `minAvailable`, but a PodSet below its admission floor currently returns every eligible task in the priority queue up to the remaining requirement. If a caller consumes that under-filled list, checkpoint/rollback is atomic only over the partial list.

This change makes the shared collector return no tasks until enough eligible tasks exist to satisfy the PodSet's remaining `minAvailable` requirement. Once the PodSet is admitted, existing elastic behavior is unchanged and continues to return at most one additional task.

The allocate action already filters jobs through `IsReadyForScheduling`. This patch preserves that guard and enforces the same invariant at the lower-level collection boundary used by other scheduler paths.

This is independent of and complementary to #2095, which fixes namespace collisions in PodGroup identity and membership.

## Related Issues

N/A. Production investigation: https://reports-a3ad46.gitlab-master-pages.nvidia.com/reports/kai-clique-pinning-prd11/report.html

## Checklist

- [x] Self-reviewed
- [x] Added/updated tests
- [x] Updated documentation (not needed; no API or configuration change)

## Breaking Changes

None.

## Additional Notes

Pre-fix red test:

- `minAvailable=2`, one pending task: returned 1, expected 0
- `minAvailable=4`, three pending tasks: returned 3, expected 0
- `minAvailable=4`, two allocated and one pending task: returned 1, expected 0 because two remaining members are required

Post-fix validation on linux/amd64:

- `go test ./pkg/scheduler/api/podgroup_info -count=1`
- `go test ./pkg/scheduler/... -timeout 30m -count=1`
- `go vet ./...`
- `make lint-go` (0 issues)
- `make build-go-amd SERVICE_NAME=scheduler`

No physical GPU is required for these scheduler-state tests.


itsomri (dismissed):

gshaibi (dismissed):

pkg/scheduler/api/podgroup_info/allocation_info.go
```diff
@@ -118,8 +118,11 @@ func collectTasksFromPodSet(ps *subgroup_info.PodSet, taskOrderFn common_info.Le
 	if taskPriorityQueue.Empty() {
 		return nil
 	}
-	maxNumOfTasksToAllocate := getNumTasksToAllocate(ps, isRealAllocation)
-	return getTasksFromQueue(taskPriorityQueue, maxNumOfTasksToAllocate)
+	numTasksToAllocate := getNumTasksToAllocate(ps, isRealAllocation)
+	if ps.GetNumActiveAllocatedTasks() < int(ps.GetMinAvailable()) && taskPriorityQueue.Len() < numTasksToAllocate {
```
davidLif: What will happen if the minAvaliable is 4, 3 tasks are active and 1 new task wasts to allocate? I would think that in this case we should try to allocate the 1 new task.

vrv3814: Isn't that how this guard works too?
With minAvailable 4, active = 3
`getNumTasksToAllocate` returns 1 (int(subGroup.GetMinAvailable()) - numAllocatedTasks)

The second half of the check doesn't fire (`taskPriorityQueue.Len() < numTasksToAllocate # 1 < 1`)

pkg/scheduler/api/podgroup_info/allocation_info.go
```diff
@@ -118,8 +118,11 @@ func collectTasksFromPodSet(ps *subgroup_info.PodSet, taskOrderFn common_info.Le
...
+		return nil
```
gshaibi, line 173: What happens when the child picked in the gang phase can't form? It returns nil now but still takes a top-K slot, so with root minSubGroup=1 and subgroups `a` (minMember=2, 1 pod), `b` (minMember=1, 1 pod), the podgroup collects zero tasks every cycle and `b` never schedules, right?
Should the gang phase skip it and try the next child, like the elastic loop does?

gshaibi: Also, this function is also consumed in preempt/consolidate for virtual preemption where we in-memory evict pods and then try to re-allocate them.
Now, a workload with `running pods < minMember` (due to completed pod for example, temporary state) will be in-memory evicted, and its pods will not be re-allocated since we don't return these pods here, right?

gshaibi (dismissed):

pkg/scheduler/api/podgroup_info/allocation_info.go
nvyutwu: Yes. With minAvailable=4, three active tasks make getNumTasksToAllocate return 1, and one pending task satisfies the floor because 1 < 1 is false. I added that exact scenario as a regression test in 8efeb01c.

pkg/scheduler/api/podgroup_info/allocation_info.go
nvyutwu: Good catches. Fixed in 8efeb01c: the gang phase now counts only children that return tasks and continues to the next prioritized child, so an underfilled child a no longer starves ready child b. The admission-floor guard now applies only to real allocation; virtual preempt/consolidate paths continue returning the available virtual tasks even when the workload is temporarily below minMember. Added regression tests for both cases; the full scheduler test suite and go vet pass.

gshaibi: Thanks. Should we make PodGroup.IsStale also minSubGroup aware so workload with a not-ready (now skipped) subgroup won't be evicted?

nvyutwu: Fixed in a88773df. PodGroupInfo.IsStale now evaluates gang satisfaction through the subgroup tree: each SubGroupSet applies its minSubGroup threshold recursively, while leaf PodSets continue to apply minMember. Added a regression test covering one satisfied running subgroup plus an optional unready subgroup.

gshaibi: @nvyutwu do you think of other areas where this might cause issues? podgroup/queue status reporting? anything else?

pkg/scheduler/api/podgroup_info/job_info.go
```diff
@@ -505,21 +505,18 @@ func (pgi *PodGroupInfo) IsStale() bool {
...
-	return false
+	return !pgi.IsGangSatisfied()
 }
 
 func (pgi *PodGroupInfo) IsGangSatisfied() bool {
-	for _, podSet := range pgi.PodSets {
-		if !podSet.IsGangSatisfied() {
-			return false
+	root := pgi.RootSubGroupSet
+	if root == nil {
```
gshaibi: In what case does this happen?

pkg/scheduler/api/podgroup_info/job_info.go
```diff
@@ -88,12 +88,12 @@ type PodGroupInfo struct {
...
 
 	// inner cache
-	allPodsMap                        *pod_info.PodsMap
-	tasksToAllocate                   []*pod_info.PodInfo
-	tasksToAllocateInitResourceVector resource_info.ResourceVector
-	PodStatusIndex                    map[pod_status.PodStatus]pod_info.PodsMap
-	activeAllocatedCount              *int
-	aliveTasksRequestedGPUs           *float64
+	allPodsMap                              *pod_info.PodsMap
+	tasksToAllocateByMode                   map[bool][]*pod_info.PodInfo
```
gshaibi: What does "by mode" mean?

pkg/scheduler/api/podgroup_info/allocation_info.go
nvyutwu: Traced this end to end. PodGroup stale evaluation is now minSubGroup-aware. Queue/resource accounting must keep partial estimates, so it now uses an explicit PartialTaskAllocation mode. The actual pending paths in preempt, reclaim, and consolidation use SimulatedTaskAllocation, which enforces the subgroup floor before JobSolver; internal solver prefixes remain partial, and victim restoration uses VictimReallocation. Added mode-routing, admission-floor, ready-sibling, and cache-isolation tests in 8889e790.

pkg/scheduler/api/podgroup_info/job_info.go
nvyutwu: There is no supported production case. Constructors and clones always initialize RootSubGroupSet. I removed the synthesized fallback; nil now fails closed, consistent with IsReadyForScheduling.

pkg/scheduler/api/podgroup_info/job_info.go
nvyutwu: The bool meant real versus virtual, but virtual was overloaded. In 8889e790 it is replaced by TaskAllocationMode: RealTaskAllocation, SimulatedTaskAllocation, PartialTaskAllocation, and VictimReallocation. The cache is partitioned by admission semantics so partial accounting cannot contaminate simulated admission.

itsomri (approved):

gshaibi (approved):

PR #2104, opened 2026-08-26
feat(scheduler): Add queuePriorityInQuotaReclaim strategy for reclaim

### LINKED ISSUE #2065: [design] design issue to allow reclaim/consolidation from lower priority queues (even within quota)

Add a flag to support workloads to use Queue priority to evict In quota preemptible workloads from lower priority queues

When higher priority Queue with in-quota workload needs resources that it can't find, it should be allowed to reclaim those resources from in-quota lower priority preemptible workloads (in-quota preemption). For over-quota workloads this is already implemented. This should be applicable for both reclaim and consolidation.

Points for consideration:

Allow only workloads that are within the queue deserved quota to reclaim workloads from queues that are in (strictly) lower queue priority even within quota
Need to adress and think about edge cases and how fairshare should be calculated to acheive this result
This behavior should not be enabled by default, consider adding a per shard configuration for that behavior
Need to consider if in all cases we want to use quota as a disruption budget or is there a mode where priority is strictly more important

### Why is this needed?

We want higher priority in-quota queues to be able to use resources on the expense of lower priority queues, even if the lower priority queues are in-quota, as long as lower priority queue workloads are preemptible.

davidLif:
## Description

Add queuePriorityInQuotaReclaim strategy for reclaim. If it's enabled (off by default) higher priority queues can reclaim in-quota resources from lower priority queues.

## Related Issues

Fixes #2065

## Checklist

- [x] Self-reviewed

### DIFF
--- a/.changes/unreleased/added-20260826-192424.yaml
+++ b/.changes/unreleased/added-20260826-192424.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: |-
+  Add queuePriorityInQuotaReclaim strategy for reclaim. If it's enabled (off by default) higher priority queues can reclaim in-quota resources from lower priority queues.
--- a/.changes/unreleased/added-20260826-192424.yaml
+++ b/.changes/unreleased/added-20260826-192424.yaml
@@ -1,3 +1,3 @@
 kind: Added
 body: |-
-  Add queuePriorityInQuotaReclaim strategy for reclaim. If it's enabled (off by default) higher priority queues can reclaim in-quota resources from lower priority queues.
+  Add queuePriorityInQuotaReclaim strategy for reclaim. If it's enabled (off by default) higher priority queues can reclaim in-quota resources from lower priority queues. This is alpha functionality, and the exact behavior of this feature might change in future releases.
--- a/docs/fairness/README.md
+++ b/docs/fairness/README.md
@@ -118,7 +118,8 @@ scheduler:
 | `false` (default) | Only over-quota resources are ever reclaimed — the quota protection guarantee holds for every queue |
 | `true` | A queue can additionally reclaim in-quota resources from any strictly lower priority queue, provided the reclaimer stays within its own deserved quota |
 
-This is alpha functionality; enable it per scheduling shard only when queue `Priority` should outrank quota protection for lower-priority queues.
+This is alpha functionality, and the exact behavior of this feature might change in future releases. 
+Enable `queuePriorityInQuotaReclaim` per scheduling shard only when queue `Priority` should outrank quota protection for lower-priority queues.
 
 ## See Also
 
--- a/docs/fairness/README.md
+++ b/docs/fairness/README.md
@@ -93,10 +93,20 @@ pluginArguments:
 
 ### Priority-Based In-Quota Reclaim
 
-Opt in to letting a strictly higher priority queue reclaim in-quota resources from a strictly lower priority queue, using `queuePriorityInQuotaReclaim`:
+Opt in to letting a strictly higher priority queue reclaim in-quota resources from a strictly lower priority queue, using `queuePriorityInQuotaReclaim`. This is a `proportion` plugin argument, set on the `SchedulingShard` this applies to (see [Scheduler Config Customization](../operator/scheduler-config-customization.md)):
 
 ```yaml
+# SchedulingShard
 spec:
+  plugins:
+    proportion:
+      arguments:
+        queuePriorityInQuotaReclaim: "true"
+```
+
+```yaml
+# Helm values — templated into the default SchedulingShard's spec.plugins
+scheduler:
   plugins:
     proportion:
       arguments:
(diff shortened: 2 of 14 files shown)


docs/fairness/README.md
```diff
@@ -89,6 +91,35 @@ pluginArguments:
...
+      arguments:
+        queuePriorityInQuotaReclaim: "true"
+```
+
+| Value | Behavior |
+|-------|----------|
+| `false` (default) | Only over-quota resources are ever reclaimed — the quota protection guarantee holds for every queue |
+| `true` | A queue can additionally reclaim in-quota resources from any strictly lower priority queue, provided the reclaimer stays within its own deserved quota |
+
+This is alpha functionality; enable it per scheduling shard only when queue `Priority` should outrank quota protection for lower-priority queues.
```
itsomri: Add here a note that behavior could change between KAI versions and to follow the release notes if using this feature

.changes/unreleased/added-20260826-192424.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Added
+body: |-
+  Add queuePriorityInQuotaReclaim strategy for reclaim. If it's enabled (off by default) higher priority queues can reclaim in-quota resources from lower priority queues.
```
itsomri: Add here as well that it's an experimental alpha feature

docs/scheduling-deep-dive/README.md
```diff
@@ -88,7 +88,7 @@ Now that both concepts have been introduced, here is how they compare — they a
 | Affects guaranteed quota? | No | No |
 | Affects over-quota distribution? | Yes — order between queues | No |
 | Affects preemption? | No | Yes — determines victims |
-| Affects reclaim? | Indirectly (via fair-share) | No |
+| Affects reclaim? | Indirectly (via fair-share); directly gates in-quota reclaim when `queuePriorityInQuotaReclaim` is enabled | No |
```
itsomri: "directly gates in-quota reclaim " is too vague and LLM-y, try to write it more humanly 
Also, I think it's fair to say that queue priority affects reclaim full stop, saying that it's "indirect" because it's due to the fair share mechanism might be correct from the internal implementation perspective, but doesn't really matter for the user. It just does.

davidLif: I actually wrote this line :) . The priority controlls the in-quota reclaims when the feature gate is enabled. I will try to think about a different way to express it, but I think it gives a concise way to understand when how reclaim is effected - it playes with the fair-share by default, and allows in-quota reclaims when the feature is enabled. 
This document is the "scheduling deep dive", so I think the reader should get a bit more details about the internal implementation.

.changes/unreleased/added-20260826-192424.yaml
davidLif: Updated.

docs/fairness/README.md
davidLif: Updated.

docs/queues/README.md
```diff
@@ -15,7 +15,7 @@ Only leaf queues (queues with no children) can be used for scheduling jobs. Pare
 | Attribute | Description | Units |
 |-----------|-------------|-------|
 | **Quota** | Guaranteed resource allocation | CPU: millicores, Memory: MB, GPU: units |
-| **Over-Quota Priority** | Resource allocation order when exceeding quota | Integer (higher = first) |
+| **Queue Priority** | Resource allocation order when exceeding quota; also gates in-quota reclaim eligibility when the `proportion` plugin's `queuePriorityInQuotaReclaim` argument is enabled — see [Fairness Policies](../fairness/README.md#priority-based-in-quota-reclaim) | Integer (higher = first) |
```
itsomri: I don't understand what "gates in-quota reclaim eligibility" means
Try simpler terms. "Allows in-quota reclaim when `queuePriorityInQuotaReclaim` is enabled" or something like that. Maybe "Prioritizes quotas when `queuePriorityInQuotaReclaim` is enabled"

docs/scheduling-deep-dive/README.md
itsomri: I don't mind who actually wrote the line, I just want it to be clear, to me it's not clear right now and I doubt a user will easily understand it. If you need to be a little more verbose to be clear, that's fine

itsomri (dismissed):

pkg/scheduler/plugins/proportion/reclaimable/filter_victims.go
```diff
@@ -29,6 +29,10 @@ func (r *Reclaimable) FilterVictim(
 		return strategies.FitsMaintainFairShare(reclaimeeQueue, reclaimeeQueue.GetAllocatedShare())
 	}
 
+	if r.queuePriorityInQuotaReclaim && reclaimerQueue.Priority > reclaimeeQueue.Priority {
+		return true
+	}
+
```
enoodle: Can this be implemented with `strategies.FitsQueuePriorityInQuotaReclaim` function to e similar to the `if` above it?

enoodle (dismissed):

docs/scheduling-deep-dive/README.md
davidLif: Updated.

docs/queues/README.md
davidLif: Updated.

pkg/scheduler/plugins/proportion/reclaimable/filter_victims.go
davidLif: Done

itsomri (approved):

Issue #1936, opened 2026-07-17
Set annotation to ignore quota occupied by workload


amy:
### What you would like to be added?

Full context: https://cloud-native.slack.com/archives/C08RKDPQPQU/p1784320123753599

Here's the scenario: We have a healthcheck pod that we don't want counted towards topology capacity. ie, if the healthcheck pod sits on a node, kai should see that node as totally free. 

This can be done by submitting the pod to have kai schedule it. But its an operational burden because the queue of the pod would need to be adjusted whenever theres upstream queue changes in the hierarchy. 

Can we submit a healthcheck pod that:
1. doesnt need to specify a queue
2. has its capacity ignored by kai 
3. when kai does schedule a normal kai workload, allow priority to preempt this healthcheck pod (we set it to negative priority)

### Why is this needed?

its an operational burden because the queue of the pod would need to be adjusted whenever theres upstream queue changes in the hierarchy.

amy:
cc/ @enoodle

mesutoezdil:
Saw this in the Slack thread. Wanna share sth from my side if i may.
1 thing missing from the issue: the annotation idea only works if you can actually touch the pod spec. Sb else in that thread had the same problem but their out-of-band pods come from the cloud provider, so they can't annotate them either.
Might be worth considering an admin-side selector instead (namespace/label) that kai checks when computing node capacity, so matching pods just get excluded from the calculation entirely. 
That covers @amy's case and the "can't touch the pod" case, no queue, no annotation, and capacity is actually ignored instead of just deprioritized. 
Preemption via negative priority could still be a fallback for anything not covered by the selector.

amy:
Yeah not tied to implementation. But I vote label or annotation or priority (not namespace)

itsomri:
Interesting issue, thanks!
For the implementation side, kai would probably need to include those pods for allocation simulations, not just ignore them - the default scheduler will not initiate the evictions for us.

enoodle:
@itsomri One opssibility to implement it is that internally KAI will create a "virtual queue" for those pods with 0 quota, infinite limit (doesn't matter it didn't schedule them anyway) and 0 over quota weight. This way all other queues will always be able to reclaim from that queue.

amy:
One question I have is... can you have disjoint hierarchies? I think yours is called root queues? So two different root queues. 

If so, wanna make sure we don't face scheduling problems.

itsomri:
[...]

PR #2108, opened 2026-08-30
feat: Background Pods

itsomri:
## Related Issues

Proposed implementation for #1936 

## Checklist

- [ ] Self-reviewed


docs/developer/designs/background-pods/README.md
```diff
@@ -0,0 +1,209 @@
...
+At session close the plugin attempts to re-place every background pod its original node.
+Those that still fit are unevicted and are not disturbed; the rest have their evictions committed, and the pipelined workloads bind in a later session once they are gone.
+
+### Selecting background pods
+
+Since external pods are deferred, the first version can rely on an opt-in that the pod itself carries, such as a `kai.scheduler/background: "true"` label.
+This keeps the blast radius small: KAI only ever ignores and evicts pods that explicitly asked to be treated this way.
+
+Open questions - possible validations needed:
+- Should we block regular users from using this label? It could be used to surpass the quota mechanisms
```
amy, line 112: I think the user themselves can probably block via like opa policy or something. I guess what mechanics do you have for rbac where you otherwise thinking? Maybe something like if the pod is from X cluster object queue, allow for pods to be labeled with the label? 

For my end, its not a high need yet. So maybe just do it if future users as for it.

docs/developer/designs/background-pods/README.md
```diff
@@ -0,0 +1,209 @@
...
+- Should we validate that `terminationGracePeriodSeconds=0` for background pods? Or just recommend it?
```
amy, line 113: Lets go with recommendation for now until people complain.

docs/developer/designs/background-pods/README.md
```diff
@@ -0,0 +1,209 @@
...
+- Should we validate that the label is not set on pods with `system-cluster-critical` or `system-node-critical` priority?
```
amy, line 114: no validation required on this front yet. Until people complain.

docs/developer/designs/background-pods/README.md
```diff
@@ -0,0 +1,209 @@
...
+This can probably be fixed with some effort if there's a need for it.
+
+
+### What is not yet designed
+
+- Background pods holding DRA resource claims.
+  Deleting the pod does not synchronously free the claim.
+- Interaction with consolidation, which now plans moves against a cluster that looks emptier than it is.
+- Observability: which events to emit, and how to keep these evictions out of the preemption and reclaim metrics so fairness reporting stays meaningful.
+- Whether background pods should still appear in their queue's externally reported `allocated`, given the goal of keeping them out of queue accounting entirely.
```
amy, line 185: probably makes sense to still appear in allocated. Given it is allocated. 

The goal is keeping out of queue accounting purely for scheduling purposes. 
But you separately still want cluster allocation to be accurate.

docs/developer/designs/background-pods/README.md
```diff
@@ -0,0 +1,209 @@
...
+## Design
+
+At the start of every scheduling session the plugin evicts all background pods from the in-memory snapshot, without committing, so their capacity is accounted as *releasing*.
+The session then runs untouched: workloads that fit in idle capacity are allocated and bound as usual, and workloads that need background-held capacity are pipelined onto it, exactly as they would be over a preemption victim.
+At session close the plugin attempts to re-place every background pod its original node.
+Those that still fit are unevicted and are not disturbed; the rest have their evictions committed, and the pipelined workloads bind in a later session once they are gone.
+
+### Selecting background pods
+
+Since external pods are deferred, the first version can rely on an opt-in that the pod itself carries, such as a `kai.scheduler/background: "true"` label.
```
amy, line 108: Do you want to make a section that's specific to user facing api/label/usage changes. 

You also mentioned a separate queue that someone would still need to create.

docs/developer/designs/background-pods/README.md
itsomri: We could use an admission webhook or even evaluate this in the scheduling cycle and not allow bg-pods to schedule on "legit" queues, for example. We'll probably leave it out for now though, unless anyone presents a good case for it

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -489,7 +502,21 @@ func (pp *proportionPlugin) allocateHandlerFn(ssn *framework.Session) func(event
...
-		job := ssn.ClusterInfo.PodGroupInfos[event.Task.Job]
+		job, found := ssn.ClusterInfo.PodGroupInfos[event.Task.Job]
+		if !found {
+			log.InfraLogger.V(7).Infof("Proportion DeallocateFunc: no job <%v> in session, skipping", event.Task.Job)
+			return
+		}
+
+		leafQueue, found := pp.queues[job.Queue]
+		if !found {
+			// The queue is absent once the plugin's session state has been torn down, and for jobs
```
gshaibi, line 513: I think this issue can occur on the numa plugin too.
So we should either fix it there, or alternatively make sure the backgroundpods deallocations happens before any plugin tear down (by either making it the first plugin or introducing a new plugin extension point that happens just before OnSessionClose, or maybe even make the restore in a last in line action instead of a plugin, I'm not sure whats best).

If we choose to keep the current form and only fix on numa, I think this deserves a short documentation for plugin users. wdut?

itsomri: We're doing both:
The change in `pkg/scheduler/framework/framework.go` makes the closing order a reverse of the opening order, so the most recently opened plugin will be the first to close. Until now it was random (iteration on map keys)

itsomri: (and since this plugin is the lowest priority, it will open last and close first)

gshaibi: But numa's priority is 0 by default and this one is 50, no?

gshaibi (approved):

gshaibi (approved):

PR #2114, opened 2026-09-02
feat(scheduler): workload-centric eviction metrics

### LINKED ISSUE #1573: metric: kai_pod_group_evicted_pods_total is hard to use — first-sample invisible, no owner label, counts pods not events

david-gang:
## Description

Replace the PodGroup eviction metric with a workload-centric shape so operators can answer "how disrupted has this JobSet/Workspace been" with standard PromQL.

- Relabel `kai_pod_group_evicted_pods_total` with owner and leaf-subgroup labels, drop the PodGroup `uid` label, and pre-initialize series at 0 on PodGroup observe so `rate()` / `increase()` work from the first eviction.
- Add `kai_pod_group_eviction_events_total`, incremented once per committed decision per affected PodGroup (statement commit for preempt/reclaim/consolidation; explicit record for stale-gang eviction).
- Seed zero-series from the configured eviction actions only (`RegisterEvictionAction`), not allocate.

Fixes #1573

## Related Issues

Fixes #1573
Design: `docs/developer/designs/eviction-metrics/README.md` (#1604)

## Checklist

- [x] Self-reviewed

## Breaking Changes

Yes. `kai_pod_group_evicted_pods_total` drops the PodGroup `uid` label and adds `owner_group`, `owner_kind`, `owner_name`, `owner_uid`, and `subgroup`. Panels that filter on `uid` will return empty. Queries that `sum by (...)` without those labels keep working.

## Additional Notes

Pre-initialization and delete handlers are wired on the PodGroup informer. Event increments must not fire for rolled-back statement operations; `Statement.Commit` records only after a successful `commitEvict`.

## Test Plan

- [ ] Unit: `go test ./pkg/scheduler/metrics ./pkg/scheduler/framework ./pkg/scheduler/scheduler ./pkg/scheduler/cache/status_updater ./pkg/scheduler/actions/stalegangeviction`
- [ ] Confirm a PodGroup that is observed then evicted once produces non-zero `increase()` on both counters
- [ ] Confirm a gang eviction of N pods on one PG increments pods by N and events by 1
- [ ] Confirm stale-gang eviction increments events with `action="stalegangeviction"`
- [ ] Confirm PodGroup delete drops the series (`DeletePartialMatch`)

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Added workload owner and subgroup labels to pod-group eviction metrics.
  * Added an eviction-events counter tracking committed eviction decisions per pod group.
  * Eviction metrics now initialize, update, and clean up as pod groups change.
* **Bug Fixes**
  * Stale-job evictions now record an event when at least one pod is successfully evicted.
* **Documentation**
  * Updated metrics documentation with the new labels and counter.

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -116,18 +116,21 @@ func registerSchedulerPodInformer(informerFactory informers.SharedInformerFactor
 func registerPodGroupEvictionMetricHandlers(
 	informer k8scache.SharedIndexInformer,
 	nodePoolLabelKey string,
+	evictionActionNames []string,
 ) error {
 	_, err := informer.AddEventHandler(k8scache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
 			if podGroup, ok := obj.(*enginev2alpha2.PodGroup); ok {
-				metrics.InitPodGroupEvictionMetrics(podGroup, nodePoolLabelKey)
+				metrics.InitPodGroupEvictionMetrics(podGroup, nodePoolLabelKey, evictionActionNames)
 			}
 		},
 		UpdateFunc: func(oldObj, newObj interface{}) {
 			oldPodGroup, oldOK := oldObj.(*enginev2alpha2.PodGroup)
 			newPodGroup, newOK := newObj.(*enginev2alpha2.PodGroup)
 			if oldOK && newOK {
-				metrics.InitPodGroupEvictionMetricsOnUpdate(oldPodGroup, newPodGroup, nodePoolLabelKey)
+				metrics.InitPodGroupEvictionMetricsOnUpdate(
+					oldPodGroup, newPodGroup, nodePoolLabelKey, evictionActionNames,
+				)
 			}
 		},
 		DeleteFunc: func(obj interface{}) {
@@ -170,6 +173,7 @@ type SchedulerCacheParams struct {
 	UpdatePodEvictionCondition  bool
 	StuckInReleasingThreshold   time.Duration
 	DiscoveryClient             discovery.DiscoveryInterface
+	EvictionActionNames         []string
 }
 
 type SchedulerCache struct {
@@ -238,6 +242,7 @@ func newSchedulerCache(schedulerCacheParams *SchedulerCacheParams) (*SchedulerCa
 	if err := registerPodGroupEvictionMetricHandlers(
 		sc.kubeAiSchedulerInformerFactory.Scheduling().V2alpha2().PodGroups().Informer(),
 		sc.schedulingNodePoolParams.NodePoolLabelKey,
+		schedulerCacheParams.EvictionActionNames,
 	); err != nil {
 		return nil, fmt.Errorf("failed to register PodGroup eviction metric handlers: %w", err)
 	}
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -68,6 +68,7 @@ import (
 	k8splugins "github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/k8s_internal/plugins"
 	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/log"
 	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/metrics"
+	"github.com/kai-scheduler/KAI-scheduler/pkg/scheduler/utils"
 )
 
 func init() {
@@ -110,6 +111,41 @@ func registerSchedulerPodInformer(informerFactory informers.SharedInformerFactor
 			filterFailedPods,
 		)
 	})
+}
+
+func registerPodGroupEvictionMetricHandlers(
+	informer k8scache.SharedIndexInformer,
+	nodePoolLabelKey string,
+) error {
+	_, err := informer.AddEventHandler(k8scache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			if podGroup, ok := obj.(*enginev2alpha2.PodGroup); ok {
+				metrics.InitPodGroupEvictionMetrics(podGroup, nodePoolLabelKey)
+			}
+		},
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			oldPodGroup, oldOK := oldObj.(*enginev2alpha2.PodGroup)
+			newPodGroup, newOK := newObj.(*enginev2alpha2.PodGroup)
+			if oldOK && newOK {
+				metrics.InitPodGroupEvictionMetricsOnUpdate(oldPodGroup, newPodGroup, nodePoolLabelKey)
+			}
+		},
+		DeleteFunc: func(obj interface{}) {
+			podGroup, ok := obj.(*enginev2alpha2.PodGroup)
+			if !ok {
+				tombstone, tombstoneOK := obj.(k8scache.DeletedFinalStateUnknown)
+				if !tombstoneOK {
+					return
+				}
+				podGroup, ok = tombstone.Obj.(*enginev2alpha2.PodGroup)
+				if !ok {
+					return
+				}
+			}
+			metrics.DeletePodGroupEvictionMetrics(podGroup.Namespace, podGroup.Name)
+		},
+	})
+	return err
 }
 
 // New returns a Cache implementation.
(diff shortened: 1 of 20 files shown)


pkg/scheduler/metrics/metrics.go
```diff
@@ -365,9 +384,115 @@ func RegisterPreemptionAttempts() {
...
+	for _, subgroup := range leafSubgroups(newPodGroup.Spec.SubGroups) {
+		if _, exists := oldLeaves[subgroup]; exists {
+			continue
+		}
+		for _, action := range evictionActionNames {
+			labels := append(podGroupEvictionLabels(newPodGroup, nodepool, action), subgroup)
+			podGroupEvictedPodsTotal.WithLabelValues(labels...).Add(0)
+		}
+	}
+}
```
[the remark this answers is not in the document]
david-gang: @enoodle asking you on this one before we change anything.

CodeRabbit is right about the mechanics: `InitPodGroupEvictionMetricsOnUpdate` only zero-inits **new leaf subgroups**, using the **current** labels. Cleanup is delete-only. If `nodepool` or the top-owner annotation changes on a live PodGroup, the new identity is not pre-initialized (so the first eviction after the change is invisible to `rate()` / `increase()`), and the old series stay until the PG is deleted.

Our read is that those labels are effectively immutable for a given PodGroup (nodepool at create, top-owner YAML written once by the podgrouper), so this is a won't-fix / follow-up rather than a merge blocker.

Do you want identity reconcile in this PR (re-init the new labels; optionally drop the old series), or are you fine leaving it?

The other CodeRabbit note (extra GoDoc / 80% docstring coverage) we plan to skip — the helpers already have short why-comments and that threshold is not a repo rule.

pkg/scheduler/metrics/metrics.go
```diff
@@ -365,9 +384,115 @@ func RegisterPreemptionAttempts() {
...
+		nodepool,
+		action,
+		owner.Group,
+		owner.Kind,
+		owner.Name,
+		owner.UID,
+	}
+}
+
 // IncPodGroupEvictedPods records a single pod eviction for a pod group.
```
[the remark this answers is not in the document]
david-gang: Skipping this. The exported helpers already have short comments that state why they exist (zero-series, committed events, new leaves, delete). Extra rationale GoDoc to hit CodeRabbit's 80% docstring check is not a repo convention.

pkg/scheduler/cache/cache.go
```diff
@@ -112,6 +113,44 @@ func registerSchedulerPodInformer(informerFactory informers.SharedInformerFactor
 	})
 }
 
+func registerPodGroupEvictionMetricHandlers(
+	informer k8scache.SharedIndexInformer,
+	nodePoolLabelKey string,
+	evictionActionNames []string,
+) error {
+	_, err := informer.AddEventHandler(k8scache.ResourceEventHandlerFuncs{
```
enoodle, line 120: @david-gang Can we somehow wrap this behind a feature flag so that it wont automatically create a metric for each submitted workload? If someone creates many it would have sever memory consumption.

enoodle (commented):
Looks good. My only concern is the memory consumption

pkg/scheduler/cache/cache.go
david-gang: @enoodle want to double-check the cut before we implement.

Reading this as: put the **OnAdd/OnUpdate zero-series registration** (`InitPodGroupEvictionMetrics`) behind a flag (default off), so we do **not** allocate Prometheus series for every submitted PodGroup. Committed eviction increments, the new owner/subgroup labels, the events counter, and delete-on-PG-delete would stay always on — series then appear only when something is actually evicted.

The other option is gating the **whole** redesign (no new labels / events counter unless the flag is on).

Is the first one what you meant?

enoodle: @david-gang I meant the first one but there is also a case for gating the whole redesign because then the metric will not change its behavior when the flag is off. I am good with either one because my main concern is the memory consumption.

enoodle: @david-gang scratch that..
It doesn't make sense to gate only the creation for zero values.
It should all be gated by the flag so that the metric will always behave the same or won't be there at all.

pkg/scheduler/cache/cache.go
```diff
@@ -112,6 +113,44 @@ func registerSchedulerPodInformer(informerFactory informers.SharedInformerFactor
...
+		AddFunc: func(obj interface{}) {
```
gshaibi, line 121: I'd suggest skipping here and on `UpdateFunc` events that belong to other scheduling shards.

pkg/scheduler/cache/cache.go
gshaibi: Also, what do you think about exposing the metric only after the Running phase? So that you will still see rate(), but won't consume memory for pending pods.

david-gang: @enoodle this was a big change. i tried to be DRY. please take a look at https://github.com/kai-scheduler/KAI-Scheduler/pull/2114/changes/badb4667e496bd0b923d0c8c9ef5e17392f031c4
@gshaibi regarding the running phase pod please look at https://github.com/kai-scheduler/KAI-Scheduler/pull/2114/changes/614c3c6728eef955b5b572312ac9065787ef91c6 if i got it right.

enoodle (approved):

pkg/scheduler/cache/cache.go
david-gang: fixed this in the latest commits

PR #2116, opened 2026-09-02
fix(scheduler): exclude Releasing pods from inter-pod affinity index

### LINKED ISSUE #2115: scheduler: Releasing pods stay in the inter-pod affinity index, so reclaim/preempt can never satisfy a required podAntiAffinity against the victims
## Summary

A pending pod with a **required** inter-pod anti-affinity against the pods that `reclaim`/`preempt` would evict can never be placed by those actions, even when eviction would fully clear a node. Every scenario fails the k8s InterPodAffinity filter with `node(s) didn't match pod anti-affinity rules`.

## Mechanism

`Statement.Evict` moves the victim to `Releasing` and calls `node.UpdateTask`, which is `RemoveTask` followed by `addTask`. `addTask` unconditionally calls `PodAffinityInfo.AddPod`, so the evicted pod goes straight back into the node's k8s `NodeInfo` that the predicates plugin hands to InterPodAffinity. The scheduler ends up with two inconsistent views of a Releasing pod:

- **resources**: future-free (`ReleasingVector`), so a pending pod can be `Pipelined` onto the node;
- **inter-pod affinity index**: present, so the pending pod's required anti-affinity rejects the node.

A `Pipelined` task is never bound in the same cycle (`commitPipeline` → `Cache.TaskPipelined`, a status update; only `commitAllocate` reaches `BindPod`), so the two pods would never actually coexist. Keeping the Releasing pod in the affinity index buys no safety and defeats reclaim/preempt for any workload that uses required anti-affinity. The symmetric check makes it worse: the victims' *own* required anti-affinity terms also reject the incoming pod while they are Releasing.

## Impact observed (v0.17.0)

An 8-node, 64-GPU gang and a 48-GPU job sat `Pending` while 157 reclaimable GPUs were held by lower-priority pods on 22 nodes that eviction would have fully cleared. Replaying the captured scheduler snapshot offline: 777 reclaim scenarios tried, 675 explicit anti-affinity rejections on nodes whose only pods were the evicted victims, `Didn't find a reclaim strategy` in 2.2s. Stripping the required anti-affinity terms from the snapshot and replaying (nothing else changed) → `Scenario solved for 8 tasks`.

## Versions

Reproduced on v0.17.0; the code path is unchanged on `main` (`pkg/scheduler/api/node_info/node_info.go`, `addTask` → `PodAffinityInfo.AddPod`).

## Proposed fix

Do not index `Releasing` tasks for inter-pod affinity, and un-index symmetrically in `RemoveTask` keyed on the stored copy's status. PR to follow with unit tests, an action-level reclaim test, and the production snapshot replay as evidence.

sam-huang1223:
## Description

**Bug.** A pending pod that carries a *required* inter-pod anti-affinity against the pods `reclaim`/`preempt` would evict can never be placed, even when eviction would fully clear a node. Every scenario fails the k8s InterPodAffinity filter with `node(s) didn't match pod anti-affinity rules`, so the action gives up in a few seconds regardless of how many victims it tries.

**Mechanism.** `Statement.Evict` moves the victim to `Releasing` and calls `node.UpdateTask`, which is `RemoveTask` followed by `addTask`. `addTask` unconditionally calls `PodAffinityInfo.AddPod`, so the evicted pod is put straight back into the node's k8s `NodeInfo` that `predicates.go` hands to the InterPodAffinity filter. The scheduler therefore holds two inconsistent views of a session-evicted pod:

| view | evicted pod is… | consequence |
|---|---|---|
| resources (`ReleasingVector`, pipelining) | future-free | pending pod can be `Pipelined` onto the node |
| inter-pod affinity index | present | pending pod fails required anti-affinity on that node |

Because a `Pipelined` task is never bound in the same cycle (`commitPipeline` only calls `Cache.TaskPipelined`; only `commitAllocate` reaches `ssn.BindPod`), the pending pod and the victim never coexist on the node. This surfaced in production (KAI v0.17.0): a 64-GPU, 8-node gang and a 48-GPU job sat `Pending` while 157 reclaimable GPUs were held by lower-tier pods on 22 fully-clearable nodes. The replayed cycle tried 777 scenarios and rejected all of them on anti-affinity.

**Fix (final form, after two review rounds).** Only tasks *this scheduling cycle has itself evicted* leave the inter-pod affinity index. The session already marks them: `Statement.Evict` sets `IsVirtualStatus` on the victim, and `Releasing && IsVirtualStatus` already means "resources freed by this cycle" elsewhere (`pod_info.go`). It just did so *after* `node.UpdateTask` had re-indexed the victim.

- `framework/statement.go` `Evict`: set `IsVirtualStatus = true` before `node.UpdateTask` (restored if the update fails). `unevict` already restores the previous value before re-indexing.
- `api/node_info` `addTask`/`RemoveTask`: skip `AddPod`/`RemovePod` when `excludedFromPodAffinity(task)`, i.e. `Status == Releasing && IsVirtualStatus`. `RemoveTask` decides on the node's stored clone, i.e. the state the pod was indexed under, so evict, the `unevict` rollback and `StuckInReleasing` transitions stay balanced and the k8s `NodeInfo.RemovePod` "not found" error cannot fire.

Pods terminating independently in the cluster (Releasing at snapshot time, no marker) stay indexed, exactly like upstream kube-scheduler, which keeps a pod until its delete event: bind-now placements and ordinary `allocate` never ignore a terminating pod. `StuckInReleasing` is not `Releasing`, so it stays indexed too. Victims that the scheduler evicts and then pipelines elsewhere (consolidation) become `Pipelined`, so they are indexed on their new node.

Cumulative change against `main`: two source files, +24/−4 lines, plus tests.

**Review history, kept as separate commits so the deltas are visible:**
1. `26f75fc8` excluded every `Releasing` pod from the single index. @gshaibi pointed out this lets `allocate` bind beside a terminating anti-affine pod when the node has idle capacity, which upstream never does. Confirmed by test (unexpected `MockCache.Bind`).
2. `92b8c119` introduced a second, post-release affinity view (own lister, InterPodAffinity instance and cycle state) used only for placements that will be Pipelined. Correct, but 22 files.
3. This commit replaces it with @enoodle's suggestion: scope the exclusion to session-selected victims via the marker that already exists. Same test matrix and the same production replay result as the dual view (identical 68-pod victim set), a fraction of the surface.

## Related Issues

Fixes #2115

## Checklist

- [x] Self-reviewed
- [ ] Updated documentation (if needed) — none required; the invariant is documented on `excludedFromPodAffinity` and at the two call sites

## Breaking Changes

None. Behavioural change: pods evicted by the current scheduling cycle no longer satisfy or contribute to inter-pod (anti-)affinity for placements made in that cycle. Pods without inter-pod affinity terms, bind-now placements, scoring, and pods terminating independently of the scheduler are unaffected.

### Testing performed

**Matrix.** The action tests were run against `main` (6d498486) and this branch. All scenarios: one 1-GPU `Releasing` pod on `node0` (terminating independently, i.e. not a session victim), one 1-GPU pending pod, `kubernetes.io/hostname` topology.

| scenario | `main` | this PR |
|---|---|---|
| A. idle GPU on the node; pending pod has required anti-affinity against the releasing pod → expect **Pending** (upstream semantics) | pass | pass |
| B. node full; only the independently terminating anti-affine pod blocks → expect **Pending** (not a victim, stays indexed) | pass | pass |
| C. as A, anti-affinity on the *releasing* pod (symmetric check) → expect **Pending** | pass | pass |
| D. as B, symmetric → expect **Pending** | pass | pass |
| E. control: no anti-affinity → expect **Binding** | pass | pass |
| reclaim: victim carries the label the pending job is anti-affine to → expect victim **Releasing**, pending **Pipelined** | **FAIL** — `Running` / `Pending` | pass |
| reclaim control: a non-victim matching pod stays → expect **Pending** | pass | pass |

Cases A–E pass on `main` by construction (the bug only affects session victims); they pin the upstream semantics so that neither of the two earlier designs can come back unnoticed — commit 1 fails A with an unexpected `Bind`, commit 2 turned B and D into `Pipelined`.

1. **`pkg/scheduler/actions/allocate/allocate_pod_anti_affinity_test.go`** — `TestAllocateWithRequiredAntiAffinityAgainstReleasingPods`, scenarios A–E.
2. **`pkg/scheduler/actions/reclaim/reclaim_pod_anti_affinity_test.go`** — `TestReclaimWithRequiredAntiAffinityAgainstVictims`, the two reclaim rows.
3. **`pkg/scheduler/api/node_info/node_info_releasing_affinity_test.go`** — `TestNodeInfoSessionEvictedPodAffinity`, gomock `NodePodAffinityInfo` with exact `.Times()`: Running add/remove (1/1); independently terminating Releasing (1/1, stays indexed); session-evicted Releasing (0/0); evict `Running→Releasing+virtual` (1/1); unevict `Releasing+virtual→Running` (1/0); `Running→StuckInReleasing` (2/1); evicted victim pipelined elsewhere (1/0, indexed again).
4. **Test utility** — `TestTaskBasic.PodAntiAffinityLabels`, so a fake pod can carry a required anti-affinity against *other* pods' labels; `nil` keeps the previous behaviour.
5. **Regression.** `go test ./pkg/scheduler/... -count=1`: 78 packages pass. `go build`, `go vet`, `gofmt` clean; golangci-lint v2.11.3 with the repo config on the changed packages: 0 issues.
6. **Production replay** — `cmd/snapshot-tool` built from v0.17.0 plus exactly this change, against the live v0.17.0 snapshot of the stuck cluster (115 nodes, 2,977 pods, 4,605 PodGroups), snapshot unmodified, two runs:
   - unpatched v0.17.0: 2,409 `didn't match pod anti-affinity rules` lines, 0 jobs reclaimed;
   - this change: **0** rejections; the 8-task gang → `Scenario solved for 8 tasks … Reclaimed resources`, all 8 workers pipelined; the stuck 48-GPU job reclaimed as well (2 workers pipelined);
   - the gang's victim set (68 pods: 21 `default`, 47 `skypilot`) is **identical** to the dual-view build's (set difference: 0).

### Design decisions

1. **Victim-scoped exclusion, using the existing marker.** A pod is invisible to affinity exactly when *this cycle* decided to evict it. That is the only case where the resource view (future-free) and the affinity view (present) disagree, and the only case where the scheduler can be sure the pending pod will never be bound while the victim is still there. Independently terminating pods are handled the way upstream handles them: wait for the delete event.
2. **`RemoveTask` keys on the stored clone.** The node keeps its own copy of each task, so the removal decision reflects what `addTask` actually indexed, even though the job's task object has already changed status when `UpdateTask` runs.
3. **`StuckInReleasing` stays indexed.** KAI refuses to pipeline onto StuckInReleasing resources, so the pod is genuinely present from the scheduler's point of view.

### Reviewer guidance

The change is `excludedFromPodAffinity` and its two call sites in `node_info.go`, plus the moved assignment in `Statement.Evict`. Everything else is tests.

### DIFF
--- a/pkg/scheduler/api/node_info/node_info.go
+++ b/pkg/scheduler/api/node_info/node_info.go
@@ -95,8 +95,22 @@ type NodeInfo struct {
 	NumaTopology *NumaTopology
 
 	PodAffinityInfo pod_affinity.NodePodAffinityInfo
+	// PipelinePodAffinityInfo is the pipeline (post-release) affinity view of this node:
+	// the same pods minus those in Releasing status. It is consulted only for placements
+	// that will be Pipelined, so a pending pod's inter-pod (anti-)affinity is evaluated
+	// against the pods that will still be there when it binds. nil when not maintained.
+	PipelinePodAffinityInfo pod_affinity.NodePodAffinityInfo
 
 	GpuSharingNodeInfo
+}
+
+// NewNodeInfoWithPipelineAffinity is NewNodeInfo with a second, pipeline-view affinity
+// index (see PipelinePodAffinityInfo). pipelinePodAffinityInfo may be nil.
+func NewNodeInfoWithPipelineAffinity(node *v1.Node, podAffinityInfo, pipelinePodAffinityInfo pod_affinity.NodePodAffinityInfo,
+	vectorMap *resource_info.ResourceVectorMap) *NodeInfo {
+	ni := NewNodeInfo(node, podAffinityInfo, vectorMap)
+	ni.PipelinePodAffinityInfo = pipelinePodAffinityInfo
+	return ni
 }
 
 func NewNodeInfo(node *v1.Node, podAffinityInfo pod_affinity.NodePodAffinityInfo, vectorMap *resource_info.ResourceVectorMap) *NodeInfo {
@@ -416,27 +430,30 @@ func (ni *NodeInfo) addTask(task *pod_info.PodInfo, allowTaskToExistOnDifferentG
 
 	ni.addTaskResources(task)
 	ni.addTaskStorage(task)
-	if !excludedFromPodAffinity(task.Status) {
-		ni.PodAffinityInfo.AddPod(task.Pod)
+	ni.PodAffinityInfo.AddPod(task.Pod)
+	if ni.PipelinePodAffinityInfo != nil && !excludedFromPipelineAffinity(task.Status) {
+		ni.PipelinePodAffinityInfo.AddPod(task.Pod)
 	}
 	return nil
 }
 
-// excludedFromPodAffinity reports whether a task in the given status is left out of
-// the node's inter-pod (anti-)affinity index.
+// excludedFromPipelineAffinity reports whether a task in the given status is left out
+// of the node's pipeline (post-release) inter-pod affinity view. The full view
+// (PodAffinityInfo) always indexes every active task, matching upstream kube-scheduler,
+// which keeps a pod until its delete event.
 //
 // A Releasing task is one the scheduler has decided to evict, or that is already
 // terminating. Its resources are already treated as future-free so a pending pod can
-// be Pipelined onto them, and a Pipelined pod is never bound while the releasing pod
-// is still on the node. Indexing it for affinity makes the two views disagree:
-// reclaim and preempt free the resources, but the k8s InterPodAffinity filter still
-// sees the victim, so a pending pod with a required anti-affinity against the victims
-// fails "didn't match pod anti-affinity rules" in every scenario, however many
-// victims are chosen.
+// be Pipelined onto them, and a Pipelined pod is never bound while the releasing pod is
+// still on the node. Evaluating a pipelined placement against the full view therefore
+// makes the resource view and the affinity view disagree: reclaim and preempt free the
+// resources, but the InterPodAffinity filter still sees the victim, and a pending pod
+// with a required anti-affinity against the victims fails "didn't match pod
+// anti-affinity rules" in every scenario. The pipeline view drops exactly those pods.
 //
-// StuckInReleasing stays indexed: its resources are not pipelinable either, so the
-// pod is treated as present in both views.
-func excludedFromPodAffinity(status pod_status.PodStatus) bool {
+// StuckInReleasing stays in both views: its resources are not pipelinable either, so
+// the pod is treated as present everywhere.
+func excludedFromPipelineAffinity(status pod_status.PodStatus) bool {
 	return status == pod_status.Releasing
 }
 
@@ -537,13 +554,17 @@ func (ni *NodeInfo) RemoveTask(ti *pod_info.PodInfo) error {
 
 	ni.removeTaskStorage(task)
 	ni.removeTaskResources(task)
-	// task is the stored copy, so its status is the one addTask indexed under. A task
-	// that was never indexed must not be un-indexed: the k8s NodeInfo.RemovePod
-	// returns "not found" and would fail UpdateTask (and the unevict rollback).
-	if excludedFromPodAffinity(task.Status) {
-		return nil
+	if err := ni.PodAffinityInfo.RemovePod(task.Pod); err != nil {
+		return err
 	}
-	return ni.PodAffinityInfo.RemovePod(task.Pod)
+	// task is the stored copy, so its status is the one addTask indexed under. A task
+	// that was never added to the pipeline view must not be removed from it: the k8s
+	// NodeInfo.RemovePod returns "not found" and would fail UpdateTask (and the
+	// unevict rollback).
+	if ni.PipelinePodAffinityInfo != nil && !excludedFromPipelineAffinity(task.Status) {
+		return ni.PipelinePodAffinityInfo.RemovePod(task.Pod)
+	}
+	return nil
 }
 
 func (ni *NodeInfo) removeTaskResources(task *pod_info.PodInfo) {
--- a/pkg/scheduler/api/node_info/node_info.go
+++ b/pkg/scheduler/api/node_info/node_info.go
@@ -416,8 +416,28 @@ func (ni *NodeInfo) addTask(task *pod_info.PodInfo, allowTaskToExistOnDifferentG
 
 	ni.addTaskResources(task)
 	ni.addTaskStorage(task)
-	ni.PodAffinityInfo.AddPod(task.Pod)
+	if !excludedFromPodAffinity(task.Status) {
+		ni.PodAffinityInfo.AddPod(task.Pod)
+	}
 	return nil
+}
+
+// excludedFromPodAffinity reports whether a task in the given status is left out of
+// the node's inter-pod (anti-)affinity index.
+//
+// A Releasing task is one the scheduler has decided to evict, or that is already
+// terminating. Its resources are already treated as future-free so a pending pod can
+// be Pipelined onto them, and a Pipelined pod is never bound while the releasing pod
+// is still on the node. Indexing it for affinity makes the two views disagree:
+// reclaim and preempt free the resources, but the k8s InterPodAffinity filter still
+// sees the victim, so a pending pod with a required anti-affinity against the victims
+// fails "didn't match pod anti-affinity rules" in every scenario, however many
+// victims are chosen.
+//
+// StuckInReleasing stays indexed: its resources are not pipelinable either, so the
+// pod is treated as present in both views.
+func excludedFromPodAffinity(status pod_status.PodStatus) bool {
+	return status == pod_status.Releasing
 }
 
 func (ni *NodeInfo) AddTasksToNode(podInfos []*pod_info.PodInfo,
@@ -517,9 +537,13 @@ func (ni *NodeInfo) RemoveTask(ti *pod_info.PodInfo) error {
 
 	ni.removeTaskStorage(task)
 	ni.removeTaskResources(task)
-	err := ni.PodAffinityInfo.RemovePod(task.Pod)
-
-	return err
+	// task is the stored copy, so its status is the one addTask indexed under. A task
+	// that was never indexed must not be un-indexed: the k8s NodeInfo.RemovePod
+	// returns "not found" and would fail UpdateTask (and the unevict rollback).
+	if excludedFromPodAffinity(task.Status) {
+		return nil
+	}
+	return ni.PodAffinityInfo.RemovePod(task.Pod)
 }
 
 func (ni *NodeInfo) removeTaskResources(task *pod_info.PodInfo) {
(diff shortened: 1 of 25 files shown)


pkg/scheduler/api/node_info/node_info.go
```diff
@@ -416,10 +416,30 @@ func (ni *NodeInfo) addTask(task *pod_info.PodInfo, allowTaskToExistOnDifferentG
...
+// is still on the node. Indexing it for affinity makes the two views disagree:
+// reclaim and preempt free the resources, but the k8s InterPodAffinity filter still
+// sees the victim, so a pending pod with a required anti-affinity against the victims
+// fails "didn't match pod anti-affinity rules" in every scenario, however many
+// victims are chosen.
+//
+// StuckInReleasing stays indexed: its resources are not pipelinable either, so the
+// pod is treated as present in both views.
+func excludedFromPodAffinity(status pod_status.PodStatus) bool {
+	return status == pod_status.Releasing
```
gshaibi: Thanks @sam-huang1223 .
I wonder if this fix will cause another (maybe less acute) bug where pod placement may not respect anti-affinity with terminating pods. 
Do you know how does the upstream scheduler handle that?

sam-huang1223: Good catch, and confirmed with a test: with this change as written, `allocate` would bind a pod next to a terminating anti-affine pod when the node also has idle capacity, which upstream never does (kube-scheduler keeps the pod cached until the delete event, and preemption waits on nominatedNodeName while victims terminate). Eviction-based actions here are unaffected since they place pipeline-only, but the allocate path is a real regression.

Reworking the PR so Releasing pods are excluded only for placements that will be Pipelined: a second post-release view at node and shared-lister level, with the bind-now path keeping today's full view (upstream semantics, and unchanged scoring). Will push as a follow-up commit with a test for exactly this scenario.

sam-huang1223: Pushed the rework as a separate commit (92b8c119) so the delta stays reviewable.

The full index is back to upstream semantics (Releasing pods included) and is what bind-now placements, `RecomputeDetailedFitErrors` and scoring use. Placements that will be Pipelined (every eviction-based action, and `allocate` onto a node whose idle resources don't fit the task) evaluate inter-pod affinity against a second, post-release view that omits Releasing pods. That view has its own InterPodAffinity instance and lister, because PreFilter reads the symmetric existing-anti-affinity counts from the plugin's own lister, plus a namespaced CycleState.

`TestAllocateWithRequiredAntiAffinityAgainstReleasingPods` case A is exactly your scenario: idle GPU next to a terminating anti-affine pod → stays Pending. Commit 1 fails it with an unexpected `MockCache.Bind`; main and this branch pass it. The PR description now has the full main / commit 1 / now matrix, the replay against the same production snapshot (identical victim set), and the trade-offs I'd like eyes on (fractional-GPU path stays conservative under allocate; one extra PreFilter per task).

enoodle: Thanks for addressing the bind-now case. I wonder whether we can keep this simpler and scope the exception to scheduler-selected victims instead of introducing a second cluster-wide affinity view.

`Statement.Evict` is already the common path for reclaim, preempt, and consolidation. Could it mark the task as a session-virtual eviction before `node.UpdateTask`, and have the existing affinity index exclude only tasks carrying that marker? The marker would be restored on `Unevict`/rollback.

That would let the speculative pipeline placement ignore exactly the victims selected by this scheduling cycle, while Pods that are independently terminating in the cluster would remain in the normal affinity index—matching kube-scheduler for bind-now and ordinary allocate paths.

It seems to solve the original reclaim/preempt issue with a much narrower semantic change, and avoids maintaining two `NodeInfo`/lister/plugin/cycle-state views.

sam-huang1223: Agreed, and thank you — that is the right design. Pushed as 3d727c24 on top of the previous commits so the deltas stay visible.

The marker already existed: `Statement.Evict` sets `IsVirtualStatus` on the victim, and `Releasing && IsVirtualStatus` already means "freed by this cycle" in `pod_info.go`. It was just set *after* `node.UpdateTask` re-indexed the victim. The change is now: mark before re-indexing (restored on failure; `unevict` already restores it before re-indexing), and `addTask`/`RemoveTask` skip the index for `Releasing && IsVirtualStatus` tasks, with `RemoveTask` deciding on the node's stored clone so evict/unevict/StuckInReleasing stay balanced. Two source files, +24/−4, plus tests; the dual view is gone.

Verification is the same as before: full `./pkg/scheduler/...` suite passes; the allocate matrix keeps upstream semantics for independently terminating pods (they stay indexed, so allocate neither binds beside them nor pipelines onto them); the reclaim test pipelines onto session victims; and the replay of the stuck production snapshot (v0.17.0 + exactly this change, two runs) reclaims the 8-node gang with the identical 68-pod victim set the dual-view build produced. Details and the three-round history are in the updated PR description.

pkg/scheduler/test_utils/tasks_fake/tasks.go
```diff
@@ -163,12 +165,15 @@ func IsTaskStartedStatus(status pod_status.PodStatus) bool {
...
-	affinity *v1.Affinity, podAffinityLabels map[string]string,
+	affinity *v1.Affinity, podAffinityLabels, podAntiAffinityLabels map[string]string,
 	podAffinityTopologyKey, podAntiAffinityTopologyKey string,
 ) *v1.Affinity {
 	if affinity == nil {
 		affinity = &v1.Affinity{}
 	}
+	if podAntiAffinityLabels == nil {
+		podAntiAffinityLabels = podAffinityLabels
+	}
```
davidLif: And what happens if a pod has both affinity and anti affinity labels? Shouldn't we merge them?

sam-huang1223: Good question, and the name was inviting it. In this helper `PodAffinityLabels` does double duty: it is put on the fake pod's labels *and* it is the default selector of both the affinity and the anti-affinity term, so a task attracts/repels its own kind. The new field only overrides the anti-affinity term's `matchLabels` and is never added to the pod's labels: the tests need a pending pod that repels another tier (e.g. `tier=preprocess`) without carrying that label itself. When both are set, the pod carries `PodAffinityLabels`, attracts them (if a podAffinity topology key is given) and repels the override set.

Merging would not work for two reasons: `matchLabels` is a conjunction, so a merged selector matches fewer pods, not more, and it cannot express two values of the same key at all — the symmetric test cases have a `tier=preprocess` pod repelling `tier=train`. The legacy self-repelling behaviour is still the default when the field is nil, so no existing test changed.

Renamed it to `PodAntiAffinitySelector` and documented the override semantics at the field and in the helper (a61e0828); test-utility only, no scheduler code touched.

enoodle (dismissed):

enoodle (approved):

davidLif (approved):

PR #2117, opened 2026-09-03
fix(scheduler): surface all errors from Pipeline

phiott:
## Description

`Statement.Pipeline` swallowed the error from `job.UpdateTaskStatus`. For a task the job no longer tracks, for example the pod was deleted mid-session, the pipeline attempt still reported success. The allocate action then saw a job with tasks left to allocate, requeued it, and repeated the same attempt for the rest of the session. The scheduler leader sat at 100% CPU and scheduled nothing else until the pod was restarted.

This change fails the attempt instead, similar to `Statement.Allocate`, and stops `Pipeline` from swallowing both `job.UpdateTaskStatus` and the `node.UpdateTask` error. The statement is discarded, the job is skipped for this session, and the next snapshot rebuilds it without the stale task.

The log line repeats at hundreds per minute while the leader is wedged:

```
Failed to update task <ns/pod> status to Pipelined in Session <id>: failed to find task <ns/pod> in job <ns/pg>
```

## Checklist

- [x] Self-reviewed

## Breaking Changes

None.

## Additional Notes

Two new tests fail on `main` and pass with this change:

- `TestStatement_Pipeline_TaskNotInJob`: calls `Pipeline` for a task whose UID is not in its job's pod map. With the change it returns an error, the task stays `Pending` with no node, the node's pod list does not gain the task, and the statement records no operation. On `main` no error is returned.
- `TestAllocateTerminatesWithStaleTaskInCache`: runs the allocate action on a job whose cached tasks-to-allocate list holds a task not in the job's pod map, while the job's real task is still pending. With the change the action returns and the job stays pending. On `main` the action requeues the job on every iteration and the test fails on its 10 s timeout.


.changes/unreleased/fixed-20260903-135006.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Fixed
+body: |-
+  Statement.Pipeline now returns the error when a task's status cannot be updated in its PodGroup (task not found in job) instead of logging it and continuing. Previously the allocate action treated such an attempt as successful and requeued the job forever, pinning the scheduler leader at 100% CPU until it was restarted.
```
enoodle: Can you rephrase it to be a simpler description to someone not versed in the scheduler internal functions? something like "Fixed a race condition that would get the scheduler pod stuck if pods are deleted mid action"

pkg/scheduler/actions/allocate/allocate_stale_task_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// Populate the job's tasks-to-allocate cache, then make it point at a task the job does not
+	// track: a clone of the pending task under a UID that is not in the job's pod map. The tracked
+	// task itself stays pending, so the job still looks like it has work to do.
+	cached := podgroup_info.GetTasksToAllocate(job, ssn.SubGroupOrderFn, ssn.TaskOrderFn, true)
+	require.Len(t, cached, 1)
+	staleTask := trackedTask.Clone()
+	staleTask.UID = common_info.PodID("pending_job-0-stale")
+	cached[0] = staleTask
+	require.Same(t, staleTask, podgroup_info.GetTasksToAllocate(job, ssn.SubGroupOrderFn, ssn.TaskOrderFn, true)[0],
```
enoodle: Can you please explain how this could happen on a real cluster?

pkg/scheduler/framework/statement.go
```diff
@@ -240,8 +240,11 @@ func (s *Statement) Pipeline(task *pod_info.PodInfo, hostname string, updateTask
 
 	previousStatus := task.Status
 	if err := job.UpdateTaskStatus(task, pod_status.Pipelined); err != nil {
+		// UpdateTaskStatus fails only when the task's UID is not in the job's pod map.
```
enoodle: Maybe we don't need those comments. I am not sure that they will be precise forever if UpdateTaskStatus is updated but it may get stuck here anyway. It is not adding much.

enoodle (changes requested):
Thank you for fixing this.
I am not sure that I understand how this could actually happen on a real cluster because the snapshot is isolated and is not getting updates from the informers. I we should merge the core change anyway but I would like to get explanation on that.
One additional review concern: the PR also returns node.UpdateTask errors, but that occurs after the job status and task.NodeName have already changed, and before the rollback operation is recorded. Discard() therefore cannot restore those earlier mutations. That part needs transactional rollback or focused coverage.

pkg/scheduler/actions/allocate/allocate_stale_task_test.go
phiott: Commented in https://github.com/kai-scheduler/KAI-Scheduler/pull/2117#issuecomment-5537352838, I'll just remove this test case since it's not accurate.

enoodle (approved):

SiorMeir (approved):

pkg/scheduler/framework/statement.go
itsomri, line 273: What about this one?

PR #2155, opened 2026-09-11
fix(binder): stop reporting a misleading Bound event after a repeated bind failure

pujitha24:
## Description

When a BindRequest's bind fails and its `Status.Phase` is already `Failed` from a previous attempt (the phase doesn't change on this reconcile), `BindRequestReconciler.updatePodCondition` decided success/failure by inspecting the error/result returned from `UpdateStatus`. `UpdateStatus` intentionally returns a `nil` error in that phase-unchanged case (to avoid a duplicate controller-runtime requeue on top of its own `RequeueAfter`/backoff), but `updatePodCondition` treated that `nil` as "the bind succeeded" and recorded a `Pod bound successfully to node X` event/condition on the very same reconcile where the bind had just failed. This is the secondary "misleading event" defect called out in the linked issue.

This PR decouples `updatePodCondition`'s success/failure decision from that reused err/RequeueAfter signal and bases it directly on `bindRequest.Status.Phase`, which `UpdateStatus` always sets correctly regardless of whether it patches the object or short-circuits. As a side benefit this also fixes the same false-"Bound"-event class of bug for the (currently unused in production) `BackoffLimit`-set path, where a real failure with `RequeueAfter > 0` previously also reported a false `Bound` event. `UpdateStatus` itself, and its interaction with controller-runtime's requeue/rate-limiter handling, is left byte-for-byte unchanged.

**Out of scope:** the issue also describes a larger, separate problem — jobs whose binds repeatedly fail keep re-winning GPU allocation every scheduling cycle, starving other schedulable pods. A maintainer explicitly said that needs more design thought, and another contributor is already iterating on a design for it in the issue thread. This PR only fixes the narrower, independent misleading-event defect and does not touch that allocation/starvation behavior, so it does not fully close the linked issue.

**Validation note:** this environment has severe, pre-existing disk space constraints unrelated to this change, so `make build` (Docker-based) and the full `make test` could not be run locally. I ran the targeted package tests, vet, and lint instead (see below) and verified the regression test fails against the pre-fix logic and passes with the fix.

```
go test ./pkg/binder/controllers -run TestBindRequest -v   # 18/18 pass
go vet ./pkg/binder/controllers                             # clean
golangci-lint run pkg/binder/controllers/...                # 0 issues
gofmt -l on both changed files                               # clean
```

## Related Issues

Relates to #2075 (does not fully fix it — see "Out of scope" above)

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. `updatePodCondition`'s signature changed (dropped unused `result`/`err` params) but it is a private method with no external callers.

## Additional Notes

CI on this repo requires a maintainer to add `ok-to-test` (if gated) — happy to address anything it surfaces. Disclosing explicitly: I could not run the Docker-based `make build` or full `make test` locally due to this environment's disk space, but the targeted unit tests directly reproduce and verify the fix for the defect described here (failing before the fix, passing after).

### DIFF
--- a/.changes/unreleased/fixed-20260911-050705.yaml
+++ b/.changes/unreleased/fixed-20260911-050705.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: |-
+  Binder no longer logs a misleading bind-success event after a bind failure on the same reconcile
--- a/pkg/binder/controllers/bindrequest_controller.go
+++ b/pkg/binder/controllers/bindrequest_controller.go
@@ -119,7 +119,7 @@ func (r *BindRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 		result, err = r.UpdateStatus(ctx, bindRequest, result, err)
 		if pod != nil {
-			r.updatePodCondition(ctx, bindRequest, pod, result, err)
+			r.updatePodCondition(ctx, bindRequest, pod)
 		}
 
 		if finalError != nil {
@@ -256,7 +256,7 @@ func (r *BindRequestReconciler) UpdateStatus(
 }
 
 func (r *BindRequestReconciler) updatePodCondition(
-	ctx context.Context, bindRequest *schedulingv1alpha2.BindRequest, pod *v1.Pod, result ctrl.Result, err error,
+	ctx context.Context, bindRequest *schedulingv1alpha2.BindRequest, pod *v1.Pod,
 ) {
 	logger := log.FromContext(ctx)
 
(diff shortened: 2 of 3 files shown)


pujitha24:
Good catch, thanks. Updated the description — "Fixes #2075" is now "Relates to #2075 (does not fully fix it — see \"Out of scope\" above)" so merging this won't auto-close the issue. No code changes here, just the description.

enoodle (approved):

itsomri (approved):

PR #2157, opened 2026-09-12
fix(binder): fail closed when hamicore cannot calculate GPU memory limit

### LINKED ISSUE #2156: Isolator fails open: LD_PRELOAD injected without CUDA_DEVICE_MEMORY_LIMIT, pod runs uncapped while logging as enforcing
### What happened?

After a GPU Operator restart, the isolator webhook injected `/etc/ld.so.preload`
into new pods but did not set `CUDA_DEVICE_MEMORY_LIMIT`. `libvgpu.so` loaded,
had no limit to enforce, and permitted the full card.

The pods ran **uncapped at 81559 MiB** instead of their 40779 MiB cap, and logged
`HAMI-core ... Initialized` exactly like their correctly-capped peers. Nothing
surfaced the fault.

## What was observed

16 engines deployed with `gpu-fraction: "0.5"` on an 8-GPU node. The first three
pods created came up uncapped; the remaining thirteen were capped correctly.

| | engine 1 (uncapped) | engine 5 (capped) |
|---|---|---|
| `/etc/ld.so.preload` | `/usr/local/vgpu/libvgpu.so` | `/usr/local/vgpu/libvgpu.so` |
| `CUDA_DEVICE_MEMORY_LIMIT` | **unset** | `40779m` |
| `nvidia-smi` inside pod | **81559 MiB** | 40779 MiB |
| `gpu-fraction` annotation | `"0.5"` | `"0.5"` |
| `runai/shared-gpu-configmap` | present | present |
| `HAMI-core ... Initialized` in log | **yes** | yes |

The annotation and the configmap were present on every pod. Only the env var
differed.

## Reproduction

The three affected engines were the first three deployed (10s stagger), shortly
after the GPU Operator's device-plugin, gpu-feature-discovery and
operator-validator pods had been restarted.

Deleting and re-applying those three, with no other change, brought all three
back capped:

| engine | before | after re-apply |
|---|---|---|
| 1 | unset / 81559 MiB | `40779m` / 40779 MiB |
| 2 | unset / 81559 MiB | `40779m` / 40779 MiB |
| 3 | unset / 81559 MiB | `40779m` / 40779 MiB |

3 of 3, deterministic. That points at a readiness race in the webhook — pods
created before the webhook is ready to set the env var get the preload injected
but no limit.

## Why this is worth fixing

The failure mode is fail-open rather than fail-closed. A pod that should have
been rejected instead runs with roughly double its memory budget, and every
available health signal says it is fine:

- the annotation is present
- the preload library is injected and initialises
- `HAMI-core ... Initialized` appears in the log, same as a capped pod
- nothing is logged as an error

The only things that distinguish enforced from attached-but-inert are the
`CUDA_DEVICE_MEMORY_LIMIT` env var and the MiB figure from `nvidia-smi` inside
the pod.

In our case this invalidated a benchmark — three engines that could allocate the
whole card are not memory-constrained like their peers and can starve the capped
co-tenant sharing their card. In a production multi-tenant setting the same
behaviour means one tenant silently consuming another's memory budget.

## Suggested fixes

Either would be an improvement on the current behavior:

- **Fail closed.** If the webhook can inject `LD_PRELOAD` it should also be able
  to set `CUDA_DEVICE_MEMORY_LIMIT`; if it cannot, reject the pod rather than
  admitting it uncapped.
- **Have libvgpu refuse to run without a limit.** If the preload loads and finds
  no `CUDA_DEVICE_MEMORY_LIMIT`, that is always a misconfiguration — erroring
  out would surface it immediately instead of silently permitting the full card.

A log line at warn level when the preload loads with no limit set would at least
make it detectable, though it would not prevent the overcommit.

A pod carrying a `gpu-fraction` annotation should either be admitted with its
memory cap enforced, or rejected.

If the isolator webhook cannot set `CUDA_DEVICE_MEMORY_LIMIT` — for any reason,
including not being ready yet — the pod should not be admitted with `LD_PRELOAD`
injected and no limit. That combination produces a pod that appears enforced by
every available signal while actually having access to the entire card.

### Environment

- KAI-Scheduler v0.17.1
- kai-resource-isolator 1.1.0-chart (docker.io/projecthami)
- `binder.plugins.hamicore.enabled: true`, `global.gpuSharing: true`
- 8x NVIDIA H100 80GB, single node
- NVIDIA driver 580.126.20
- NVIDIA GPU Operator v25.10.1
- microk8s v1.35.6, 1 node
- Workload: vLLM serving Qwen3-8B, one engine per pod, `gpu-fraction: "0.5"`
  (40779 MiB cap)

pujitha24:
## Description

The `hamicore` binder plugin's `PreBind` swallowed any error from
`calculateCudaDeviceMemoryLimit` and returned `nil` (success). When the
calculation fails — for example because the node is momentarily missing the
`nvidia.com/gpu.memory` label right after the GPU Operator's
gpu-feature-discovery pod restarts — the plugin now used to silently skip
setting `CUDA_DEVICE_MEMORY_LIMIT` and let the bind proceed anyway.

Since the admission webhook injects that env var from a ConfigMap key with
`Optional: true`, a pod could end up admitted with `LD_PRELOAD` (HAMI-core)
active but no memory limit enforced, running uncapped while logging
`HAMI-core ... Initialized` just like a correctly capped pod.

This PR makes `PreBind` propagate the error instead of swallowing it, so the
bind fails and is retried instead of completing uncapped.

## Related Issues

Fixes #

## Approach

`pkg/binder/plugins/hamicore/hami_core.go`: `PreBind` now returns a wrapped
error when `calculateCudaDeviceMemoryLimit` fails, instead of `return nil`.
`Binder.Bind` (`pkg/binder/binding/binder.go`) returns the `PreBind` error
before patching bind annotations or creating the `Binding` subresource, so the
pod is never bound to the node in this case — the bind request controller
rolls back and the reconcile is requeued, rather than the pod starting
uncapped.

## Validation

- Added `TestPreBindFailsClosedWhenMemoryLimitCannotBeCalculated` in
  `pkg/binder/plugins/hamicore/hami_core_test.go`: builds a node missing the
  `nvidia.com/gpu.memory` label and asserts `PreBind` returns an error. Verified
  it **fails on the old code** (`got nil error`) and **passes with the fix**.
- `go build ./...` — passes.
- `go test ./pkg/binder/plugins/hamicore/...` — passes.
- `go test ./pkg/binder/...` — all packages pass except
  `pkg/binder/controllers/integration_tests`, which fails in this environment
  only because the envtest control-plane binary (`/usr/local/kubebuilder/bin/etcd`)
  isn't installed here; that suite doesn't touch the changed package and the
  failure is pre-existing/environmental, not caused by this change.
- `go vet ./pkg/binder/...` — clean.
- `golangci-lint run ./pkg/binder/plugins/hamicore/...` — 0 issues.
- `gofmt -l` on both changed files — clean.
- Not run: a live cluster reproduction with a real GPU Operator restart (no
  hardware/cluster available here). The fix is a straightforward error
  propagation whose effect is proven by the unit test and by reading
  `Binder.Bind`'s control flow, which confirms no path binds the pod after a
  `PreBind` error.

## Checklist

- [x] Self-reviewed

## Breaking Changes

None. Behavior only changes for the narrow failure case where
`CUDA_DEVICE_MEMORY_LIMIT` cannot be calculated (e.g. missing GPU memory node
label): the pod bind now fails and retries instead of completing without
memory enforcement.

## Additional Notes

Root cause and fix suggested by the issue reporter ("fail closed" option).
Reported originating scenario is a readiness race after a GPU Operator
restart; this fix addresses the observed fail-open code path rather than the
underlying provider timing itself.

Fixes #2156


enoodle (approved):
LGTM

itsomri (approved):

PR #2167, opened 2026-09-15
fix(admission): align queue quota validation across create and update

### LINKED ISSUE #1881: Parent/child quota validation sums children quotas without handling the -1 unlimited sentinel

### LINKED ISSUE #1923: queue validator: validateParentChildQuota only sums CPU across siblings
In pkg/admission/webhook/queuehooks/queue_validator.go, validateParentChildQuota checks an individual child against its parent for CPU, GPU and memory, but the sibling total-sum over-subscription check only accumulates and compares CPU. GPU and memory sibling sums are not validated on the queue create path.

The other function, validateChildrenQuotaSum (used on the update path), already sums CPU, GPU and memory. So the create and update paths are inconsistent: a set of sibling queues whose combined GPU or memory quota exceeds the parent is caught on update but not on create.

Suggested fix: accumulate the GPU and memory sibling totals in validateParentChildQuota too and compare them with the parent using the existing -1-aware helpers (quotaExceeds / addQuota), mirroring validateChildrenQuotaSum.

Found during review of #1883.

eladb:
## Description

Queue quota admission validation was inconsistent between the create and update paths. This aligns them:

- Sum sibling GPU and memory quotas against the parent on create. Previously only CPU was summed; GPU/memory were checked per-child but never totaled, so siblings could over-subscribe the parent on create while being rejected on update.
- Treat `-1` (unlimited) correctly via `quotaExceeds`/`addQuota` helpers instead of comparing and summing it as a plain number. Unlimited prints as `unlimited`.
- Tolerate float drift so fractional GPU quotas that sum to the parent (e.g. `0.1 + 0.2` vs `0.3`) don't spuriously warn.
- Emit per-child GPU/memory warnings on the update path, matching create.

No API or CRD changes.

## Related Issues

Fixes #1923
Fixes #1881

## Checklist

- [x] Self-reviewed
- [x] Added/updated tests
- [x] Added changelog fragment

## Testing Notes

`queue_validator_test.go` covers the sibling sums, `-1` semantics on both paths, and the fractional-sum case. The `-1` and fractional tests fail before the change and pass after.


eladb:
Yes, warnings only — both `ValidateCreate`/`ValidateUpdate` return warnings with a nil error, so oversubscription is still admitted. Added test `216ccd9` to lock that in.

enoodle (approved):

itsomri (approved):

eladb:
@enoodle @itsomri Would you be able to approve the workflow runs?

PR #2176, opened 2026-09-16
perf(scheduler): guard disabled verbose logs

enoodle:
## Description

Avoid allocations from disabled scheduler verbose logs.

`log.InfraLogger.V(...).Infof(...)` evaluates formatting arguments before the
logger can discard a disabled message. This is expensive in hot scheduling
paths, particularly task-by-node loops.

This change:
- Guards every production and command `InfraLogger` log at V(5)+ with
  `VerboseLogger.Do`.
- Defers formatting and diagnostic-only work, including node-name collection,
  device rendering, detailed fit errors, and resource/vector stringification.
- Adds `SchedulerLogger.Infof` so guarded logs preserve the existing caller
  attribution while avoiding a second verbosity lookup.
- Keeps log levels and message contents unchanged.
- Adds coverage that disabled guards skip argument evaluation and enabled logs
  retain caller attribution.

## Related Issues

Fixes #

## Checklist

- [x] Self-reviewed
- [x] Added/updated tests
- [ ] Updated documentation (not needed)

## Breaking Changes

None - all logs should stay the same when enabled.


pkg/scheduler/framework/session_plugins.go
```diff
@@ -445,7 +447,7 @@ func logNodeSetsPluginResult(subsetNodesFn api.SubsetNodesFn, podGroup *podgroup
 			}
 			nodeSetNames = append(nodeSetNames, names)
 		}
-		log.InfraLogger.V(7).Infof(
```
SiorMeir, line 448: a typo? why remove the severity level?

SiorMeir (dismissed):

pkg/scheduler/framework/session_plugins.go
enoodle: This is inside a clause (starting line 439) of 
```
log.InfraLogger.V(7)Do(func() { 
```
So it is just redundant. I wanted it to be similar to the rest of the changes.

itsomri (dismissed):

itsomri (approved):

itsomri (approved):

PR #2181, opened 2026-09-16
feat(chart): replace crd-upgrader image with distroless helm-hooks binary

gshaibi:
## Description

All KAI images are now based on `nvcr.io/nvidia/distroless/go`.

The `crd-upgrader` image was the only exception: `alpine` + bash + a downloaded `kubectl`, reused by four Helm hook Jobs (crd-upgrader, kai-config-deployer, topology-migration, post-delete-cleanup) as a shell toolbox. It is replaced by a `helm-hooks` Go binary with four subcommands (`apply-crds`, `apply-config`, `migrate-topologies`, `cleanup`), built through the same `cmd/<name>` + root `Dockerfile` path as every other service. The alpine Dockerfile, the shell scripts and the special Makefile targets are removed.

Details:
- CRDs are compiled in via the existing `deployments/kai-scheduler/crds` embed package, so the binary always ships the CRDs of its own commit (`make manifests` + `make validate` keep them in sync).
- CRD apply keeps `kubectl` as the field manager, so upgrades take over the managedFields entry left by earlier releases instead of adding a second owner. Verified on a kind cluster upgraded from v0.17.2: managers stay `kubectl/Apply`, `helm/Apply`.
- Topology migration now creates `v1alpha1` KAI Topologies. The old script emitted `kai.scheduler/v1`, which the CRD never served, so the hook failed whenever a Kueue Topology existed.
- The Kueue Topology API version is read from the CRD storage version instead of being hardcoded.
- The `-fips` variant of the hook image is now actually FIPS-built (before it was a prebuilt upstream kubectl).
- The `crd-upgrader` Job, `crdupgrader` values key and ServiceAccount keep their names; only the image name changes.

## Related Issues

N/A

## Checklist

- [x] Self-reviewed

## Breaking Changes

The `crd-upgrader` image is no longer published. Anyone overriding a hook `image.name` back to it, or using it as a kubectl toolbox, must switch to `helm-hooks`. Pinning an older hook `image.tag` with this chart will not work since the old image has no subcommands.

## Additional Notes

Verified end to end on a throwaway kind cluster: installed v0.17.2, added the Kueue Topology CRD and a Topology, upgraded to this chart with the new image. All hooks completed, the topology was migrated with the `kai.scheduler/migrated-from` annotation, and `helm uninstall` ran `cleanup`, removing operator-managed deployments and `kai-config`. Also `make test-chart`, unit tests, golangci-lint and `go vet` pass. The image has no `/bin/sh`.

Not included (possible follow-ups): `fipsOnlyEnv` on the other three hook Jobs; polling CRDs for `Established` after apply.


deployments/kai-scheduler/templates/hooks/post/kai-config-deployer/job.yaml
```diff
@@ -66,21 +66,7 @@ spec:
...
-          # Strip leftover Helm hook annotations from previous chart versions
-          # where kai-config was managed as a Helm pre-install,pre-upgrade
-          # hook (#1536). No-op when the CR doesn't exist or annotations are
-          # already absent.
-          kubectl annotate config kai-config \
-            helm.sh/hook- helm.sh/hook-weight- 2>/dev/null || true
-          kubectl apply --server-side --force-conflicts \
-            --field-manager=kai-config-deployer \
-            -f /manifest/kai-config.yaml
+        args: ["apply-config", "--file=/manifest/kai-config.yaml"]
```
SiorMeir, line 69: Something that I have a hard time understanding:
How do you call helm-hooks here? I can see the args but you need to build a complete command e.g:
`helm-hooks apply-config --file...` 
but I can't see that anywhere

What am I missing?

gshaibi: Good question.
In the common Dockerfile, the `helm-hooks` binary is the entrypoint of the image, so whoever runs it can pass just the arguments and it automatically appends them to the `helm-hooks` binary.

SiorMeir: cool, so it is there it's just obscure a bit. 
Sounds good to me, I'm approving

SiorMeir (approved):

PR #2193, opened 2026-09-17
build: use scratch base image instead of distroless

gshaibi:
## Description

Switches the production image base from `nvcr.io/nvidia/distroless/go` to `scratch`.

- All services except `resourcereservation` are cgo-free. They are now built with `CGO_ENABLED=0` (statically linked) and shipped on `scratch`, with only the Debian CA bundle copied in from the `golang` image already used by the debug stage.
- `resourcereservation` imports go-nvml, which needs cgo and dlopens the host NVML library at runtime, so it requires a libc. It stays on distroless via a new `prod-cgo` Dockerfile target.
- `CGO_SERVICES` in `build/makefile/base.mk` (default: `resourcereservation`) selects both the cgo flag and the Docker target per service. Explicit `CGO_ENABLED=1` overrides still work.
- The global `CGO_ENABLED?=1` default in `golang.mk` is removed; it made every binary dynamically linked.

Verified locally by cross-compiling all 15 services with `CGO_ENABLED=0 GOOS=linux`: 14 build, `resourcereservation` fails on go-nvml as expected. `make -n` dry runs pick `--target prod`/cgo off for regular services and `--target prod-cgo`/cgo on for `resourcereservation`.

## Related Issues

Fixes #

## Checklist

- [x] Self-reviewed

## Breaking Changes

None expected. Images have no shell already. `scratch` drops tzdata and `/etc/passwd` compared to distroless; no code uses `time.LoadLocation` or `os/user`, and the numeric `USER 65532` needs neither.

## Additional Notes

Image builds were not run locally (no Docker daemon available); relying on CI build + e2e for that.


Dockerfile
```diff
@@ -17,7 +17,25 @@ USER 65532:65532
 
 ENTRYPOINT ["/go/bin/dlv", "exec", "--headless", "-l", ":10000", "--api-version=2", "/workspace/app", "--"]
 
-FROM nvcr.io/nvidia/distroless/go:v3.2.1 AS prod
+FROM golang:1.26.3 AS certs
```
enoodle: Maybe should use `golang:1.26.3-bookworm` so that it will be debian and we will know for sure that this is the location of the certificates. It is not critical, because thats the same image, but for clarity.

enoodle (dismissed):
I have one comment that you should consider, but it is very minor...

enoodle (approved):

PR #2195, opened 2026-09-17
fix(scheduler): match GPU resources by exact name instead of suffix

### LINKED ISSUE #2194: Scheduler treats any resource ending in "gpu" as a GPU, so HAMi's nvidia.com/vgpu inflates node capacity
## Summary

The scheduler classifies any resource whose name ends in `gpu` as a GPU. On clusters that also run HAMi, nodes advertise `nvidia.com/vgpu` (one slice-count per physical GPU, 80 on an 8-GPU node) next to `nvidia.com/gpu: 8`. KAI folds `nvidia.com/vgpu` into the GPU slot of the node's resource vector, so it believes the node has 80 GPUs and binds pods onto nodes whose 8 real GPUs are fully allocated. Kubelet then rejects the pod:

```
UnexpectedAdmissionError: Pod was rejected: Allocate failed due to requested number of devices unavailable for nvidia.com/gpu. Requested: 1, Available: 0, which is unexpected
```

## Version

v0.17.0. The code is unchanged on `main` and in v0.17.1.

## Cause

`pkg/scheduler/api/resource_info/resource_vector.go`:

```go
func isGpuResource(resourceName v1.ResourceName) bool {
	return strings.HasSuffix(string(resourceName), constants.GpuResource) // "gpu"
}

func normalizeResourceName(resourceName v1.ResourceName) v1.ResourceName {
	if isGpuResource(resourceName) {
		return v1.ResourceName(constants.GpuResource)
	}
	return resourceName
}
```

`ResourceVectorMap.GetIndex` and `AddResource` go through `normalizeResourceName`, so `nvidia.com/vgpu` resolves to `GPUIndex`. `Resource.ToVector` then does `vec.Set(idx, ...)` for scalar resources, which overwrites the real `nvidia.com/gpu` count already placed at `GPUIndex` when `nvidia.com/vgpu` is iterated after it.

By contrast, `ResourceFromResourceList` in `resource_info.go` matches GPUs exactly (`case GPUResourceName, amdGpuResourceName`), so the two paths disagree.

## Evidence

The proportion plugin logs, on a 23-GPU-node cluster:

```
Total allocatable resources are <CPU: 2114.835 (cores), memory: 13787.47 (GB), Gpus: 1839837>, number of nodes: <32>
```

Sum of `nvidia.com/gpu` allocatable across those nodes: 178. Sum of `nvidia.com/vgpu`: 1840.

Observed placement: a Grove PodGang (Dynamo DGD, decode leader requesting `nvidia.com/gpu: 1`) was bound to a node with 8/8 GPUs allocated to other pods; kubelet rejected it. Deleting the Failed pod and retrying reproduced the same result until a GPU was actually released on that node.

## Impact

- Gang members land on full nodes and go to `Failed` rather than `Pending`, so Grove-managed gangs need manual pod deletion to recover.
- The normal "wait for the terminating pod to release its GPU" behaviour is lost: KAI re-binds immediately because it never saw the node as full.
- Any other extended resource ending in `gpu` would trigger the same misaccounting.

## Expected

Only `nvidia.com/gpu` and `amd.com/gpu` (already the exact names in `resource_requirment.go`), plus the internal `gpu` key, should map to `GPUIndex`.

## Proposed fix

Replace the suffix check with an exact match on those names. PR to follow with a regression test.

Abdelsalam-Abbas:
## What

`isGpuResource` matched any resource name ending in `gpu`. Replace it with an exact match on `nvidia.com/gpu`, `amd.com/gpu`, and the internal `gpu` key, consistent with `ResourceFromResourceList`.

## Why

On clusters that also run HAMi, nodes advertise `nvidia.com/vgpu` (80 on an 8-GPU node) next to `nvidia.com/gpu: 8`. The suffix match folded `nvidia.com/vgpu` into `GPUIndex`, and `Resource.ToVector`'s `Set` overwrote the real count, so the scheduler saw 80 GPUs per node and bound pods onto fully allocated nodes. Kubelet rejected them with `UnexpectedAdmissionError`. Details and log evidence in #2194.

## Testing

- New Ginkgo cases in `resource_vector_test.go`: exact names map to `GPUIndex`; `nvidia.com/vgpu` and `example.com/mygpu` do not.
- `go test ./pkg/scheduler/...` passes.

## Changelog

Fragment added under `.changes/unreleased/`.

Fixes #2194


pkg/scheduler/api/resource_info/resource_vector.go
```diff
@@ -189,8 +189,15 @@ func convertResourceToFloat64(rName v1.ResourceName, rQuant resource.Quantity) f
 	return float64(rQuant.Value())
 }
 
+// Exact match only. A suffix match also captured unrelated resources such as
+// HAMi's nvidia.com/vgpu (80 slices per 8-GPU node), which then overwrote the
+// node's real GPU capacity and let the scheduler bind pods onto full nodes.
```
enoodle: explaining past code is redundant

enoodle (approved):
@Abdelsalam-Abbas Lets fix the sign off signature on the commit so that we can merge it

enoodle (approved):

PR #2200, opened 2026-09-19
perf(scheduler): accelerate greedy GPU matching

enoodle:
## Description

Replace repeated linear scans in `greedyMatchRequirements` with a compact max segment tree.

This is a performance-only change. It preserves current fixed-order first-fit behavior:

- Requirements and holders remain sorted descending.
- Each requirement selects earliest original-order holder with sufficient remaining capacity.
- Virtual-allocation arithmetic remains unchanged.
- Existing greedy false negatives remain false; this does not introduce exact bin packing, best-fit, or heap-based matching.

The segment tree stores maximum remaining capacity per holder range. Each lookup descends left first to find same earliest fitting holder, then updates residual capacity in O(log N).

Updated matcher coverage with fixed named trace cases that assert both exact selected-holder sequence and final result:

- First-fit holder order
- Known greedy false negative
- Fractional GPU requirements
- Zero tail
- Equal-capacity leftmost tie

## Related Issues

N/A

## Checklist

- [x] Self-reviewed
- [x] Added/updated tests
- [ ] Updated documentation (not needed)
- [x] Added changelog fragment via `make changelog`

## Breaking Changes

None.

## Additional Notes

| Measure | Baseline / linear | Segment tree | Change |
|---|---:|---:|---:|
| 2,000x2,000 triangular benchmark | 3.178 ms/op | 250.7 us/op | 12.7x faster |
| Allocations | 16 KiB / 1 | 64 KiB / 3 | Higher memory |
| Scheduler-state replay reclaim | 176.84 s | 47.09 s | 3.76x faster; 73.4% less |
| Scheduler-state replay total | 178.25 s | 48.58 s | 3.67x faster |

Scheduler-state replay preserved same scheduling outcome while exercising repeated accumulated-scenario filtering. It validates replay performance and decision parity, not live Kubernetes eviction or controller behavior.


pkg/scheduler/actions/common/solvers/accumulated_scenario_filters/idle_gpus/common.go
```diff
@@ -36,33 +38,81 @@ func greedyMatchRequirements[K comparable](
...
+	if len(requirements) == 0 || requirements[0] == 0 {
+		return true
+	}
+
+	totals := make([]float64, len(holders))
+	for i, holder := range holders {
+		totals[i] = capacity(holder)
+	}
+	allocated := make([]float64, len(holders))
+	tree := newMaxSegmentTree(totals)
```
gshaibi, line 54: Great optimization for large clusters!

Since the requirements and holders are already sorted, I would suggest also adding a fast path before building the tree - let R be requirement count, if there are at least R holders and the R-th largest capacity fits the largest requirement , return true to avoid allocations for easy matches.
This way we won't pay the price of tree allocations for small jobs.

gshaibi (dismissed):

pkg/scheduler/actions/common/solvers/accumulated_scenario_filters/idle_gpus/common.go
enoodle: @gshaibi The difference is about 0.36% for small / medium clusters (according to the benchmark results above) - does it worth the extra code complication of having both implementations just to save it for small / medium clusters?

gshaibi: In that case, no. I imagined it would have a bigger impact.

gshaibi: Though now that I think about it, the improvement would be for *small workloads* on large clusters, not for small/medium clusters as the benchmark measures. 
I'm not sure we have benchmark for that though.
Anyway, this can be done later on upon need.

gshaibi (approved):

pkg/scheduler/actions/common/solvers/accumulated_scenario_filters/idle_gpus/common.go
enoodle: Fixed.

